PR #277, opened 2015-05-29
Fix inverse stereo projection on an ellipsoid.

### LINKED ISSUE #266: Bug with inverse stereographic projection
**Reported by qulogic on 2 Mar 2015 02:00 UTC**
[This bug](http://osgeo-org.1560.x6.nabble.com/Problem-with-inverse-stereo-projection-td3841805.html) was reported ''nearly 5 years ago'' with a patch, but seems to have been forgotten and not fixed. I am posting it here so that hopefully it gets the attention it deserves.

massimo costantini:

> Hi,
> someone can tell me why if I set a center o stereo projection on lat 0.0 and lon 0.0 the inverse projection (stereo to geo), return always 0E 0N.

Glynn Clements:

> It works for a sphere, but not an ellipsoid.
> 
> The problem is that the setup function doesn't initialise all of the
> values that it needs for the ellipsoid, inverse, equatorial case.
> 
> In the forward and spherical inverse cases, the equatorial case
> (|phi0| < 1e-10) is a simplification of the oblique case, and requires
> fewer parameters. However, for the ellipsoid inverse case (e_inverse),
> the oblique calculation is used, but the setup function doesn't
> initialise all of the parameters.
> 
> Modifying the setup() function in PJ_stere.c so that EQUIT and OBLIQ
> use the same code should fix the problem.

```
--- PJ_stere.c 2001-02-07 17:55:05.000000000 +0000
+++ PJ_stere.c.tmp 2010-05-14 18:55:26.000000000 +0100
@@ -195,8 +195,6 @@
  }
  break;
  case EQUIT:
- P->akm1 = 2. * P->k0;
- break;
  case OBLIQ:
  t = sin(P->phi0);
  X = 2. * atan(ssfn_(P->phi0, t, P->e)) - HALFPI;
```

Migrated-From: https://trac.osgeo.org/proj/ticket/266

QuLogic:
The problem is that the setup function doesn't initialise all of the
values that it needs for the ellipsoid, inverse, equatorial case.

In the forward and spherical inverse cases, the equatorial case
(|phi0| < 1e-10) is a simplification of the oblique case, and requires
fewer parameters. However, for the ellipsoid inverse case (e_inverse),
the oblique calculation is used, but the setup function doesn't
initialise all of the parameters.

Fixes #266, with patch pulled from there / on the mailing list.

### DIFF
--- a/src/PJ_stere.c
+++ b/src/PJ_stere.c
@@ -192,8 +192,6 @@ setup(PJ *P) { /* general initialization */
 			}
 			break;
 		case EQUIT:
-			P->akm1 = 2. * P->k0;
-			break;
 		case OBLIQ:
 			t = sin(P->phi0);
 			X = 2. * atan(ssfn_(P->phi0, t, P->e)) - HALFPI;


rouault:
Merged but an added test case would have been appreciated

QuLogic:
Sorry. don't have one; I just pulled the patch from the mailing list.

PR #278, opened 2015-05-30
Fix small memory leak in pj_init.

QuLogic:
Patch from #149.


QuLogic:
Sorry, forgot to indicate this will close #149, so you'll have to do that manually.

PR #280, opened 2015-05-31
Several warning fixes

### LINKED ISSUE #4: Patch for warnings and perhaps a bug
**Reported by dcthomp on 3 Jun 2008 15:19 UTC**
I've made some changes to PROJ.4 to get it compiling without
any warnings even setting CFLAGS="-W -Wall". I've attached a
patch.

Most of the changes are straightforward, a few are tedious, and
one of the warnings appears to actually point to a bug. Specifically,
in src/PJ_imw_p.c it looks like the argument "double\* yc" is
shadowed by a local variable below and thus never used. The patch
includes what I believe to be a fix, but it would be nice if someone
reviewed it.

The only other real problem I had was with the declaration and
initialization of "static char EMESS_H_ID[]" inside src/emess.h.
Since this header file is included multiple times it results in
EMESS_H_ID being instantiated in multiple object files. I removed
the variable since it doesn't appear to be used.

Migrated-From: https://trac.osgeo.org/proj/ticket/4

QuLogic:
The code will now compile without _extra_ warnings (i.e, with `CFLAGS='-Wall -Wextra -Werror'`) on both gcc 5.1.1 and clang 3.5.0.

Supersedes and fixes #4.


QuLogic:
Now compiles with optimization too (`CFLAGS='-g -O2 -Wall -Wextra -Werror'`) on gcc and clang (except for some false-positives in system headers with clang).

QuLogic:
OK, fixed your comment (and it passes on the new Travis too!)

PR #282, opened 2015-06-28
Remove generated files and minor build updates

QuLogic:
Generated files really should not be in a VCS. It is especially more annoying with git where you will not be able to do things because of a dirty workspace (e.g., no merging if both changed generated files, no rebase, no bisect, etc.)

Since there will be no generated files in git, one needs to run `./autogen.sh` (it already existed) when running from git. Note also that means that tarballs that GitHub automatically generates are **not** valid _release_ tarballs. But that's fine; you can run `make dist` to produce one, and you need to test the tarball before officially releasing it anyway.


rouault:
I believe this should be discussed on the proj mailing list first

QuLogic:
This should be back in sync again.

PR #289, opened 2015-08-13
Fix the constants used for degree/radian conversions.

cffk:
In dmstor.c an inaccurate constant was used for pi/180.  This meant that
90deg converted to radians and back to degrees gave a result which was
larger than 90deg.  The constant has been replaced by DEG_TO_RAD
(defined in proj_api.h).

In rtodms.c, the trailing digits in CONV were wrong; the constant has
been truncated so that all the digits are corrent.


cffk:
Needless to say that some of the tests have the bad value of pi/180
baked in.  So someone needs to verify the new (presumably more accurate)
values.  I believe that sticking with the old buggy value of pi/180 is
not an option.  It's really bad news that cos(90 deg) is negative and
than 0N 0E and 0N 180E are not on the same meridian.

hobu:
@cffk Please merge/update this when you are comfortable.

cffk:
I recommend slipping this fix to after the release of 4.9.2.  I'll make the merge then to
give people time to address the failing tests.

hobu:
> I recommend slipping this fix to after the release of 4.9.2

Yes, that was my intention. I presume there will be some ripples...

cffk:
I propose to wait a few days before doing a merge, in case anyone wants to
weigh in on the changes I made to the tests...

rouault:
@cffk I think this commit caused the following test failure on MacOSX : https://travis-ci.org/rouault/proj.4/builds/83738157

```
310c310

< -140.100000 -87.000000 987122.418330284 -14429896.539530911 0.000000000

---

> -140.100000 -87.000000    987122.418330284    -14429896.539530909 0.000000000

```

cffk:
OK, I'll fix...

PR #290, opened 2015-08-16
Drop in the latest geodesic library from GeographicLib (version 1.44).

cffk:
See

  http://geographiclib.sourceforge.net/1.44/C/index.html

The changes are:
- Improve accuracy of calculations by evaluating trigonometric
  functions more carefully and replacing the series for the reduced
  length with one with a smaller truncation error.
- The allowed ranges for longitudes and azimuths is now unlimited; it
  used to be [-540d, 540d).
- Enforce the restriction of latitude to [-90d, 90d] by returning NaNs
  if the latitude is outside this range.
- The inverse calculation sets s12 to zero for coincident points at
  pole (instead of returning a tiny quantity).

This commit also includes a work-around for an inaccurate value for
pi/180 in dmstor.c (see the definitions of DEG_IN and DEG_OUT in
geod_interface.c).

### DIFF
--- a/man/man1/geod.1
+++ b/man/man1/geod.1
@@ -30,16 +30,22 @@ file[s]
 ]
 file[s]
 .SH DESCRIPTION
-.I Geod
+.I geod
 (direct) and
 .I invgeod
 (inverse)
-perform geodesic (\*(lqGreat Circle\*(rq) computations for determining
+perform geodesic ("Great Circle") computations for determining
 latitude, longitude and back azimuth of a terminus point
 given a initial point latitude, longitude, azimuth and distance (direct) or
 the forward and back azimuths and distance between an initial and
 terminus point latitudes and longitudes (inverse).  The results are
 accurate to round off for |\fIf\fR| < 1/50, where \fIf\fR is flattening.
+.B invgeod
+may not be available on all platforms; in this case call
+.B geod
+with the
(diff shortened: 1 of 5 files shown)


cffk:
How is the coverage being measured?  I have several tests of the geodesic routines from the C++ library that I can drop into proj.4.  Would that help?

rouault:
I don't think you should take care much about this tiny decreasing coverage percentage. This could probably be modified in coveraslls settings. It must just be a consequence of more lines of code added that aren't triggered by existing tests

cffk:
The pull request addresses issues #291.

cffk:
Any chance of this pull request being accepted for version 4.9.2?

rouault:
@cffk I assumed you had direct commit rights to OSGeo/proj.4, haven't you ?

cffk:
Oh, probably; let me check.

cffk:
The pull request page says (for me)

"This branch is up-to-date with the base branch
Only those with write access to this repository can merge pull requests."

which I assume means that I don't have write access.  (I also tried a "manual" push
and this fails.)

rouault:
@cffk I've just created a proj-4-committers team, with write access to OSGeo/proj.4, and invited you to join. So after you accept, that should work, hopefully

PR #296, opened 2015-09-06
Fix 'flaky' typo, causing `make dist-all` failure.

sebastic:
`nad/Makefile.am` uses `nad/testflaky`, but the file was named `nad/testflakey`.

This caused `make dist-all` to fail with:
`make[1]: *** No rule to make target 'testflaky', needed by 'distdir'.  Stop.`


rouault:
I've also added 'make dist-all' in .travis.yml so that this can be spotted earlier : https://github.com/OSGeo/proj.4/commit/c8d1b0431c73f44b352d4120e5d98a2f98127ec0

PR #299, opened 2015-09-07
.travis.yml: do make, make install, make dist-all from the tarball generated by a first make dist-all

rouault:


rouault:
"Merged" in a cleaner way in master

PR #301, opened 2015-09-08
Don't include files in proj dist, also included in proj-datumgrids dist.

sebastic:
As discussed in the [Missing proj_def.dat in 4.9.1 release](http://lists.maptools.org/pipermail/proj/2015-September/007229.html) thead on the proj list,
`make dist` for proj shouldn't include the datumgrid files also included in the proj-datumgrid distributions.

The following files shouldn't be included in proj, because they're in proj-datumgrids already:

```
 nad/ntf_r93.gsb        (included in proj-datumgrids <= 1.6RC1)
 nad/ntv1_can.dat       (included in proj-datumgrids <= 1.6RC1)
 nad/null.lla           (included in proj-datumgrids <= 1.5)
 nad/nzgd2kgrid0005.gsb (included in proj-datumgrids <= 1.6RC1)
```

This change comprises two (partial) reverts:

1) Revert "nad/Makefile.am: add missing 'null.lla ntf_r93.gsb nzgd2kgrid0005.gsb ntv1_can.dat' to EXTRA_DIST"

This reverts commit a4c7892a94667e62d62caa053ac4b51ef790741c.

2) Partial Revert "add files to EXTRA_DIST so they are removed by make uninstall"

This partially reverts commit 0a7eda807da1e17f60558acebd89a08872b5b6e9.

### DIFF
--- a/nad/Makefile.am
+++ b/nad/Makefile.am
@@ -16,14 +16,15 @@ TESTIGN = $(NADPATH)/testIGNF
 pkgdata_DATA = GL27 nad.lst proj_def.dat nad27 nad83 world epsg esri \
 		esri.extra other.extra \
 		CH IGNF testIGNF proj_outIGNF.dist \
-		null.lla ntf_r93.gsb nzgd2kgrid0005.gsb ntv1_can.dat makefile.vc CMakeLists.txt
+		makefile.vc CMakeLists.txt
 
 EXTRA_DIST = GL27 nad.lst proj_def.dat nad27 nad83 pj_out27.dist pj_out83.dist td_out.dist \
 		test27 test83 world epsg esri tv_out.dist tf_out.dist \
 		testflaky testvarious testdatumfile testntv2 ntv2_out.dist \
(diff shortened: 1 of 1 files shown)


sebastic:
I've added the second commit to this PR to ease merging, but it should have been a separate PR.

This still keeps the changes from 0a7eda8 to include the IGNF test files, the makefile.vc, and CMakeLists.txt in the distribution tarballs. `make install` just won't install them.

rouault:
@sebastic Looks good to me. Thanks. Confirms my post-mortem analysis (http://lists.maptools.org/pipermail/proj/2015-September/007231.html). And I guess that the revert of https://github.com/OSGeo/proj.4/commit/ba3454ec0f9d1d33e165d0ace70f766c8283892e will also be needed to please Travis, as this was in fact the reason for those changes. I let @hobu double check as it is late here and my brain becomes foggy.

rouault:
@hobu note the 2 above follow-up commits

Issue #302, opened 2015-09-09
Problems building on FreeBSD 10, PTHREAD_MUTEX_RECURSIVE


turepalsson:
I'm having problems building on FreeBSD 10. In `src/pj_mutex.c`, the following lines

```
#ifndef PTHREAD_MUTEX_RECURSIVE
        pthread_mutexattr_settype(&mutex_attr, PTHREAD_MUTEX_RECURSIVE_NP);
#else
        pthread_mutexattr_settype(&mutex_attr, PTHREAD_MUTEX_RECURSIVE);  
#endif
```

blow up, because `PTHREAD_MUTEX_RECURSIVE` is an `enum` constant, not a `#define`:d value (and `PTHREAD_MUTEX_RECURSIVE_NP` is not defined at all).

rouault:
Fixed per https://github.com/OSGeo/proj.4/pull/303

PR #303, opened 2015-09-09
Use autoconf for PTHREAD_MUTEX_RECURSIVE (#302)

turepalsson:
On FreeBSD 10, PTHREAD_MUTEX_RECURSIVE exists, but as an enum constant,
not a #define macro, so it cannot be detected using #ifdef.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -98,6 +98,10 @@ if test "$with_mutex" = yes -o x"$with_mutex" = x ; then
         THREAD_LIB="-lpthread"
     fi
     AC_CHECK_LIB(pthread,pthread_mutexattr_settype,,,)
+    AC_CHECK_DECL(PTHREAD_MUTEX_RECURSIVE,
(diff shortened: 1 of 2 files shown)


turepalsson:
Oh, bother, it breaks on Windows. No idea how to fix that, unfortunately. :)

rouault:
Perhaps including proj_config.h only within a #ifndef _MSC_VER #endif

turepalsson:
Ah, I was worrying that I didn't know what PTHREAD_\* stuff to use on Windows, but now I realize that Windows doesn't use pthreads at all.  :)

How about moving the include inside the existing #ifndef _WIN32, would that make sense?

rouault:
> How about moving the include inside the existing #ifndef _WIN32, would that make sense?

Should be fined too

hobu:
To be included in 4.9.2RC2...

rouault:
Merged in master

PR #311, opened 2015-09-24
Extension of proj API with fwd3d and inv3d

piyushrpt:
Hi,
    I have made appropriate changes to include a fw3d and inv3d API functions. And included a new projection system called the SCH system, often used in radar imaging by NASA. 

Thanks
Piyush


src/PJ_sch.c
```diff
@@ -0,0 +1,206 @@
+/*
+** Author : Piyush Agram
+** This code is in the public domain and implements the SCH coordinate system first developed at JPL.
```
QuLogic: Public domain is not really a valid license. And it's going to be confusing if it's not under the same license as the rest of the code.

hobu: I believe the contributor is a U.S. federal employee, and therefore cannot define a license other than public domain for contributions like this (or at least not easily without some bureaucratic mountain moving). I think there are many situations (see GEOTRANS) where code is contributed in this fashion.

rouault: See how FrankW dealt with similar situation in GDAL : https://github.com/OSGeo/gdal/blob/trunk/gdal/frmts/pds/pdsdataset.cpp

piyushrpt: I'm making this contribution as a volunteer (in my own free time). The appropriate documentation for the SCH coordinate system has been officially approved for public release. I'm happy to change the headers to read something like the example rouault pointed me to.

piyushrpt:
Hi,
    I made significant changes to the header of PJ_sch.c file and made changes to make sure that my changes did not break anything using travis-ci and appveyor. Do let me know if more changes are needed for this functionality to get included. 

Thanks
Piyush

hobu:
I'm :+1: to allow inclusion of this PR, but I want to hear from someone who knows proj4 better than me. I've nagged @rouault and @warmerdam to give things a once over, but if we don't hear from them in a week or so, I will simply merge it regardless.

Thanks for your contribution!

rouault:
Looks good to me with the following remarks :
- it would be great to see SCH forward and inverse tested in nad/testvarious
- pj_inv(), pj_fwd(), pj_inv3d(), pj_fwd3d() should have checks to make sure the function pointer called is not NULL
- a further step would be to retrofit the geocent projection to the new 3D framework (but that doesn't necessarily have to be part of this PR)

piyushrpt:
I agree. Geocent would seamlessly fit into the fwd3d, inv3d scheme. Maybe for completeness, when we check for existence of functions pointers - in the fwd3d, inv3d we could also revert to calling fwd, inv if the 3d functions were not defined while retaining the same z values. That way fwd3d, inv3d would also be available for all sets of projections.

rouault:
> Maybe for completeness, when we check for existence of functions pointers - in the fwd3d, inv3d we could also revert to calling fwd, inv if the 3d functions were not defined while retaining the same z values. That way fwd3d, inv3d would also be available for all sets of projections.

I'm somewhat ambivalent about the fallback. pj_inv3d(), pj_fwd3d() should in most cases be only internal functions of proj.4 called by pj_transform(), so I don't think we need much user friendliness here.

piyushrpt:
This contribution has now officially been approved for unlimited public release. I will updated the latest files later today after making sure that it doesnt break anything.

piyushrpt:
Hi,
   I have completed the updates to my branch. The contribution has been officially approved and made available for distribution without restrictions. It would be great if you could merge this with the main branch.

Thanks
Piyush

rouault:
It doesn't look like the following of my above comments have been addressed :
- it would be great to see SCH forward and inverse tested in nad/testvarious
- pj_inv(), pj_fwd(), pj_inv3d(), pj_fwd3d() should have checks to make sure the function pointer called is not NULL

(And personnaly, I believe if would probably be appropriate to have a "clean" PR with clean commits instead of the incremental ones)

piyushrpt:
Great. Will include the tests and revert. Feel free to delete this request. Will make a fresh request when ready.

QuLogic:
For future reference, you could have left this PR open and rebased to squash the extra commits. Also, I see this PR is based on your `master` branch; GitHub works a lot better if you create a separate branch for each PR (just to keep things self-contained), though it probably won't harm anything.

piyushrpt:
Still a little new to git/github. Will keep your advice in mind for future changes. 
Thanks.

PR #315, opened 2015-10-05
avoid read failure for minimal GTX files #269

hobu:


QuLogic:
Can close #269 now, then?

PR #317, opened 2015-10-14
Avoid strcpy of overlapping strings.

QuLogic:
The source and destination for `strcpy` must not overlap, and failure to ensure that's true causes a crash on OS X. `memmove` does not have such a restriction.

### DIFF
--- a/src/rtodms.c
+++ b/src/rtodms.c
@@ -57,13 +57,15 @@ rtodms(char *s, double r, int pos, int neg) {
 		(void)sprintf(ss,format,deg,min,sec,sign);
 	else if (sec) {
 		char *p, *q;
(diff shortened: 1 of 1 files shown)


QuLogic:
Hrm, wondering if maybe this should go into one of the 4.9 branches (why are there two?) as a bugfix instead of `master`?

rouault:
The 4.9 is obsolete and was based on the abandonned 4.9.0.
4.9.2-maintenance was branched off 4.9.2 release and would be appropriate for backport

rouault:
Cherry-picked in 4.9.2-maintenance as b9fe7472186b8d37f16aa6b7c6039e26b49cf6e4

PR #319, opened 2015-10-17
Fix configure.ac to add config.guess / config.sub

rouault:
On my system (Ubuntu 10.04), ./autogen.sh doesn't copy config.guess
and config.sub, which cause ./configure to fail


rouault:
Both commits merged together in 1906ed3c70d1cf3474add6909b30276a5b0d7c8c

PR #322, opened 2015-10-22
API extension and SCH support

piyushrpt:
Hi,
    Pull request with following features:
1. Support for new API functions - fwd3d and inv3d. This will allow users to define more general transformations between coordinate systems. 
2. Support for a sensor-centric coordinate system called SCH - commonly used in radar imaging, that uses this new functionality. Including tests. 
3. NULL pointer checks in pj_fwd(), pj_inv(), pj_fwd3d(), pj_inv3d() which returns HUGE_VAL, when corresponding transformation functions fwd(), inv(), fwd3d() and inv3d() are undefined. 

Do let me know, if you need further changes.
Piyush


piyushrpt:
I have tried to follow @QuLogic 's suggestion and rebase everything into a single commit. Do let me know if there is any other information that I can provide or if there is something I'm missing.

src/PJ_sch.c
```diff
@@ -0,0 +1,228 @@
+/******************************************************************************
+ * $Id$
```
QuLogic, line 2: You can leave this `$Id$` out; it's a subversion thing.

PR #329, opened 2015-12-12
inverse for hammer projection

jswhit:
Implements the inverse projection for Hammer and Eckert-Greifendorff based on
http://mathworld.wolfram.com/Hammer-AitoffEqual-AreaProjection.html

submitted this back in 2010 (http://osgeo-org.1560.x6.nabble.com/patch-to-PJ-hammer-c-for-inverse-projection-td3841509.html), but it apparently slipped through the cracks.


QuLogic:
The indent in the new lines is an odd mix of tabs and spaces, but I'm not totally sure what the convention is here.

rouault:
Could use add a new test case in nad/testvarious and expected result in nad/tv_out.dist ? Untested code is future broken coden.

jswhit:
OK - I've cleaned up the spaces and tabs, and added a test in testvarious (with expected output in tv_out.dist).

PR #348, opened 2016-01-26
Increase maximum path size for grid shift loading

drhaynes:
On systems with long paths (e.g. iOS, which uses UUIDs for bundle
directories), the existing maximum path length of 128 resulted in path
truncation, and caused the grid shift file loading to fail.

This change increases the buffer size for the path to 1024 chars,
enabling the nadgrids file to be loaded on such systems.

### DIFF
--- a/src/pj_gridlist.c
+++ b/src/pj_gridlist.c
@@ -43,6 +43,7 @@
 #endif /* _WIN32_WCE */
 
 static PJ_GRIDINFO *grid_list = NULL;
+#define PJ_MAX_PATH_LENGTH 1024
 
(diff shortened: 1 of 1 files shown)


hobu:
Please fix up so the windows builds compile and run successfully, and I'll be happy to merge.

https://ci.appveyor.com/project/OSGeo/proj-4/build/1.0.116/job/drlu8xqd7bsy9m45#L78

drhaynes:
Who doesn't love pleasing the MS C compiler :-)

Are we ok with the coveralls numbers? Looks like the reduction in coverage is <0.1%

PR #347, opened 2016-02-05
unnecessary assignment in Lambert Azimuthal Equal Area

bastik:
In [PJ_laea.c](https://github.com/OSGeo/proj.4/blob/04058618706b306520f1a4656a521ba555601b3b/src/PJ_laea.c#L118), line 118 and 121, there are assignments to the local variable q, but q is never used afterwards. I verified, that the calculation is correct nevertheless, the multiplication and assignment can simply be dropped (keeping the `ab = ...` part).


bastik:
Do you like me to make a pull request?

hobu:
Please do.

hobu:
@rouault can you look at this one and :+1:?

rouault:
LGTM. .Just merged it

Issue #354, opened 2016-02-15
updates to the geodesic(3)


cffk:
proj version 4.9.2 incorporates the geodesic library from GeographicLib
1.44.  I'd like to upgrade it to the latest GeographicLib version, 1.46.
Here are the changes:

Version 1.45 http://geographiclib.sourceforge.net/1.45/C
- This fixes the bug where the solution of the inverse problem with nan
  and a point on equator did not return all nans.  This is illustrated
  by the following call to geod (version 4.9.2)
  
  ``` bash
  $ echo 91 0 0 90 | geod +ellps=WGS84 -I -f %.5f
  90.00000 -90.00000 10018754.171
  ```
  
  This should return "nan nan nan", since latitude 91&deg; is treated as a
  nan.
- A test suite is included.
- Add geod_polygon_clear().

Version 1.46 http://geographiclib.sourceforge.net/1.46/C
- More accurate inverse solution when longitude difference is close to
  180&deg;.
- Add s13 and a13 to the geod_geodesicline struct.
- Add geod_directline, geod_gendirectline, and geod_inverseline.

The last two changes in 1.46 simplify the process of generating
waypoints on a geodesic specified by its endpoints.

The changes in 1.45 and and 1.46 are both source-code compatible with
1.44.  However, version 1.46 requires an increment to the proj library
version, since the geod_geodesicline struct has increased in size.

So the question is: should I upgrade just to 1.45 (no change in library
version) or should I go all the way to 1.46 (which will require a change
in library version)?

My recommendation would be to opt for the full upgrade unless a release
is anticipated in the next month or so.

hobu:
I am planning to issue a 4.9.3 release candidate during the [OSGeo Paris Code Sprint](https://wiki.osgeo.org/wiki/Paris_Code_Sprint_2016) next week. I don't know if any one else was going to tackle proj.4 issues next week. I see no reason why you can't go to 1.46 and increment the soversion'ing as necessary in support of it.

cffk:
Done!  See pull request #355

PR #355, opened 2016-02-15
Upgrade geodesic library from GeographicLib 1.46.

cffk:
- upgrade geodesic.[ch3]
- add test suite geodtest.c and invoke via cmake's add_test
- increment version to 4.9.3 and library version to 11.0.0

This addresses issue #354.


QuLogic:
If I'm not mistaken, you've only _added_ to the ABI? Then according to [libtools docs](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html), you'd also need increment the last field as well (i.e., 11:0:1)

cffk:
@QuLogic Here's my understanding of the libtool rules:

Since one of the structs has changed size, any function that accepted
such a struct as an argument counts as "changed interface" which
requires that "age" be set to 0.  This means that the library version is
reset to NN.0.0.

It took me a while to beat src/proj.def into shape -- and I noticed that
some geod_\* entries were missing in 4.9.2.  However, I believe this is
now up to date.  (It's not clear to me that the ordinal designations
at-MM in this file are necessary.  Microsoft says "We don't recommend
exporting functions by ordinal unless your DLL's clients need it for
legacy support."  But perhaps we should let sleeping dogs lie.)

The test suite is an executable, geodtest.  With a cmake build, this is
compiled (but not installed), and invoked by add_test.  I haven't added
it to the autoconf configuration (I rather wish this would go away);
probably having the tests in just the cmake build suffices to catch
problems?  As far as I can tell the automated builds do not run the
cmake tests.  Shouldn't they?

I took the opportunity to fix 2 "double to int" warnings in geod_set and
etmerc.  (My ideal is to have compilations with no warnings -- but
that's hard to achieve on legacy code.)

Ignoring the coverage check, this patch is now passing.  I'll wait a day
for any additional feedback and then merge.

QuLogic:
Ah sorry,  I missed the updated struct in the summary in #354; carry on then.

I think testing is a bit sparse overall; if you want to enable it in more places (and it works correctly), I'm sure the powers that be would not complain.

Issue #337, opened 2016-01-02
cs2cs man page should be maintained in its original nroff format


jidanni:
https://github.com/OSGeo/proj.4/wiki/Mancs2cs should be maintained in its original nroff format. It should say CS2CS(1) on its corners. And should have an updated URL at bottom, and have a BUGS section.

micahcochran:
I've done all of those changes with the exception of adding a BUGS section.  I have submitted a pull request.

micahcochran:
PR #382 has been merged.  Closes this issue.

PR #356, opened 2016-02-19
Man Page Updates

micahcochran:
cs2cs header was updated.  HOME PAGE links were changed.  Converted man pages to html using man2html program using option "-r".  Any changes in the html are changes that already existed in the original man page format.

This is in regard to issue #337 .
[ci skip]

### DIFF
--- a/html/man_cs2cs.html
+++ b/html/man_cs2cs.html
@@ -1,6 +1,9 @@
-<HTML><HEAD><TITLE>Manpage of PROJ</TITLE>
+Content-type: text/html; charset=UTF-8
+
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
+<HTML><HEAD><TITLE>Man page of CS2CS</TITLE>
 </HEAD><BODY>
-<H1>PROJ</H1>
+<H1>CS2CS</H1>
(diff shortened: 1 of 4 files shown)


micahcochran:
Thanks!

cffk:
For the record, #355 updated geod.1 but not the html version.  Should I do this?
In an ideal world the html file would not be checked into git, but would be created
when a release is packaged (as is the ./configure script, for example).

hobu:
I'm trying to get a process together to generate stuff from Sphinx too, but i'm not quite there yet

http://proj4.org

micahcochran:
When I was updating this, I was thinking it would be nice to have a shell script that generated the html.  I think there needs to be an overhaul of this.

I was simply running:

```
man2html -r man/man1/geod.1 > html/man_geod.html
```

I tried to regenerate pj_init.3 and man2html choked on the input, but man is fine with pj_init.3.  I did figure out how to use groff to do this, but I didn't like groff's output.

I should be able to get a shell script that creates an html version of the man pages.  It won't be compatible with Windows, but it might be an improvement.

cffk:
Assuming this is done at the packaging stage, there's no need for compatibility with windows.
(Presumably autogen is being run on a Unix-like system.)

PR #358, opened 2016-02-23
Add support for MISR SOM projection

mbull2016:
This adds a new projection type 'misrsom' with '+path=' argument for MISR path numbers.  It is similar to existing 'lsat' projection type.


hobu:
Thanks for the contribution, but this pull request fails both compilation tests. It will have to be working and tested before we can merge it.

mbull2016:
I'm not sure how to fix the automatic compilation test.  I get a successful build on Fedora 20 with the following:

```
sh autogen.sh
./configure CFLAGS=-std=c99 
make
```

mbull2016:
Compilation error fixed

html/man_cs2cs.html
```diff
@@ -1,6 +1,9 @@
-<HTML><HEAD><TITLE>Manpage of PROJ</TITLE>
+Content-type: text/html; charset=UTF-8
```
QuLogic, line 1: This shouldn't be here, I think.

mbull2016:
Closing to start again with a fresh branch

PR #359, opened 2016-02-24
Generate HTML versions of man pages using a shell script

micahcochran:
Adds gen_htmlman.sh shell script, which makes directories and runs man2html.  This is based on discussion from PR #356. I got the HTML files to install on CMake. It will conditionally install the html file based on if the directories exist. 

Removed the file `html/man_*.html`.

I was NOT brave enough to even think about attempting to have autotools install the html man pages.  This does not break the autotool installation.

I did not touch HOWTO-RELEASE, which should be updated to add a step to run the script ./gen_htmlman.sh.


QuLogic:
I'm not sure this shell script is the best way to go. It would be better to create a custom command in CMake/autotools, and then set up the targets for each file. This way, dependencies are handled automatically.

micahcochran:
This script intended to only be ran when creating a distribution for release.  I'm just trying to make a small improvement.

If there is a better way, please by all means create a better competing patch.  I have no background in CMake/autotools (with the exception of the CMake I learned for this patch), so I cannot be of any help.

I forgot to these lines to `.gitignore`:

```
# html
/html/dl
/html/robots.txt
# the below lines need to be added
/html/man1/*.html
/html/man3/*.html
```

Also, the line at the top in the HTML Content-type: text/html; charset=UTF-8` is some slightly odd behavior being created by Debian's patches to man2html.

hobu:
@micahcochran At the [OSGeo Code Sprint](https://wiki.osgeo.org/wiki/Paris_Code_Sprint_2016), I have started trying to collect all of proj.4's documentation into a `./docs` directory, to be generated and served by [Sphinx](http://www.sphinx-doc.org/en/stable/). My plan is as follows:
1. Use [github pages](https://pages.github.com/) to host the proj.4 site at http://proj4.org
2. Collect as much wiki, trac, postscript and other documentation as possible into the Sphinx-organized tree
3. Generate HTML, PDF, and man documentation from the Sphinx configuration
4. Configure Travis-CI to build and push the documentation every time a commit is made to the maintenance branch for the release (`4.9.2-maintenance` at the moment).

The current branch is `hobu-docs-refactor` in the main `proj.4` github tree.

micahcochran:
@hobu I'm beginning to understand that this a complete overhaul of the documentation.  I assume the goal is to have living documentation that is updated when the code is updated.  reStructuredText seems to be a good fit since it is easy to edit in github.  I assume Travis and Sphinx will be used to parse the documentation every time and flag syntax errors in a pull request.  I think that I now understand, and can get behind what you are doing.  (This PR is pretty useless to accomplishing your goal.)

> Collect as much wiki, trac, postscript and other documentation as possible into the Sphinx-organized tree

So the man pages will need to be converted from groff syntax into reStructuredText.  Other documentation, if still relevant, will also need to be converted.

BTW I'm not the author of parameters.rst that was just a copy of a page from the old wiki https://trac.osgeo.org/proj/wiki/GenParms .  Frank Warmerdam appears to be the author (see the [revision history of that wiki](https://trac.osgeo.org/proj/wiki/GenParms?action=history)).

hobu:
>  I assume the goal is to have living documentation that is updated when the code is updated

Exactly. 

> So the man pages will need to be converted from groff syntax into reStructuredText. Other documentation, if still relevant, will also need to be converted.

Yes, I think so. I wasn't quite sure if there was a prescribed format for reStructuredText Sphinx pages to convert to man format. I assume there is, but looking around, I didn't find any clear examples.

> BTW I'm not the author of parameters.rst that was just a copy of a page from the old wiki

I will update... Thanks for the notice, and thanks for the documentation contributions. It will be really nice to collect all this stuff up into one management structure.

micahcochran:
> Yes, I think so. I wasn't quite sure if there was a prescribed format for reStructuredText Sphinx pages to convert to man format. I assume there is, but looking around, I didn't find any clear examples.

I looked a little and found sphinx-build, [github](https://github.com/sphinx-doc/sphinx/tree/master/doc/man) and [html](http://www.sphinx-doc.org/en/stable/man/sphinx-build.html).  I've not seen the groff output.

Thank you.

hobu:
I will merge my current docs branch into master to make it convenient for you to continue to contribute. Thanks so much for your efforts!

PR #368, opened 2016-03-27
Add Patterson Cylindrical Projection

micahcochran:
Implements issue #252 to add this projection.   (migrated from https://trac.osgeo.org/proj/ticket/252  )

The tests points were projected into Patterson Cylindrical using [Flex Projector software](http://www.flexprojector.com/), which uses the [Java Map Projection Library](https://github.com/OSUCartography/JMapProjLib) (the algorithm reference implementation).   Note that Java Map Projection Library uses 6371008.7714 meter diameter for it SPHERE ellipsoid, so the tests reflect that.


micahcochran:
Thank you!

PR #372, opened 2016-03-29
Move math constants to projects.h.

micahcochran:
Any constants used more than once in the project were moved to projects.h. If name was not obvious, I looked mostly looked in "libproj4: A Comprehensive Library of Project Functions" for a formula match to confirm that it is the same constant.  Add pseudocode description for those constants (example: HALFPI means pi/2).  Ensure plenty of precision for constants by using Python gmpy2 library to calculate constants to many digits of precision.  I've have the python console output, if anyone wants to see it.

I tried to make changes to PJ_isea.c, which caused compilation errors.  So, I left it alone.

I did some extra testing in order to ensure that these changes didn't mess anything up.  I used an 80 coordinate test [testpat2.txt](https://github.com/OSGeo/proj.4/files/194462/testpat2.txt).  With the bash script [test_old_new.txt](https://github.com/OSGeo/proj.4/files/194463/test_old_new.txt).

The tests all match.   However, I didn't test the omerc projection nor the JNI code.   I, also, just remember that I didn't test inverse projections (which should be done).  I did have to change the test to exclude latitude of 90, and longitude of 180, which caused tolerance errors for a few projections.

If wanted, the bash script could fairly easily be modified into a unit test.  I think I would be willing to do that.

### DIFF
--- a/src/PJ_aitoff.c
+++ b/src/PJ_aitoff.c
@@ -32,13 +32,6 @@
 #define PJ_LIB__
 #include <projects.h>
 
-#ifndef M_PI
-#  define M_PI 3.14159265358979323846
-#endif
-#ifndef M_PI_2
-#  define M_PI_2 1.57079632679489661923
-#endif
-
 PROJ_HEAD(aitoff, "Aitoff") "\n\tMisc Sph";
 PROJ_HEAD(wintri, "Winkel Tripel") "\n\tMisc Sph\n\tlat_1";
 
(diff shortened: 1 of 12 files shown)


micahcochran:
I added inverse tests (for the projections with inverse functions), which passed and gave consistent results.   I added two more projections to the tests.

updated test [test_old_new2.txt](https://github.com/OSGeo/proj.4/files/196114/test_old_new2.txt)

hobu:
Please notify the mailing list that you've done this. I don't see any problem with it, but others may have an opinion.

micahcochran:
I have applied to mailing list membership.  I will post I will do this as soon as I get approval.

hobu:
> applied to mailing list membership

You should be able to join the mailing list with no approval. Did it require you to wait in an approval queue?

micahcochran:
When I signed up, it gave me ambiguous message that depending on mailing list settings it *might\* require approval.   I tried to send an email to the list, and it sent it back "Your email has been rejected because it was posted from an email address that is not subscribe to the PROJ mailing list."

I did sign up from a vanity email address, which I know is a flag for spammers.  I could sign up with my work email address.

QuLogic:
`M_PI` is the non-standard standard name. That is the reason for the `#ifdef` in the first file here. See [GNU libc docs](https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html) and [MSVC docs](https://msdn.microsoft.com/en-us/library/4hwaceh6.aspx).

I don't know if it's better to use these names instead and only define them if undefined.

micahcochran:
I'm not opposed to using the `M_PI` and like names.  I think the naming in of some of the `M_PI` #defines are a little clunky, but it is easy enough to get help on the math.h header.  The `M_` prefix does make it have something akin to a namespace for the constants.

I was taking the existing constants and logically extending those with choices that have already been made.

I was thinking that it might actually be better to have "const double" variables.  That is a bit more of a radical change, and I'm not really sure if that is a good idea or not.

PR #373, opened 2016-04-01
First steps toward simplified macros/internals

busstoptaktik:
The brief version::

In an attempt to make proj.4 code slightly more secure and much easier to read and maintain, I'm trying to eliminate a few unfortunate design decisions from the early days of proj.4

The work will be _very_ intrusive, especially in the PJ_xxx segment of the code tree, but great care has been taken to design a process that can be implemented stepwise and localized, one projection at a time, then finalized with a relatively small and concentrated work package.

@hobu: The (very) long version: See the comments in PJ_minimal.c


busstoptaktik:
(@kbevers, @hobu:  This is the text from pj_minimal.c, I was referring to in the pull request header)

**A verbose justification for some highly intrusive code surgery**

**The brief version:**

In an attempt to make proj.4 code slightly more secure and much easier to read and maintain, I'm trying to eliminate a few unfortunate design decisions from the early days of proj.4

The work will be _very_ intrusive, especially in the _PJ_xxx_ segment of the code tree, but great care has been taken to design a process that can be implemented stepwise and localized, one projection at a time, then finalized with a relatively small and concentrated work package.

**The (very) long version:**

Gerald I. Evenden's original design for the proj.4 projection system is a beautiful example of software architecture, where a very limited set of policy rules leads to a well defined hierarchical structure and
a high degree of both encapsulation and internal interoperability.

In the proj.4 code, the policy rules are _enforced_ by a system of preprocessor macros for building the scaffolding for implementation of a new projection.

While this system of macros undeniably possesses the property of both reducing repetitive code and enforcing policy, unfortunately it also possesses two much less desirable properties:

First, while enforcing policy, it also _hides_ policy: The _beauty in simplicity_ of Gerald's design is hidden behind layers of macros, whose architectural clarity do not match that of proj.4 in general.

Second (and related), the macros make the source code look like something only vaguely related to C, making it hard to read (an effect that gets amplified to the tune of syntax highlighters getting confused by the macros).

While the policy rule enforcement macros can be eliminated in relatively non-intrusive ways, a more fundamental flaw in the proj.4 use of macros is found in the _PJ_xxx.c_ files implementing the individual projections: The use of internal redefinition of `PJ`, the fundamental proj data object, through the use of the `PROJ_PARMS__` macro, makes the `sizeof (PJ)` fundamentally unknown to the calling `pj_init` function.

This leads to code that is probably not in full conformance with the C standard.

It is also a memory management catastrophe waiting to happen.

But first and foremost, it leads to some very clumsy initialization code where `pj_init` (the constructor function), needs to start the constsruction process by asking the `PJ_xxx` function to do the memory allocation (because `pj_init` does not know the size of the `PROJ_PARMS`-mangled `PJ` object being instantiated).

Then, after doing some initialization work, `pj_init` returns control to `PJ_xxx`, asking it to finalize the initialization with the projection specific parameters specified by the `PROJ_PARMS__` macro.

Behind the scenes, hidden by two layers of macros, what happens is even worse, as a lot of the initialization code is duplicated in every _PJ_xxx_ file, rather than being centralized in the `pj_init` function.

**Solution procedure:**

Evidently, the way to eliminate this clumsyness will be to introduce an opaque object, that is managed by tne individual PJ_xxx projection code, and represented as a simple `void`-pointer in the `PJ` object.

This can be done one projection code file at a time, working through the code base as time permits (it will take at least a month).

When a PJ_xxx file is on the surgical bench, it will also have its `ENTRYA/ENTRY0/ENTRY1/ENTRY2/ENDENTRY/`etc. etc. macro-guts torn out and replaced by the `PROJECTION` macro (introduced in `projects.h`).

This leads to code that looks a lot more like real C, and hence is much less confusing to both syntax higlighters and humans. It also leads to code that, after all projections have been processed, with a final sweep over the code base can be brought into the style of the code in `PJ_minimal.c`

In my humble opinion the result wil be a code base that is not only easier to maintain, but also more welcoming to new contributors.

And if proj is to expand its strong basis in projections into the fields of geodetic transformations and general geometric geodesy, we will need to be able to attract quite a few expert geodesist contributors.

And since expert geodesists are not necessarily expert coders, a welcoming code base is a real asset (to put the icing on the cake of the already welcoming user- and developer community).

Note that the entire process does not touch the algorithmic/mathematical parts of the code at all - it is actuallly an attempt to make this part stand out more clearly.

---

The attached material (pj_minimal.c) is an attempt to show what happens if we remove the layers of macros, and introduce a more centralized approach to memory allocation and initialization.

Please note, however, that the level of centralization achieved here is not yet fully supported by the proj.4 infrastructure: It is an example, intended to show what can be achieved through a smooth,
gradual and safe refactoring of the existing layered macro system.

In my humble opinion, this version makes the beauty of Gerald's design much more evident than the current layered-macro-version.

Thomas Knudsen, thokn@sdfe.dk, 2016-03-31

hobu:
One comment I have immediately on this is you're using `O` for the opaque object variable name. This is very easily confused with `0`. 

The changes look fine to me. Extensive and intrusive, yes, but it looks like improvement to me. I would like you to bring it up with the mailing list, even if just as a courtesy announcement, in case someone with some strong opinions can chime in. 

Is there any way to increase the coveralls percentage and test that these changes are going to cause regressions?

src/projects.h
```diff
@@ -226,12 +226,18 @@ typedef struct ARG_list {
 	/* base projection data structure */
 
 
+#ifdef PJ_LIB__
+    /* we need this forward declaration in order to be able to add a
+       pointer to struct opaque to the typedef struct PJconsts below */
+    struct opaque;
```
rouault: Shouldn't struct opaque be prefixed with pj_ or PJ to avoid potential "namespace" clashes with code including proj headers ?

rouault: Forget my above comment. The #ifdef PJ_LIB__ likely only exposes that to proj internals and not user code.

busstoptaktik: @rouault: Actually you are right Even, and I will correct this at the next commit to the branch: As I'm trying to eliminate the polymorphic aspect of the PJ-object, the PJ_LIB__ guard you refer to will, as the branch matures, be eliminated, and PJ will be uniform across the entire library.

Thanks for noting this!

busstoptaktik: @hobu: OK - I'll change the variable name from O to Q: I wanted something related to the word opaque, while also being in the same ballpark of the alphabet as the P used for the PJ object in most of the PJ_xxx.c files. But going from just-before-P to just-after-P, and from Opaque to opaQue should not make much of a difference (except for the intended of avoiding a potential for misreading).

I'll post a note about my intentions to the mailing list tomorrow.

Also, I'm right now working on adding rudimentary unit tests to all PJ_xxx.c files: I need this to make sure I do not make stupid blunders of the kind repaired in commits 83cdf16 and afebe65.

I suppose this will up the test coverage percentage significantly.

micahcochran:
> Is there any way to increase the coveralls percentage and test that these changes are going to cause regressions?

@hobu I got some of the [Geospatial Integrity of Geoscience Software GIGS testing](http://www.iogp.org/Geomatics#2521115-gigs) 51xx series working this weekend using pyproj.  The results were pretty promising.  Most all of the tests passed.  These could be used for coverage tests.

I'm going to start a new issue to just to discuss this.

busstoptaktik:
@hobu - I believe this PR is ready for merging. It co-habitates nicely with the existing code, and builds a foundation for further work in a straightforward manner, by updating a handful of projections at a time for a series of comming pull requests from myself and @kbevers.

hobu:
@busstoptaktik is this one ready for merge?

kbevers:
@hobu Technically, yes, you can merge now without any problems, but it might be better to wait a little longer. @busstoptaktik and I are not that far from having applied the changes to all of the projections. In my [fork](https://github.com/kbevers/proj.4/tree/fix-projs-with-l) there's around 40 left. It should be possible to finish them all within the next week or so.

busstoptaktik:
@hobu: I believe this is ready for merging now.

@kbevers and I have now eliminated all instances of the ENTRY-family of macros. Next step in this effort will be to eliminate the ERROR-family.

But before that we have an upcoming project introducing support for multi-step transformations ("projection pipelines").

The multi-step work is rather extensive, but not at all as intrusive as this PR. So before going on with the multi-step, It would be very nice to see this succesfully merged.

Note that during the work on this PR, we have expanded test coverage from around 30% to 72%.

PR #382, opened 2016-05-23
Add BUGS section to manpages & update old URLs.

sebastic:
The common BUGS section is missing as reported by Dan Jacobson in [Debian Bug #809690](https://bugs.debian.org/809690).

Some manpages also still references the homepage on remotesensing.org.

### DIFF
--- a/man/man1/cs2cs.1
+++ b/man/man1/cs2cs.1
@@ -202,4 +202,4 @@ USGS Bulletin 1532).
 A list of known bugs can found at https://github.com/OSGeo/proj.4/issues
 where new bug reports can be submitted too.
 .SH HOME PAGE
-https://github.com/OSGeo/proj.4/wiki
+http://proj4.org/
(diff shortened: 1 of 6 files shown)


kbevers:
I think you should change the website to proj4.org instead of the wiki pages on github. Otherwise it looks good.

sebastic:
Is proj4.org official now? Last I heard of it, it was still a work-in-progress.

hobu:
In progress, but not yet done enough to be official IMO.

sebastic:
Shall I revert 26eac9b then, or can you cherry-pick a8460af?

hobu:
> Shall I revert 26eac9b then, or can you cherry-pick a8460af?

Oh, I missed that :)

Let's just leave it. It's not any worse than directing them to the wiki at this point anyway.

kbevers:
Yeah, that was exactly my point and hopefully proj4.org will improve over time. Speaking of which, we should probably do a bit more to facilitate that. The next time I have few hours to spare I'll try to pick up the pieces and lay out a plan for how to get this going again. "Improved documentation" would be a nice bullet point in the next release notes :-)

Issue #374, opened 2016-04-04
GIGS testing


micahcochran:
I got a few of the [Geospatial Integrity of Geoscience Software GIGS testing](http://www.iogp.org/Geomatics#2521115-gigs) 51xx series working this weekend using pyproj.  The results were pretty promising.  Most all of the tests passed.  These could be used for coverage tests.  A few failed, but it might be a good opportunity look at what is going wrong.  (It might even be pyproj or Python related.)

These will be a little bit harder to get working in only Proj.4.  Some Python code should be able to be leveraged in some manner.

The 51xx series of is basically tables of points in lat/long (in many various definitions with WGS 84 only being used once) and projected northings/eastings (or southings/westings).  The results are  compare the results that within an error range of at most 50 mm or 0.002 seconds. The first test is a single conversion.  The other does 1000 roundtrips.

Projections that are tested by 51xx series tests:  aea, cass, laea, lcc, merc, omerc, somerc, sterea, tmerc, and utm.

micahcochran:
GIGS testing is more extensive testing than is in the current Proj.4 test suite for these few projections that are tested.

I converted the Excel spreadsheets into csv.   Then, I used python code to convert the extend the testing for GIGS. 

The only test fails is test 5102 part 2.  This is a Lambert Conic Conformal (1SP) projection.  This is converting between EPSG:4807 and EPSG:27572. Both projections set the prime meridian to paris.  I haven't done much investigation to try to understand why it fails.

Everything else looks like rounding errors.  I did try to change to floor rounding, but that didn't quite work either.

The test suite in PR #373 will work better for this type of testing because it uses a numeric tolerance instead of using rounding.  There is also a 1,000 round trip test that is prescribed by GIGS 51xx, which could be added in the future.  My initial pyproj tests showed fairly good results.

I've added [gigs branch](https://github.com/micahcochran/proj.4/tree/gigs) with the files [nad/testgigs](https://github.com/micahcochran/proj.4/blob/gigs/nad/testgigs) and [nad/tg_out.dist](https://github.com/micahcochran/proj.4/blob/gigs/nad/tg_out.dist)

If you run it, I suggest using a tool like [Meld](http://meldmerge.org/) or [icdiff](https://github.com/jeffkaufman/icdiff) that provides a side-by-side comparison with coloration to highlight the difference between the files.

I want to get this done so that there might be an option for me to clean this up for merging before PR #373, if desired.  Or it might be better to commit this after using the new test suite.

[...]

PR #386, opened 2016-05-24
Add GIGS tests.

micahcochran:
This is to implement #374 

This is a Python framework that uses JSON formatted test files.  It needs pyproj to be installed.

It does 2 different kinds of tests.  A simple conversion test and 1,000 roundtrips.

**Notes:**
~~\* Python 3 seems to pass more of the tests, so I assume that it is more precises, where Python 2.7 seems to have some slight rounding errors with the 1,000 roundtrip tests.  I assume that this is some issue with Python or pyproj.  So, Python 3.3+ is recommended for testing. :smiley:    This was a very recent discovery.~~
- Some of the tests fail and they are not included in Travis.  Still they are in the commit for testing.  The [TESTNOTES.md](https://github.com/micahcochran/proj.4/blob/py-gigs-test4/test/gigs/TESTNOTES.md) notes how badly they were off.  :arrow_backward: that needs to be reformatted.
- There could still be issues with the testing methods itself.  GIGS uses Degrees Minutes Seconds for lat/long, where I've used Decimal Degrees for lat/long.
- Using cs2cs was attempted, but there were some precision errors and it is a few orders of magnitude slower.  (It seems like a few thousand Python Popen calls per test is expensive.)  18 minute runtime for cs2cs testing, versus the 10 second runtime for pyproj.  The cs2cs code is still in the code. 

I'd like this to be reviewed at least a little bit before merging, so **this is not yet ready for merging**.

I have the question:  Should this be a test suite/framework that is outside of Proj.4 project?  BUT I need to ask the International Association of Oil & Gas Producers how their material can be reproduced appropriately.

**EDIT**: Python 2.x & 3.x now work the same for testing purposes.  This was due to the zip function being different in Python 3.x.  I've worked around this.


QuLogic:
Can you set up the JSON so that they are all formatted the same way? Sometimes the coordinates are on multiple lines, and sometimes all on one (very long) line.

test/gigs/TESTNOTES.md
```diff
@@ -0,0 +1,227 @@
...
+ * pyproj (optional but highly recommended), this speeds up tests, makes
+    results more precise but has the trade-off of making installation a
+    little more compilated
+
+
+
+### Installing pyproj ##################################################
+
+ * Install `pip` if not installed, should already be installed on new versions 
+      (Python 2 >=2.7.9 or Python 3 >=3.4) or if using a virtual environment for python.
```
QuLogic: I thought you recommended 3.3?

test/gigs/TESTNOTES.md
```diff
@@ -0,0 +1,227 @@
...
+```
+         $ pip install cython
+```
+    - install latest release
+        - set PROJ_DIR environment variable to an installed version of PROJ4
+          library.  This should have the directories include/ lib/ & 
+          share/proj/ underneath it.
+    - Linux  (default ./configure)
+```
+         $ sudo PROJ_DIR=/usr/local pip install pyproj
```
QuLogic: `sudo` should not be encouraged, especially if using a virtual environment.

micahcochran:
Thanks @QuLogic . I have reformatted the JSON files, but I will admit they aren't perfect.  I've removed `sudo` from the TESTNOTES example.  I fixed a bug that make Python 2.x & 3.x work the same.

QuLogic:
I suggest running `pep8` on the Python files to check for consistent formatting.

micahcochran:
Thanks @QuLogic. I updated the formatting based on `flake8` and fixed most of the problems.  I only left some of the greater than 79 character lines.

I just sent an email to IOGP about this PR.

micahcochran:
I have given up on getting a response from IOGP.  I am considering creating a separate project, so that this project won't have to deal with the licensing issues of for using test data that is IOGP property.  Also, this test framework could be useful for other coordinate transformation libraries.

Any suggestions?

PR #387, opened 2016-05-31
Change math constants, similar to PR #372.  Use M_ namespace with the…

micahcochran:
… de facto standard M_PI and its ilk.  

Change names that are widely used in the project to be in the M_ namespace, so HALFPI becomes M_HALFPI. HALFPI is #defined as M_PI_2 (the de facto standard name).  
# defines _USE_MATH_DEFINES for MS Visual Studio (I didn't personally test this part, but Appveyor will not build otherwise).

The response to PR #372 was no strong opinions with perhaps indifference.

Note: this would open the door to allowing the use of M_PI_2 and the other de facto standard math constants defined by the compiler.

### DIFF
--- a/src/projects.h
+++ b/src/projects.h
@@ -101,7 +101,7 @@ extern double hypot(double, double);
 #define M_FORTPI         M_PI_4                   /* pi/4 */
 #define M_HALFPI         M_PI_2                   /* pi/2 */
 /* M_PI                                               pi */
-#define M_PI_HALFPI      4.71238898038468985766   /* 1.5*pi */
+#define M_PI_HALFPI      4.71238898038468985769   /* 1.5*pi */
 #define M_TWOPI      	 6.28318530717958647693   /* 2*pi */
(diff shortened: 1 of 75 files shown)


src/projects.h
```diff
@@ -92,11 +92,21 @@ extern double hypot(double, double);
...
+/* enable predefined math constants M_* for MS Visual Studio workaround */
+#ifdef _MSC_VER
+#define _USE_MATH_DEFINES
+#endif
+
+/* some more useful math constants and aliases */
+#define M_FORTPI         M_PI_4                   /* pi/4 */
+#define M_HALFPI         M_PI_2                   /* pi/2 */
+/* M_PI                                               pi */
+#define M_PI_HALFPI      4.71238898038468985766   /* 1.5*pi */
```
cffk: 1.5*pi should be 4.71238898038468985769 (change in last digit)

micahcochran: Good catch @cffk! I pushed another commit to reflect that.

micahcochran:
Thanks @kbevers

PR #388, opened 2016-06-01
Transformation pipelines

busstoptaktik:
With the completion of the moving of projection specific parameters to opaque objects, the size of the PJ object is now constant and deterministic.

This means that pj_init, which returns a pointer to the PJ object may, syntactically speaking, just as well return a pointer to an array of PJ objects.

This means that we may implement transformations as pipelines of steps, just by introducing one new projection, which handles the handover between a series of operations, each implemented as individual projections.

That is the approach taken in this pull request, which introduces a new 3D transformation program “tran”, the “pipeline” projection handler, the “step” keyword, and reinroduces the “inv” qualifier.

It also introduces a number of transformation primitives (each implemented as a PJ-style projection):

+proj=cart: For geographic / 3D cartesian transformation
+proj=helmert: For 7 parameter Helmert shifts
+proj=add: For 3 parameter shifts

(more will follow)

And implements a 3D transformation program "tran". Details, User’s Guide and binary downloads are available over at http://thomasknudsen.net/proj4.html

The pull request also introduces a more semantically meaningful approach to type-punning than the “use preprocessor redefinitions to make both LPZ and XYZ look like an UVW”  approach used currently: The new approach introduces the COORDINATE union (which should perhaps be renamed TRIPLET or PJ_TRIPLET), which allows us to have all the flexibility of the UVW-approach, while still keeping the semantic meaning which is so much more evident when the data type actually reflects the data contained.

This is work in progress, only 10 days into its existence. The pull request is certainly not ready for merging at this stage, but there is a number of things I would like to discuss, and find a PR the most straightforward way of starting that discussion.

Discussion points:

1.
Is this at all interesting? In my book, it introduces some very useful functionality in a almost entirely non-intrusive way. But different people, different bananas...

2.
ABI breakage: It has become clear that a number of additions to the PJ data type would be beneficial. Would it be OK to introduce these in 4.10.0? Currently I have implemented workarounds, but added elements would simplify the code, and open up new possibilities.

3.
If this is interesting, what about the syntax? Any advice for improvements? One thing could be to make +step= a synonym for +proj, so the syntax could be simplified from “+step +proj=foo...”  to  “+step=foo...”
4.
Your concern here...


rouault:
This would be material worth posting on the mailing list !

A few remarks and questions :
- Just for my education: do you have examples where using this approach of not using the WGS84 pivot result in increased accuracy for example ? or perhaps it enables doing some transforms that were not possible to do or hard to do with proj before ?
- Despite reading the header of PJ_cart.c, I'm not completely sure to have understood how different it is from geocent. My understanding is that it converts between geodetic and geocentric coordinates as +proj=geocent but as proj standard way of operating is essentially 2D you had to reimplement that ? Regarding the name "cart", it is less intuitive than geocent, but we probably cannot do better
- "+proj=pipeline" seemded a bit awkward at first. Wondering if "+pipeline" wouldn't be better (but I think that would require a particular case in pj_init() ?) ? No strong opinion.
- ABI breakage: it has already been broken by the addition of pj_fwd3d / pj_inv3d in 757a2c8f946faccf9d094d76cb79e6ebe0006564 (not yet in a release)
- the public API of proj is in proj_api.h so I guess some of PJ_pipeline.h should be moved there. Which would require not using PJ as it is defined in projects.h which is labelled as something between public/private use.
- wondering if there's an interest of keeping the projection formalism for pipelines. I guess that keeping projection formalism could enable uses of "cs2cs +proj=pipeline..... +to +proj=pipeline ...." but are there such uses ? If not then a dedicated C API would probably make more sense.

busstoptaktik:
Even @rouault , thank you for some very useful remarks - I will try to answer and/or comment.

First things first... regarding pivot datum or not: when going from cartographic level accuracy to geodetic accuracy, it is actually rather the primary case not to have a pivot datum - the national realizations of ETRS89 are based on differing realizations to differing epochs, and you run into rather hairy operations in order to transform between those, even though they are all official ETRS89 realizations.

Also (at least my experience from Danish, Faroese and Greenlandic systems), transformations between older systems can be in the form of polynomial corrections taking the older plane coordinate system to ED50/utm32 - which can then afterwards be transformed further to ETRS89 given access to the relevant Helmert parameters for the area in question (ED50 was rather jerky and transformation parameters are only sufficiently accurate for fairly small, typically national scale or less, areas).

The use of the designation “WGS84” for the pivot datum (through the +towgs... keyword) is also somewhat unfortunate - because which of the WGS84 realizations, e.g. WGS84(G730, G873, G1150, G1674), is the pivot? It indicates that for cartographic purposes we may take ITRFxx=WGS84(Gxxxx)=ETRFxx=IGSxx - and at some time in history and for some applications that is certainly true. But 30 years after the first realization of WGS84, your average continent has moved by around 1 m, indicating that the identity between the realizations mentioned is now only true if you are happy with meter level accuracy.

And for most civil uses (but probably not including road navigation), that is certainly not good enough. Hence, for greater accuracy, you need to know what you do. And while cs2cs works wonders in guessing the right incantation, I believe that as the accuracy requirements grow, we will need to provide a more tunable interface (not in order to take over the cs2cs and proj corners of the cartographic ecosystem, but rather to expand the range of validity and usefulness for libproj).

Regarding PJ_cart.c and the question about geocentric etc: pj_geocent is a no_op - it just returns what was given to it and the actual transformation is carried out inside the stomach of pj_transform.c, using the routines in geocent.c, which can also be used from PJ_cart.c.

I decided to implement my own routines as well, because I could find no solid references to the algorithm used in geocent.c, so I used the Bowring stuff, mentioned in the Hofmann-Wellenhof & Moritz reference in the code. It is well known and fairly simple to implement, so to have a “second opinion” from an algorithm I trusted, I implemented that one.

But remember that this is work in progress, and I see no deep reason to keep my code in the codebase if I can find evidence that the geocent.c stuff is more trustworthy. A quick scan of Wikipedia does, however, indicate that both Bowring’s algorithm, and the one in geocent.c has been superseded by newer, faster, and more accurate algorithms - so perhaps it is worth reconsidering the contents of geocent.c

As regards your comment that “geocentric” is a better name than “cartesian”, I tend to disagree: Cartesian indicates directly that the coordinates refer to a rectangular system, while geocentric just says something about the origin of the system: It could equally well be a system using geocentric latitude, longitude and height.

Full semantic coverage is first reached by calling it “geocentric, cartesian”, which I’m sure we’ll both agree is way too long.

I chose “cart” because in geodetic software, the indicators cart and crt are common - but to me “crt” still means cathode ray tube, and cart came with the added bonus that you can actually pronounce it as a word - so cart it was :-)

About +proj=pipeline: It seems awkward to you at first, and it certainly also did to me: My first intention was to introduce a new keyword “+tran” for transformation (all projections are transformations, but not all transformations are projections), but since it would really just be a synonym for +proj anyway, I decided to defer it, thinking that the ugliness would provoke some good thoughts for improved syntax.

That being said, the current implementation is technically very attractive, being simple and non-intrusive: Everything is done inside the PJ_pipeline.c file. I would prefer having at least a comparable degree of simplicity for any newly proposed syntax.

About the public/private limit: I think we can greatly improve the public facing interface
simply by forward declaring struct PJconsts, and moving the typedef to proj_api.h:

struct PJconsts;
typedef struct PJconsts PJ;

That way, for user code, PJ will be entirely opaque, not posing any risk of accidentally breaking anything - user code will only have to ferry around pointers to PJ-objects, so it does not need to know the internals, which will be declared in projects.h.

Obviously, for compatibility, we will have to keep the projPJ typedef, although I personally find the name rather ugly, and believe we should try to stay in the PJ/pj namespace, except for the XYZ, LPZ, etc. data types, which are perfectly generic geodetic datatypes, which should not be considered private to libproj, but as a general service to the entire community, delivered through a lightweight proj_api.h

But while this is related, it is also somewhat orthogonal to the pipeline stuff, where I agree that parts of PJ_pipeline.h should be moved to proj_api.h - and the rest to projects.h: I really only introduced PJ_pipeline.h to keep the experimental stuff somewhat localized.

Your last comment about C-API versus “+proj=pipeline”: The code is prepared for cs2cs-style use. But my assumption is that the ordinary user does not know about the details of pipelines, but rather just uses an init-file provided by his national geodetic authority, such that e.g. the Danish transformation from the old System34 to ETRS89/utm32 can be invoked as “tran +init=DK:s34j_utm32etrs”, or some such.

In my mental model, the C-API to pipelines is simply pj_init, and the pipeline syntax, and the pipeline primitives, make that API configurable. Does that sound reasonable to you?

Thank you again for some very interesting and constructive comments

/thomas

busstoptaktik:
Got some build system blues here. For development, I use a cmake mingw build, which works fine. The same is the case for the Travis cmake based build. But the autotools-build bombs with the message

config.status: error: cannot find input file: `src/Makefile.in

(see https://travis-ci.org/OSGeo/proj.4/builds/135358444).

Anyone here who speaks autotools fluently, and can help with debugging this?

src/Makefile.am
```diff
@@ -73,6 +75,8 @@ libproj_la_SOURCES = \
 	geocent.c geocent.h pj_utils.c pj_gridinfo.c pj_gridlist.c \
 	jniproj.c pj_mutex.c pj_initcache.c pj_apply_vgridshift.c geodesic.c \
 	pj_strtod.c
+	\
```
QuLogic, line 78: This line continuation doesn't make sense here. All it does is join the next line with this one, so you've just placed a couple file names here all by themselves.

The backslash needs to be on the previous line, or you need to just put the file names there directly.

I don't know why Travis doesn't just give up earlier, though.

QuLogic:
Autotools errors are noted [here](https://travis-ci.org/OSGeo/proj.4/builds/135358444#L202-L207), but I've commented on one of them.

src/Makefile.am
QuLogic, line 78: > I don't know why Travis doesn't just give up earlier, though.

And the answer to that would be, [because it's commented out for some reason](https://github.com/OSGeo/proj.4/blame/master/autogen.sh#L35).

src/tran.c
```diff
@@ -0,0 +1,353 @@
...
+            }
+
+            /* Option arg inline */
+            opt_arg[(int) c] = argv[i] + j + 1;
+            break;
+        }
+    }
+
+    /* Process all '+'-style options */
+    for (i = i; i < argc; i++) {
```
kbevers: i = i? Maybe i=1 or i=0 here?

kbevers:
@busstoptaktik 

You need to add tran in this line, otherwise it wont build unless you explicitly run 'make tran:
https://github.com/OSGeo/proj.4/pull/388/files#diff-480477e89f9b6ddafb30c4383dcdd705R3

Also se my other comment (that should be referenced here now). With those to changes it compiles just fine with autotools for me.

src/tran.c
busstoptaktik: No - actually correct albeit somewhat funky. It just goes on where '-' style option processing left. Have added comment clarifying this

busstoptaktik:
@kbevers @QuLogic: Thanks, guys! The Travis builds are now going smoothly. The nmake builds under Windows are also OK, while the cmake based ones hiccups - probably because I need to add a few more symbols to proj.def

busstoptaktik:
Norm Olsen describes a related effort for the CS-Map library over at https://trac.osgeo.org/csmap/wiki/CsMapRfc2 - worth reading!

rhuitl:
Hi @busstoptaktik, that is very interesting work you're doing! I've been looking into ways to define a coordinate system which is Gauß-Krüger Zone 4 plus an affine 3D transform (to define an arbitrarily oriented local tangent space) and found this branch. I compiled proj, recompiled stuff that depends on it (e.g., PostGIS) and played with the "tran" tool and the Helmert transform.

While I could finally get +proj=helmert to apply my affine transform (which is defined by a 2x2 rotation matrix and a translation vector, so it was necessary to compute the angle rz, convert the angle to arcseconds, as well as deriving the proper scale), I would really love to see a generic 3D affine transform that allows specifying the six unique elements of a 3x3 rotation matrix R00, R01, R02, R11, R12, R22 and the 3-vector for translation. That would allow you to easily concatenate multiple transformations in local Cartesian spaces until a final transform like tmerc describes how the datum is transformed to WGS-84.

So having "tran" work, I was curious if this is going to work in PostGIS. I defined a custom SRID with a pipeline to see if I can add an extra Helmert transformation on top of GK-Z4. So I define a spatial reference system and try to use it with ST_Transform:

``` sql
INSERT INTO spatial_ref_sys VALUES(1000, 'test', 1000, NULL, '+proj=tmerc +lat_0=0 +lon_0=12 +k=1 +x_0=4500000 +y_0=0 +ellps=bessel +datum=potsdam +units=m +no_defs');
SELECT ST_AsText( ST_Transform( ST_GeomFromText('POINT(4470814.627454 5640426.090921)', 1000), 4326) );
>>> "POINT(11.5836498596428 50.898421786718)"
```

(not exactly what I expected: 11.585083007813, 50.899658203125, but close enough for now...)

As soon as I turn that into a pipeline (even with only a single transform), the result is very different:

``` sql
INSERT INTO spatial_ref_sys VALUES(1000, 'test', 1000, NULL, '+proj=pipeline +step +proj=tmerc +lat_0=0 +lon_0=12 +k=1 +x_0=4500000 +y_0=0 +ellps=bessel +datum=potsdam +units=m +no_defs');
SELECT ST_AsText( ST_Transform( ST_GeomFromText('POINT(4470814.627454 5640426.090921)', 1000), 4326) );
>>> "POINT(-13.4108026989403 0.00376315356350652)"
```

Do you have any hints for me if what I want to achieve is at all possible with your modifications and how to get this to work?

kbevers:
@rhuitl Thanks for your comments. And thanks for testing the transformation pipelines in the real world. We haven't even reached that point yet.
You should be able to do what you want when this pull request have been merged, but for now, please note that this is still very early work and reliable transformations can't be expected.

Please continue to test the code when new commits appear. Bug reports and suggestions on how to improve the interface etc. are much appreciated! Your idea about using the rotation matrix directly in the pipeline is definitely worth considering.

busstoptaktik:
@rhuitl: Thanks for stress testing this attempt at proj.4 enhancement: I really appreciate that, as this PR is essential for implementing transformations for historical Danish, Faroese, and Greenlandic systems.

Unfortunately I am busy with other tasks for the moment, and will not be back on this project for the next few weeks. I will do my best to follow up some time before mid-August.

/Thomas

rhuitl:
@busstoptaktik That's no problem, I'll keep watching this PR and can redo my tests when you think it makes sense :)

hobu:
Do you guys have some cycles to polish this up and push it in to 4.9.3, or do you want to wait?

busstoptaktik:
It has to wait: It is actually quite intrusive, and needs extensive testing
(and discussion) before pushing.

I do not want to delay 4.9.3 by attempting to get this merged.

I will, however, continue the work in a few weeks time.

/Thomas

Den 17/08/2016 16.11 skrev "Howard Butler" notifications@github.com:

busstoptaktik:
Closing this PR - refactoring and resubmitting as a series of smaller PRs

busstoptaktik:
@rhuitl - This PR has now been (mostly) succeeded by PR #453 "Plumbing for Pilelines". I would be grateful if you would redo the testing you did back in July, with the new and _much_ improved implementation

haugwitz:
What is the status on implementing S34s/j in Proj4? Since TrLib is deprecated it would be nice to move to Proj.4. 
If there is a still a need for help with debugging the conversions/transformations I would like to assist if I can.

busstoptaktik:
@haugwitz - yes, we have a working implementation of s34 (actually of s45, the s34 for Bornholm, but the code is identical to s34j/s, only the constants differ.

I'm fully booked for the next two weeks, but please contact me directly at thokn AT sdfe DOT dk for more info about testing.

PR #392, opened 2016-06-10
SELFTEST: Make proj -VC selftest return exit code of number of test failures.

micahcochran:
This change would make Travis-CI and Appveyor mark tests as failures if any of the `proj -VC` selftests fail.  On Travis and Appveyor, `proj -VC` completes running it will abort testing.

Currently, proj.c only returns a zero (0) exit code.  (Some really bad problem might cause a non-zero exit code.)

**EDIT:** Late breaking idea.  This should probably be documented in the man page, too.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -45,6 +45,11 @@ build_script:
 
 test_script:
   - echo test_script
+  - if "%BUILD_TYPE%" == "nmake" cd C:\PROJ\bin
+  - if "%BUILD_TYPE%" == "cmake" cd C:\projects\proj-4\bin
+  - echo "Contents of current directory:"
+  - dir
+  - proj.exe -VC
 
 deploy: off
 
--- a/src/proj.c
+++ b/src/proj.c
@@ -268,8 +268,7 @@ int main(int argc, char **argv) {
                 bin_in = bin_out = 1;
                 continue;
               case 'C': /* Check - run internal regression tests */
-                pj_run_selftests (very_verby);
-                return 0;
+                return pj_run_selftests (very_verby);
                 continue;
               case 'v': /* monitor dump of initialization */
                 mon = 1;


kbevers:
Yeah, I've been meaning to do this for a while. I believe the reason Thomas ( @busstoptaktik ) forced it to return zero was that windows might do windowsy things when getting a non-zero return code. I haven't gotten round to testing if that actually happens, but I a have little more faith in the geniuses from Redmond than Thomas have, so I suspect it will do just fine. I'll do that first thing monday morning when I'm in front of a windows computer again.

micahcochran:
I decided to make some test fail to see what would happen.  I took wink1 and wink2 and replaced all the test cases to make them fail.  [Changes made in order to make this happen.](https://github.com/OSGeo/proj.4/compare/master...micahcochran:selftestexitcode-failure)  My [failing branch](https://github.com/micahcochran/proj.4/tree/selftestexitcode-failure).

CI-Testing:
[Travis](https://travis-ci.org/micahcochran/proj.4/builds/136770188) - test fails as expected
[Appveyor](https://ci.appveyor.com/project/micahcochran/proj-4/build/1.0.1) - unexpectedly passing

Is it intentional that Appveyor does not run the `proj -C` selftests?

kbevers:
Appveyor doesn't do any testing at the moment. Probably because until the self-test were implemented tests were unix only. Try adding proj -C and so we can see how appveyor responds to that.

micahcochran:
Under Windows 64-bit, gs50 projection is failing.  Everything else is passing.  

I don't have the set up to debug on Windows.  Appveyor does have deploy options, which could be used to make a binary.

kbevers:
The appveyor setup seems fine. Thanks for doing this, Micah. I'll have a look at gs50 tomorrow.

kbevers:
I have tested the non-zero return code on windows 7 and it works fine for me. I think we should pull this to master, as it has already proved itself useful. Anyone disagree?

**Regarding gs50 failing on 64 bit windows** 
I believe it is numerical instability caused by the test coordinates being hugely out of range for the given projection. See the example below. We saw a similar thing with alsk (which shares code with gs50) on unix vs windows and adjusted the tolerance accordingly. I think it's time to find some proper test-coordinates.

```
Spherical forward calculation:

  Expected.x 14161556831.844320000000, Expected.y -11410194658.020117000000
  Received.x 14161556831.844309000000, Received.y -11410194658.020121000000
```

micahcochran:
Thanks Kristian!

If we are modifying PJ_mod_ster.c,  "Stererographics" is misspelled in the heading (and it has been for a while), it should probably be "Stereographic".

One approach is to just [allow the 64-bit Windows tests to fail in Appveyor](https://www.appveyor.com/docs/build-configuration#allow-failing-jobs).  In this case, I don't like that approach.

Here's my exploration of trying to find some "in range" test coordinates using iPython (jupyter) [in a notebook](https://gist.github.com/micahcochran/761082f4d40d8e7b1bea1ab5d3b06433).  The functions in it transform the points to the given projection.  For the projection, I used the simplest PROJ.4 projection string, so for example `+proj=gs50`.  I chose points of (mainly) cities in these projections.    Is this on the right track?

kbevers:
I don't like that either! I've been working on my own in-range set of coordinates. I'll make a new pull request later. I might include yours as well (I went with the slightly more complicated +proj=xxx +a=...).

I'll make sure to fix the name while I'm at it. Nice catch btw.

micahcochran:
I just found in USGS Open-File Report 90-284 (PDF page 80), that these projections are intended to be used with Clark 1866 ellipsoid or a sphere radius of 6,370,997m.  Once I added `+ellps=clrk66`, the roundtrip difference improved quite a bit.  I updated the notebook based on that.

kbevers:
Those values area actually set directly in the code. Must be the coordinate scaling that's playing games. Happens in other projections as well. It's not really useful in case a projection is hardwired to a certain ellipsoid or radius. Will have to think a bit about this...

I'll adjust test-coordinates accordingly.

kbevers:
gs48, gs50 and alsk with better test data: https://github.com/OSGeo/proj.4/pull/395

I'll merge this one and let discussion continue in the comments of my new PR.

micahcochran:
Thanks Kristian!

PR #399, opened 2016-07-13
Eliminate load time non-computable initializers in healpix

kbevers:
This takes care of https://github.com/OSGeo/proj.4/issues/376

All warnings from `gcc -ansi -W -Wall -Wextra -pedantic -I. -c PJ_healpix.c`  are now silenced.
I did it in two steps. First I eliminated warnings related to undefined M_PI\* constants when compiling with gcc's  `-ansi` flag (those constants where first introduced with C99). I did this by adding a conditional section in  `projects.h` that defines the M_PI\* constants if they are not already defined by the compiler.

The next step was to initialize a few arrays in many statements instead of just one.
That is,

```
double a[2];
a[0] = x; a[1] = y;
```

instead of 

```
double a[] = {x, y};
```

Most notable in the function `in_image` which has a rather large two-dimensional array.

While I was at it, I changed a lot of occurrences of `M_PI/2.0` and `M_PI/4.0` to their constant equivalents `M_HALFPI` and `M_FORTPI` and refactored some duplicate code.
I also cleaned up `projects.h` since it had mixed use of tabs and spaces. I removed all tabs in order to have consistency in white-space from top to bottom.

### DIFF
--- a/src/PJ_healpix.c
+++ b/src/PJ_healpix.c
@@ -49,9 +49,9 @@ PROJ_HEAD(rhealpix, "rHEALPix") "\n\tSph., Ellps.\n\tnorth_square= south_square=
 # define EPS 1e-15
 
 struct pj_opaque {
-    int north_square; \
-    int south_square; \
-    double qp; \
+    int north_square;
+    int south_square;
+    double qp;
     double *apa;
 };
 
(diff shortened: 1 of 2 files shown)


kbevers:
@busstoptaktik Are you happy with this solution?

PR #400, opened 2016-07-27
Implementation of Times projection

kbevers:
Michał Siwicki requested on the [mailing list](http://lists.maptools.org/pipermail/proj/2016-July/007440.html) that the Times projection was added to proj.4. This PR takes care of that. The implementation is based on Snyder's "Flattening the Earth".

Here's how it looks:
out


src/makefile.vc
```diff
@@ -4,146 +4,146 @@
...
+CFLAGS  = /nologo -I. -DPROJ_LIB=\"$(PROJ_LIB_DIR)\" \
+          -DHAVE_STRERROR=1 -DHAVE_LOCALECONV=1 $(OPTFLAGS)
 
-default:	all
+default:    all
 
 all: proj.lib $(PROJ_EXE) $(CS2CS_EXE) $(GEOD_EXE) $(NAD2BIN_EXE)
 
-proj.lib:	$(LIBOBJ)
-	if exist proj.lib del proj.lib
```
QuLogic, line 83: I don't think you really wanted to remove all these tabs.

kbevers, line 83: Oh, it was definitely intentional - tabs has no place in modern computer code. I'm slowly getting rid of them whenever I come across a file with tabs. There is quite a lot of files that mixes tabs and spaces in indentations which makes the code hard to read for most people. 

But thanks for pointing it out. Actually you made me think about the consequences, so I've made a recommitted the file without removing tabs since a few of the other open PR's will be harder to merge with the tabs->spaces conversion.

hobu, line 83: > tabs has no place in modern computer code

Please be careful about changing the "shape" of lots of existing code for whitespace purity. I don't care for tabs so much either, but if the code is consistently one or the other, IMO it should be left alone. If it isn't consistent, fix it 😄

kbevers, line 83: I agree, which is also why I did not do anything about Makefile.am. It has consistent use of tabs for indentation. Most files in proj are a mixture, and it always looks horrible. I try as best as possible to keep the intended indentation levels but that can be difficult as well; often there's several different styles in the same file.

cffk, line 83: With Makefiles, it's not so much consistency that's the issue but the
**requirement** that the commands for a rule be indented with a TAB
character.

Also if tabbed code is looking horrible, it's possible that your editor
needs to be adjusted for a different width tab.  On Unix machines (and
earlier), a TAB was 8 spaces by default.  MS Visual Studio seems to use
a TAB of 4 spaces.

hobu:
@kbevers how are you making this picture, and is it something we could automate for the generation of figures for the website materials?

src/makefile.vc
kbevers, line 83: > With Makefiles, it's not so much consistency that's the issue but the
> requirement that the commands for a rule be indented with a TAB
> character.

Ah, I was not aware of that! I've never made one of those from scratch, so never had a reason dig into all the details.

I am well aware that you can have different tab widths. That's why tabs are mostly bad in my opinion. Code looks different on different setups, which is fine if tabs are used consistently. That's rarely the case for the code in proj and leads to people indenting code with spaces to whatever tab width is in their particular editor. In my editor I have tabs showing up as "^I" to make them super obvious (I like to know what is going on, and have been fooled by stray tabs before...). This also makes mixed indentations look just a bit more horrible and scream at you to be fixed :)

kbevers:
@hobu It's a python script I've been working on. My motivation was precisely what you are suggesting: Automating figure generation for documentation.

I've made a copy here: https://gist.github.com/kbevers/9fa5c4708ab37c465313f7bc17ced252

It's still work in progress. For now it only works for some of the simpler projections like the Times projection showed above. Healpix for instance does not produce a pretty picture at all. I'll probably follow a different path later and use some of the in-build stuff in matplotlib. I believe it is possible to link it with proj and the use the cartographic features. Should be a better way!

julien2512:
@kbevers  I am not sure if 
http://spatialreference.org/ref/sr-org/7670/
is equivalent with yours ? It seems to use a rimes projection already.

julien2512:
"times" projection of course. My smartphone can't stand it !

kbevers:
@julien2512 It definitely seems to be referring to the same projection. It's not been implemented in proj before though. Also, if you follow the link to the proj-definition you get an empty page which I guess is spatialreference.org's way of saying that it can't be done...

julien2512:
You are probably right ! I guess they will change it after the  master commit (or I will tell them).

rouault:
Note: spatialreference.org is no longer maintained and hasn't been updated in years. It reflects the states of GDAL, proj.4 from several years ago.

julien2512:
@rouault So to make me to maintain it. Many users still using it ! I will check it up later. Tks.

PR #403, opened 2016-08-20
The definition of EPSG:28992 is wrong. For some reason there has been…

julien2512:
reported by @bbrala on proj4php

See 
https://oegeo.wordpress.com/2008/05/20/note-to-self-the-one-and-only-rd-projection-string/

I am going to tell https://epsg.io/28992 too.

### DIFF
--- a/nad/epsg
+++ b/nad/epsg
@@ -9076,7 +9076,7 @@
 # Amersfoort / RD Old
 <28991> +proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 +x_0=0 +y_0=0 +ellps=bessel +towgs84=565.4171,50.3319,465.5524,-0.398957,0.343988,-1.87740,4.0725 +units=m +no_defs  <>
 # Amersfoort / RD New
-<28992> +proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel +towgs84=565.4171,50.3319,465.5524,-0.398957,0.343988,-1.87740,4.0725 +units=m +no_defs  <>
+<28992> +proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.999908 +x_0=155000 +y_0=463000 +ellps=bessel +units=m +towgs84=565.2369,50.0087,465.658,-0.406857330322398,0.350732676542563,-1.8703473836068,4.0812 +no_defs  <>
 # SAD69 / Brazil Polyconic (deprecated)
 <29100> +proj=poly +lat_0=0 +lon_0=-54 +x_0=5000000 +y_0=10000000 +ellps=GRS67 +towgs84=-57,1,-41,0,0,0,0 +units=m +no_defs  <>
 # SAD69 / Brazil Polyconic


micahcochran:
Reoccurrence of issue #96

It looks like the definition was correct before https://github.com/OSGeo/proj.4/commit/832329a9aadbfcb7f46919c903f7672e719d2a82 , which updated the epsg definitions was updated in January. 

Also, it looks like EPSG:28991 has the same PROJ string.  It has also was changed by that update.

julien2512:
I sent a request to epsg.org.

Regards,
Julien2512

QuLogic:
I don't understand the commit message: "For some reason there has been..." what?

micahcochran:
>  I don't understand the commit message: "For some reason there has been..." what?

It looks like a copy&paste error for proj4php commit that corrects this issue https://github.com/proj4php/proj4php/pull/50.

julien2512:
That's right : the sentence was completed in  the  issue's description.

Le 21 août 2016 3:09 AM, "Micah Cochran" notifications@github.com a
écrit :

bbrala:
Yeah sorry, i used the default PR title, which is always the first X chars for a title. Thanks for reporting it here also @julien2512

rouault:
Patching only the epsg file isn't appropriate for a lasting solution. This would be lost again at the next upgrade. A proper fix is too patch https://trac.osgeo.org/geotiff/browser/trunk/libgeotiff/csv/datum_shift_pref.csv to add a mapping from the Amersfoort datum (EPSG:4289) to the appropriate datum shift entry in https://trac.osgeo.org/geotiff/browser/trunk/libgeotiff/csv/datum_shift.csv, which seems to be 15934 with small differences due to the precision into which values are stored.

558,15934,4289,4326,"Parameter values from Amersfoort to ETRS89 (3) (tfm code 15739) assuming that ETRS89 is equivalent to WGS 84 within the accuracy of the transformation. Replaces Amersfoort to WGS 84 (2) (code 1672). Replaced by Amersfoort to WGS 84 (4) (tfm code 4833).","Approximation at the +/- 1m level.",1275,50.75,53.7,3.2,7.22,1,0,9607,565.2369,50.0087,465.658,0.406857,-0.350733,1.87035,4.0812,0

Hopefully that shouldn't make a real difference for most applications. Otherwise a more involved patching would be needed

bbrala:
I've contacted someone I know who at a dutch osgeo group,  they will have look at a proper patch for the dutch projection hopefully.

fsteggink:
This issue was discussed at the [osgeo.nl mailing list](https://lists.osgeo.org/pipermail/dutch/2016-August/001465.html).

I have some question as for the origin of the parameters. Martijn's blog post is from 2008. The EPSG parameters for RD have been updated since. Also the scale factor has a rounding error. The correct value is 0.9999079. See http://www.epsg-registry.org/, and look up code 28992. Unfortunately no permalink. When copying, all formatting is lost.

Today I've downloaded a fresh copy from [RDNAPTRANS](https://www.kadaster.nl/web/formulier/Rijksdriehoeksmeting-formulieren/Aanvraag-download-RDNAPTRANS2008.htm), to have a look at the "towgs84" parameters. The link is in Dutch only, and you have to fill in contact details.

Here are the transformation parameters from RD to **ETRS89**:

```
Transformation parameters from RD(Bessel) to ETRS89
Pivot point: center of the ellipsoid
alfa    1,9342  *10^-6 rad
beta    -1,6677     *10^-6 rad
gamma   9,1019  *10^-6 rad
delta   4,0725  *10^-6
tx  565,4171    m
ty  50,3319     m
tz  465,5524    m
```

Note that alfa/beta/gamma are in millionths of radians, not arc seconds.
Also note that the transformation parameters are to ETRS89, and not WGS84, since the Netherlands is part of the EU. I'm pretty sure they are identical, especially not because continental drift is still going on. WGS84 is not mentioned anywhere in the downloaded ZIP file at all.

fsteggink:
@rouault: Why do you suggest datum shift 15934 should be used? Yes, it is the last one in datum_shift.csv, but they don't appear to be in chronological order, rather sorted by coord_op_code. It is also mentioned in the comments that it was superseded by version4. The previous one (apparently mistakenly named version 5), seems to be the most recent. It also states that it replaces version 3 (identified by 15934). No remarks that it has been superseded, although it might have been superseded in the EPSG database by now. I didn't check that.

rouault:
I'm confused. I suggested datum shift 15934 since it is the entry that has the closest values to the ones suggested by @julien2512 in his commit https://github.com/OSGeo/proj.4/pull/403/commits/2a3f99ce209bacecbc9ef706535a5fc6cb04f79f , ie +towgs84=565.2369,50.0087,465.658,-0.406857330322398,0.350732676542563,-1.8703473836068,4.0812 . But @fsteggink seems to suggest that the parameters that came with the v8.9 update are correct vs the  RDNAPTRANS database ? Who is right :-) ?

julien2512:
Thank you all for all these  answers !

I am OK about to dismiss the pull request because it is not the good way to
change projection defs.

But back to the aim, as Even Rouault tells US, how could we know who's
right ?

The map tells the projection is wrong. May we keep two versions of the
projection def ?
I did not have an answer from EPSG I asked the support on by know.

Regards,
Julien2512

julien2512:
I close the pull request. I will make it back on the proper way later !

fsteggink:
Update: we've been discussing on on the Dutch OSGeo mailing list. An expert from the Cadastre, Lennard Huisman, has provided some answers: http://lists.osgeo.org/pipermail/dutch/2016-August/001471.html. Unfortunately, since it was right before the FOSS4G, I didn't think of updating the pull request sooner, sorry for that.

His remarks (translated) plus my comments:
1. 15934 was superseded, and has been replaced by 4833. This is not related to the fact that RD is linked to ETRS89, and not to WGS84.
Since this transformation assumes an approximation of about 1 meter, we can assume that ETRS89 and WGS84 are the same. This is also quoted by Lennard Huisman. See also point 3.
Although RDNAPTRANS2008 is the official "standard", using the parameters of 4833 should give accurate coordinates within 0,25 meter. This is called the "approximate transformation". I assume this is the case when using the parameters mentioned in 4833 for a transformation to ETRS89, and not to WGS84.
2. There is also a datum shift grid file for RD. This file is entered into the EPSG database with code 7000. When using this transformation, instead of 4833, you'll get the "improved approximate transformation" at ground level, except close to the validity contour of RD (which is outside the Netherlands). With height difference related to the ground level, there is a deviance of 1 mm per 50 m height difference.
3. With regard to the relation between WGS84, ETRS89, and RD, he proposes to give an additional explanation during a meeting, as well as plans to redefine the transformation. (I recall a discussion on LinkedIn about this a few years ago.) The difference between WGS84 and ETRS89 is between 60 and 70 cm, and is increasing yearly. It is not yet 5 meter (which I assumed).

I realize that this answer might raise more questions. I still do have some questions. Since this is the case for more people, I'm organizing a meeting about this topic soon. This will be held some time in October. More details will be posted to the Dutch mailing list soon. Maybe there will be a new pull request, now with the parameters everyone, including the experts, are agreeing on.

For now my conclusion is that the parameters of transformation 4833 can be used for transformation from the Amersfoort datum to both ETRS89, as well as WGS84. You'll get coordinates within 1 meter accuracy, which is more than the difference between ETRS89 and WGS84 (60 / 70 cm).
When using the grid shift file in transformation 7000, you'll get coordinates which are accurate within 0,25 meter, as compared to RDNAPTRANS. However, this is true for ETRS89 only. Because of the current difference of WGS84, they are accurate within 0,25 + 0,70 = 0,95 meter, so there is no benefit using this way when going to WGS84. I don't know how to get more accurate coordinates in WGS84, but this is not only the case in the Netherlands, but in whole Europe.

Regards,

Frank

bbrala:
Thanks for updating here. It would be great to see this lingering problem fixed after all these years. Kudos to you guys on the dutch osgeo list for talking this through and working to a solution. :+1

julien2512:
I've got the answer from EPSG too (below).

Roger Lott says that +k=0.9999079 and +k=0.999908 are the same value, but one is only badly rounded. So that it is not to be changed : we need to keep +k=0.9999079 value.

He also says that there is no need to change +wgs84 values because they are not a part of the projection definition, and subject to change by time. They are here for the convenience of some uses, but not all.

@bbrala, I am not sure about what this mean, but may be you will ever need to change the +wgs84 values for your own case.

I am trying to make Roger Lott to help us on this issue.

> Julien,
> 
> You wrote to us saying
> 
> Ø  Hello, An user of proj4php reports me that EPSG::28992 might be wrong. Its very clear on this blog : https://oegeo.wordpress.com/2008/05/20/note-to-self-the-one-and-only-rd-projection-string/ Are you agree with him ? Regards, Julien2512
> 
> EPSG::28992 (Amersfoort / RD New for the Netherlands) as defined in the EPSG Dataset is correct. The proj strings quoted in the blog are correct (or almost so – the value of the scale factor given in one of the blog strings is incorrectly rounded as 0.999908 when the true value, as given in the EPSG Dataset and the other blog string, is 0.9999079).
> 
> To overlay any coordinate data, including but not limited to RD New, on Web Mercator first requires the [RD New] data to be transformed to WGS 84, the Virtual Earth default coordinate reference system (CRS). This is no different to having two data sets in for example UTM zone 31 and UTM zone 32 and merging these together without conversion to a common coordinate reference system: without the conversion the imported data will plot in the wrong place.
> 
> The towgs84 elements mentioned in the blog is doing the transformation to WGS 84. What is not correct is to assume that it the EPSG::28992 definition is wrong because it does not have the towgs84 string. It is a computing convenience to use WGS 84, but it is not essential. If you were merging Dutch data with German data, one could be converted to the other directly – WGS 84 is not necessary. It is actually technically incorrect to require the transformation to WGS 84 to be part of a CRS definition. Indeed, in Europe, using WGS 84 is not strictly legal: ETRS89 should be used, not WGS 84.
> 
> So the error is not in EPSG::28992 but in the bloggers understanding of CRS definitions and merging of different CRSs.
> 
> Roger Lott
> 
> For IOGP Geodesy Subcommittee

PR #405, opened 2016-08-22
Changed self-test build behaviour from opt-out to opt-in

kbevers:
This PR changes the self-test build process according to @rouault's suggestion on the [mailing list](http://lists.maptools.org/pipermail/proj/2016-May/007385.html). With these changes you now have to specifically ask for the self-test to be compiled, whereas it was compiled by default before. The idea behind the opt-in for the self-test is to keep the binary size as small as possible in "production" builds.

Since proj can be build in a few different ways here's some examples:

```
# cmake
cmake proj.4/ -DSELFTEST=ON && make

# autotools
CFLAGS="-DPJ_SELFTEST" ./configure && make

# nmake
nmake /f makefile.vc SELFTEST=1
```

I'm in no way an expert in any of the build systems so suggestions for better solutions are welcome.

The build flags should be documented properly at some point. For now they are documented in .travis.yml and appveyor.yml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,9 +91,9 @@ install:
   - cd ..
 # autoconf build with grids and coverage
   - if [ $TRAVIS_OS_NAME == "osx" ]; then
-      CFLAGS="--coverage" ./configure;
+      CFLAGS="-DPJ_SELFTEST --coverage" ./configure;
     else
-      CFLAGS="--coverage" LDFLAGS="-lgcov" ./configure;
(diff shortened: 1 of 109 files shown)


hobu:
@kbevers I think something is up with the passing of the variable to appveyor https://ci.appveyor.com/project/OSGeo/proj-4/build/1.0.304/job/1w89fadfs3ff3nxq#L14

kbevers:
@hobu Yes, I caught that as well. It's fixed now and all tests pass.

PR #411, opened 2016-09-02
Docs projection - eqc

julien2512:
A first pull request with formula for eqc from proj4 sources (PJ_eqc.c).

Note that proj4 do not use lambda0 as in https://en.wikipedia.org/wiki/Equirectangular_projection.


docs/source/projections/eqc.rst
```diff
@@ -8,3 +8,31 @@ Equidistant Cylindrical (Plate Caree)
    :scale: 50%
    :alt:   Equidistant Cylindrical (Plate Caree)  
 
+Marinus of Tyre was attributed this equirectangular projection :
+
+.. math::
+
+   x = (\lambda - \lambda0) cos \phi0
```
QuLogic, line 15: I think all these 0s should probably be subscripts.

QuLogic:
There appear to be several commits from your previous PR that should not be included here.

hobu:
> There appear to be several commits from your previous PR that should not be included here.

Yes, that will need to be cleaned up before this can be merged. Please be sure to use Gerald's original PostScript files as the source information for these pages. Do not take from other sources, because it is not clear that proj.4's codebase actually matches that.

https://github.com/OSGeo/proj.4/tree/master/docs/old

docs/source/projections/eqc.rst
julien2512: Sorry, I need to remove \lambda0, as proj4 do not implement it.

I can change \phi0 to \phi1 if it sounds better, or something ?

julien2512:
You are right. I need to pratice a bit more. Then I will pull request it again.

PR #412, opened 2016-09-07
EQC projection doc

julien2512:
Here is my second test, I think it will be accepted.
The references I used for are :
https://github.com/OSGeo/proj.4/blob/master/docs/old/libproj.pdf
ftp://ftp.ams.org/pub/tex/doc/amsmath/amsldoc.pdf
make html ;)

Tell me if it's ok, then I will make more.

I have a first question : in proj4 source code, there an optionnal \phi_0 that do not appear in the old documentation formula, but only in the parameters.
May I fix it in the new docs ?

It leads me to a second question about the png generated in that page.
There is "+eqc", but it does not show the parameters [+lat 0= | +lat ts=] as in the old documentation.
How can I improve it ?

### DIFF
--- a/docs/source/projections/eqc.rst
+++ b/docs/source/projections/eqc.rst
@@ -8,3 +8,23 @@ Equidistant Cylindrical (Plate Caree)
    :scale: 50%
    :alt:   Equidistant Cylindrical (Plate Caree)  
 
+The simplist of all projections. Standard parallels (0° when omitted) may be specified that determine latitude of true scale (k=h=1).
+
+.. math::
+
+   x = \lambda cos \phi_{ts}
+
+.. math::
+
+   y = \phi
+
+Reverse operation :
+
+.. math::
+
+   \lambda = x / cos \phi_{ts}
+
+.. math::
+
+   \phi = y
+
--- a/docs/source/projections/eqc.rst
+++ b/docs/source/projections/eqc.rst
@@ -12,11 +12,11 @@ The simplist of all projections. Standard parallels (0° when omitted) may be sp
 
 .. math::
 
-   x = \lambda cos \phi_{ts}
+   x = \lambda \cos \phi_{ts}
 
 .. math::
 
-   y = \phi
+   y = \phi - \phi_0
 
 Reverse operation :
 
@@ -26,5 +26,5 @@ Reverse operation :
 
 .. math::
 
-   \phi = y
+   \phi = y + \phi_0
 
--- a/docs/source/projections/merc.rst
+++ b/docs/source/projections/merc.rst
@@ -8,3 +8,87 @@ Mercator
    :scale: 50%
    :alt:   Mercator  
 
+Scaling may be specified by either the latitude of true scale (:math:`\phi_{ts}`) or setting :math:`k_0` with :math:`+k=` or :math:`+k_0=`.
+
+The spherical form :
+
+.. math::
+
+   x = k_0 \lambda
+
+.. math::
+
+   y = k_0 * \ln( \tan(\pi/4 + \phi/2))
+
+.. math::
+
+   k_0 = \cos(\phi_{ts})
+
+The spherical form's reverse operation :
+
+.. math::
+
+   \lambda = x/k_0
+
+.. math::
+
+   \phi = \pi/2 - 2 \arctan(e^{-y/k_0})
+
+.. math::
+
+   k_0 = \cos(\phi_{ts})
+
+
+The elliptical form :
+
+.. math::
+
+   x = k_0 \lambda
+
+.. math::
+
+   y = k_0 \ln(t(\phi))
+
+.. math::
+
+   k_0 = m(\phi_{ts})
+
+The elliptical form's reverse operation :
+
+.. math::
+
+  \lambda = x / k_0
+
+.. math::
+
+  \phi = t^{-1} (e^{-y/k_0})
+
+where :math:`t()` is the isometric latitude kernel function :
+
+.. math::
+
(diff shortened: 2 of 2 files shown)


rouault:
Yes I think it makes sense to update the formulas to take into account phi_0.

(note: A bit out of scope of this PR, but would probably be good to have some generic intro page for all projections to present the naming conventions of all those common variables. I had no idea what k and h meant until I found paragraph "3.8.1 Scale factors")

docs/source/projections/eqc.rst
```diff
@@ -8,3 +8,23 @@ Equidistant Cylindrical (Plate Caree)
    :scale: 50%
    :alt:   Equidistant Cylindrical (Plate Caree)  
 
+The simplist of all projections. Standard parallels (0° when omitted) may be specified that determine latitude of true scale (k=h=1).
+
+.. math::
+
+   x = \lambda cos \phi_{ts}
```
QuLogic: I believe `\cos` should be possible here.

hobu: Not specific to @julien2512's contribution (it is greatly appreciated!), but this highlights the tricky part about this endeavor -- it is really easy to miss math symbols when porting things over.

hobu:
@kbevers's plotting script is now merged into `./docs/plot/plot.py` but it will need to be updated to a little bit to make it convenient to replace a single image.

kbevers:
> but it will need to be updated to a little bit to make it convenient to replace a single image.

Yeah, I have a few ideas on how to do that. I will get to it soon(ish). Among other things I want to be able to define an area of interest. For most projections it is fine to just plot the whole world, but in some cases projections are only valid for a certain area. Eventually we will probably also want to have more than one image in the docs for a particular projection. I'll make some sort json-setup file where you can set the proj-string, area, filename, etc.  The plotting script can then parse that and make better figures for the doc-pages.

julien2512:
@rouault @QuLogic  done. Thank you for the feedback ! I will start other projections soon.

@kbevers  Is there a way to get your settings from the rst file in a comment or something rather than an other json-setup file ? It can be more convenient !

kbevers:
I don't know. There might be some magic in sphinx that can do that. For now I would rather keep it simple and just have it in one file. Don't want to over-engineer this from the start :)

julien2512:
Yes you are right, it is better to make it works !
I had a look in sphinx doc, but can't really stand how to do it... I though you were an expert ! ;)

julien2512:
Dear all,

I tryed myself on tmerc projection (see the pull request).

It's harder than eqc and I ve got many asks :
- the formula on https://github.com/OSGeo/proj.4/blob/master/docs/old/libproj.pdf do not exactly match the proj4 formula. I meant PI - 2 arctan became PI/2 - 2 arctan in proj4 sources
- there is two totally different mercator projection (spherical and elliptical) but defined in the same class. It ends with a hard page full of maths.
- the elliptical formula uses math functions like the isometric latitude kernel function which might be use in other projections.
- I can't understand the first sentence : "Scaling may be specified by either the latitude of true scale (φ ts ) or setting k 0 with +k= or +k 0=.". May it be "... or setting k_0" ?

julien2512:
@hobu You seems to have make tmerc already !
So I will try to match your template for the eqc projection ... and start an other pull request again.

hobu:
Please let developers close pull requests so we know if stuff has been merged or not.

> the formula on https://github.com/OSGeo/proj.4/blob/master/docs/old/libproj.pdf do not exactly match the proj4 formula. I meant PI - 2 arctan became PI/2 - 2 arctan in proj4 sources

That's curious and should be investigated further.

kbevers:
I noticed that as well when I wrote the Mercator documentation. The libproj text is wrong. Snyder's book is in agreement with the implementation in proj4. I am sure you can find other sources as well. The docs on proj4.org uses formula from Snyder.

kbevers:
http://mathworld.wolfram.com/MercatorProjection.html uses the same formula as Snyder (more or less...). Just for fun, I also derived the formula on paper and I get the same result. I am now 100% certain that the spherical inverse mercator function is correct. And the docs reflect what's in the code.

julien2512:
Thank you.
I won't close pull requests any more !
Eventhough, I need this one  to be  close in  order  to push eqc matching
the projection doc  template.

On an other side, I understand the new doc  will respect the  proj4
implémentation, but we may loose  information from the  old docs. If so, I
can  put  it on wikipedia as I do with eqc. As an example, I have set extra
link on their page to the  new doc for eqc. Is there any rule between
wikipedia and OSGeo ? (I see a page on osgeo.fr named "wikipedia links").

Le 13 sept. 2016 4:57 PM, "Kristian Evers" notifications@github.com a
écrit :

http://mathworld.wolfram.com/MercatorProjection.html uses the same formula
as Snyder (more or less...). Just for fun, I also derived the formula on
paper and I get the same result. I am now 100% certain that the spherical
inverse mercator function is correct. And the docs reflect what's in the
code.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
https://github.com/OSGeo/proj.4/pull/412#issuecomment-246709803, or mute
the thread

kbevers:
> Eventhough, I need this one  to be  close in  order  to push eqc matching
> the projection doc  template.

Just create a new branch, make your changes and put up another PR on the new branch.

julien2512:
@kbevers  Thank you, i will do it the day after tomorrow (edit ;) ).

julien2512:
@hobu I do not close it on purpose this time /) It was automatic after the branch deletion on my repo.

The new pull request is done :) !

PR #415, opened 2016-09-15
eqc documentation matching template (version 3)

julien2512:
Here is the third version, matching the template proposal.

As I got a better understanding of proj4 implementation, I think there might be a mistake in eqc implementation.

In the description of wikipedia, nor wolfram, there is only one parameter \phi_ts. Evenden do not use any.
But in proj4, we got two : \phi_ts and \phi_0.

That's why I tryed to document it on this pull request.
Although, I can't find how this value is valuated in proj4, and I find no documentation (wikipedia, evenden, wolfram, epsg) talking about a parameter phi_0 different than phi_ts.

May be phi_0 value is simply always 0 in proj4, and just leave in the code as a reminder ?

### DIFF
--- a/docs/source/projections/eqc.rst
+++ b/docs/source/projections/eqc.rst
@@ -1,10 +1,10 @@
 .. _eqc:
 
 ********************************************************************************
-Equidistant Cylindrical (Plate Caree)
+Equidistant Cylindrical (Plate Carrée)
 ********************************************************************************
 
-The simplist of all projections. Standard parallels (0° when omitted) may be specified that determine latitude of true scale (k=h=1).
+The simplest of all projections. Standard parallels (0° when omitted) may be specified that determine latitude of true scale (k=h=1).
 
 +---------------------+----------------------------------------------------------+
 | **Classification**  | Conformal cylindrical                                    |
@@ -13,18 +13,18 @@ The simplist of all projections. Standard parallels (0° when omitted) may be sp
 +---------------------+----------------------------------------------------------+
 | **Defined area**    | Global, but best used near the equator                   |
 +---------------------+----------------------------------------------------------+
-| **Implemented by**  | Marinus of Tyre                                          |
+| **Implemented by**  | Gerald I. Evenden                                        |
 +---------------------+----------------------------------------------------------+
 | **Options**                                                                    |
 +---------------------+----------------------------------------------------------+
 | `+lat_ts`           | Latitude of true scale. Defaults to 0.0                  |
 +---------------------+----------------------------------------------------------+
-| `+lat_0`            | Center of the map                                        |
+| `+lat_0`            | Center of the map : latitude of origin                   |
 +---------------------+----------------------------------------------------------+
 
 .. image:: ./images/eqc.png
    :scale: 50%
-   :alt:   Equidistant Cylindrical (Plate Caree)  
+   :alt:   Equidistant Cylindrical (Plate Carrée)  
 
 Usage
 ########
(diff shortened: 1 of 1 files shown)


docs/source/projections/eqc.rst
```diff
@@ -4,7 +4,95 @@
 Equidistant Cylindrical (Plate Caree)
 ********************************************************************************
 
+The simplist of all projections. Standard parallels (0° when omitted) may be specified that determine latitude of true scale (k=h=1).
```
QuLogic: simplest

docs/source/projections/eqc.rst
```diff
@@ -4,7 +4,95 @@
...
+| **Options**                                                                    |
++---------------------+----------------------------------------------------------+
+| `+lat_ts`           | Latitude of true scale. Defaults to 0.0                  |
++---------------------+----------------------------------------------------------+
+| `+lat_0`            | Center of the map                                        |
++---------------------+----------------------------------------------------------+
+
 .. image:: ./images/eqc.png
    :scale: 50%
    :alt:   Equidistant Cylindrical (Plate Caree)
```
QuLogic: Carree, or even more correct, Carrée, especially because that's used later.

docs/source/projections/eqc.rst
```diff
@@ -4,7 +4,95 @@
...
+########
+
+Because of the distortions introduced by this projection, it has little use in navigation or cadastral mapping and finds its main use in thematic mapping. In particular, the plate carrée has become a standard for global raster datasets, such as Celestia and NASA World Wind, because of the particularly simple relationship between the position of an image pixel on the map and its corresponding geographic location on Earth.
+
+The following table gives special cases of the cylindrical equidistant projection. 
+
++---------------------------------------------------------+--------------------------+
+| Projection Name                                         | (lat ts=) :math:`\phi_0` |
++---------------------------------------------------------+--------------------------+
+| Plain/Plane Chart                                       | 0◦                       |
```
QuLogic: This is the "White Bullet" symbol, not the degree sign, cf. ◦ vs. °.

docs/source/projections/eqc.rst
```diff
@@ -4,7 +4,95 @@
...
++---------------------------------------------------------+--------------------------+
+| Simple Cylindrical                                      | 0◦                       |
++---------------------------------------------------------+--------------------------+
+| Plate Carrée                                            | 0◦                       |
++---------------------------------------------------------+--------------------------+
+| Ronald Miller—minimum overall scale distortion          | 37°30'                   |
```
QuLogic: The "apostrophe" should be a "prime", cf. ' vs. ′ (though I can't really see a difference with my default font here, so it's rather minor, and probably not a big deal.)

docs/source/projections/eqc.rst
```diff
@@ -4,7 +4,95 @@
 Equidistant Cylindrical (Plate Caree)
```
QuLogic: Carree or Carrée

QuLogic (changes requested):
Mostly minor tweaks to stuff.

docs/source/projections/eqc.rst
```diff
@@ -4,7 +4,95 @@
...
+Example using latitude of true scale::
+
+    $ echo 2 1 | proj +proj=eqc +a=6400000 +lat_1=0.5 +lat_2=2
+    223402.14       111701.07
+
+Mathematical definition
+#######################
+
+The formulas describing the Equidistant Cylindrical projection are all taken from proj4 sources.
+
```
kbevers, line 74: This section would be more clear if some of the variables are defined here. Like phi_ts is the latitude of true scale and can be set with +lat_ts.

kbevers (changes requested):
Good work @julien2512. Thanks for doing this.

I've pointed out a few things I'd like to see adressed before I merge this. On top of this I think that @QuLogic has some valid points as well. 

Regarding the phi_0 variable. It is the latitude of origin and can be set with +lat_0.

docs/source/projections/eqc.rst
```diff
@@ -4,7 +4,95 @@
...
+The simplist of all projections. Standard parallels (0° when omitted) may be specified that determine latitude of true scale (k=h=1).
+
++---------------------+----------------------------------------------------------+
+| **Classification**  | Conformal cylindrical                                    |
++---------------------+----------------------------------------------------------+
+| **Available forms** | Forward and inverse                                      |
++---------------------+----------------------------------------------------------+
+| **Defined area**    | Global, but best used near the equator                   |
++---------------------+----------------------------------------------------------+
+| **Implemented by**  | Marinus of Tyre                                          |
```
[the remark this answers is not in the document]
julien2512: done.

docs/source/projections/eqc.rst
```diff
@@ -4,7 +4,95 @@
...
+| Ronald Miller Equirectagular                            | 50◦30'                   |
++---------------------------------------------------------+--------------------------+
+| E.Gradarend and A. Niermann minimum linear distortion   | 61°7'                    |
++---------------------------------------------------------+--------------------------+
+
+
+Example using latitude of true scale::
+
+    $ echo 2 1 | proj +proj=eqc +a=6400000 +lat_1=0.5 +lat_2=2
+    223402.14       111701.07
```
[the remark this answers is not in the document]
julien2512: Alright.

docs/source/projections/eqc.rst
julien2512: Really ? I hope parsers will recognise both.
done.

docs/source/projections/eqc.rst
julien2512: It was from the old doc !
done.

docs/source/projections/eqc.rst
julien2512: Even better !
done.

docs/source/projections/eqc.rst
julien2512: There is a point here. The plate carrée is just one case of equidistant cylindrical, isn't it ? So why using it on the title ...

done.

docs/source/projections/eqc.rst
julien2512: done

julien2512:
@kbevers 

> Regarding the phi_0 variable. It is the latitude of origin and can be set with +lat_0.

Can you help me to know where the lat_0 is taking a value in proj4 ? It can make me ask many less questions !

kbevers:
> Can you help me to know where the lat_0 is taking a value in proj4 ? It can make me ask many less questions !

Sure. It's done when calling pj_init. So when you initialize a projection.

julien2512:
Thank you, I found :
`PIN->phi0 = pj_param(ctx, start, "rlat_0").f;`

julien2512:
@QuLogic thank you for your help.

docs/source/projections/eqc.rst
julien2512: I tryed something with the parameters.
You may have a look on it ?
I feel it is important to evaluate the quality before stepping to an other projection doc.

docs/source/projections/eqc.rst
```diff
@@ -71,8 +71,10 @@ Mathematical definition
 #######################
 
 The formulas describing the Equidistant Cylindrical projection are all taken from proj4 sources.
-:math:`lat_ts` is the latitude of true scale, that mean the standard parallels where the scale of the projection is true.
-:math:`lat_0` is the latitude of origin that match the center of the map.
+
+:math:`phi_{ts}` is the latitude of true scale, that mean the standard parallels where the scale of the projection is true. It can be set with "lat_ts".
```
kbevers, line 75: I think there's a backslash missing here.  Same for the line below.

julien2512:
This is working for me without backslashes. Is there a better way to format
these lines ?

Le 16 sept. 2016 23:27, "Kristian Evers" notifications@github.com a
écrit :

kbevers:
Really? Do you build the docs with sphinx?

julien2512:
Of course ! 
Eventhough, I am open to your proposal.

kbevers:
Great. I'll merge your PR once you change it.

kbevers:
Thanks for writing this, Julien. It is very much appreciated!

julien2512:
Let's start other ones !

julien2512:
It will be Cassini.

Andre-J:
There is a bug in the example: 
Example using Plate Carrée projection with true scale at latitude 30° and central meridian 90°W:
$ echo -88 30 | proj +proj=eqc +lat_ts=30 +lat_0=90w

Central meridian should be lon_0, which is not included in the formulas, but changing values for lon_0 different from zero resolves to different results. Seems that Proj can do more than Snyder wrote in his formulas. Maybe this should be pointed out on the manpage.

kbevers:
Good catch. Indeed it should be `+lon_0`. Fixed in https://github.com/OSGeo/proj.4/commit/89af2753651f8c21eae9df631a71c0a3b03bcc3b

> Seems that Proj can do more than Snyder wrote in his formulas. Maybe this should be pointed out on the manpage.

The `+lon_0` adjust the longitude before it reaches the projection fwd/inv functions, as seen here: https://github.com/OSGeo/proj.4/blob/84c2fed374da46ba0ca3cae1125350b1aa3f73fd/src/pj_fwd.c#L86

I agree that this should be documented better. It's on my list for the next large overhaul of the docs.

PR #418, opened 2016-09-26
Remove unused code in PJ_isea.c:isea_hex().

schwehr:
It is possible that there was a missing if before the first "return 1;".
However, there has never been one since this initial commit by Warmerdam of
Nathan Wagner's code.

https://github.com/OSGeo/proj.4/commit/c5c96ad2237e89871e8b0c3f734f8bef95b23aab#diff-e4763731fad7050d4fb7e0916748f654


schwehr:
Invalid fix.

PR #422, opened 2016-09-29
Cassini projection docs (first try)

julien2512:
3 remarks :
- the use of the M function and how to document it
- the implementation in proj4 is very far from Evenden proposal for the Elliptical version
- Wikipedia's page do not tell about the elliptical version at all, and there is no French page for the cassini projection ! I shall fix that.

### DIFF
--- a/docs/source/projections/cass.rst
+++ b/docs/source/projections/cass.rst
@@ -1,10 +1,142 @@
 .. _cass:
 
 ********************************************************************************
-Cassini
+Cassini (Cassini-Soldner)
 ********************************************************************************
+
+Although the Cassini projection has been largely replaced by the Transverse Mercator, it is still in limited use outside the United States and was one of the major topographic mapping projections until the early 20th century.
+
++---------------------+-------------------------------------------------------------------------+
+| **Classification**  | Transverse and oblique cylindrical                                      |
++---------------------+-------------------------------------------------------------------------+
+| **Available forms** | Forward and inverse, Spherical and Elliptical                           |
++---------------------+-------------------------------------------------------------------------+
+| **Defined area**    | Global, but best used near the central meridian with long, narrow areas |
++---------------------+-------------------------------------------------------------------------+
+| **Implemented by**  | Gerald I. Evenden                                                       |
++---------------------+-------------------------------------------------------------------------+
+| **Options**                                                                                   |
++---------------------+-------------------------------------------------------------------------+
+| `+lat_0`            | Center of the map : latitude of origin                                  |
++---------------------+-------------------------------------------------------------------------+
 
 .. image:: ./images/cass.png
    :scale: 50%
    :alt:   Cassini  
+
+Usage
+########
+
+There has been little usage of the spherical version of the Cassini, but the ellipsoidal Cassini-Soldner version was adopted by the Ordnance Survey for the official survey of Great Britain during the second half of the 19th century (Steers, 1970, p. 229). Many of these maps were prepared at a  scale of 1:2,500. The Cassini-Soldner was also used for the detailed mapping of many German states during the same period. 
+
+
+Example using EPSG 30200 (Trinidad 1903, units in clarke's links)::
+
+    $ echo 0.17453293 -1.08210414 | proj +proj=cass +lat_0=10.44166666666667 +lon_0=-61.33333333333334 +x_0=86501.46392051999 +y_0=65379.0134283 +a=6378293.645208759 +b=6356617.987679838 +to_meter=0.201166195164 +no_defs 
+    66644.94	82536.22
+
+Example using EPSG 3068 (Soldner Berlin)::
+
+    $ echo 13.5 52.4 | proj +proj=cass +lat_0=52.41864827777778 +lon_0=13.62720366666667 +x_0=40000 +y_0=10000 +ellps=bessel +datum=potsdam +units=m +no_defs
+    31343.05	7932.76
+
+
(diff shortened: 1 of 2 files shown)


cffk:
GeographicLib implements the ellipsoidal version using geodesics.  The
same approach would be possible for proj.4.  (This is how the azimuthal
equidistant projection was generalized to the ellipsoidal case.)

docs/source/projections/cass.rst
```diff
@@ -1,10 +1,142 @@
...
++---------------------+-------------------------------------------------------------------------+
+
 .. image:: ./images/cass.png
    :scale: 50%
    :alt:   Cassini  
 
+Usage
+########
+
+There has been little usage of the spherical version of the Cassini, but the ellipsoidal Cassini-Soldner version was adopted by the Ordnance Survey for the official survey of Great Britain during the second half of the 19th century (Steers, 1970, p. 229). Many of these maps were prepared at a  scale of 1:2,500. The Cassini-Soldner was also used for the detailed mapping of many German states during the same period.
```
kbevers: I think this should be referenced properly. Add it to references.rst as `[Steers1970]` and reference it in the text with `[Steers1970]_`. Not sure how to reference a specific page, but maybe the Sphinx docs can help with that.

docs/source/projections/cass.rst
```diff
@@ -1,10 +1,142 @@
...
++---------------------+-------------------------------------------------------------------------+
+| **Available forms** | Forward and inverse, Spherical and Elliptical                           |
++---------------------+-------------------------------------------------------------------------+
+| **Defined area**    | Global, but best used near the central meridian with long, narrow areas |
++---------------------+-------------------------------------------------------------------------+
+| **Implemented by**  | Gerald I. Evenden                                                       |
++---------------------+-------------------------------------------------------------------------+
+| **Options**                                                                                   |
++---------------------+-------------------------------------------------------------------------+
+| `+lat_0`            | Center of the map : latitude of origin                                  |
```
kbevers: I may be a big pedant, but I think it's more correct to keep it at "Latitude of origin" as the center is made up of both a latitude and a longitude.  Also, please add the default value of zero.

docs/source/projections/cass.rst
```diff
@@ -1,10 +1,142 @@
...
+Forward projection
+------------------
+
+.. math::
+
+  x = \arcsin(\cos(\phi) * \sin(\lambda))
+
+.. math::
+
+  y = \arctan(\tan(\phi), \cos(\lambda)) - \phi_0
```
kbevers: This is supposed to be arctan2, right?

docs/source/projections/cass.rst
```diff
@@ -1,10 +1,142 @@
...
+
+Spherical form
+==============
+
+Forward projection
+------------------
+
+.. math::
+
+  x = \arcsin(\cos(\phi) * \sin(\lambda))
```
kbevers: The \* here looks kinda un-mathematical when rendered. I think it would look better if it is left out, or maybe instead use something like \cdot or \times depending on how the end result turns out.

kbevers (changes requested):
Thanks, Julien. There's a few things I'd like you to add, the rest is just suggestions. I haven't checked if the maths are correct, I trust that you've got it right.

Regarding the m-functions, we need to define those elsewhere since they are used quite frequently. I think there needs to be a background or theory section somewhere. I am working on it, and will finish it soon(ish)

docs/source/projections/cass.rst
julien2512: done

docs/source/projections/cass.rst
julien2512: That's right.

docs/source/projections/cass.rst
```diff
@@ -1,10 +1,142 @@
...
+
+.. math::
+
+  x = N ( A - T \frac{A^3}{6} - (8-T+8C)T\frac{A^5}{120})
+
+.. math::
+
+  y = M(\phi) - M(\phi_0) + N \tan(\phi)(\frac{A^2}{2} + (5-T+6C)\frac{A^4}{24})
+
+and M() is the meridinial distance function.
```
QuLogic: I think this should be meridional.

docs/source/projections/cass.rst
julien2512: Well, I tried something.

docs/source/projections/cass.rst
julien2512: Yes, it's better without.

docs/source/projections/cass.rst
julien2512: even better ! Thank you.

docs/source/projections/cass.rst
kbevers: Yeah, and it sort of worked. If you put the reference in docs/source/references.rst it's possible to reference the same text from other pages as well. I can do it for you as well if you don't want to play around with it anymore.

julien2512: Now it fits your mind !
Shall we include references inside the page ?

kbevers: Thanks! It would be cool to also have the references listed on each page, but I am not sure sphinx can do that. At least not without cluttering other types of output like pdf.

kbevers (approved):

julien2512:
@kbevers I am a little busy right now, but the next will be "transverse Mercator" !

PR #428, opened 2016-10-09
add a copy of Snyder to the proj.4 repository and use it in doc references

hobu:


kbevers:
@hobu  It seems like a no brainer to merge this. Is there a reason you haven't merged this already?

PR #434, opened 2016-10-12
Add BuildRequires and run autogen.sh

volvo850fahre:
I added the autogen.sh at first the build step. Autogen.sh requires libtool (libtoolize) so i added the line: BuildRequires: libtool

### DIFF
--- a/proj.spec
+++ b/proj.spec
@@ -25,10 +25,10 @@ data with a wide range of selectable projection functions.
 
 %prep
 %setup -D -n proj-4.9.3
+%{_builddir}/%{name}-%{version}/autogen.sh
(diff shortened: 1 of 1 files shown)


proj.spec
```diff
@@ -28,6 +28,7 @@ data with a wide range of selectable projection functions.
 %configure
 
 %build
+%{_builddir}/%{name}-%{version}/autogen.sh
```
QuLogic: That should be done before configuration.

volvo850fahre: Okey, 
I will move the line just under the %prep tag, and try it out.

Thank you!

PR #437, opened 2016-10-17
Transverse Mercator first projection doc shot

julien2512:
First shot for transverse Mercator projection doc, for the spherical form only.

Because I don't get if proj4 have a full cover of the ellipsoidal form.
Evenden's doc tells about Gauss Boaga as an other projection, but it is not in proj4. EPSG docs tells Gauss Boaga and Gauss Kruber are the same transverse Mercator with different parameters.
But the proj4 sources seems to match exactly the Gauss Kruber.
Do the table with all the special cases I set in the doc a reality in proj4 ?

### DIFF
--- a/docs/source/projections/tmerc.rst
+++ b/docs/source/projections/tmerc.rst
@@ -66,7 +66,7 @@ Example using Gauss Boaga on Italy area (EPSG:3004) ::
 Mathematical definition
 #######################
 
-The formulas describing the Transverse Mercator are all taken from Evenden's formula.
+The formulas describing the Transverse Mercator are all taken from Evenden's [Evenden2005]_.
 
 :math:`\phi_0` is the latitude of origin that match the center of the map. It can be set with ``+lat_0``.
 
--- a/docs/source/projections/tmerc.rst
+++ b/docs/source/projections/tmerc.rst
@@ -72,6 +72,8 @@ The formulas describing the Transverse Mercator are all taken from proj4 sources
 
 :math:`k_0` is the scale factor at the natural origin (on the central meridian). It can be set with ``+k_0``.
 
+:math:`M(\phi)` is the meridianal distance.
+
 Spherical form
 **************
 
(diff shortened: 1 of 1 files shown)


julien2512:
I guess the Gauss Boaga is a case where (λcosφ)^7 is about zero, that's why Evenden reports it as a distinct projection.
I am going to end the doc with the Gauss Kruber Elliptical projection.

julien2512:
My remark is true for the forward projection, but it seems there is true differences for the inverse projection that do not appear in proj4 sources ?
As an example (1 + η_1^2) do not appear in Gauss Kruber inverse projection, but appears in Gauss Boaga in Evenden maths.

julien2512:
tmerc doc complete for Gauss Kruber only.

docs/source/projections/tmerc.rst
```diff
@@ -1,10 +1,195 @@
...
+
+Example using Gauss Boaga on Italy area (EPSG:3004) ::
+
+    $ echo 15 42 | proj +proj=tmerc +lat_0=0 +lon_0=15 +k=0.9996 +x_0=2520000 +y_0=0 +ellps=intl +units=m +no_defs
+    2520000.00	4649858.60 
+
+Mathematical definition
+#######################
+
+The formulas describing the Transverse Mercator are all taken from proj4 sources.
```
kbevers: I believe the formulas used in tmerc are based on Snyders work. So credit were credit is due...

It might actually be a good idea to have a look at those and see if they describe the maths more clearly. Most times what's in the code has been rewritten because of numerical concerns.

kbevers (commented):
Sorry for not getting to this sooner. See the inline comments.

docs/source/projections/tmerc.rst
julien2512: There is a perfect match with Evenden's formula, despite of the rewritten.

But I will check Snyders too.

julien2512: It definitely better match Evenden's formulas rather than Snyders.

kbevers: Okay. Please mention that then, and add a citation for the relevant Evenden document.

julien2512:
Yes !

The next will be the Gall stereographic projection.

But I need to end wikipedia's review first. It's a bit harder than in proj4 !

julien2512:
I'm stuck on Gall projection. I guess the formula come from Snyder. But I definitely need to buy Flattening the Earth to confirm it, cos I can't find it for free ! So to say I need a little more time for this !

kbevers:
@julien2512 It's lying on my coffee table right now. What do you need?

julien2512:
I have to check that snyder's match proj4 gall projection maths.
Even though I already made the order !

julien2512:
Alright, Gall proj4 math come from Snyder's "Flattening the earth" !
I got to pull request gall doc as soon as possible !

PR #443, opened 2016-10-27
Fixing both scaling and extreme coordinate issues mentioned in #166

kbevers:
This fixes both issues mentioned in #166.

The original issues is dealt with by applying the fix supplied by the submitter.

The problem introduced by using `DEG_TO_RAD` in `dmstor.c` was a tad more difficult to understand. Basically what happened was that because of the an inaccurate value of `PI/180.0` the input for lam_1 and lam_2 where not precisely translated to radians which in turn caused `cos(lam1) == -0.0`when `+lam_1=-90`where used in the proj-string. The negative zero eventually results in `P->lam0 == 0` when using the inaccurate degrees to radians conversions factor and `P->lam0 == PI` when using the proper `DEG_TO_RAD`.

This has been fixed by manually fixing the wrap-around when `+lat_1=-90` is used. I believe this to be the only case where it is necessary to handle wrap-around, but another set of eyes on this would be great to confirm that this is indeed the correct way to fix this.

### DIFF
--- a/src/PJ_ocea.c
+++ b/src/PJ_ocea.c
@@ -18,7 +18,6 @@ static XY s_forward (LP lp, PJ *P) {           /* Spheroidal, forward */
     XY xy = {0.0,0.0};
     struct pj_opaque *Q = P->opaque;
     double t;
-
     xy.y = sin(lp.lam);
     t = cos(lp.lam);
     xy.x = atan((tan(lp.phi) * Q->cosphi + Q->sinphi * xy.y) / t);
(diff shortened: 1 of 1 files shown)


busstoptaktik:
Checks fail due to regression in ocea - probably should update expected values.

Wrt. additional numerical gotchas, I think it would be excellent if @cffk could take a look...

kbevers:
Yes, I forgot to update the test-values... I'll do that when we're sure that this is the correct fix

kbevers:
Doesn't seem to be any objections to this fix. I'll let it sit for a few more days, and merge at the end of the week if no one comments.

PR #444, opened 2016-10-28
Hungarian grid moved to free grids

zsiki:


micahcochran:
This simply moves the section from Non-Free Grids to a Free grid.  The grid appears to be licensed under GPL.  This seems like a good change.

PR #445, opened 2016-10-30
Pipeline plus api - in continuation of the Genereic Coordinates pull request

busstoptaktik:
This PR continues the work on a rationalized (i.e. better namespaced) API, and also, as another step down the road to implementing transformation pipelines, reintrodices the PJ_cart.c cartesian/ellipsoidal converter, previously discussed in a former PR.

Regarding the API improvements, the full story is in the comments in proj.h, which I, for completeness, include below:

```
       Purpose:  Revised, experimental API for PROJ.4, intended as the foundation
       for added geodetic functionality.

       The original proj API (defined in projects.h) has grown organically
       over the years, but it has also grown somewhat messy.

       The same has happened with the newer high level API (defined in
       proj_api.h): To support various historical objectives, proj_api.h
       contains a rather complex combination of conditional defines and
       typedefs. Probably for good (historical) reasons, which are not
       always evident from today's perspective.

       This is an evolving attempt at creating a re-rationalized API
       with primary design goals focused on sanitizing the namespaces.
       Hence, all symbols exposed are being moved to the pj_ namespace,
       while all data types are being moved to the PJ_ namespace.

       Please note that this API is *orthogonal* to  the previous APIs:
       Apart from some inclusion guards, projects.h and proj_api.h are not
       touched - if you do not include proj.h, the projects and proj_api
       APIs should work as they always have.

       A few implementation details:

       Apart from the namespacing efforts, I'm trying to eliminate three
       proj_api elements, which I have found especially confusing.

       FIRST and foremost, I try to avoid typedef'ing away pointer
       semantics. I agree that it can be occasionally useful, but I
       prefer having the pointer nature of function arguments being
       explicitly visible.

       Hence, projCtx has been replaced by PJ_CONTEXT *.

       SECOND, I try to eliminate cases of information hiding implemented
       by redefining data types to void pointers.

       I prefer using a combination of forward declarations and typedefs.
       Hence:
           typedef void *projCtx;
       Has been replaced by:
           struct projCtx_t;
           typedef struct projCtx_t PJ_CONTEXT;
       This makes it possible for the calling program to know that the
       PJ_CONTEXT data type exists, and handle pointers to that data type
       without having any idea about its internals.

       (obviously, in this example, struct projCtx_t should also be moved
       to struct pj_ctx some day...)

       THIRD, I try to eliminate implicit type punning. Hence this API
       introduces the OBSERVATION data type, for generic coordinate and
       ancillary data handling.

       It includes the PJ_SPATIOTEMPORAL and PJ_TRIPLET unions
       making it possible to make explicit the previously used
       "implicit type punning", where a XY is turned into a LP by
       re#defining both as UV, behind the back of the user.

       The bare essentials API presented here follows the PROJ.4
       convention of sailing the coordinate to be reprojected, up on
       the stack ("call by value"), and symmetrically returning the
       result on the stack. Although the OBSERVATION object is 4 times
       as large as the traditional XY and LP objects, timing results
       have shown the overhead to be very reasonable.

       See pj_proj_test.c for an example of how to use the API.

       Author:   Thomas Knudsen, <thokn@sdfe.dk>
```


src/proj.h
```diff
@@ -182,27 +262,93 @@ PJ_OBSERVATION pj_apply (PJ *P, enum pj_direction direction, PJ_OBSERVATION obs)
 /* Measure internal consistency - in forward or inverse direction */
 double pj_roundtrip(PJ *P, enum pj_direction direction, int n, PJ_OBSERVATION obs);
 
-
 int pj_show_triplet (FILE *stream, const char *banner, PJ_TRIPLET point);
 
-/* Constructor for the OBSERVATION object */
+PJ *pj_pj (PJ_CONTEXT *ctx, char *definition);
```
rouault: 1. const char\* would be better
2. pj_pj() isn't really explicit. pj_create_from_def() maybe ?
3. This is a more general question, but how should we document functions and structures. Informally here as comments, or with Doxygen formalism in the .c file (or in .h) ? Update: I see we have https://github.com/OSGeo/proj.4/tree/master/docs/source/development now, so that would be the place for that

src/proj.h
```diff
@@ -182,27 +262,93 @@ PJ_OBSERVATION pj_apply (PJ *P, enum pj_direction direction, PJ_OBSERVATION obs)
...
 #define TORAD(deg)  ((deg)*M_PI/180.0)
 #endif
 
-#endif /* ndef PROJECTS_H */
+
+
+
+/* This is mostly a direct copy of (parts of) the proj_api header, but with cleaned up name space */
+
+#define PJ_VERSION 493
```
rouault: This is a bit error prone to have PJ_VERSION defines in 2 header files (but we probably don't want to have a proj_version.h with just that #define ?). HOWTO-RELEASE will have to be updated.

src/Makefile.am
```diff
@@ -74,7 +74,7 @@ libproj_la_SOURCES = \
 	jniproj.c pj_mutex.c pj_initcache.c pj_apply_vgridshift.c geodesic.c \
 	pj_strtod.c \
 	\
-	pj_observation.c
+	pj_observation.c pj_cart.c
```
rouault: Not sure if makefiles are filename case sensitive. But if so, should be PJ_cart.c.
Or should PJ_cart.c be in lowercase itself? (I tend to prever lower case for filenames but proj is a mix of both...)

src/pj_proj_test.c
```diff
@@ -0,0 +1,39 @@
...
+    /* Null projection */
+    a = pj_apply (p, PJ_IDENT, a);
+    printf ("%15.9f %15.9f\n", TODEG(a.coo.lpz.lam), TODEG(a.coo.lpz.phi));
+
+    /* Invalid projection */
+    a = pj_apply (p, 42, a);
+    printf ("%15.9f %15.9f\n", a.coo.lpz.lam, a.coo.lpz.phi);
+    err = pj_ctx_get_errno (pj_ctx (p));
+    printf ("pj_ctx_errno: %d\n", err);
+    return 0;
```
rouault: Would be a good practice to free p

src/proj_api.h
```diff
@@ -49,20 +49,21 @@ extern "C" {
  *   are padded with leading zeros if necessary); e.g., PJ_VERSION=401000
  *   for version 4.10.0.
  */
+#ifndef PROJ_H
```
rouault: It seems that there's a provision to do #include "proj.h" and then #include "proj_api.h", but the reverse wouldn't work well, right ?

src/proj.h
```diff
@@ -182,27 +262,93 @@ PJ_OBSERVATION pj_apply (PJ *P, enum pj_direction direction, PJ_OBSERVATION obs)
...
+    int      (*fseek)(PJ_FILE *file, long offset, int whence);
+    long     (*ftell)(PJ_FILE *file);
+    void     (*fclose)(PJ_FILE *file);
+};
+
+void        pj_ctx_set_fileapi (PJ_CONTEXT *, PJ_FILEAPI *);
+PJ_FILEAPI *pj_ctx_get_fileapi (PJ_CONTEXT *);
+PJ_FILEAPI *pj_get_default_fileapi();
+
+PJ_FILE  pj_ctx_fopen  (PJ_CONTEXT *ctx, const char *filename, const char *access);
```
rouault: Do we need those pj_ctx_fopen() and the like in the public API ? Looks like they should be used only by the proj.4 implementation

src/proj.h
```diff
@@ -182,27 +262,93 @@ PJ_OBSERVATION pj_apply (PJ *P, enum pj_direction direction, PJ_OBSERVATION obs)
...
+int      pj_ctx_fseek  (PJ_CONTEXT *ctx, PJ_FILE *file, long offset, int whence);
+long     pj_ctx_ftell  (PJ_CONTEXT *ctx, PJ_FILE *file);
+void     pj_ctx_fclose (PJ_CONTEXT *ctx, PJ_FILE *file);
+char    *pj_ctx_fgets  (PJ_CONTEXT *ctx, char *line, int size, PJ_FILE *file);
+
+PJ_FILE  pj_open_lib(PJ_CONTEXT *, const char *, const char *);
+
+
+/* Logging API */
+void pj_log (PJ_CONTEXT *ctx, int level, const char *fmt, ... );
```
rouault: pj_log() probably doesn't need to be public

src/pj_observation.c
```diff
@@ -112,3 +168,25 @@ int pj_show_triplet (FILE *stream, const char *banner, PJ_TRIPLET point) {
...
+    if (0==ctx)
+        return pj_init_plus_ctx (pj_get_default_ctx(), definition);
+    return pj_init_plus_ctx (ctx, definition);
+}
+
+PJ *pj_pj_argv (PJ_CONTEXT *ctx, int argc, char **argv) {
+    return pj_init_ctx (ctx, argc, argv);
+}
+
+PJ_CONTEXT *pj_ctx (PJ *P) {
```
rouault: Why would one need to get the context of a PJ\* object ?

src/pj_observation.c
```diff
@@ -112,3 +168,25 @@ int pj_show_triplet (FILE *stream, const char *banner, PJ_TRIPLET point) {
...
+    if (0==P)
+        return pj_get_default_ctx ();
+    return pj_get_ctx (P);
+}
+
+void pj_ctx_set (PJ *P, PJ_CONTEXT *ctx) {
```
rouault: Why would one need to set the context of a PJ\* object ?

src/proj.h
```diff
@@ -182,27 +262,93 @@ PJ_OBSERVATION pj_apply (PJ *P, enum pj_direction direction, PJ_OBSERVATION obs)
...
+PJ_FILEAPI *pj_get_default_fileapi();
+
+PJ_FILE  pj_ctx_fopen  (PJ_CONTEXT *ctx, const char *filename, const char *access);
+size_t   pj_ctx_fread  (PJ_CONTEXT *ctx, void *buffer, size_t size, size_t nmemb, PJ_FILE file);
+int      pj_ctx_fseek  (PJ_CONTEXT *ctx, PJ_FILE *file, long offset, int whence);
+long     pj_ctx_ftell  (PJ_CONTEXT *ctx, PJ_FILE *file);
+void     pj_ctx_fclose (PJ_CONTEXT *ctx, PJ_FILE *file);
+char    *pj_ctx_fgets  (PJ_CONTEXT *ctx, char *line, int size, PJ_FILE *file);
+
+PJ_FILE  pj_open_lib(PJ_CONTEXT *, const char *, const char *);
```
rouault: I know this isn't a new function that you added, but the naming is quite bad. It doesn't open the lib, but a file.
OK so if we keep that one public (could potentially be useful if people wants to parse the 'epsg' file at hand?), we need to keep the fread(), fseek() etc public as well

src/proj.h
kbevers: I agree with 1 and 2. Even though the most obvious name is occupied it must be possible to come out with something more explicit tha`pj_pj`. `pj_setup` maybe?
3. Yes, I added that recently but I am not happy with it yet. For now a proj_h.rst or something like that would be good enough. I think eventually we should use a system like Doxygen.

src/pj_proj_test.c
```diff
@@ -0,0 +1,39 @@
+/* Tiny test of an evolving new API. Thomas Knudsen, 2016-10-30 */
```
kbevers: Since this file is not really a part of the main library I think it would be better to move it to either the `test` folder or a new `examples`. This way it is more clear what is what.

src/Makefile.am
kbevers: Keeping with the tradition the file should be `PJ_cart.c` since all files with "projections" are named that way, and the rest uses the lower-cases nomenclature. I don't like it either (I'd prefer to organise the files in directories instead).

src/pj_observation.c
kbevers: You can do the same with the API in `proj_api.h` as documentend here: http://proj4.org/development/threads.html
This, and the one below, is just to get consistent naming across all `pj_ctx*`functions. I am not really able to comment on why you would need that functionality though...

kbevers (changes requested):
My initial comments are inlined. I'll probably have more after my second read-through of the code. In general it looks good.
The build fails that obviously has to change before merging...

src/proj.h
```diff
@@ -116,20 +205,17 @@ typedef struct { double lam, phi,  z; }  LPZ;
 typedef struct { double d, m,  s; }  PJ_DMS;
 
 /* Geoid undulation (N) and deflections of the vertical (z, e) */
-typedef struct { double   N,   z,  e; }  PJ_NZE;
+typedef struct { double   z,  e, N; }  PJ_ZEN;
```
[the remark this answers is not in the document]
busstoptaktik: No, not really - and I have changed it again now to EZN, since eta is the east-west component, zeta the north-south, and N is related to heights, so EZN makes most sense

src/proj.h
busstoptaktik: I have changed it now, so it lives in proj_api.h only - but making it work requires some ugly defines.

src/PJ_cart.c
```diff
@@ -0,0 +1,253 @@
...
+    geocent.c, at least thay haven't been through as heavy
+    use as their geocent sisters. Some care has been taken
+    to avoid singularities, but extreme cases (e.g. setting
+    es, the squared eccentricity, to 1), will cause havoc.
+
+**************************************************************/
+
+
+static void freeup (PJ *P) {
+    pj_freeup_plain (P);
```
kbevers, line 102: This function does not appear to be defined anywhere. `pj_dealloc` would usually be used here. Is this meant to be an alias?

src/proj.h
busstoptaktik: If I remember Frank's explanation when they were introduced, they were actually intended as user accessible, for making the proj.4 library play nice with other environments.

Personally I wouldn't mind moving them to a lower level API, but I do not think thay are intended as proj.4 internals (on the other hand, I have never seen them in use...)

src/proj.h
busstoptaktik: I agree with 1 & 2 and changed the char \* arg to const char *, and the name to pj_create which I believe strikes a balance between brevity and expressivity :-)

src/proj.h
busstoptaktik: The name is horrible... but I think this API should live its life inside the library and evolve a bit on the trip to the next release, and for a new API we do not need to keep the old names.

I will probably remove some of these, and reintroduce them when need (and better ideas) arise

src/proj.h
busstoptaktik: It has been up until now, but I agree that keeping things compact from start would be a good idea

src/Makefile.am
busstoptaktik: Yes, that was a bug. Actually, I think that 8 years ago, Gerald had planned to move all projection files into a proj_NAME.c namespace (cf. proj_etmerc.c and proj_rouss.c)

Doing that would also make it way easier to find the right file in a lexicographically ordered file open dialog.

But it would probably make it slightly harder to follow the development history of the code in git.

src/pj_proj_test.c
busstoptaktik: Done - thanks for spotting that!

src/pj_observation.c
busstoptaktik: Probably in order to pass it on to pj_init_ctx in order to initialize another PJ running in the same thread?

src/pj_observation.c
busstoptaktik: Probably reason closely related to the above mentioned.  Basically I have never seen it in use.

This might be another of the "drop them now, and reintroduce if necessary later" crowd

src/pj_observation.c
```diff
@@ -49,22 +50,77 @@ PJ_OBSERVATION pj_observation (
...
+};
+
+const PJ_OBSERVATION pj_observation_null = {
+    {{0, 0, 0, 0}},
+    {{0, 0, 0}},
+    0, 0
+};
+
+PJ_OBSERVATION pj_fwdobs (PJ_OBSERVATION obs, PJ *P) {
+#if 0
```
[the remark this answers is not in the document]
busstoptaktik: Done

src/pj_observation.c
```diff
@@ -49,22 +50,77 @@ PJ_OBSERVATION pj_observation (
...
+    if (0!=P->fwd) {
+        obs.coo.xy  =  pj_fwd (obs.coo.lp, P);
+        return obs;
+    }
+    pj_ctx_set_errno (P->ctx, EINVAL);
+    return pj_observation_error;
+}
+
+PJ_OBSERVATION pj_invobs (PJ_OBSERVATION obs, PJ *P) {
+#if 0
```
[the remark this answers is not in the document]
busstoptaktik: Done

src/pj_proj_test.c
busstoptaktik: Introduced the examples folder now

src/proj.h
kbevers: [This](http://lists.maptools.org/pipermail/proj/2013-June/006611.html) is Franks introduction to the context file API. He mentions that it is used in the init and gridshift file accessors. I am not sure if that is actually true since I am having a hard time finding said code. I am not very familiar with that part of the code base though, so it is likely that I have missed something.

I wouldn't be surprised if no one has ever used that functionality.

src/proj.def
```diff
@@ -92,4 +92,11 @@ EXPORTS
     pj_observation                  @90
     pj_show_triplet                 @91
     pj_apply                        @92
-    
+    pj_fwdobs                       @93
+    pj_invobs                       @94
+    pj_roundtrip                    @95
+    pj_pj                           @96
+    pj_pj_argv                      @97
+    pj_roundtrip                    @98
```
kbevers: defined twice

kbevers:
@busstoptaktik Your last commit, [Minor extensions to new API](https://github.com/OSGeo/proj.4/pull/445/commits/185973eb6a2f1ae6486cce6d402871cc4694357c), adds 1800+ lines of code. That doesn't seem very minor to me :) Was it your intention to add `PJ_helmert.c`, `PJ_pipeline.c` and `horner.c` at this stage? Or were they included by accident?

busstoptaktik:
Definitely a "git add" blunder. Removed now.

src/PJ_cart.c
busstoptaktik: It is defined in pj_malloc.c and has a slightly different prototype than pj_dealloc: It takes a PJ as arg and frees its opaque object. It can be used to simplify/eliminate the freeup code for a large number of projections.

busstoptaktik:
@kbevers - I think I have taken care of all your comments.

@rouault - following your review, and a study by @kbevers revealing the sparsity of use of the context interface, I have now cleaned thing up dramatically, cf. e.g. examples/pj_proj_test.c for some descriptive comments

kbevers (approved):

busstoptaktik:
OK - this PR has been announced on the mailing list (see thread "A re-rationalized API for PROJ.4"), without generating any negative response. It passes all checks, and has "no conflicts with the base branch". I find it ready for merging, and will do so in a moment.

Thanks for reviewing, @kbevers  and @rouault - your comments and change requests took the PR in a different, but much better, direction

PR #452, opened 2016-11-17
Updated CMakeLists to only set the debug postfix if the target is being built

tfili:
If you set the options `BUILD_CS2CS`, etc to `OFF`, cmake processing fails on Windows because the targets don't exist.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -24,23 +24,29 @@ endif ()
 
 if(BUILD_CS2CS)
  include(bin_cs2cs.cmake)
+ set(BIN_TARGETS ${BIN_TARGETS} cs2cs)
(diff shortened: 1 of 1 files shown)


tfili:
Just wondering if anyone had a chance to look at this? Is there anything else that needs to be done?

hobu:
I was going to test this before merging, but it looks ok at the moment.

tfili:
@hobu Ok, thanks. Just wanted to make sure there wasn't anything else I needed to do.

tfili:
@hobu Any update on this? Obviously not critical, but it would be nice to get it in at some point.

rouault:
@hobu I've merged it. The appveyor tests pass, so at least that doesn't break the default setup

PR #453, opened 2016-11-18
Plumbing for pipelines

busstoptaktik:
This is a much improved reimplementation of the ideas behind PR #388 "Transformation Pipelines".

This PR implements the library code only. An upcomming PR will reintroduce the TRAN command line interface, and a handful of elementary transformations typically used in geodetic transformation pipelines.


kbevers (changes requested):
I actually haven't got very many comments to the code. It seems quite well thought-out to me. Good job, Thomas.

The pipeline concept is very powerful, but also rather complicated compared to what has been possible in PROJ.4 up until now. This needs to be documented thoroughly. I suggest a new section on the webpage, "Coordinate transformation" where the concept of pipelines can be introduced. It will also be a nice place to document the geodetic transformations that are coming from you soon. Additionally the new  `proj.h`/`pj_obs_api.c` API should be documented in the "Development" section. This doesn't have to be a part of this PR, but should be written at least before the next release and preferably a lot sooner so first-movers has a chance to try things out.

src/PJ_pipeline.c
```diff
@@ -0,0 +1,547 @@
...
+    current_argv = argv_params (P->params);
+    if (0==current_argv)
+        return pipeline_freeup (P, ENOMEM);
+
+    /* Do some syntactic sanity checking */
+    for (i = 0;  i < argc;  i++) {
+        if (0==strcmp ("step", argv[i])) {
+            if (-1==i_pipeline) {
+                pj_log_error (P, "Pipeline: +step before +proj=pipeline");
+                return pipeline_freeup (P, 51);
```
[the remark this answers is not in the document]
busstoptaktik: This part is definitely a bug - thanks for spotting it. The pj_strerrno values must be negative. They are, however, mostly used by proj, geod, and cs2cs, so I'm not sure it's worth the effort to introduce symbolic constants. I'll consider whether it makes sense to open an enhancement proposal to implement this.

src/PJ_pipeline.c
```diff
@@ -0,0 +1,547 @@
+/***********************************************************************
+
+               Transformation pipeline manager
+
+                 Thomas Knudsen, 2016-05-20/2016-11-14
+
+************************************************************************
+
+**The Problem**
```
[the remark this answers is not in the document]
kbevers: Works for me too. Most of the text can be used as a general introduction in the docs as well.

src/PJ_pipeline.c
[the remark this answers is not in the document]
kbevers: > They are, however, mostly used by proj, geod, and cs2cs, so I'm not sure it's worth the effort to introduce symbolic constants. I'll consider whether it makes sense to open an enhancement proposal to implement this.

Me neither. Usually though, I have no idea what a given error number means and I'll have to look it up. Could just as well be solved by comments in the code. Labelling the error numbers with good names fixes that and makes the code more readable. But that is outside the scope of this PR i think :)

kbevers (approved):

src/PJ_pipeline.c
```diff
@@ -0,0 +1,603 @@
...
+        return 0;
+
+    argc = argc_params (P->params);
+    argv = argv_params (P->params);
+    if (0==argv)
+        return pipeline_freeup (P, ENOMEM);
+
+    /* The elements of current_argv are not used - we just use argv_params */
+    /* as allocator for a "large enough" container needed later            */
+    current_argv = argv_params (P->params);
```
QuLogic, line 365: That's two extra iterations through `P->params` when you already have `argc` and can just `calloc` it, no?

src/PJ_pipeline.c
```diff
@@ -0,0 +1,603 @@
...
+    if (0==current_argv)
+        return pipeline_freeup (P, ENOMEM);
+
+    /* Do some syntactic sanity checking */
+    for (i = 0;  i < argc;  i++) {
```
QuLogic, line 370: The calculation of `argc` or `argv` could have been folded into here.

QuLogic (commented):
Alas, this was merged before I finished reading it, but I think the parsing could perhaps have been simplified a bit.

src/PJ_pipeline.c
```diff
@@ -0,0 +1,603 @@
...
+
+*********************************************************************/
+
+
+
+
+static PJ_OBS pipeline_forward_obs (PJ_OBS point, PJ *P) {
+    int i, first_step, last_step, incr;
+
+    first_step = 1;
```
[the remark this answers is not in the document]
busstoptaktik: It's actually mostly for historical reasons (in the previous version of the pipeline stuff, P[0] was the pipeline driver itself, and a sentinel was needed in the far end of the pipeline, because P was an array of PJs, not of PJ *.

I intend to clean this up at a later date, but I need to go on with some material actually using the pipeline functionality in order to stress test the implementation

src/PJ_pipeline.c
```diff
@@ -0,0 +1,603 @@
...
+    P->fwd    =  pipeline_forward;
+    P->inv    =  pipeline_reverse;
+    P->pfree  =  pipeline_freeup_wrapper;
+
+    P->opaque = pj_calloc (1, sizeof(struct pj_opaque));
+    if (0==P->opaque)
+        return 0;
+
+    argc = argc_params (P->params);
+    argv = argv_params (P->params);
```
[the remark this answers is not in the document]
busstoptaktik: Thanks - implementing this improvement as part of comming PR

src/PJ_pipeline.c
busstoptaktik: Thanks again - have implemented it for next PR

busstoptaktik:
Sorry for fat-fingering a merge (due to a momentarily sluggish mouse response), while you were reviewing. I appreciate your comments and will implement most of them in comming PRs

src/PJ_pipeline.c
QuLogic: I think it should be possible in just two iterations to do everything; I can open a PR for it.

busstoptaktik: Yes, in two iterations. I think I misunderstood: Thought you wanted to fold everything info the syntax check loop  - which can be done, but not very elegantly. And as this part is not at all a speed concern, I preferred to just implement the first two suggestions, eliminating two trips through the stack, and Still keeping the code clean.

QuLogic: No, definitely not just one loop, as that would require assuming some maximum size of each step or some messy `realloc`ing.

Two loops should work though, first loop counts `nstep` and maximum args-per-step, and second loop populates `argv` for parsing each step's arguments.

busstoptaktik: See https://github.com/busstoptaktik/proj.4/blob/Horner-and-Helmert/src/PJ_pipeline.c#L359-L388 for the current implementation, which I think captures most of your intentions

PR #461, opened 2016-12-06
Tmerc docs projection - @kbevers remark on typo

julien2512:
@kbevers remark on typo


julien2512:
problem with this pull request. starting a fresh one !

PR #464, opened 2016-12-12
Add a scripts/fix_typos.sh script and correct reported errors

rouault:


kbevers:
Cool. This could be implemented as a test on travis. Fail on typos in new code. Or is that a bit too over the top? :)

hobu:
> Fail on typos in new code

American dictionary exceptionalism it is then?

rouault:
Note that the dictionaries used evolve over time so depending when the script is run results might change

hobu:
I don't think we should force them for tests to pass, but maybe as part of the docs `make html` process or something?

kbevers:
That's not a bad idea. I think most contributors these days aren't native english speakers. Spelling errors are bound to happen (I've introduced my fair share it seems). This is a good way to catch them.

PR #465, opened 2016-12-12
Link to NEWS for Release Notes.

micahcochran:
This is a small clean up to the documentation here http://proj4.org/download.html

This an attempt to make the project slightly easier to maintain by just pointing users to all the release notes contained within NEWS file.


micahcochran:
Thanks.

PR #468, opened 2016-12-18
.travis.yml: add multi config setup

rouault:
Setup 4 configs: Linux/GCC, Linux/CLang, OSX CLang and mingw32
OSX allowed to fail, since it fails in the proj -VC step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,132 +1,55 @@
 # This is the config file for building proj.4 and running its test suite
 # with Travis-ci.org
 
-language: c
-
-services:
-    docker
-
-compiler:
-  - gcc
-#  - clang
-
-os:
-  - linux
-#  - osx
+cache:
+    directories:
+        - $HOME/.ccache
 
 matrix:
-  exclude:
-    - os: osx
+  fast_finish: true
+  include:
+    - os: linux
       compiler: gcc
+      language: c
+      sudo: required
+      services:
+        - docker
+      env:
+        - BUILD_NAME=linux_gcc
+        - DETAILS="linux, gcc"
+    - os: linux
+      compiler: clang
+      language: c
+      sudo: required
+      services:
+        - docker
+      env:
+        - BUILD_NAME=linux_clang
+        - DETAILS="linux, clang"
+    - os: osx
+      language: c
+      env:
+        - BUILD_NAME=osx
+        - DETAILS="osx"
+    - os: linux
+      compiler: gcc
+      language: c
+      sudo: required
+      env:
+        - BUILD_NAME=mingw32
+        - DETAILS="mingw32"
 
-before_install: ./travis/before_install.sh
+#  allow_failures:
+#    - os: osx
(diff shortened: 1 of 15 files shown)


kbevers:
Why does it fail on OS X? I have no problems running ```proj -VC``` on OS X.

rouault:
@kbevers I've no idea why it fails (the output doesn't really tell if it some subtle precision difference for example), but it does. See https://travis-ci.org/rouault/proj.4/jobs/184953944

kbevers:
Actually... I see the same on my mac. Must have been introduced recently. I'll try to pin-point when.

And yes, the self-tests aren't very useful when it comes to output. I have been thinking a lot about creating an entirely new unified testing system. I would base it on the GIGS testing framework that @micahcochran made and have json-files with inputs and expected returns. On top of that I would make some C program that tests the public API thoroughly. With that in place I would scrap the self-test and the tests in ```nad/``` (the test coordinates would be re-used). It would be a simpler and more modern system that can be executed on all systems. Thoughts?

kbevers:
And by the way, I like this setup. I am not sure I understand the linux/mingw32 stuff though. Is that emulating windows on linux?

kbevers:
44c611c574f001267cd39c904c081e8f1f0d8c96 is the culprit. I'll see if I can eliminate some of the errors. I think you should remove the ```allow_failures``` tag for osx, this is a regression we could have avoided had travis been configured like suggested here.

rouault:
> I am not sure I understand the linux/mingw32 stuff though. Is that emulating windows on linux?

i586-mingw32msvc is a gcc based cross-compiling chain that targets Windows (this is a rather ancient one by the way. there are others). And in that config I also install wine, the Windows user land emulator, so as to be able to run .exe

kbevers:
That's a neat idea. I knew about wine but never thought about using it this way. Do you think this approach is better than running windows tests on appveyor?

rouault:
> Do you think this approach is better than running windows tests on appveyor?

It's complementary. It can catch some Windows specific issues, with free software only, but not all. And there are people also using mingw (generally on native Windows OS rather than on cross compiling) instead of MSVC to compile their stuff, so it is a valid target (although admitedly somewhat marginal).

kbevers:
I am all in favour of testing as many configurations as possible.

rouault:
I guess I should wait for the OSX failures to be fixed before merging ? (otherwise that's going to confuse potential new PR)

kbevers:
Yeah, I think #469 needs to be resolved before merging this. I am looking at it right now. I am hoping that it is something trivial.

micahcochran:
@kbevers I intended the GIGS testing system to be simple enough that it could serve as a general purpose unit test system.   Anyway, I'd be willing to help in whatever way that I can.

kbevers:
@rouault tests should now pass on OSX, so I guess you can merge this whenever you are ready.

PR #493, opened 2017-02-17
Pushing fix for thread safety issue 478

paulhanssen:
Hi,

I'm from the same company as the person that raised the issue in https://github.com/OSGeo/proj.4/issues/478.

This is our fix for it ... it compiles and runs "make tests" successfully.

Hope it helps.

### DIFF
--- a/src/pj_init.c
+++ b/src/pj_init.c
@@ -685,6 +685,19 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
         PIN = 0;
     }
 
+    /* load gridlist */
+    if (pj_param(ctx, PIN->params,"snadgrids").s != 0) {
+      PIN->gridlist = pj_gridlist_from_nadgrids( ctx, 
+        pj_param(ctx, PIN->params,"snadgrids").s, 
+        &(PIN->gridlist_count) );
+    }
+
+    if (PIN->has_geoid_vgrids) {
+      PIN->vgridlist_geoid = pj_gridlist_from_nadgrids( ctx,
+        pj_param(ctx, PIN->params,"sgeoidgrids").s,
+        &(PIN->vgridlist_geoid_count) );
+    }    
+
     return PIN;
 }
 


rouault:
Apparently it fails the multithreading stress test:
```
Segmentation fault      (core dumped) PROJ_LIB=../nad src/multistresstest
```

kbevers:
@paulhanssen Have you lost interest in this, or do you want to try to fix the problem pointed out by the tests?

paulhanssen:
Hi @kbevers I did not make this change, only pushed the pull request, but let me try giving it a go over the next week.

kbevers:
Cool, thanks!

liuj03:
I wrote the fix originally.

We were trying to call pj_transform on the same projPJ instance from multiple threads. But encountered a race condition which caused pj_transform to crash.

https://github.com/OSGeo/proj.4/issues/478

Our previous workaround was to mutex every call to proj4, which eventually became too big a bottleneck.

The fix appears to have resolved the race condition and crash is no longer happening. But I may not have sufficient understanding of proj4 to ensure all cases are covered.

Is the latest version of proj4 able to handle multiple calls to pj_transform on the same projPJ instance from multiple threads?

kbevers:
> Is the latest version of proj4 able to handle multiple calls to pj_transform on the same projPJ instance from multiple threads?

You shouldn't do that, no. I understand what you were trying to do, and appreciate that you have a working setup in your environment, but if we are to merge this change at least the continuous integration tests should pass. They don't at the moment. 

If you don't want to pursue a fix that passes our tests, then that is perfectly fine. If so, let me know and I'll close this PR.

liuj03:
What is the correct way to call pj_transform from multiple threads?
Do I need to have a separate instance of projPJ per thread?

busstoptaktik:
In brief: Yes.

A little less brief: I think it has never been discussed in public, but
apparently, the "context" element of proj has to do with thread safety
(albeit in ways that are not totally clearly described or argued), and each
PJ object is connected to a specific context (which may be shared among
many PJs)

In the OBS API, which I am currently developing (cf. pj_obs_api.c and
proj.h), one of the intentions is to make this more clear (an upcoming PR
will address this in more depth).

A major problem in touching the proj context related code is that it is not
really clear who uses it, hence very hard to devise evolutionary paths that
do not interfere with established use cases.

/thomas


2017-06-02 11:24 GMT+02:00 Jason Liu <notifications@github.com>:

busstoptaktik:
Closing, as this is likely only relevant for the use case in operation in the project by @paulhanssen  and @liuj03 (see comments above regarding why)

@paulhanssen  and @liuj03: If you believe closing is wrong, please reopen, and explain why

PR #495, opened 2017-02-25
Fix warnings related to -Wshadow -Wnull-dereference -Wfloat-conversion -Wmissing-prototypes -Wmissing-declarations

rouault:


src/PJ_robin.c
```diff
@@ -20,47 +20,47 @@ struct COEFS {
...
-    {0.8679, -0.00609363, -0.000113898, 3.32484e-06},
-    {0.835, -0.00698325, -6.40253e-05, 9.34959e-07},
-    {0.7986, -0.00755338, -5.00009e-05, 9.35324e-07},
-    {0.7597, -0.00798324, -3.5971e-05, -2.27626e-06},
-    {0.7186, -0.00851367, -7.01149e-05, -8.6303e-06},
-    {0.6732, -0.00986209, -0.000199569, 1.91974e-05},
-    {0.6213, -0.010418, 8.83923e-05, 6.24051e-06},
-    {0.5722, -0.00906601, 0.000182, 6.24051e-06},
-    {0.5322, -0.00677797, 0.000275608, 6.24051e-06}
+    {1, 2.2199e-17f, -7.15515e-05f, 3.1103e-06f},
```
QuLogic: Didn't change this `1` like the other below.

src/p_series.c
```diff
@@ -4,6 +4,10 @@
 #include <string.h>
 #define NF 20 /* length of final format string */
 #define CUT 60 /* check length of line */
+
+/* FIXME: put the declaration in a header. Also used in gen_cheb.c */
+void p_series(Tseries *T, FILE *file, char *fmt);
+
 	void
```
QuLogic: Maybe fix this extra indent like the other one, too?

src/rtodms.c
```diff
@@ -49,9 +49,9 @@ rtodms(char *s, double r, int pos, int neg) {
 	r = floor(r * CONV + .5);
 	sec = fmod(r / RES, 60.);
 	r = floor(r / RES60);
-	min = fmod(r, 60.);
+	min = (int)fmod(r, 60.);
         r = floor(r / 60.);
-        deg = r;
+        deg = (int)r;
```
QuLogic: What's up with the odd spacing here? Mixing tab and space?

src/test228.c
```diff
@@ -59,9 +60,13 @@ void* thread_main(void* unused)
 int main(int argc, char* argv[])
```
QuLogic, line 60: Could just be `int main(void)`.

rouault: Good point. Fixed in master now

PR #496, opened 2017-02-28
Enable cppcheck and fix related mostly false-positive warnings

rouault:


src/geodesic.c
```diff
@@ -85,7 +85,10 @@ static void Init() {
     tolb = tol0 * tol2;
     xthresh = 1000 * tol2;
     degree = pi/180;
-    NaN = sqrt(-1.0);
+    {
```
kbevers, line 88: Perhaps this should be changed upstream?

rouault: @cffk See above. One of the versions of cppcheck complains that sqrt(-1.0) is undefined, but later no longer.

cffk: Thanks, I'll backport these changes.  Also I note that Visual Studio gives two warnings
that S12 "may" be uninitialized.  It's wrong! but I initialize it to 0 in those places together
with a comment /* Initialize S12 to stop Visual Studio warning */

src/PJ_labrd.c
```diff
@@ -60,14 +60,18 @@ static LP e_inverse (XY xy, PJ *P) {          /* Ellipsoidal, inverse */
     xy.y +=   Q->Cb * V1 - Q->Ca * V2 - Q->Cd * V3 + Q->Cc * V4;
     ps = Q->p0s + xy.y / Q->kRg;
     pe = ps + P->phi0 - Q->p0s;
-    for ( i = 20; i; --i) {
+
+    V1 = Q->A * log(tan(M_FORTPI + .5 * pe));
```
kbevers: Wouldn't a `do {} while ()` be the more elegant solution here?

src/PJ_labrd.c
```diff
@@ -46,7 +46,7 @@ static XY e_forward (LP lp, PJ *P) {          /* Ellipsoidal, forward */
 static LP e_inverse (XY xy, PJ *P) {          /* Ellipsoidal, inverse */
     LP lp = {0.0,0.0};
     struct pj_opaque *Q = P->opaque;
-    double x2, y2, V1, V2, V3, V4, t, t2, ps, pe, tpe, s;
+    double x2, y2, V1, V2, V3, V4, t = 0.0, t2, ps, pe, tpe, s;
```
cffk: I recommend tagging (with a comment) any unnecessary initializations performed to suppress a warning.  Then a developer 5 years from now won't be left wondering whether the value t = 0.0 ever gets used.

rouault: Good point. Comment added

PR #503, opened 2017-04-18
Fix aea projection return value in case of invalid parameters

maximchurilin:
In aea projection setup(P) will free P pointer and return 0 (inside E_ERROR macro) if any of projection parameters are invalid. Right now return value of setup(P) is ignored that could lead to attempt to free P pointer once more and segmentation fault. 
This commit fixes described issue.


busstoptaktik:
Very well spotted - thanks!

PR #504, opened 2017-04-25
Put the cmake exported targets in a namespace

leolchat:
It is good practice to have imported targets be in the project namespace (https://cmake.org/cmake/help/v3.8/manual/cmake-developer.7.html?highlight=namespace). Even more when the target names are so simple and confusing as `proj`


kbevers:
Seems to be functionally equivalent to https://github.com/OSGeo/proj.4/pull/451. Is that correct?

leolchat:
Sorry, yes. That is a duplicate.

kbevers:
Allright, I'll close this. But thanks for submitting anyway, at least we know that this is something that is of interest within the community.

PR #507, opened 2017-05-02
Extended Helmert transformation to 14-parameters.

kbevers:
This PR extends the Helmert transformation to the fourth dimension and
enables spatio-temporal datum shifts in PROJ.4. On top of the usual 7
parameters (`+x`, `+y`, `+z`, `+s`, `+rx`, `+ry`, `+rz`) the rates of change of the seven
parameters can now be used as well. The new parameters are called `+dx`,
`+dy`, `+dx`, `+ds`, `+drx`, `+dry` and `+drz`. To keep track of the datum epoch and
coordinate epoch two additional parameters have been added to the
Helmert transformation, one of which is mandatory in the 14-parameter
case. The mandatory datum epoch is controlled with `+epoch` (given in
YYYY.yyyy format) and the coordinate, or observation, epoch is either
controlled in the proj-string with `+tobs` or by using 4D-coordinates
when transformating coordinates with pj_trans().

Full examples of how to use the 14-parameter version of `PJ_helmert` can be found
in the test functions. Here's is a quick example of a proj-string that transforms from
ITRF2000@2017.0 to ITRF93@2017.0. 

```
+proj=helmert
+x=0.0127 +y=0.0065 +z=-0.0209 +s=0.00195
+rx=-0.00039 +ry=0.00080 +rz=-0.00114
+dx=-0.0029 +dy=-0.0002 +dz=-0.0006 +ds=0.00001
+drx=-0.00011 +dry=-0.00019 +drz=0.00007
+epoch=1988.0 +tobs=2017.0 +transpose
```

Effectively this is a first step towards supporting dynamic datums like the Australian [ATRF](http://www.icsm.gov.au/gda2020/index.html) which will be introduced in a couple of years. It is also a step in the right direction in terms of supporting WKT2 datum descriptions in open source GIS in general, since datum shifts can be described more generally with the 14-parameter Helmert than what the deprecated `TOWGS84` field that the first version of WKT offers.

### DIFF
--- a/src/PJ_helmert.c
+++ b/src/PJ_helmert.c
@@ -6,7 +6,8 @@
 
 ************************************************************************
 
-  Implements 3- and 7-parameter Helmert transformations for 3D data.
+  Implements 3-, 7- and 14-parameter Helmert transformations for 3D
+  data.
 
   Primarily useful for implementation of datum shifts in transformation
   pipelines.
@@ -14,7 +15,8 @@
 ************************************************************************
 
 Thomas Knudsen, thokn@sdfe.dk, 2016-05-24/06-05
-Last update: 2016-11-24
+Kristian Evers, kreve@sdfe.dk, 2017-05-01
+Last update: 2017-05-01
 
 ************************************************************************
 * Copyright (c) 2016, Thomas Knudsen / SDFE
@@ -46,7 +48,7 @@ Last update: 2016-11-24
 #include <assert.h>
 #include <stddef.h>
 #include <errno.h>
-PROJ_HEAD(helmert, "3- and 7-parameter Helmert shift");
+PROJ_HEAD(helmert, "3-, 7- and 14-parameter Helmert shift");
 
 static XYZ helmert_forward_3d (LPZ lpz, PJ *P);
 static LPZ helmert_reverse_3d (XYZ xyz, PJ *P);
@@ -81,10 +83,17 @@ struct pj_opaque_helmert {
     Projection specific elements for the "helmert" PJ object
 ************************************************************************/
     XYZ xyz;
+    XYZ xyz_0;
+    XYZ dxyz;
     PJ_OPK opk;
+    PJ_OPK opk_0;
+    PJ_OPK dopk;
     double scale;
+    double scale_0;
+    double dscale;
     double R[3][3];
-    int no_rotation;
+    double epoch, t_obs;
+    int no_rotation, approximate, transpose;
 };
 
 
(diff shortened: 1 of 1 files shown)


src/PJ_helmert.c
```diff
@@ -101,7 +110,206 @@ struct pj_opaque_helmert {
...
+    Q->xyz.z = Q->xyz_0.z + Q->dxyz.z * dt;
+
+    Q->opk.o = Q->opk_0.o + Q->dopk.o * dt;
+    Q->opk.p = Q->opk_0.p + Q->dopk.p * dt;
+    Q->opk.k = Q->opk_0.k + Q->dopk.k * dt;
+
+    Q->scale = Q->scale_0 + Q->dscale * dt;
+
+    /* debugging output */
+    pj_log_trace(P, "Transformation parameters for observation epoch %g:", Q->t_obs);
```
busstoptaktik: Since this is a large block of logging calls, I suggest surrounding them with
"if (pj_err_level (P, PJ_ERR_TELL) >= PJ_LOG_TRACE) {...}". It will eliminate 8 function calls (although, since this is setup-code, it will happen only once. The situation is worse if it's part of the projection code)

src/PJ_helmert.c
```diff
@@ -245,196 +468,102 @@ PJ *PROJECTION(helmert) {
...
+    /* Use "other" rotation sign convention? */
+    if (pj_param (P->ctx, P->params, "ttranspose").i)
+        Q->transpose = 1;
+
+    Q->xyz    =  Q->xyz_0;
+    Q->opk    =  Q->opk_0;
+    Q->scale  =  Q->scale_0;
+
+    /* Let's help with debugging */
+    pj_log_debug(P, "Helmert parameters:");
```
busstoptaktik: Same comment as above, regarding enclosing in if (PJ_LOG...) { ...}.

I agree that this logging info is very important when implementing new transformations, so the comment about helping with debugging is highly appropriate

src/PJ_helmert.c
```diff
@@ -46,7 +48,7 @@ Last update: 2016-11-24
 #include <assert.h>
 #include <stddef.h>
 #include <errno.h>
-PROJ_HEAD(helmert, "3- and 7-parameter Helmert shift");
+PROJ_HEAD(helmert, "3-, 7- and 14-parameter Helmert shift");
```
busstoptaktik, line 51: For future work, we should also add the 4-parameter 2D Helmert transform (note to self. Will add to issue tracker)

busstoptaktik (approved):
Neat code, nice test additions, good work!

I have only one suggestion about enclosing large blocks of logging calls with a conditional to see whether they will actually result in any output anyway. It is really not very important when part of the initialization code, but it could serve as a nice go-to example of how to do it, e.g. in replies on the mailing list.

PR #510, opened 2017-05-08
Completing the Helmert transformation with the 4-parameter shift

kbevers:
Completing the Helmert driver with the 4-parameter shift that handles
the 2D transformation. The implementation is written in such a way that
not only 2D-points but also 3D- and 4D-points can be transformed with
the 4-parameter Helmert. The four parameters that can be set in this
mode are +x, +y (translations), +s (scale) and +theta (rotation). The
presence of the +theta parameter activates the 2D-helmert code,
irregardless of the input data's dimensions.

The units are meters for the translations as in all the other versions
of the Helmert transform. The rotation unit is arcseconds.
The units of the scale differ from the 3-, 7- and 14-parameter shift
where the unit is ppm. Here it is instead given directly and is as such
unitless.

The 4-parameter case can also be extended to an 8-parameter shift in the
same way as the 7-parameter shift extens to the 14-parameter shift. This
might be a bit silly and will probably never be used, but nonetheless, I
have included it for the sake of completeness. The rates of change are
givens as +dx, +dy, +ds and +dtheta.

### DIFF
--- a/src/PJ_helmert.c
+++ b/src/PJ_helmert.c
@@ -1,12 +1,12 @@
 /***********************************************************************
 
-             Helmert transformations, 3- and 7-parameter
+           Helmert transformations, 3-, 4-, 7- and 14-parameter
 
                     Thomas Knudsen, 2016-05-24
 
 ************************************************************************
 
-  Implements 3-, 7- and 14-parameter Helmert transformations for 3D
+  Implements 3-, 4-, 7- and 14-parameter Helmert transformations for 3D
   data.
 
   Primarily useful for implementation of datum shifts in transformation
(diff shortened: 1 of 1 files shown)


busstoptaktik (approved):
My comments are minor, and only related to comments and descriptions - consider or ignore: Getting this code operational is really a huge improvement to the pipeline capabilities.

I'm sure your introduction of the 4 parameter shift will be a lifesaver also in many cases where old and new planar data of varying spatial accuracy is to be merged. The corresponding 8-parameter version is probably more of a rarity, but I agree that including it for completness' sake is worth it (rather than spending time explaining why it's not, even just once)

PR #514, opened 2017-05-16
Unit conversion pseudo-projection.

kbevers:
A pseudo-projection that can be used to convert units of input and
output data. Primarily useful in pipelines.

Unit conversion is performed by means of a pivot unit. The pivot unit
for distance units are the meter and for time we use the modified julian
date. A time unit conversion is performed like

    Unit A -> Modified Julian date -> Unit B

distance units are converted in the same manner, with meter being the
central unit.

The modified Julian date is chosen as the pivout unit since it has a
fairly high precision, goes sufficiently long backwards in time, has no
danger of hitting the upper limit in the near future and it is a fairly
common time unit in astronomy and geodesy. Note that we are using the
Julian date and not day. The difference being that the latter is defined
as an integer and is thus limited to days in resolution. This approach
has been extended wherever it makes sense, e.g. the GPS week unit also
has a fractional part that makes it possible to determine the day, hour
and minute of an observation.

In- and output units are controlled with the parameters

    +xy_in, +xy_out, +z_in, +z_out, +t_in and +t_out

where xy denotes horizontal units, z vertical units and t time units.

Distance units are converted similar to what is already in use in PROJ.4.
To ease usage of the already defined conversion factors a new column
with the factors defined as doubles has been added to the pj_units array.
This simplifies the code significantly, since parsing the defined strings
can be avoided.

A quick example to convey the idea behind this. Observations from a GPS station will usually have timestamps in the GPS week format. When transforming the observed coordinates from a ITRF reference frame to a local frame with a Helmert shift, the required time format is decimalyears. 

```
+proj=pipeline +step
+proj=unitconvert +t_in=gps_week +t_out=decimalyear
+proj=helmert
+x=0.0127 +y=0.0065 +z=-0.0209 +s=0.00195
+rx=-0.00039 +ry=0.00080 +rz=-0.00114
+dx=-0.0029 +dy=-0.0002 +dz=-0.0006 +ds=0.00001
+drx=-0.00011 +dry=-0.00019 +drz=0.00007 
+epoch=1988.0 +transpose
```

Hence the need for a unit conversion tool like this. For completeness unit conversion of horizontal and vertical unit conversion has been included as well.

For now only decimalyears, GPS weeks and the modified Julian date has been implemented. When the need arises other time units should be implemented as well, for instance unix timestamps, various Julian date incarnations, GPS time (from the LAS pointcloud format), etc., etc.


kbevers:
AppVeyor might have gone mad. It is clearly a timeout thing, but it puzzles me as to why it takes so long to finish and why it fails in the exact same place i two tests. I have even replicated it in several builds by recommitting the latest changes a few times.

PR #516, opened 2017-05-20
Memleak fixes

rouault:
@busstoptaktik I'm not completely sure about the change in pj_init(). It could be a problem if a (*proj)(PIN) function needed the ->geod member, but it doesn't look like this is the case. Annther issue, which existed before, is that in the unlikely case PIN->geod = pj_calloc(...) would fail, we would return a valid PIN pointer. So code later that would need geod could crash

### DIFF
--- a/src/PJ_wag3.c
+++ b/src/PJ_wag3.c
@@ -28,13 +28,8 @@ static LP s_inverse (XY xy, PJ *P) {           /* Spheroidal, inverse */
 	return lp;
 }
 
-
-static void *freeup_new (PJ *P) {                       /* Destructor */
-    return pj_dealloc(P);
-}
-
 static void freeup (PJ *P) {
-    freeup_new (P);
+    pj_freeup_plain (P);
     return;
 }
 
@@ -43,7 +38,10 @@ PJ *PROJECTION(wag3) {
 	double ts;
     struct pj_opaque *Q = pj_calloc (1, sizeof (struct pj_opaque));
     if (0==Q)
-        return freeup_new (P);
+    {
+        freeup(P);
+        return 0;
+    }
     P->opaque = Q;
 
 	ts = pj_param (P->ctx, P->params, "rlat_ts").f;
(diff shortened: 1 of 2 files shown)


src/PJ_wag3.c
```diff
@@ -30,6 +30,11 @@ static LP s_inverse (XY xy, PJ *P) {           /* Spheroidal, inverse */
 
 
 static void *freeup_new (PJ *P) {                       /* Destructor */
```
busstoptaktik: I suggest using the simplified approach used in PJ_cart.c: Remove freeup_new(), and replace freeup() with the version from PJ_cart.c:
```
static void freeup (PJ *P) {
    pj_freeup_plain (P);
    return;
}
```

Later on, I expect to improve the PROJECTION macro, to replace freeup with pj_freeup_plain. It will, however require one pass over the code, cleaning up every PJ_*.c file (but it will result in removing most cases of freeup-functions, and a bit of init-code for each projection, at no cost at all (in the few cases where freeup does more advanced things than freeing P->opaque, then freeing P, we will have to add one line: P->pfree = freeup;)

src/pj_init.c
```diff
@@ -685,6 +680,12 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
             }
         PIN = 0;
     }
+    else {
```
busstoptaktik, line 683: Yes - this seems to repair a genuine blunder (although it would be nice if geod_init included a check for null pointers, but I think, given that the geod-code seems to be autotranslated from the C++-code of GeographicLib, would be too much to ask for, so this solution is definitely preferrable)

busstoptaktik (commented):
Two interesting bugs, and two nice and simple repairs. I do, however, suggest to change the freeup reapir to be more in line with my ongoing attempts to simplify the overall memory management (see comment inlined in the patch)

src/PJ_wag3.c
rouault: OK, I pasted that from another file. Fixed from your suggestion

PR #517, opened 2017-05-23
Fix possible zero division [OSS-Fuzz issue 1801]

kbevers:
For some extreme values of eccentricity a zero divison can occur. In
those rare cases we return HUGE_VAL to indicate something went wrong
while still returning a defined value.

Credit to OSS Fuzz.

I still need to confirm this against OSS-Fuzz, just checking that the usual test-suite is happy.

### DIFF
--- a/src/pj_tsfn.c
+++ b/src/pj_tsfn.c
@@ -2,9 +2,15 @@
 #include <math.h>
 #include <projects.h>
 
-	double
-pj_tsfn(double phi, double sinphi, double e) {
-	sinphi *= e;
-	return (tan (.5 * (M_HALFPI - phi)) /
-	   pow((1. - sinphi) / (1. + sinphi), .5 * e));
+double pj_tsfn(double phi, double sinphi, double e) {
+    double denominator;
+    sinphi *= e;
+
+    /* avoid zero division, fail gracefully */
+    denominator = 1.0 + sinphi;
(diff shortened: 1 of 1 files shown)


src/pj_tsfn.c
```diff
@@ -2,9 +2,15 @@
...
-	sinphi *= e;
-	return (tan (.5 * (M_HALFPI - phi)) /
-	   pow((1. - sinphi) / (1. + sinphi), .5 * e));
+double pj_tsfn(double phi, double sinphi, double e) {
+    sinphi *= e;
+
+    /* sinphi * e == -1 results in zero-division below,
+       instead we return a defined value that indicates
+       something went wrong. */
+    if (sinphi == -1.0)
```
rouault: I don't trust floating point logic, so I would replace this test by 1. + sinphi == 0.0 to have exactly the same expression as below. Or create a divisor = 1. + sinphi variable and use it in both places

rouault:
Would be good to add a URL to the bugs.chromium.org ticket in the commit message as well

src/pj_tsfn.c
kbevers: Good suggestion. I'll be right on it!

kbevers:
> Would be good to add a URL to the bugs.chromium.org ticket in the commit message as well

I shall add that as well

PR #518, opened 2017-05-23
Avoid potentially very long loop when normalizing longitudes around long_wrap_center

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=1809

Credit to OSS Fuzz

### DIFF
--- a/src/pj_init.c
+++ b/src/pj_init.c
@@ -559,7 +559,16 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
     /* longitude center for wrapping */
     PIN->is_long_wrap_set = pj_param(ctx, start, "tlon_wrap").i;
     if (PIN->is_long_wrap_set)
+    {
         PIN->long_wrap_center = pj_param(ctx, start, "rlon_wrap").f;
+        /* Don't accept excessive values otherwise we might perform badly */
+        /* when correcting longitudes around it */
+        if( !(fabs(PIN->long_wrap_center) < 10 * M_TWOPI) )
+        {
+            pj_ctx_set_errno( ctx, -14 );
+            goto bum_call;
+        }
+    }
 
     /* axis orientation */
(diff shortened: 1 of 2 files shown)


src/pj_init.c
```diff
@@ -559,7 +559,16 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
     /* longitude center for wrapping */
     PIN->is_long_wrap_set = pj_param(ctx, start, "tlon_wrap").i;
     if (PIN->is_long_wrap_set)
+    {
         PIN->long_wrap_center = pj_param(ctx, start, "rlon_wrap").f;
+        /* Don't accept excessive values otherwise we might perform badly */
+        /* when correcting longitudes around it */
+        if( !(fabs(PIN->long_wrap_center) < 10 * M_TWOPI) )
```
kbevers, line 566: Is this not equivalent to `if ( fabs(PIN->-long_wrap_center) > 10 * M_TWOPI )` ? A little easier to read I think.

rouault: Not completely... In case long_wrap_center is NaN.   fabs(NaN) > bla is equivalent to NaN > bla , and comparisons involving NaN are always false

kbevers: Bloody NaN's! Allright, carry on :-)

busstoptaktik: Another indication that we need a perl-like "unless" construct in C (as in #define unless(x) if (x){;} else, or (less safe) #define unless(x) if (!(x))

There are quite a few expressions which are way easier to grasp mentally if stated in negated logic :-)

cffk: Perhaps a formulaic comment "Backwards test to allow NaNs to satisfy the condition" is called for.

rouault:
> Perhaps a formulaic comment "Backwards test to allow NaNs to satisfy the condition" is called for.

Agreed. Done

PR #519, opened 2017-05-23
Handle zero division errors reported by OSS-Fuzz

kbevers:
I'll try to take care of alle the zero division errors in this PR.

The relevant issues are:
 
- [x] 1814
- [x] 1800
- [ ] 1798
- [x] 1797
- [x] 1795
- [x] 1793
- [x] 1791
- [x] 1789
- [x] 1787

### DIFF
--- a/src/PJ_aea.c
+++ b/src/PJ_aea.c
@@ -30,8 +30,8 @@
 #define PJ_LIB__
 #include <projects.h>
 
-# define EPS10	1.e-10
-# define TOL7	1.e-7
+# define EPS10  1.e-10
+# define TOL7   1.e-7
 
 PROJ_HEAD(aea, "Albers Equal Area") "\n\tConic Sph&Ell\n\tlat_1= lat_2=";
 PROJ_HEAD(leac, "Lambert Equal Area Conic") "\n\tConic, Sph&Ell\n\tlat_1= south";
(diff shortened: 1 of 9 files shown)


kbevers:
According to cppcheck this 

```
    if (EN->C == 0.0)
        return NULL;
``` 

introduces a memory leak. How so?

QuLogic:
`en` was allocated on line 48 above.

kbevers:
> en was allocated on line 48 above.

Yes. I don't see what the problems with that is though?

QuLogic:
Returning early means it's not freed.

kbevers:
Doh! Obviously. Some how I was blind to that... Thanks.

kbevers:
@cffk 

OSS-Fuzz experienced a zero-division in the function below from `geodesic.c`
```
static real copysignx(real x, real y) {
  return fabs(x) * (y < 0 || (y == 0 && 1/y < 0) ? -1 : 1);
}
```
What's a good value to return when it fails? Or can this perhaps be solved better in another part of the code?

cffk:
For CPUs supporting IEEE floating point (all of them at this point?),
division by floating point zero is well defined (returning +/- inf or
nan).

The copysignx code uses 1/y to distinguish between +/- 0.0 and in some
cases this distinction is important.  I recommend retaining the code;
maybe there's some way of telling OSS-Fuzz that this has been checked.

An uglier (and more expensive version) of this code is

    return fabs(x) * (y < 0 || (y == 0 && atan2(y, -1.0) < 0) ? -1 : 1);

Btw, copysign is part of the standard C++ library (since C++11) and is
available in C's libm on some platforms.  If it's known that copysign is
available then you can just use

    return copysign(x, y);

Later... I see in http://en.cppreference.com/w/c/numeric/math/copysign that
this function appeared in the C99 standard.  Does the configuration for proj.4
check for C99 and provide a macro to check?  If so, the code could check
that macro (and this would apply to other functions, e.g., cbrtx).  If it doesn't
I can add the functionality (maybe) by testing compilation of test code.

PR #520, opened 2017-05-24
Remove confusing macros.

QuLogic:
The upper-case macro `VAR` casts the lower-case variable `var` to an expected type, but that usage obfuscates the variables to the casual reader. This is especially prevalent in the allocation function in which `var` is the correct type already, so both `var` and `VAR` are valid *and* in use.

In any case, for those functions that don't obtain `var` in the correct type from their parameters, it's much simpler to just cast once at the beginning.

### DIFF
--- a/src/pj_gauss.c
+++ b/src/pj_gauss.c
@@ -34,7 +34,6 @@ struct GAUSS {
     double e;
     double ratexp;
 };
-#define EN ((struct GAUSS *)en)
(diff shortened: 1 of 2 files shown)


kbevers:
Those defines certainly tripped me up. I think this is a good change, and no one else has objected to it so I'll merge it. Thanks @QuLogic.

PR #521, opened 2017-05-31
Disallow +s=0 when in 4-param. mode.

kbevers:
Avoids zero-division in PJ_helmert.c

Fixes
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=1956

Credit to OSS-Fuzz.

### DIFF
--- a/src/PJ_helmert.c
+++ b/src/PJ_helmert.c
@@ -520,12 +520,15 @@ PJ *PROJECTION(helmert) {
     if (pj_param (P->ctx, P->params, "ttheta").i) {
         Q->theta_0 = pj_param (P->ctx, P->params, "dtheta").f * ARCSEC_TO_RAD;
         Q->fourparam = 1;
+        Q->scale_0 = 1.0; /* default scale for the 4-param shift */
     }
 
     /* Scale */
-    if (pj_param (P->ctx, P->params, "ts").i)
+    if (pj_param (P->ctx, P->params, "ts").i) {
         Q->scale_0 = pj_param (P->ctx, P->params, "ds").f;
-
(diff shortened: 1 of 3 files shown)


src/PJ_helmert.c
```diff
@@ -520,12 +520,15 @@ PJ *PROJECTION(helmert) {
...
+        Q->scale_0 = 1.0; /* default scale for the 4-param shift */
     }
 
     /* Scale */
-    if (pj_param (P->ctx, P->params, "ts").i)
+    if (pj_param (P->ctx, P->params, "ts").i) {
         Q->scale_0 = pj_param (P->ctx, P->params, "ds").f;
-
+        if (pj_param (P->ctx, P->params, "ttheta").i && Q->scale_0 == 0.0)
+            return freeup_msg(P, -52);
```
rouault: I hate this old habit of hardcoding error codes. Shouldn't we extend the list in project.h with a #define PJD_ERR_SCALE_CANNOT_BE_ZERO -52 ? and in the comment of the new line in pj_strerrno, mention this symbolic name as well

In that case, shouldn't we actually generalize the error to PJD_INVALID_SCALE if negative or other values are found to be non sensical at some point ?

kbevers: I don't like it either. If the list in projects.h is extended I think it should be done for all error codes. It would definitely make it easier to read the code. Perhaps it should be done in a separate PR?

For now I will change the error message to "Invalid scale".

rouault: Converting everything might be good indeed but tedious... But I see someone (likely @busstoptaktik if it is not you) added a few PJD_ERR_ constants in projects.h, so I think we should at least use that new good habit for new codes.

kbevers: Good point. I'll add a constant to projects.h

PR #524, opened 2017-06-13
PJ_vgridshift.c: fix memleak in initialization if grid loading fail

rouault:
When gridloading fail for one grid, but return a gridlist, we leak the grid list array.
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2230. Credit to OSS Fuzz

Also use the P->vgridlist_geoid member consistently to try reducing confusion between
P->gridlist and P->vgridlist_geoid

### DIFF
--- a/src/PJ_vgridshift.c
+++ b/src/PJ_vgridshift.c
@@ -29,8 +29,8 @@ static XYZ forward_3d(LPZ lpz, PJ *P) {
         /* Only try the gridshift if at least one grid is loaded,
          * otherwise just pass the coordinate through unchanged. */
         pj_apply_vgridshift( P, "sgrids",
-                             &(P->vgridlist_geoid),
-                             &(P->vgridlist_geoid_count),
+                             &(P->gridlist),
+                             &(P->gridlist_count),
                              1, 1, 0,
                              &point.xyz.x, &point.xyz.y, &point.xyz.z );
     }
@@ -47,8 +47,8 @@ static LPZ reverse_3d(XYZ xyz, PJ *P) {
         /* Only try the gridshift if at least one grid is loaded,
          * otherwise just pass the coordinate through unchanged. */
         pj_apply_vgridshift( P, "sgrids",
-                             &(P->vgridlist_geoid),
-                             &(P->vgridlist_geoid_count),
+                             &(P->gridlist),
+                             &(P->gridlist_count),
                              0, 1, 0,
                              &point.xyz.x, &point.xyz.y, &point.xyz.z );
     }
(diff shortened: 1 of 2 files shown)


src/PJ_vgridshift.c
```diff
@@ -76,14 +76,19 @@ PJ *PROJECTION(vgridshift) {
...
     /* Build gridlist. P->gridlist can be empty if +grids only ask for optional grids. */
-    P->gridlist = pj_gridlist_from_nadgrids( P->ctx, pj_param(P->ctx, P->params, "sgrids").s,
-                                             &(P->gridlist_count) );
+    P->vgridlist_geoid = pj_gridlist_from_nadgrids( P->ctx, pj_param(P->ctx, P->params, "sgrids").s,
+                                             &(P->vgridlist_geoid_count) );
 
     /* Was gridlist compiled properly? */
     if ( pj_ctx_get_errno(P->ctx) ) {
         pj_log_error(P, "vgridshift: could not find required grid(s).");
+        pj_dalloc(P->vgridlist_geoid);
```
rouault: Here's the real fix for the issue spotted by OSS Fuzz

kbevers:
I deliberately used `P->gridlist` in favour of `P->vgridlist_geoid` because the grids are not necessarily geoids grids. They could be a number of other types of grids, most notably various chart datums at sea (e.g. LAT, MLWS, etc.). And since we know that the `P` object is only used for vertical grid offsets in this case I considered it best thinking of it as a "gridlist", plain and simple. 
It is also an attempt at moving away from the old nomenclature of the `+nadgrids` and `+geoidgrids` keywords, that are in many (most?) cases misleading.

So to sum up, I'd rather not make the change to use `P-vgridlist_geoid`. However, I appreciate the fix for the  memleak that OSS-Fuzz found!

kbevers (changes requested):
I deliberately used `P->gridlist` in favour of `P->vgridlist_geoid` because the grids are not necessarily geoids grids. They could be a number of other types of grids, most notably various chart datums at sea (e.g. LAT, MLWS, etc.). And since we know that the `P` object is only used for vertical grid offsets in this case I considered it best thinking of it as a "gridlist", plain and simple. 
It is also an attempt at moving away from the old nomenclature of the `+nadgrids` and `+geoidgrids` keywords, that are in many (most?) cases misleading.

So to sum up, I'd rather not make the change to use `P-vgridlist_geoid`. However, I appreciate the fix for the  memleak that OSS-Fuzz found!

rouault:
@kbevers  OK, well noted for gridlist vs vgridlist_geoid. However currently at https://github.com/OSGeo/proj.4/blob/master/src/PJ_vgridshift.c#L32 and https://github.com/OSGeo/proj.4/blob/master/src/PJ_vgridshift.c#L50 , vgridlist_geoid is used. That looks like a bug / suboptimal.

kbevers:
Well spotted! I think we have a case of bad cut and paste here... It works because `pj_apply_vgridshift()` allocates the gridlist if it is empty. So yeah, definitely suboptimal although it works. Feel free to fix it!

In general the code regarding grids are rather messy and it is something I would like to take a deeper look at changing in the future.

rouault:
Just forced pushed a new version using gridlist everywhere

kbevers (approved):

PR #525, opened 2017-06-15
Better plots for documentation

kbevers:
Updated plots and plotting script.

Fixes https://github.com/OSGeo/proj.4/issues/513 and many more plots. On top of that the plots are now more visually appealing (to me, anyway). The downside is that this adds ~30MB to the repository, which is a bit on the heavy side.


mwtoews:
You can spare a few MBs by post-processing the images with `pngcrush` in a subprocess. Testing a few of the new files, I see potential reductions of 10-15%.

kbevers:
I already ran the [greyscale.sh](https://github.com/OSGeo/proj.4/blob/master/docs/plot/greyscale.sh) script which reduced it to half the size. I'll play around with `pngcrush` as well. Thanks for the suggestion, @mwtoews.

The files themselves are also quite large: 2400x1800 pixels. This could be reduced, although I think it is preferable to have higher resolution when we get round to producing a PDF-file of the docs. Thoughts?

micahcochran:
A little different direction might be SVG output.  All [major web browsers support it](https://caniuse.com/#search=SVG).  Matplotlib [supports SVG](https://matplotlib.org/faq/usage_faq.html#what-is-a-backend).

A cursory look at Sphinx (for the doc) appears that it also supports SVG images.

kbevers:
> All major web browsers support it.

I wasn't aware of that. I'll re-upload the files as svg. I think that is the best way to go.

cffk:
If you're moving to SVG, then you can shrink the size of the files (sometimes by a lot!) by simplifying the polylines.  An example of the Visvalingam-Whyatt line simplification algorithm in operation is given at https://bost.ocks.org/mike/simplify/.

hobu:
Sphinx+PDFlatex output of the website won't support SVG, and I'm 👎 on SVG because of that. Is it possible to get Sphinx+PDFlatex to support SVG?

kbevers:
@cffk also a good point! I tried it for the Airy-plot, it goes from 519kb as a PNG to 63kb as a SVG with simplified features. Definitely the way to go. I'll try to find the time to do a proper update during the weekend.

Thanks for all the suggestions.

micahcochran:
@hobu 
> Sphinx+PDFlatex output of the website won't support SVG

It looks like there might be an extension that works around for that problem, [sphinx.ext.imgconverters](http://www.sphinx-doc.org/en/stable/ext/imgconverter.html?highlight=SVG).  That requires Imagemagick, but I don't see that as a huge requirement.

kbevers:
I just tried building the website with the imgconverter extension and it didn't go well:

```
sh: line 1: 80496 Trace/BPT trap: 5       'inkscape' '/var/folders/br/ws53j18550zgc2wj7cdqblqh0000gn/T/magick-8049495UsldavyfqC' --export-png='/var/folders/br/ws53j18550zgc2wj7cdqblqh0000gn/T/magick-804942ZwkbqjX8zC7' --export-dpi='96,96' --export-background='rgb(100%,100%,100%)' --export-background-opacity='1' > '/var/folders/br/ws53j18550zgc2wj7cdqblqh0000gn/T/magick-804949VmApmHfag1N' 2>&1
convert: non-conforming drawing primitive definition `path' @ error/draw.c/DrawImage/3269.

Extension error:
convert exited with error:
[stderr]
None
[stdout]
b''
make: *** [html] Error 
```
Any ideas? My knowledge of imagemagick is near zero. I can reproduce the same error by a simple `convert airy.svg airy.png`. Maybe some extra arguments are needed for `convert`. They can be added in the sphinx setup if needed.

micahcochran:
Is Inkscape installed?

kbevers:
yes

micahcochran:
Does [this](http://www.wizards-toolkit.org/discourse-server/viewtopic.php?t=26010) help?

QuLogic:
Are you really running the `plot.sh` script? This does not work on Python 3 due to various Python 2 assumptions.

docs/plot/plot.py
```diff
@@ -1,39 +1,132 @@
...
+    y = np.array(y)
+
+    l = len(x)
+    dsts = np.sqrt((x[0:-1]-x[1:l])**2 + (y[0:-1]-y[1:l])**2)
+    I = dsts > tol
+
+    offset = 0
+    xy = np.ndarray(shape=(2, 0))
+    for i, b in enumerate(I):
+        xy = np.append(xy, (x[i], y[i]))
```
QuLogic: This is kind of inefficient; it would be better to group things, e.g., make a list where each entry is either a) a bunch of existing points that are below tolerance or b) an interpolated list of points. Then concatenate them at the end. I would try this out, but I can't yet get this script to run.

docs/plot/plot.py
```diff
@@ -63,7 +156,14 @@ def project(coordinates, proj_string, in_radians=False):
...
     out = np.frombuffer(stdout, dtype=np.double)
-    return np.reshape(out, coordinates.shape)
+    return np.reshape(out, (-1, 2))
+
+def project_xy(x, y, proj_string):
+    '''
+    Wrapper for project() that works with shapely.ops.transform().
+    '''
+    a = project(zip(x, y), proj_string)
+    return zip(*a)
```
QuLogic: Need `list()` around both of these `zip`s.

docs/plot/plot.py
```diff
@@ -83,55 +183,113 @@ def parallel(lat, lon_min, lon_max):
...
+            plt.plot(x, y, color=COLOR_COAST, linewidth=0.5)
+
     # Plot frame
+    frame = []
+    frame.append(parallel(plotdef['latmin'], plotdef['lonmin'], plotdef['lonmax']))
+    frame.append(parallel(plotdef['latmax'], plotdef['lonmin'], plotdef['lonmax']))
     for line in frame:
-        line = project(line, proj_str)
+        line = project(line, plotdef['projstring'])
         x, y = zip(*line)
```
QuLogic: `line` is an array now, I believe:
```python
x = line[:, 0]
y = line[:, 1]
```

docs/plot/plot.py
```diff
@@ -83,55 +183,113 @@ def parallel(lat, lon_min, lon_max):
...
+        plotdef['lonmax'],
+        plotdef['latmin'],
+        plotdef['latmax'],
+    )
 
     # Plot graticule
-    graticule = project(graticule, proj_str)
     for feature in graticule:
+        feature = project(feature, plotdef['projstring'])
         x, y = zip(*feature)
```
QuLogic: `feature` is an array now, I believe:
```python
x = feature[:, 0]
y = feature[:, 1]
```

docs/plot/plot.py
```diff
@@ -83,55 +183,113 @@ def parallel(lat, lon_min, lon_max):
...
-    plt.suptitle(proj_str, fontdict=font)
+    font = {
+        'family': 'serif',
+        'color': 'black',
+        'style': 'italic',
+        'size': 12
+    }
+    plt.suptitle(plotdef['projstring'], fontdict=font)
 
     plt.autoscale(tight=True)
```
QuLogic: `axes`

docs/plot/plot.py
```diff
@@ -83,55 +183,113 @@ def parallel(lat, lon_min, lon_max):
...
-    name = proj_str.split()[0].split('=')[1]
-    plt.savefig(outdir + '/' + name + '.png', dpi=300)
+    if not os.path.exists(outdir):
+        os.makedirs(outdir)
+    plt.savefig(outdir + '/' + plotdef['filename'], dpi=300)
```
QuLogic: `fig`

QuLogic (commented):
The medium GeoJSON is pretty heavy; if these files don't get rebuilt very often, maybe it would be useful to put them in a gist/separate repo and download at build time?

QuLogic:
@kbevers OK, I figured out why it didn't work; `PROJ_LIB` is not set and it tries to use the data out of `/usr` instead of the non-installed version, which are of course not consistent when using the built version.

I've opened a PR to your branch fixing that and applying the optimizations I noted above.

kbevers:
@QuLogic I've incorporated your suggested changes (squashed a few of the commits). 

I'll try to get the svg/sphinx setup working at some point, so for now I have discarded the png-files I initially added to this PR. I will investigate the SVG/Sphinx setup a bit further, and once I come to a conclusion I will make sure to add the updated plots in this PR.

QuLogic:
It doesn't look like you removed the first extra copy.

kbevers:
> It doesn't look like you removed the first extra copy.

You are correct. I thought I already had removed those. I'll try again.

Update: That's a task that is right on the edge of my git abilities. It's too late in the evening for me to do this correctly :)

hobu:
Does this work on the PDF output too?

kbevers:
I would image - they are still PNGs!

hobu:
> I would image - they are still PNGs!

oh, sorry. I thought you found a way to imagemagick them as SVGs for the website as PNGs for the latexpdf.

kbevers:
I tried many things, but never found a solid solution. In the end I think it came down to some weird problems with imagemagick that I couldn't solve. In theory it should be possible though!

PR #527, opened 2017-06-21
Correct calculation of meridian convergence for non-conformal projections

### LINKED ISSUE #526: Equations for meridian convergence and partial derivatives reviewed (corrected)
I have found strange results for meridian convergence in non-conformal projections. As we know from map projection books (with Gauss-Bomford convention for convergence), meridian convergence should be calculated by formula **fac->conv = - atan2(fac->der.x_p , fac->der.y_p)**. 

Current formula used in Proj4 is fac->conv = - atan2(fac->der.y_l , fac->der.x_l), which can be useful only for conformal projections (due to Cauchy-Riemann equations), and in fact should be without minus sign, i.e. fac->conv = atan2(fac->der.y_l , fac->der.x_l). For non-conformal projections we now get angle between parallel and x-axis, instead of meridian convergence.

Minus sign currently used in formula comes from fact that fac->der.x_p and fac->der.y_l are calculated with opposite signs.

Corrected equations are:

**pj_deriv.c**

**Line 14**
instead of:
der->x_l = t.x; der->y_p = t.y; der->x_p = -t.x; der->y_l = -t.y;
should be:
der->x_l = t.x; der->y_p = t.y; **der->x_p = t.x; der->y_l = t.y;**

**Line 19**
instead of:
der->x_l += t.x; der->y_p -= t.y; der->x_p += t.x; der->y_l -= t.y;
should be:
der->x_l += t.x; der->y_p -= t.y; **der->x_p -= t.x; der->y_l += t.y;**

**Line 23**
instead of:
der->x_l -= t.x; der->y_p -= t.y; der->x_p += t.x; der->y_l += t.y;
should be:
der->x_l -= t.x; der->y_p -= t.y; **der->x_p -= t.x; der->y_l -= t.y;**

**Line 27**
instead of:
der->x_l -= t.x; der->y_p += t.y; der->x_p -= t.x; der->y_l += t.y;
should be:
der->x_l -= t.x; der->y_p += t.y; **der->x_p += t.x; der->y_l -= t.y;**

**pj_factors.c**

**Line 72**
instead of:
fac->conv = - atan2(fac->der.y_l, fac->der.x_l);
should be:
**fac->conv = - atan2(fac->der.x_p, fac->der.y_p);**

One can question the practical use of meridian convergence in non-conformal projections, but nevertheless, it would be nice to have it calculated correctly.

kbevers:
Resolves #526

### DIFF
--- a/src/pj_factors.c
+++ b/src/pj_factors.c
@@ -2,90 +2,96 @@
 #define PJ_LIB__
 #include <projects.h>
 #include <errno.h>
+
 #ifndef DEFAULT_H
 #define DEFAULT_H   1e-5    /* radian default for numeric h */
 #endif
+
 #define EPS 1.0e-12
-	int
-pj_factors(LP lp, PJ *P, double h, struct FACTORS *fac) {
-	struct DERIVS der;
-	double cosphi, t, n, r;
 
-        der.x_l = 0.0;
-        der.x_p = 0.0;
-        der.y_l = 0.0;
-        der.y_p = 0.0;
+int pj_factors(LP lp, PJ *P, double h, struct FACTORS *fac) {
+    struct DERIVS der;
+    double cosphi, t, n, r;
 
-	/* check for forward and latitude or longitude overange */
-	if ((t = fabs(lp.phi)-M_HALFPI) > EPS || fabs(lp.lam) > 10.) {
-                pj_ctx_set_errno( P->ctx, -14);
-		return 1;
-	} else { /* proceed */
-		errno = pj_errno = 0;
+    der.x_l = 0.0;
+    der.x_p = 0.0;
+    der.y_l = 0.0;
+    der.y_p = 0.0;
+
+    /* check for forward and latitude or longitude overange */
+    if ((t = fabs(lp.phi)-M_HALFPI) > EPS || fabs(lp.lam) > 10.) {
+        pj_ctx_set_errno( P->ctx, -14);
+        return 1;
+    } else { /* proceed */
+        errno = pj_errno = 0;
                 P->ctx->last_errno = 0;
 
-		if (h < EPS)
-			h = DEFAULT_H;
-		if (fabs(lp.phi) > (M_HALFPI - h))
-                /* adjust to value around pi/2 where derived still exists*/
-		        lp.phi = lp.phi < 0. ? (-M_HALFPI+h) : (M_HALFPI-h);
-		else if (P->geoc)
-			lp.phi = atan(P->rone_es * tan(lp.phi));
-		lp.lam -= P->lam0;	/* compute del lp.lam */
-		if (!P->over)
-			lp.lam = adjlon(lp.lam); /* adjust del longitude */
-		if (P->spc)	/* get what projection analytic values */
-			P->spc(lp, P, fac);
-		if (((fac->code & (IS_ANAL_XL_YL+IS_ANAL_XP_YP)) !=
-			  (IS_ANAL_XL_YL+IS_ANAL_XP_YP)) &&
-			  pj_deriv(lp, h, P, &der))
-			return 1;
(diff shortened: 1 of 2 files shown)


src/pj_factors.c
```diff
@@ -72,7 +72,7 @@ int pj_factors(LP lp, PJ *P, double h, struct FACTORS *fac) {
 
         /* convergence */
         if (!(fac->code & IS_ANAL_CONV)) {
-            fac->conv = - atan2(fac->der.y_l, fac->der.x_l);
+            fac->conv = - atan2(fac->der.x_p, fac->der.y_p);
```
rouault: I'm not sure why I'm looking at that as I've absolutely no background on this :-) , but this change looks surprising since atan2 arguments are documented to be in y,x order.

kbevers: My knowledge on this is also limited, just wanted to test it out. @dtutic can you elaborate a bit on this?

dtutic: Please find attached the discussion and equations for meridian convergence.
[Meridian convergence in map projections.pdf](https://github.com/OSGeo/proj.4/files/1096686/Meridian.convergence.in.map.projections.pdf)
.

kbevers: Thanks, I think I get it now. I guess this implies a change in convention? Are the old an new calculations equivalent for conformal projections? In that case I think it is safe to merge. If not, we should consider adding this as a seperate derivative measure of coordinates.

Did you write this your self? Can we include it in the documentation?

dtutic: I write it myself with a help of colleague who is a mathematician. You can include it in documentation, but it can be shortened or expanded depending on the purpose of documentation. The definition of meridian convergence is taken from [https://www.ngs.noaa.gov/CORS/Articles/SolerFuryASCE.pdf] , which is again taken from Geodetic Glossary by National Geodetic Survey (U.S.). The derivation of equations is an application of mathematical analysis, and it can be written in even simpler form.

The convention is the same (positive when meridian is west of grid north) , and calculations are same for conformal projections because the partial derivatives are also corrected to all have same sign convention which was not the case before, and was a reason for extra minus in previous formula.

kbevers: Thanks. Both for taking time to explain and for letting us use your text in the docs.

kbevers:
I am now convinced that this is a good idea to merge. I would like a second opinion though, just to be on the safe side. @busstoptaktik, @cffk ?

cffk:
My preferred "definition" of meridian convergence is that it is the
bearing of grid north given as degrees clockwise (the "conventional"
sense) from true north.  This is consistent with @dtutic's definition.
Needless to say, it's always a good idea to do sanity checks measuring
the meridian convergence by computing the forward (resp. reverse)
projection of two points on the same meridian (resp. on x = const).

dtutic:
Equation for meridian convergence in LCC projection seems to give correct sign only when standard parallels are on south hemisphere. I guess that the reason is that Q->n is always positive between 0 and 1 (ratio of angle between meridians in projection and on sphere/ellipsoid). 

It is now calculated in PJ_lcc.c with:
**fac->conv = - Q->n * lp.lam;**

It should change sign depending on standard parallels,  easy to see when changing standard parallel on this animation (http://www.kartografija.hr/tl_files/konusne_final_en.xml). I will try to figure it out. There is already and issue reporting this problem [https://github.com/OSGeo/proj.4/issues/16](url)

kbevers:
@dtutic have you had a chance to look at the meridian convergence problem in LCC?

dtutic:
Not yet. Hope to do it by end of the next week.


________________________________

kbevers:
> Not yet. Hope to do it by end of the next week.

Okay, then I'll merge this one now. You are welcome to add your findings to the existing ticket (#16) when you've had a closer look. Thanks for your help.

PR #534, opened 2017-07-03
Let pj_init() expand more than one +init argument.

### LINKED ISSUE #479: PJ_pipeline doesn't expand +init arguments when using +inv
When using ```pj_create``` from the new API this doesn't work:
```
+proj=pipeline 
+step +inv +init=epsg:25832
+step +init=epsg:25833
```
while this does:
```
+proj=pipeline 
+step +inv +proj=utm +zone=32 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs
+step +init=epsg:25833"
```

```+init``` arguments need to be expanded when ```+inv``` is used.

### PR DESCRIPTION
**Pull request by kbevers**
With the introduction of transformation pipelines it is now a valid
use case to have more than one +init argument in a proj string. 
This makes sure that all +init's are expanded.

Fixes #479.

### DIFF
--- a/src/PJ_pipeline.c
+++ b/src/PJ_pipeline.c
@@ -610,6 +610,27 @@ int pj_pipeline_selftest (void) {
     if (fabs (110574.388554153 - dist) > 1e-4)
         return 4002;
 
+
+    /* test a pipeline with several +init steps */
+    P = pj_create(
+        "+proj=pipeline "
+        "+step +init=epsg:25832 +inv "
+        "+step +init=epsg:25833  "
+        "+step +init=epsg:25833 +inv "
+        "+step +init=epsg:25832 "
+    );
+    if (0==P)
+        return 5000;
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rouault** [comment]
Probably deserves a test case ?

**Comment by kbevers** [comment]
> Probably deserves a test case ?

Good point. I'll add one to the pipeline driver.

**Comment by kbevers** [comment]
The new code fails when a proj-string only consists of a `+init` argument. Will fix it tomorrow.

```
$ echo 16.33 48.20 |./cs2cs +proj=latlong +datum=WGS84 +to +init=epsg:31284
Using to definition: init=epsg:31284
Rel. 4.9.3, 15 August 2016
<cs2cs>:
cause: projection not named
program abnormally terminated

$ echo 16.33 48.20 |./cs2cs +proj=latlong +datum=WGS84 +to +noop +init=epsg:31284
595710.37	5357598.46 -44.50
```

PR #546, opened 2017-07-14
Expose pj_factors and pj_deriv in public API.

### LINKED ISSUE #98: move pj_factors() into proj_api.h
**Reported by hamish on 22 Feb 2011 08:18 UTC**
Hi,

wrt OSGeo4W bug # 34 (https://trac.osgeo.org/osgeo4w/ticket/34), pj_factors() is currently housed in projects.h which is apparently supposed to be a private header file.

If projects.h goes away, we need a public way of getting at pj_factors(), or an alternate way for end-user apps to get at the convergence angle value. As an example, the code in question is GRASS's g.region/printwindow.c:

https://trac.osgeo.org/grass/browser/grass/trunk/general/g.region/printwindow.c#L586

FWIW the method I used to get at fact.conv was directly inspired by the code for the `proj -v` executable.

to avoid any gaps it would be best if this could happen before projects.h was removed from the install.

thanks,
Hamish

Migrated-From: https://trac.osgeo.org/proj/ticket/98

kbevers:
`proj_derivatives_retrieve()` and `proj_factors_retrieve()` has been
introduced to the public `proj.h` API. The new functions are simply
wrappers for the original functions. The h argument of the two functions
has been omitted in the new wrappers as it does not seem to be used very
much (if at all?). The wrapper functions allocates a new struct
themselves which users are expected to free again with `proj_release()`.

Fixes #98.

I have followed the retrieve/release pattern introduced with `proj_definition_retrieve()`. Potentially it results in many allocations/deallocations, which perhaps is unwanted? Another solution is to let the user allocate and deallocate the `PJ_FACTORS`/`PJ_DERIVS` structs themselves. That is how `pj_factors()` and `pj_deriv()` works and might be preferable if derivatives or factors are calculated for many points. Or it could simply just return the struct like `PJ_FACTORS proj_factors(PJ *P, LP lp)` with a PJ_FACTORS struct being allocated on the stack and then returned.

Comments on this are appreciated!

### DIFF
--- a/src/PJ_cart.c
+++ b/src/PJ_cart.c
@@ -236,8 +236,8 @@ int pj_cart_selftest (void) {
     PJ *P;
     PJ_OBS a, b, obs[2];
     PJ_COORD coord[2];
-    PJ_DERIVS *derivs;
-    PJ_FACTORS *factors;
+    PJ_DERIVS derivs;
+    PJ_FACTORS factors;
     int err;
     size_t n, sz;
     double dist, h, t;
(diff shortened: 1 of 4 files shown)


kbevers:
After having thought about this for a while, I decided to change the function to just return the structs instead of pointers. I think it is better this way since no memory need to be freed. Changed the function names to `proj_derivatives` and `proj_factors`.

Issue #538, opened 2017-07-05
Add a proj_get_def() to proj.h


kbevers:
@rouault mentions in https://github.com/OSGeo/proj.4/issues/529 that QGIS and GDAL uses `pj_get_def()`. Smilar functionality should be available in the new `proj.h` API. It is just an alias for `pj_get_def()`.

Issue #300, opened 2015-09-07
Add get function for searchpath (trying to protect against repeats of the proj_def.dat disaster)


rsbivand:
Please add

```
/************************************************************************/
/*                            pj_get_searchpath_count()                       */
/************************************************************************/

int pj_get_searchpath_count(void)
{
    return(path_count);
}

/************************************************************************/
/*                            pj_get_searchpath()                       */
/************************************************************************/

char *pj_get_searchpath(int i)
{
    if (path_count > 0) return(search_path[i]);
    else return(NULL);
}
```

to src/pj_open_lib.c and

```
int pj_get_searchpath_count(void);
char *pj_get_searchpath(int);
```

to src/proj_api.h to provide a way of checking that the searchpath is set. 

Are there threading issues with the searchpath? In:

```
#include <stdio.h>
#include <string.h>
#include <proj_api.h>

void main(void) {
  PAFile fp;
  projCtx ctx;
  projPJ test;
  char s[255];
  char ss[255];
  const char **paths;
  strcpy(ss, "+proj=utm +zone=32");
  test = pj_init_plus(ss);
  if (!test) printf("Step 1: %s\n", pj_strerrno(*pj_get_errno_ref()));
  else {
    pj_free(test);
    exit(1);
  }
  ctx = pj_get_default_ctx();
  fp = pj_open_lib(ctx, "proj_def.dat", "rb");
  if (fp == NULL) printf("Step 2: no proj_def.dat found\n");
  paths = (const char **) malloc(sizeof(const char **));
  strcpy(s, "/tmp");
  paths[0] = s;
  printf("Step 3: set searchpath to %s\n", paths[0]);
  pj_set_searchpath(1, paths);
  free(paths);
  for (int i=0; i<pj_get_searchpath_count(); i++)
    printf("Step 4: path[%d]: %s\n", i, pj_get_searchpath((i)));
  fp = pj_open_lib(ctx, "proj_def.dat", "rb");
  if (fp == NULL) {
    printf("Step 5: no proj_def.dat found\n");
  } else {
    printf("Step 5: proj_def.dat found\n");
    pj_ctx_fclose(ctx, fp);
  }
  test = pj_init_plus(ss);
  if (!test) {
    printf("Step 6: %s\n", pj_strerrno(*pj_get_errno_ref()));
  } else {
    printf("Step 6: %s\n", pj_get_def(test, 0));
    pj_free(test);
  }
  exit(0);
}
```

[...]

PR #551, opened 2017-07-24
Adding info functions to proj.h API.

### LINKED ISSUE #471: Add public API method to query active EPSG db version
Add something like  `char* pj_get_epsg_version()` to return a string representing the currently active EPSG db.

kbevers:
Four new functions are added with this commit: `proj_info()`, `proj_pj_info()`,
`proj_grid_info()` and `proj_init_info()`. Additionally four new data types are
added: `PJ_INFO`, `PJ_PROJ_INFO`, `PJ_GRID_INFO` and `PJ_INIT_INFO`. The functions
return the corresponding data types.

These functions allows users of the PROJ.4 library to get information about
various PROJ.4 entities and the library itself. The new data types are structs
that contain specific information about either the library instance, a `PJ`
instance, a grid or an init file. Together the four new functions cover a big
part of the functionality in the semi-public `projects.h` API and should hopefully
make it easier for user to migrate their code to the `proj.h` API in the future.

Besides covering already existing functionality in the old API, this commit
introduces the ability to add metadata to init-files. This is primarily added
to give users a way of knowing which version of the EPSG database they are
using, but it also comes in handy for other init-files. The init-file metadata
is added directly to the init-file as a special "projection" called "metadata".
The info projection of the epsg init-file is thus described as:

    <metadata> +version=9.0.0 +origin=EPSG +lastupdate=2017-01-10

The `proj_init_info()` function uses the internal `pj_param()` to read the
metadata. As a consequence, "metadata" will not be available as a the name of
a projection in the future. This is a reasonable price to pay considering the
ease of the implementation of adding metadata to init-files this way, and of
course that "metadata" is a very unlikely name for a projection in any case.
A metadata tag has been added to all init-files in the nad-directory. For most
only a subset of the possible parameters has been added.

This PR removes a few functions that was recently introduced to the `proj.h` API. Namely `proj_has_inverse()`, `proj_definition_retrieve()` and `proj_release()` since they provide the
same functionality as `proj_pj_info()`. Except for `proj_release()`, but with `proj_definition_retrieve()` gone, it no longer has a purpose.

I think this way of providing information for the users of PROJ.4 is a better than having
a function for each specific piece of info a users might want. By using structs with fixed
length strings memory handling issues should be at a minimum, although at the cost of
a slightly larger memory consumption.

This PR resolves #471, #300 and #538 (which was already fixed, but now handled differently).

------------------------------------------------------------------------------

To demonstrate the info functions I have written a simple program:

```
#include <stdio.h>
#include <proj.h>

int main(void) {
    /*PJ *P = proj_create(0, "+proj=pipeline +step +init=epsg:25833"); */
    PJ *P = proj_create(0, "+proj=merc +lat_0=24 +lon_0=53");
    PJ_GRID_INFO grid_info = proj_grid_info("BETA2007.gsb");

    printf("\n");
    printf("                            proj_info()\n");
    printf("--------------------------------------------------------------------\n");
    printf("Major:       %d\n", proj_info().major);
    printf("Minor:       %d\n", proj_info().minor);
    printf("Patch:       %d\n", proj_info().patch);
    printf("             %s\n", proj_info().release);
    printf("version:     %s\n", proj_info().version);
    printf("Search path: %s\n", proj_info().searchpath);
    printf("\n");
    printf("                            proj_pj_info()\n");
    printf("--------------------------------------------------------------------\n");
    printf("Projection id: %s\n", proj_pj_info(P).id);
    printf("Description:   %s\n", proj_pj_info(P).description);
    printf("Definition:    %s\n", proj_pj_info(P).definition);
    printf("Has inverse:   %d\n", proj_pj_info(P).has_inverse);
    printf("\n");
    printf("                            proj_grid_info()\n");
    printf("--------------------------------------------------------------------\n");
    printf("Grid name:    %s\n", grid_info.gridname);
    printf("File name:    %s\n", grid_info.filename);
    printf("Format:       %s\n", grid_info.format);
    printf("Lower left:   (%f, %f)\n", proj_todeg(grid_info.lowerleft.lam),
                                       proj_todeg(grid_info.lowerleft.phi));
    printf("Upper right:  (%f, %f)\n", proj_todeg(grid_info.upperright.lam),
                                       proj_todeg(grid_info.upperright.phi));
    printf("Grid size:    (%d, %d)\n", grid_info.n_lon, grid_info.n_lat);
    printf("\n");
    printf("                            proj_init_info()\n");
    printf("--------------------------------------------------------------------\n");
    printf("Init name:   %s\n", proj_init_info("epsg").name);
    printf("File name:   %s\n", proj_init_info("epsg").filename);
    printf("Version:     %s\n", proj_init_info("epsg").version);
    printf("Origin:      %s\n", proj_init_info("epsg").origin);
    printf("\n");

    proj_destroy(P);
    return 0;
}
```

Which on my computer generates the following output:

```
                            proj_info()
--------------------------------------------------------------------
Major:       10
Minor:       0
Patch:       0
             Rel. 4.9.3, 15 August 2016
version:     10.0.0
Search path: C:\Users\b012349;C:\OSGeo4W64\\share\proj

                            proj_pj_info()
--------------------------------------------------------------------
Projection id: merc
Description:   Mercator
        Cyl, Sph&Ell
        lat_ts=
Definition:    +proj=merc +lat_0=24 +lon_0=53 +ellps=WGS84
Has inverse:   1

                            proj_grid_info()
--------------------------------------------------------------------
Grid name:    BETA2007.gsb
File name:    C:\OSGeo4W64\\share\proj\BETA2007.gsb
Format:       ntv2
Lower left:   (5.500000, 47.000000)
Upper right:  (15.833333, 55.400000)
Grid size:    (62, 84)

                            proj_init_info()
--------------------------------------------------------------------
Init name:   epsg
File name:   C:\OSGeo4W64\\share\proj\epsg
Version:     9.0.0
Origin:      EPSG
```

### DIFF
--- a/nad/epsg
+++ b/nad/epsg
@@ -1,3 +1,4 @@
+<metadata> +version=9.0.0 +origin=EPSG +lastupdate=2017-01-10
 # HD1909
 <3819> +proj=longlat +ellps=bessel +towgs84=595.48,121.69,515.35,4.115,-2.9383,0.853,-3.408 +no_defs  <>
 # TWD67
--- a/src/pj_obs_api.c
+++ b/src/pj_obs_api.c
@@ -510,25 +510,198 @@ void proj_context_destroy (PJ_CONTEXT *ctx) {
 }
 
 
-/* Build a fully expanded proj_create() compatible representation of P */
-char *proj_definition_retrieve (PJ *P) {
-    if (0==P)
-        return 0;
-    return pj_get_def(P, 0);
+/*****************************************************************************/
+PJ_INFO proj_info(void) {
+/******************************************************************************
+    Basic info about the current instance of the PROJ.4 library.
+
+    Returns PJ_INFO struct. Searchpath member of the struct is truncated to 512
+    characters.
+
+******************************************************************************/
+    PJ_INFO info;
+    const char **paths;
+    char *tmpstr;
+    int i, n;
+
+    info.major = PROJ_VERSION_MAJOR;
+    info.minor = PROJ_VERSION_MINOR;
+    info.patch = PROJ_VERSION_PATCH;
+    sprintf(info.version, "%d.%d.%d", info.major, info.minor, info.patch);
+    strcpy(info.release, pj_get_release());
+
+    paths = proj_get_searchpath();
+    n = proj_get_path_count();
+    /* build search path string */
+    tmpstr = getenv("HOME");
+    if (tmpstr != NULL) {
+        sprintf(info.searchpath, "%s", tmpstr);
+    }
+
+    tmpstr = getenv("PROJ_LIB");
+    if (info.searchpath != NULL) {
+        /* $HOME already in path */
+        if (tmpstr != NULL)
+            sprintf(info.searchpath, "%s;%s", info.searchpath, tmpstr);
+    } else {
+        /* path is empty */
+        if (tmpstr != NULL)
+            sprintf(info.searchpath, "%s", getenv("HOME"));
+    }
+
+    for (i=0; i<n; i++) {
+        if (strlen(info.searchpath)+strlen(paths[i]) >= 512)
+            continue;
+
+        if (info.searchpath != NULL)
+            sprintf(info.searchpath, "%s;%s", info.searchpath, paths[i]);
+        else
+            sprintf(info.searchpath, "%s", paths[i]);
+    }
+
+    return info;
 }
 
-/* ...and get rid of it safely */
-void *proj_release (void *buffer) {
-    return pj_dealloc (buffer);
+/*****************************************************************************/
+PJ_PROJ_INFO proj_pj_info(const PJ *P) {
+/******************************************************************************
+    Basic info about a particular instance of a projection object.
+
+    Returns PJ_PROJ_INFO struct.
+
+******************************************************************************/
+    PJ_PROJ_INFO info;
+    char *def;
+
+    /* projection id */
+    if (pj_param(P->ctx, P->params, "tproj").i)
+        strncpy(info.id, pj_param(P->ctx, P->params, "sproj").s, sizeof(info.id));
+
+    /* projection description */
+    strncpy(info.description, P->descr, sizeof(info.description));
+
+    /* projection definition */
+    def = pj_get_def((PJ *)P, 0); /* pj_get_def takes a non-const PJ pointer */
+    strncpy(info.definition, &def[1], sizeof(info.definition)); /* def includes a leading space */
+    pj_dealloc(def);
+
+    /* this does not take into account that a pipeline potentially does not */
+    /* have an inverse.                                                     */
+    info.has_inverse = (P->inv != 0 || P->inv3d != 0 || P->invobs != 0);
+
+    return info;
+}
+
+
+/*****************************************************************************/
+PJ_GRID_INFO proj_grid_info(const char *gridname) {
+/******************************************************************************
+    Information about a named datum grid.
+
+    Returns PJ_GRID_INFO struct.
+
+******************************************************************************/
+    PJ_GRID_INFO info;
+
+    /*PJ_CONTEXT *ctx = proj_context_create(); */
+    PJ_CONTEXT *ctx = pj_get_default_ctx();
+    PJ_GRIDINFO *gridinfo = pj_gridinfo_init(ctx, gridname);
+    memset(&info, 0, sizeof(PJ_GRID_INFO));
+
+    /* in case the grid wasn't found */
+    if (gridinfo->filename == NULL) {
+        pj_gridinfo_free(ctx, gridinfo);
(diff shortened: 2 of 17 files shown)


nad/epsg
```diff
@@ -1,3 +1,4 @@
+<metadata> +version=9.0.0 +origin=EPSG +lastupdate=2017-01-10
```
rouault, line 1: This change will be lost at the next upgrade of EPSG database. Perhaps https://github.com/OSGeo/gdal/blob/trunk/gdal/swig/python/scripts/epsg_tr.py should be altered to add it. But it is called several times as mentionned towards the end of https://trac.osgeo.org/geotiff/browser/trunk/libgeotiff/csv/README . Or perhaps just this README file should be modified to mentionn adding this header line

src/pj_obs_api.c
```diff
@@ -510,25 +510,198 @@ void proj_context_destroy (PJ_CONTEXT *ctx) {
...
+
+    /* in case the grid wasn't found */
+    if (gridinfo->filename == NULL) {
+        pj_gridinfo_free(ctx, gridinfo);
+        strcpy(info.format, "missing");
+        return info;
+    }
+
+    /* name of grid */
+    strncpy(info.gridname, gridname, sizeof(info.gridname));
```
rouault: Caution (here and in all other places where strncpy() is used). This will not zero terminate the buffer if the source argument length is >= sizeof(ret_value). I'd suggest using a proj_strlcpy() that would call strncpy() and force zero termination.

src/pj_obs_api.c
```diff
@@ -510,25 +510,198 @@ void proj_context_destroy (PJ_CONTEXT *ctx) {
...
+    info.patch = PROJ_VERSION_PATCH;
+    sprintf(info.version, "%d.%d.%d", info.major, info.minor, info.patch);
+    strcpy(info.release, pj_get_release());
+
+    paths = proj_get_searchpath();
+    n = proj_get_path_count();
+    /* build search path string */
+    tmpstr = getenv("HOME");
+    if (tmpstr != NULL) {
+        sprintf(info.searchpath, "%s", tmpstr);
```
rouault: Caution (here in other call places): sprintf() can overflow the output buffer

src/proj.h
```diff
@@ -269,6 +282,42 @@ struct PJ_OBS {
...
+struct PJ_PROJ_INFO {
+    char        id[16];             /* Name of the projection in question           */
+    char        description[128];   /* Description of the projection                */
+    char        definition[512];    /* Projection definition                        */
+    int         has_inverse;        /* 1 if an inverse mapping exists, 0 otherwise  */
+};
+
+struct PJ_GRID_INFO {
+    char        gridname[32];       /* name of grid                         */
+    char        filename[128];      /* full path to grid                    */
```
rouault: Wondering if 128 is this big enough

nad/epsg
kbevers: It is probably easiest to just to do it manually, although it would be cool to have the script to it automatically.

What do you think about this way of providing the version number of the EPSG database?

src/proj.h
kbevers: So did I. I'll change it to something higher if you think it is needed.

I guess it will always be an issue with fixed length strings. Perhaps an errno should be set in case the path is too long to store in the available space?

src/pj_obs_api.c
kbevers: Good point. I was almost certain that I would do this in an unsafe way. I'll find a way to avoid the problem.

src/pj_obs_api.c
kbevers: Good point again. I guess `snprintf` would solve the problem... I'll put in a safe-guard that avoid overflows.

rouault: Beware of snprintf() on Windows : it is not available before _MSC_VER >= 1900. Previously _snprintf() is available, but not POSIX compliant. In particular _snprintf() won't zero terminate

kbevers: I don't think it is ANSI C either, so I won't use it. I think I can get around the calls to sprintf in other ways.

src/proj.h
QuLogic: On Windows, `MAX_PATH` is 260, so it should be at least that long, but this limit only applies to old-style file paths. NT-style paths are limited to 32K bytes...

kbevers: I'll change the length of filename to `MAX_PATH` in both `PJ_INIT_INFO` and `PJ_GRID_INFO`. That should be enough for all but the ridiculous cases. I'll just do it as `char filename[260]` since I don't want to clutter the header too much with a bunch of #defines at this time.

In `PJ_INFO` the searchpath is 512 at the moment. I think it is highly unlikely that it will be maxed out, but of course it could be changed as well. Maybe to 1024? Opinions?

PR #555, opened 2017-08-06
Add tests for C99 math functions, hypot, atanh, cbrt, etc.

cffk:
This tests whether the functions are declared in <math.h>.  If they are,
then -DHAVE_C99_MATH=1 is added to the C flags.  The intention is that
this flag is only seen when building proj.4 and shouldn't be referenced
in any of the installed include files.  The next update to geodesic.c
will use this flag.

Left unaddressed is what to do if HAVE_C99_MATH is 0.  The strategy in
geodesic.c is to assume that the missing functions will need to be
defined explicitly.  A less safe alternative is to assume that the
functions are in fact available in libm and that all that needs to be
done is to declare the functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,28 @@ if (HAVE_PTHREAD_MUTEX_RECURSIVE_DEFN)
     add_definitions(-DHAVE_PTHREAD_MUTEX_RECURSIVE=1)
 endif()
 
+include (CheckCSourceCompiles)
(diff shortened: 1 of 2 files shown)


cffk:
I suppose an additional test could be made where math.h is missing these functions to see if libm has them defined nevertheless.  Probably it would be best to formulate a general strategy first.  Presumably at some point we can declare that proj.4 requires a C99 (or later) compliant compiler.

PR #562, opened 2017-08-29
Release candidate for geodesic library version 1.49.

cffk:
Only substantial changes are (1) testing the HAVE_C99_MATH flag and
acting accordingly and (2) adding a couple of tests.


cffk:
What's the mechanism for getting code included in the coverage tests?  geodtest.c is run by cmake's make test and exercises most of geodesic.c.  Presumably there's some magic that gets this run by the coverage tests.

kbevers:
[These](https://github.com/OSGeo/proj.4/blob/master/travis/install.sh#L75-L80) are the relevant lines in the Travis setup script. Is the geod test only included in the cmake test? If so, that might be the reason.

Issue #559, opened 2017-08-23
proj_create_crs_to_crs(): need for area of use ?


rouault:
Jumping from QGIS discussion https://github.com/qgis/QGIS-Enhancement-Proposals/issues/100 to here, I'm wondering if proj_create_crs_to_crs() shouldn't have a parameter to express the area of use, so that we can select a better transform from the EPSG database when we will consult it. Or perhaps this will be a extented  proj_create_crs_to_crs_ex() version that will do that ?

kbevers:
I can see the value in such functionality but I am not sure how feasible it is to implement in practice. For instance, transformations from ITRF20xx to ETRS89 are defined differently in Scandinavia than in, say, Germany. As far as I am aware this sort of level of detail is not available in the EPSG-dataset. Of course it might be in the future. 

This example is maybe a bit esoteric - are there better and more plausible scenarios where an area of interest is needed for the correct transformation? 

How would you change the function prototype in practice? 
 
```PJ  *proj_create_crs_to_crs (PJ_CONTEXT *ctx, const char *srid_from, const char *srid_to, double lat_min, double_latmax, double lon_min, double lon_max)```

perhaps? Of course there's the issue of what the bounding box coordinates refer to, but I guess it is okay as long as they are in roughly the desired area.

**Comment by rouault** [comment]
> As far as I am aware this sort of level of detail is not available in the EPSG-dataset

It is. The EPSG database proposes different transformations for the same (source_crs, target_crs) tuple, based on the area of use. Here for Pulkovo 42 to WGS84, you can see different transformation per country (typically different towgs84 parameters, see https://www.epsg-registry.org/export.htm?wkt=urn:ogc:def:coordinateOperation:EPSG::1645 and https://www.epsg-registry.org/export.htm?wkt=urn:ogc:def:coordinateOperation:EPSG::15496 )

```
$ ogrinfo pg:dbname=epsg -sql "select coord_op_code, coord_op_name, area_of_use_code, a.area_name, a.area_south_bound_lat, a.area_north_bound_lat, a.area_west_bound_lon, a.area_east_bound_lon from epsg_coordoperation c join epsg_area a on c.area_of_use_code = a.area_code where c.source_crs_code = 4179 and c.target_crs_code = 4326" -q

Layer name: sql_statement
OGRFeature(sql_statement):0
  coord_op_code (Integer) = 1645
  coord_op_name (String) = Pulkovo 1942(58) to WGS 84 (1)
[...]

PR #565, opened 2017-09-06
Addition of an "area of use" argument to proj_transform_crs_to_crs

kbevers:
In #559 we discussed adding an "area of use" argument to the function `proj_transform_crs_to_crs`, in order to, in the future, be able to select the best possible transformation when doing lookup in the EPSG dataset.

This PR adds a `PJ_AREA` data type and the additional `area` argument to the transform function. I've added the extra argument in anticipation of changes in the future that will likely need the area description. By adding it now we can hopefully avoid adding an extra function like `proj_transform_crs_to_crs_area` in the future.


src/pj_obs_api.c
```diff
@@ -384,14 +384,23 @@ PJ  *proj_create_crs_to_crs (PJ_CONTEXT *ctx, const char *srid_from, const char
     function can be extended to support "late-binding" transformations in the
     future without affecting users of the function.
 
+    An "area of use" can be specified in area. In the current version of this
+    function is has no function, but is added in anticipation of a
+    "late-binding" implementation in the future. The idea being, that if a user
+    supplies an area of use, the correct transformatin between to given systems
```
rouault: "the correct transformation" (typo: missing o at end of transformation): there might be several candidates. So I'd say "a more accurate transformation"

rouault (approved):

busstoptaktik (changes requested):
Adding a BBOX is not enough to make things work: For example, the BBOXes for Denmark and Sweden overlap - a large part of southern Sweden is situated inside the BBOX surrounding Denmark, due to the Danish islands in the Baltic Sea.

However, the transformations needed for Denmark and Sweden are different: Even though the Danish geoid model overlaps southern Sweden, it is not correct to use it in Swedish jurisdictions, since the height systems differ. Also, due to differences in the national implementations of ETRS89, the correct 7 parameter shift cannot be selected from BBOX alone.

Hence, we need a more explicit way of stating what’s to be done. This is not easy, and I think we need more discussion than can be fit into a code review.

But one way to manage this at the API level, for now, could be to let the argument be more opaque, not making it hint directly at BBOXitude, define it in projects.h, and provide a forward declaration and typedef only, in proj.h.

Unfortunately, geodesy is not just geophysics - it is also a product of local and national borders and historical decisions. This complicates things immensely.

kbevers:
I guess this boils down to level of ambition. My initial goal was to just return the best guess for a given transformation, but I agree that we can do better than that. It's probably not going to be easy :-)

> But one way to manage this at the API level, for now, could be to let the argument be more opaque, not making it hint directly at BBOXitude, define it in projects.h, and provide a forward declaration and typedef only, in proj.h.

I'll try to come up with a suggestion for a way to do this in practice. A quick browse of the EPSG database reveals that there is a handfull of ways to search for an area: area name ("Denmark"), bounding box and a few different ISO area codes ("DK", "DNK"). Not all of those are particularly easy to search for, so perhaps a method of providing the different options for transformation setup between to CRS's is needed.

> This is not easy, and I think we need more discussion than can be fit into a code review.

Let's try to keep the discussion in the original issue then.

kbevers:
Okay, here is a different approach to the area of use problem. The opaque struct `PJ_AREA` is introduced. Alongside it comes two setter-functions, `proj_area_bbox()` and `proj_area_description()`. The bounding box can be set in the first and a descriptive area identifier in the latter. Once values have been assigned to a `PJ_AREA` it can be used with `proj_create_crs_to_crs()`.  `proj_create_crs_to_crs()` should have the ability to assign an area ID from the EPSG database to a `PJ_AREA` based on the values in the `PJ_AREA` struct. This way, if the `PJ_AREA` struct is reused it will not be necessary to do the (potential) expensive database lookup again.

The text description of an area can be 2- and 3-letter country codes as well as descriptors suchs as "Europe" and "Australia".

I think this covers everything we've discussed so far while still being fairly compact API wise. Thoughts?

src/pj_obs_api.c
```diff
@@ -387,8 +387,8 @@ PJ  *proj_create_crs_to_crs (PJ_CONTEXT *ctx, const char *srid_from, const char
     An "area of use" can be specified in area. In the current version of this
     function is has no function, but is added in anticipation of a
     "late-binding" implementation in the future. The idea being, that if a user
-    supplies an area of use, the correct transformatin between to given systems
-    can be chosen.
+    supplies an area of use, the more accurate transformation between to given
```
rouault: typo: between *two* given

kbevers:
I feel like this is a good addition to the new API and would like to merge. @rouault @busstoptaktik Any objections?

rouault:
No objection

busstoptaktik:
I think it's a nice step in a fruitful direction, carefully designed to keep a range of options open, so no objections here...

busstoptaktik (approved):

kbevers:
Perfect. As it stands now, I have added a few prototypes for setting the `PJ_AREA` object, but have not implemented them yet since the area-functionality is not there yet. Would it better to leave the prototypes out until actual functionality is in place? Maybe we don't want to promise too much just yet?

Issue #568, opened 2017-09-12
Always true tests in PJ_cart.c


rouault:
```
PJ_cart.c:489:14: warning: address of array 'info.version' will always evaluate to 'true' [-Wpointer-bool-conversion]
    if (info.version) {
    ~~  ~~~~~^~~~~~~
PJ_cart.c:494:15: warning: address of array 'info.release' will always evaluate to 'true' [-Wpointer-bool-conversion]
    if (!info.release)    return 56;
        ~~~~~~^~~~~~~
PJ_cart.c:495:15: warning: address of array 'info.searchpath' will always evaluate to 'true' [-Wpointer-bool-conversion]
    if (!info.searchpath) return 57;
        ~~~~~~^~~~~~~~~~
```

kbevers:
Yeah, I have noticed those. If I remeber correctly I struggled a bit with finding a way to check if these fields are set or not. I recall getting some warning for doing ```if (info.version[0] != \0)``` or something similar. What's the correct way of doing this?

rouault:
```if( info.version[0] != '\0' )``` should be fine

kbevers:
I would think so too. Can't remember the specifics of why I ended up not doing it that way. I'll make the change some time later today. Thanks for reminding of this problem.

PR #569, opened 2017-09-12
Fix for #568

kbevers:


src/PJ_cart.c
```diff
@@ -486,13 +486,13 @@ int pj_cart_selftest (void) {
     /* proj_info()                                                            */
     /* this one is difficult to test, since the output changes with the setup */
     info = proj_info();
-    if (info.version) {
+    if (info.version[0] != '\0' ) {
         char tmpstr[64];
         sprintf(tmpstr, "%d.%d.%d", info.major, info.minor, info.patch);
         if (strcmp(info.version, tmpstr)) return 55;
     }
```
QuLogic, line 493: I see that if `info.version` is non-empty, it should match the expected version. But should it ever be empty?

Issue #378, opened 2016-05-04
Compiling with CMake yields different results than configure script


pmattern:
Compiling current VCS checkouts of PROJ.4 yields different results running CMake and the configure script respectively.

Leaving the usual files in `/usr/share/cmake` aside which obviously are created only when CMake is used the following files

```
/usr/include/org_proj4_PJ.h
/usr/include/org_proj4_Projections.h
/usr/lib/pkgconfig/proj.pc
/usr/share/proj/ntv1_can.dat
/usr/share/proj/null
```

get created running the configure script only.

Seen compiling 1384b97 on Arch Linux. Both `configure` and `cmake` invoked without any option except the respective one to tweak the path.

EDIT Add `/usr/share/proj/null` as seen with 9133b3d.

hobu:
Whatever the configure script does is historically correct, and CMake needs to follow along because we don't want to change any behavior.

pmattern:
@kbevers 
Problem persists in 9133b3d, in addition a file `/usr/share/proj/null` is built by the configure script only. Different ways to build the code yielding among other different headers seems pretty odd.
So mind dropping a note why you closed this one?

kbevers:
Frankly I thought that it was fixed already. I've been going through all the reported issues and have closed quite a bit either because they were fixed, not applicable any more or if it seemed like no one cared about the issue. I am sure I've been too aggresive and have closed issues not shouldn't have been. A nice side effect from that is that it brings people out of the bush and makes it clear that there is still unresolved issues. So thank you for that!

If you have a fix to this please submit a pull request.

hobu:
Fix attempt in #578. Please reopen if still an issue.

PR #578, opened 2017-09-25
CMake updates

hobu:
* Ensure Proj4 data is installed in proper location for CMake builds
  on unix platforms

* Remove dead CMake policy


kbevers:
While you're fiddling with CMake, could you take a look at  https://github.com/OSGeo/proj.4/issues/378 too?

hobu:
I think I may have fixed #378 with this one. My intention was to get the PROJ_DATA install paths correct when the CMake build is used.

PR #579, opened 2017-09-26
Add proj_list_* functions that exposes various internal lists

### LINKED ISSUE #173: Please add API call to get access to available datum names
**Reported by tvrusso on 15 May 2012 16:17 UTC**
I have a program that uses pj_get_datums_ref to obtain a list of all supported datums in proj.4, which my program then uses to create a GUI pull-down menu of datums.  This program was written by patterning its access of the datum list off of the loop done in cs2cs when "-ld" is specified.  

Since the deprecation/removal of projects.h from the install of proj.4, this function (and the internal PJ_DATUMS structure) is no longer accessible.

I would like to see an API call that allows me to get at the same information, without reliance on private data.  My program only uses the datum name out of the PJ_DATUMS structure and uses the PJ_DATUMS structure only for the looping, so I don't _really_ need access to the internals, but this wass the only way to get at the full list of supported datum names.  Once I have them, these names are used in construction of a pj_init string as needed when the user chooses from the pull-down list, so for my own purposes an API call that returned a list of supported datum names without unnecessarily exposing private data would be ideal.  

Migrated-From: https://trac.osgeo.org/proj/ticket/173

kbevers:
Fixes #173, #187 and #220

### DIFF
--- a/src/proj.h
+++ b/src/proj.h
@@ -205,6 +205,23 @@ typedef struct PJ_GRID_INFO PJ_GRID_INFO;
 struct PJ_INIT_INFO;
 typedef struct PJ_INIT_INFO PJ_INIT_INFO;
 
+/* Data types for list of operations, ellipsoids, datums and units used in PROJ.4 */
+struct PJ_LIST;
+typedef struct PJ_LIST PJ_OPERATIONS;
+
+struct PJ_ELLPS;
+typedef struct PJ_ELLPS PJ_ELLPS;
+
+struct PJ_DATUMS;
+typedef struct PJ_DATUMS PJ_DATUMS;
+
+struct PJ_UNITS;
+typedef struct PJ_UNITS PJ_UNITS;
+
+struct PJ_PRIME_MERIDIANS;
+typedef struct PJ_PRIME_MERIDIANS PJ_PRIME_MERIDIANS;
+
+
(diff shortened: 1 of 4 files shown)


src/proj.h
```diff
@@ -410,6 +427,13 @@ PJ_PROJ_INFO proj_pj_info(PJ *P);
 PJ_GRID_INFO proj_grid_info(const char *gridname);
 PJ_INIT_INFO proj_init_info(const char *initname);
 
+/* List functions: */
+/* Get lists of operations, ellipsoids, datums, units and prime meridians. */
+PJ_OPERATIONS       *proj_list_operations(void);
```
rouault: I'd suggest to prefix the return types with const so that the user isn't tempted to modify them

kbevers: Good point. I'll fix it right away.

kbevers:
I am considering leaving out the datums list, since I think the ``+datum``  parameter in many cases makes life more complicated for users. It was meant to make life simpler, but with the way PROJ.4 has evolved recently users can quite easily create transformations that they think are correct but are in fact completely wrong.
Setting ``+datum`` implies that you want you output data in that particular datum (in the forward case). This will only work if your input data is in WGS84. This was fine when every transformation was forced through WGS84, but now that is not always the case.

PR #583, opened 2017-09-28
Introduce a default destructor for PJ objects

busstoptaktik:
Most  `PJ` objects contain no additionally allocated memory beyond the `PJ->opaque` level. Hence, most PJs can be deallocated using a default destructor.

This PR introduces  `pj_default_destructor` and modifies all `PJ_*.c` files to use it. It results in a simplified deallocation and error messaging code flow, reduces the overall weight of the library by approximately 1400 LOC, and makes it easier to introduce new projections (and other coordinate operations).


src/pj_init.c
```diff
@@ -415,6 +415,7 @@ pj_init_plus_ctx( projCtx ctx, const char *definition )
 }
 
 
+void *dealloc_params (projCtx ctx, paralist *start, int errlev);
```
QuLogic: I'm not sure this makes sense here. If it's used elsewhere, it should be in a header file; if it's only used in this file, then it should be `static` and no prototype is needed.

busstoptaktik: Thanks, @QuLogic - I actually needed `dealloc_params()` (now `pj_dealloc_params()`) in `pj_init_ctx()` as well as in `pj_default_destructor()`, but had to take a break half way through, as it got very late :-)

`pj_dealloc_params()`is now in `projects.h`

busstoptaktik:
@QuLogic @hobu @rouault - and everyone else with better “memory leak fu” than myself: I think this PR is getting ready for merging, but @kbevers and I have introduced some rather intrusive and extensive refactorings to the memory management aspect of PROJ.4, so a few new memory leaks would probably not be surprising, although the new state of things should help avoiding leaks in the future.

I have checked the library by running the debug build of `proj -VC` through `DrMemory`, which found nothing suspicious. If any of you have more serious superpowers and/or power tools to throw at the analysis, I would be grateful for any results and comments.

Note: We know already that leaks, related to grid management remain, cf. comment in `pj_malloc.c`:

```
/* We used to call pj_dalloc( P->catalog ), but this will leak */
/* memory. The safe way to clear catalog and grid is to call */
/* pj_gc_unloadall(pj_get_default_ctx()); and pj_deallocate_grids(); */
/* TODO: we should probably have a public pj_cleanup() method to do all */
/* that */
```

rouault:
> Note: We know already that leaks, related to grid management remain, cf. comment in pj_malloc.c:
> 
> /* We used to call pj_dalloc( P->catalog ), but this will leak */
> /* memory. The safe way to clear catalog and grid is to call */
> /* pj_gc_unloadall(pj_get_default_ctx()); and pj_deallocate_grids(); */
> /* TODO: we should probably have a public pj_cleanup() method to do all */
> /* that */

Those are not real "leaks". They are memory still reachable through proj.4 static variables. They will appear as "still reachable" in a Valgrind output, to be opposed to "definitely lost" blocks.

What could help to track memory mis-use/leaks is to have a Travis-CI target that would do a build with -fsanitize=address (requires a gcc/clang recent enough), that way we could detect issues in code that is tested by the test suit.  and/or run it through Valgrind

busstoptaktik:
> What could help to track memory mis-use/leaks is to have a Travis-CI target that would do a build with -fsanitize=address (requires a gcc/clang recent enough)

Good idea. But since my .travis-fu is even weaker than my memleak-fu, I will check it by doing a commit simply adding -fsanitize=address to CFLAGS in the travis setup, letting the tests run, and then removing that change

src/PJ_ob_tran.c
```diff
@@ -89,76 +89,112 @@ static void *destructor(PJ *P, int errlev) {
...
+
+/* turn paralist into argc/argv style argument list */
+static ARGS ob_tran_target_params (paralist *params) {
+    char **argv;
+    size_t i = 0;
+    ARGS args = {0, 0};
+    size_t argc = paralist_params_argc (params);
+    if (0==argc)
+        return args;
+    args.argv = pj_calloc (argc, sizeof (char *));
```
rouault: As far as I can see this args.argv is never deallocated

src/PJ_ob_tran.c
```diff
@@ -89,76 +89,112 @@ static void *destructor(PJ *P, int errlev) {
...
+    if (0==args.argv)
+        return args;
+    args.argc = (int) argc - 1; /* all args except the proj_ob_tran */
+    
+    /* Copy all args *except* the proj=ob_tran arg to the argv array */
+    for (i = 0;  params != 0;  params = params->next) {
+        if (0==strcmp (params->param, "proj=ob_tran"))
+            continue;
+        args.argv[i++] = params->param;
+        if (0==params->next)
```
rouault: useless test. the for() test condition will take care of that

src/PJ_ob_tran.c
busstoptaktik: Oops! Thanks! Correcting immediately

src/PJ_ob_tran.c
busstoptaktik: Very well spotted... remnants from debugging instrumentation - removed now

busstoptaktik:
This PR seems to build and test OK now, and I intend to merge it tonight.

Following advice from @rouault, I added `-fsanitize=address` to the linux-clang Travis build, during the memleak hunting. It helped tracking down not only the leak I was searching for, but also additional leaks in the `multistresstest.c` code. This is evidently useful, so I have kept it in, for improved future memleak detection.

While chasing the leak in the `ob_tran` code, I also enabled the use of ellipsoidal projections with ob_tran.  Originally ob_tran shared parts of the PJ object with its target projection, which was somewhat hacky.

Now, the target projection sits in a separate PJ, and while using spherical formulas to rotate an ellipsoidal  projection is **formally** wrong, it really just amounts to defining a slightly different compound projection. Output data are consistent, so while we could return an error message, I have, for now, decided to let it claim success, and return a PJ, so you can get what you ask for if you know what you do.

Thanks go to @rouault, @QuLogic for some useful reviews and comments, and to @kbevers, who did much of the initial hard work of touching up every single `PJ_xxx.c` file to work with the centralised default destructor.

PR #586, opened 2017-10-03
Pro forma PR to force a check of ob_tran memleak. DO NOT MERGE!!!

busstoptaktik:
Basically, this is a case of checkout master, then add address sanitizer to linux_clang build.

There appears to be a hard to trace memleak in PR #583, and I need to see whether it was present even before that work, but have no accessible way of running a -fsanitize=address enabled build, other than through travis.


busstoptaktik:
Closing this immediately.

PR #588, opened 2017-10-04
Kinematic gridshifting

kbevers:
Some transformations are defined with the use of a velocity model. Use of the kinematic gridshift is  similar to the regular gridshift already in use in PROJ.4 today. The difference is that the amount a coordinate is shifted is dependent on how much time as passed since the grid-epoch. The grid values are defined in units of m/year, and by multiplying with an elapsed time the grid-values are translated into a shift in meters.

Some refactoring of the existing grid-handling code has been necessary to be able to read single grid values in the forward/reverse functions of the operation.

No new grid formats are introduced. Vertical kinematic gridshifts are done with gtx-formats and the horizontal grids can be handled with CTable/CTable2.


src/pj_list.h
```diff
@@ -144,6 +144,7 @@ PROJ_HEAD(vandg2, "van der Grinten II")
 PROJ_HEAD(vandg3, "van der Grinten III")
 PROJ_HEAD(vandg4, "van der Grinten IV")
 PROJ_HEAD(vitk1, "Vitkovsky I")
+PROJ_HEAD(velocity, "Kinematic grid shift")
```
rouault: Perhaps try to uniformize to one of both terms: velocity or kinematic ? And perhaps the projection name should have gridshift in it for consistency with the hgridshift and vgridshift names (hum, but that would be a long name, so perhaps not a good idea)

src/pj_apply_vgridshift.c
```diff
@@ -212,14 +220,75 @@ int pj_apply_vgridshift( PJ *defn, const char *listname,
...
+    name that is used in the proj-string to
+    specify the grids to load, e.g. "+grids".
+    The + should be left out here.
+
+    Returns the number of loaded grids.
+
+***********************************************/
+
+    /* prepend "s" to the "grids" string to allow usage with pj_param */
+    char *sgrids = (char *) pj_malloc( (strlen(grids)+1) *sizeof(char) );
```
rouault: you need strlen(grids)+1+1 for the 's' and nul terminator

test/gie/velocity.gie
```diff
@@ -0,0 +1,68 @@
...
+OPERATION   +proj=velocity +xy_grids=alaska +t_epoch=2016.0 +t_obs=2000.0
+-------------------------------------------------------------------------------
+ACCEPT      -3004295.5882503074 -1093474.1690603832 5500477.1338251457
+EXPECT      -3004295.5888766116 -1093474.1688513425 5500477.1338251457
+ROUNDTRIP   1000
+
+-------------------------------------------------------------------------------
+Test using only vertical grid and +tobs parameter
+-------------------------------------------------------------------------------
+OPERATION   +proj=velocity +z_grids=egm96_15.gtx +t_epoch=2016.0 +t_obs=2000.0
```
rouault: Does this make sense to use egm96_15.gtx ? This does not contain velocity values, unless I'm missing something. I can understand this is practical for testing, but some comment should mention this is not meant to be a real use case, just a technical one
(perhaps the same with the alaska grid ?)

kbevers: No. Except that it doesn't matter in a test setting. There's a comment about precisely that point in the top of the file. I don't want to clutter the repository with too many grid files, since we already have another way of distributing those.

src/pj_list.h
kbevers: I've tried a bunch of different variations and it is quite hard to have a short and descriptive name. The best I could come up with was `velocity`, since it somewhat hints at what's at play. The description expands on that, although with some different words. The operation could be called `kgridshift` for "*kinematic gridshift*", but that is not a very descriptive name I think. I am open to suggestions :)

src/pj_apply_vgridshift.c
kbevers: I'll take care of that. Thanks for catching that.

test/gie/velocity.gie
rouault: Ah sorry for missing the comment. I never read file headers...

kbevers: No worries.

The whole point with this stuff is to implement scandinavian transformations that uses a velocity model. When I've got this code sorted (there's some failing tests I need to fix), then I'll work on some actual data and hopefully confirm that this works as intended. Maybe I'll include that as test cases in the end.

kbevers: I am actually already using GDAL for that. A few weeks ago I wrote [this small utility](https://github.com/Kortforsyningen/kmsgrid/blob/master/kmsgrid.py) with the intend of translating various grids files from the Danish TrLib system to formats that PROJ.4 understands.

Good point regarding creating a specific test grid. I'll do that once I am happy with the code.

src/pj_list.h
busstoptaktik: @kbevers: "deformation" could be a possibility, since the result of the operation is to model what geodetically is called intraplate deformation. This would also lead to the possibility of using `dgridshift`, for *"deformation gridshift",* in a symmetrical naming scheme.

kbevers: I like `deformation` better than `dgridshift`. Partly because I am seriously considering merging `hgridshift` and `vgridshift` (I didn't realize the possibility of seperating horizontal and vertical grids in the same `PJ` at the time). This would then simply be called `gridshift`. The symmetry would then be lost.

I'll think about it about and merge this once I have decided on a name.

busstoptaktik: deformation is certainly descriptive - and merging gridshifts sounds like a laudable rationalization

busstoptaktik:
@kbevers - this must be en route from WIP to a 100% PR? Is it time for another round of reviews?

kbevers:
> this must be en route from WIP to a 100% PR? Is it time for another round of reviews?

Yes. I am happy with this myself, but another pair of eyes would be great!

src/PJ_deformation.c
```diff
@@ -0,0 +1,294 @@
...
+
+        if (delta.x == HUGE_VAL)
+            break;
+
+        dif.x  = out.x - dt*delta.x - input.x;
+        dif.y  = out.y + dt*delta.y - input.y;
+        out.x += dif.x;
+        out.y += dif.y;
+
+    } while ( i-- && fabs(dif.x) > TOL && fabs(dif.y) > TOL );
```
busstoptaktik: If I understand the algorithm correctly, the dif.x,y cases should be or'ed, rather than and'ed: Otherwise you can exit the loop as soon you reach the right coordinate in one axis, while still being kilometers off along the other.

Addititonally, it might be slightly better to use hypot (dif.x, dif.y) > TOL (and perhaps adjust TOL accordingly), rather than the dual fabs: Since we can have no idea about the local gradients, we really want to make sure we're "inside the disk", while checking each dimension individually only ensures we're "inside the square".

Finally, I think `i--` should be changed to `--i`, otherwise `i = MAX_ITERATIONS` would result in `MAX_ITERATIONS + 1` iterations.

src/PJ_deformation.c
```diff
@@ -0,0 +1,294 @@
...
+    if (0==Q)
+        return pj_default_destructor(P, ENOMEM);
+    P->opaque = (void *) Q;
+
+    Q->cart = proj_create(P->ctx, "+proj=cart");
+    if (Q->cart == 0)
+        return destructor(P, ENOMEM);
+
+    /* copy ellipsoid parameters from P to Q->cart (simpler than guessing how */
+    /* the ellipsoid was specified in original definition)                    */
```
busstoptaktik: Perhaps put a comment about this in PJ_cart.c, since this will bomb if algorithms are changed in PJ_cart in a way so additional ellipsoidal parameters are used.

Probably, this indicates the need for an internal `pj_inherit_ellipsoid_defs()`  function.

src/pj_apply_vgridshift.c
```diff
@@ -28,174 +28,182 @@
...
+
+
+    return value;
+}
 
 /************************************************************************/
 /*                        pj_apply_vgridshift()                         */
 /*                                                                      */
-/*      This implementation takes uses the gridlist from a coordinate    */
+/*      This implementation takes uses the gridlist from a coordinate   */
```
busstoptaktik, line 136: Remove either "takes" or "uses"

busstoptaktik (commented):
There may be a serious problem in a loop exit condition - or I may just not understand the algorithm properly.

Remaining comments are on minor issues

src/PJ_deformation.c
kbevers: This was taken more or less verbatim from the implementation in `nad_cvt()`. I've changed the variable names, but otherwise it is the same algorithm. I think your suggestions are sensible and seem to be working fine as well. I don't think the decrement method of `i` matters all that much since convergence is reached within the first few passes of the loop but we might as well do it as correct as possible.

src/PJ_deformation.c
kbevers: I added the suggested function. No need to add further comments to PJ_cart.c.

PR #595, opened 2017-10-08
Resolve #594 and OSS-Fuzz-3569

### LINKED ISSUE #594: Be more defensive in pj_ctx
pj_ctx offers ample opportunity to shoot oneself in the foot by following null pointers. As these functions are seldom called, they do not need to be so fast as to not checking a pointer against NULL, and doing something sensible (e.g. nothing) in that case.

This is the great example shown in the C standard library `free()` function, which checks for NULL, and does nothing if handed a null pointer. In general it simplifies the control flow of the application level code much, if you can rely on the library handling this. And in cases where the app level code needs to be so super fast it cannot afford an unnecessary  function call, no harm is done: It can just do the check itself, and drop the call if not needed.

(I'm working on this)

busstoptaktik:
OSS-Fuzz bug 3569 resulted from incomplete deallocation code in PJ_geos.c. Repairing that fits naturally into the general work needed to resolve #594.

### DIFF
--- a/src/PJ_geos.c
+++ b/src/PJ_geos.c
@@ -40,20 +40,10 @@ struct pj_opaque {
     double radius_g;
     double radius_g_1;
     double C;
-    char *sweep_axis;
     int flip_axis;
 };
 
 PROJ_HEAD(geos, "Geostationary Satellite View") "\n\tAzi, Sph&Ell\n\th=";
-
-
-static void *destructor (PJ *P, int errlev) {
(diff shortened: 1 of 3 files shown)


busstoptaktik:
Remaining AppVeyor build error is a fetch error on one target - would probably have built flawlessly, as all the other builds did.

Resolves #594 
Resolves OSS-Fuzz Issue 3569

Credit to OSS-Fuzz

PR #596, opened 2017-10-09
Resolve OSS Fuzz issue 3595 division by zero by replacing atan(y/x)

busstoptaktik:
with atan2(y,x) in geocent.c line 433

Credit to OSS Fuzz

### DIFF
--- a/src/geocent.c
+++ b/src/geocent.c
@@ -430,7 +430,7 @@ void pj_Convert_Geocentric_To_Geodetic (GeocentricInfo *gi,
     while (SDPHI*SDPHI > genau2 && iter < maxiter);
 
 /*	ellipsoidal (geodetic) latitude */
-    *Latitude=atan(SPHI/fabs(CPHI));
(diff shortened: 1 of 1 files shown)


mwtoews:
Searching "[atan](https://github.com/OSGeo/proj.4/search?q=atan)" I see potentially a dozen or so cases that may potentially have a similar issue with a zero divisor, where atan2(y, x) is more suited. I'm not suggesting these be fix with this PR, just making a comment for now.

busstoptaktik:
> Searching "atan" I see potentially a dozen or so cases that may potentially have a similar issue with a zero divisor, where atan2(y, x) is more suited.

I agree - it a horribly common error.

> I'm not suggesting these be fix with this PR, just making a comment for now.

I would probably prefer to get rid of the material in geocent.c and implement one of the newer and very fast algorithms - also just a comment for now: not planning to do it in the next couple of weeks...

busstoptaktik:
Travis CI not completing is a Travis issue with OSX setup, not an issue with this PR. Merging immediately.

PR #616, opened 2017-10-20
Handle allocation failure when loading grid files

aaronpuchert:
Continuing #606, we tackle the same issues in the loading and processing of grid files. This should fix potential crashes and memory leaks, and makes sure the global lock is always released.

### DIFF
--- a/src/pj_gridinfo.c
+++ b/src/pj_gridinfo.c
@@ -212,7 +212,9 @@ int pj_gridinfo_load( projCtx ctx, PJ_GRIDINFO *gi )
         ct_tmp.cvs = (FLP *) pj_malloc(gi->ct->lim.lam*gi->ct->lim.phi*sizeof(FLP));
         if( row_buf == NULL || ct_tmp.cvs == NULL )
         {
-            pj_ctx_set_errno( ctx, -38 );
+            pj_dalloc( row_buf );
+            pj_dalloc( ct_tmp.cvs );
+            pj_ctx_set_errno( ctx, ENOMEM );
             pj_release_lock();
             return 0;
         }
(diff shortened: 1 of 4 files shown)


src/pj_gridinfo.c
```diff
@@ -568,10 +579,23 @@ static int pj_gridinfo_init_ntv2( projCtx ctx, PAFile fid, PJ_GRIDINFO *gilist )
         else
         {
             gi = (PJ_GRIDINFO *) pj_malloc(sizeof(PJ_GRIDINFO));
+            if (!gi) {
+                pj_dalloc(ct);
+                pj_gridinfo_free(ctx, gilist);
+                pj_ctx_set_errno(ctx, ENOMEM);
+                return 0;
+            }
             memset( gi, 0, sizeof(PJ_GRIDINFO) );
```
busstoptaktik: You can eliminate this by switching to `pj_calloc` in line 581. While slightly outside the scope of this PR, it is still about memory handling, and in general, I think using calloc when needing a 0-initialized buffer is the right thing to do.

src/pj_gridinfo.c
```diff
@@ -864,9 +896,18 @@ PJ_GRIDINFO *pj_gridinfo_init( projCtx ctx, const char *gridname )
 /*      cannot be loaded.                                               */
 /* -------------------------------------------------------------------- */
     gilist = (PJ_GRIDINFO *) pj_malloc(sizeof(PJ_GRIDINFO));
+    if (!gilist) {
+        pj_ctx_set_errno(ctx, ENOMEM);
+        return NULL;
+    }
     memset( gilist, 0, sizeof(PJ_GRIDINFO) );
```
busstoptaktik: Same as for line 581 - use pj_calloc

busstoptaktik (commented):
Suggest to change pj_malloc/memset combos to pj_calloc

busstoptaktik:
All checks passed, merging and closing.

PR #618, opened 2017-10-20
Replace some asserts by compile-time assertions

aaronpuchert:
Decisions regarding the size of certain types are always made at compile time, hence this is where the assertion should take place.

If the assertions fail, there is now a compiler error.

### DIFF
--- a/src/nad2bin.c
+++ b/src/nad2bin.c
@@ -1,10 +1,10 @@
 /* Convert bivariate ASCII NAD27 to NAD83 tables to NTv2 binary structure */
 #include <stdio.h>
 #include <stdlib.h>
-#include <assert.h>
 
 #define PJ_LIB__
-#include <projects.h>
+#include "proj_internal.h"
+#include "projects.h"
 #define U_SEC_TO_RAD 4.848136811095359935899141023e-12
 
 /************************************************************************/
@@ -177,8 +177,8 @@ int main(int argc, char **argv) {
             exit(2);
 	}
 
-        assert( MAX_TAB_ID == 80 );
-        assert( sizeof(int) == 4 ); /* for ct.lim.lam/phi */
+        STATIC_ASSERT( MAX_TAB_ID == 80 );
+        STATIC_ASSERT( sizeof(pj_int32) == 4 ); /* for ct.lim.lam/phi */
 
         memset( header, 0, sizeof(header) );
 
@@ -267,13 +267,13 @@ int main(int argc, char **argv) {
         {
             unsigned char achHeader[11*16];
             double dfValue;
-            int nGSCount = ct.lim.lam * ct.lim.phi;
+            pj_int32 nGSCount = ct.lim.lam * ct.lim.phi;
             LP ur;
 
             ur.lam = ct.ll.lam + (ct.lim.lam-1) * ct.del.lam;
             ur.phi = ct.ll.phi + (ct.lim.phi-1) * ct.del.phi;
 
-            assert( sizeof(nGSCount) == 4 );
+            STATIC_ASSERT( sizeof(nGSCount) == 4 );
 
             memset( achHeader, 0, sizeof(achHeader) );
 
--- a/src/pj_gridinfo.c
+++ b/src/pj_gridinfo.c
@@ -28,20 +28,11 @@
 
 #define PJ_LIB__
 
-#include <projects.h>
+#include "proj_internal.h"
+#include "projects.h"
 #include <string.h>
 #include <math.h>
 #include <errno.h>
-
-#ifdef _WIN32_WCE
-/* assert.h includes all Windows API headers and causes 'LP' name clash.
- * Here assert we disable assert() for Windows CE.
- * TODO - mloskot: re-implement porting friendly assert
- */
-# define assert(exp)	((void)0)
-#else
-# include <assert.h>
-#endif /* _WIN32_WCE */
 
 /************************************************************************/
 /*                             swap_words()                             */
@@ -431,23 +422,8 @@ static int pj_gridinfo_init_ntv2( projCtx ctx, PAFile fid, PJ_GRIDINFO *gilist )
     int num_subfiles, subfile;
     int must_swap;
 
-    assert( sizeof(int) == 4 );
-    assert( sizeof(double) == 8 );
-#ifdef _MSC_VER
-#pragma warning( push )
-/* disable conditional expression is constant */
-#pragma warning( disable : 4127 )
-#endif
-    if( sizeof(int) != 4 || sizeof(double) != 8 )
-#ifdef _MSC_VER
-#pragma warning( pop )
-#endif
-    {
-        pj_log( ctx, PJ_LOG_ERROR,
-             "basic types of inappropraiate size in pj_gridinfo_init_ntv2()" );
-        pj_ctx_set_errno( ctx, -38 );
-        return 0;
-    }
+    STATIC_ASSERT( sizeof(pj_int32) == 4 );
+    STATIC_ASSERT( sizeof(double) == 8 );
 
 /* -------------------------------------------------------------------- */
 /*      Read the overview header.                                       */
@@ -541,8 +517,8 @@ static int pj_gridinfo_init_ntv2( projCtx ctx, PAFile fid, PJ_GRIDINFO *gilist )
         ct->del.lam = *((double *) (header+9*16+8));
         ct->del.phi = *((double *) (header+8*16+8));
 
-        ct->lim.lam = (int) (fabs(ur.lam-ct->ll.lam)/ct->del.lam + 0.5) + 1;
-        ct->lim.phi = (int) (fabs(ur.phi-ct->ll.phi)/ct->del.phi + 0.5) + 1;
+        ct->lim.lam = (pj_int32) (fabs(ur.lam-ct->ll.lam)/ct->del.lam + 0.5) + 1;
+        ct->lim.phi = (pj_int32) (fabs(ur.phi-ct->ll.phi)/ct->del.phi + 0.5) + 1;
 
         pj_log( ctx, PJ_LOG_DEBUG_MINOR,
                 "NTv2 %s %dx%d: LL=(%.9g,%.9g) UR=(%.9g,%.9g)\n",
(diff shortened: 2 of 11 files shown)


aaronpuchert:
There is a [keyword for static assertions](http://en.cppreference.com/w/c/language/_Static_assert) in C11, but in C89 macros must be used. I found a nice compilation of [variants](http://www.pixelbeat.org/programming/gcc/static_assert.html), maybe someone likes another variant more than the one I chose.

busstoptaktik:
That's a beautiful, readable, and very compact implementation of a static assert you have provided, @aaronpuchert!

In my opinion it is an excellent step towards "failing early and noisily", and it probably should go in `projects.h` and/or `proj_internal.h`, for its general usefulness.

But since the use of static `assert`ions introduces a new coding aspect into the PROJ.4 code base, I think a few others (like @rouault, @kbevers, and @hobu) should voice their opinion, before merging (especially since it would mean build failures on platforms where `sizeof int != 4`)

kbevers:
I agree that the STATIC_ASSERT should be placed somewhere central. I prefer `proj_internal.h` (to keep it private to PROJ.4).

> especially since it would mean build failures on platforms where sizeof int != 4

I guess the impact on this is that folks with differently sized integers will fail to build PROJ.4, whereas today they get a usable library except if trying to load grids, yes? Are there any modern systems that do not have 4 byte integers?

aaronpuchert:
> I guess the impact on this is that folks with differently sized integers will fail to build PROJ.4, whereas today they get a usable library except if trying to load grids, yes?

Correct. However, I think that in these cases it would be better to get a compiler error, see that something is missing, and decide how to proceed. Maybe this entails providing a fallback implementation.

Regarding `sizeof(double) == 8`, this is pretty much fixed by the IEEE 754 standard. I don't see a problem with removing that.

> Are there any modern systems that do not have 4 byte integers?

Do people use PROJ.4 on embedded systems? Then `sizeof(int)` could be smaller.

In [C99](http://en.cppreference.com/w/c/types/integer) we could just use `int32_t` and be done.

busstoptaktik:
@aaronpuchert said:
> Do people use PROJ.4 on embedded systems? Then sizeof(int) could be smaller.

And on SPARC64 it could be wider.

> In C99 we could just use int32_t and be done.

And it would be a wonderful thing to be able to do. Unfortunately, the geniuses from Redmond have, until recently, refused to support even the slightest shade of C99 (and even to provide full C89 support, by e.g. making `HUGE_VAL` compile time constant). Not being able to compile on MSVC is, unfortunately, not an option, so we have to live with the maldecisions of our Redmond Overlords. [Stupidity is universal](https://qz.com/967554/the-five-universal-laws-of-human-stupidity/).

src/nad2bin.c
```diff
@@ -3,6 +3,7 @@
 #include <stdlib.h>
 
 #define PJ_LIB__
+#include <proj_internal.h>
```
kbevers: `proj_internal.h` is private to PROJ.4 so should be included as such with `#include "proj_internal.h"`. `projects.h` is going to be removed from the public API in the near future, so we are changing the include style of that too whenever we come across it.

src/pj_gridinfo.c
```diff
@@ -28,6 +28,7 @@
 
 #define PJ_LIB__
 
+#include <proj_internal.h>
```
kbevers: Same as above

src/nad2bin.c
aaronpuchert: No problem. What about the order of the includes? That is a bit inconsistent. In `nad2bin.c` it would then be:

```cpp
#include <stdio.h>
#include <stdlib.h>

#define PJ_LIB__
#include "proj_internal.h"
#include "projects.h"
```

In `pj_gridinfo.c`:

```cpp
#define PJ_LIB__

#include "proj_internal.h"
#include "projects.h"
#include <string.h>
#include <math.h>
#include <errno.h>
```

kbevers: I tend to put the `""`-includes at the bottom of the list, but as you've discovered there's not much consistency in the existing code.

busstoptaktik:
Perhaps a solution could be to compile time check that `INT_MAX==2147483647`, then providing errlev-setting stub implementations of the library functions for platforms where the 32 bit assertion is not met. This would result in a "mostly functional" library on these platforms.

`nad2bin` would also need a code block announcing that it is not supported on this platform, cf. the conditionally compiled "windows block" in `test228.c`:

    #ifdef _WIN32
    int main(int argc, char* argv[]) {
        printf("Test not yet ported on Win32\n"); 
        return 0;                                 
    }                                             
    #else   
    ...

aaronpuchert:
There is also the possibility to [utilize CMake](https://cmake.org/Wiki/CMake:How_To_Write_Platform_Checks). We could detect the sizes of `short int`, `int`, and `long int` at configuration time and then choose an appropriate type:

```cpp
#if __SIZEOF_INT__ == 4
typedef int int_32;
typedef unsigned int uint_32;
#elif __SIZEOF_LONG__ == 4
typedef long int_32;
typedef unsigned long uint_32;
#elif __SIZEOF_SHORT__ == 4
typedef short int_32;
typedef unsigned short uint_32;
#else
# error "No native data type can represent a 32-bit integer."
#endif
```

This is what I've seen in other projects.

busstoptaktik:
@aaronpuchert said:
> There is also the possibility to utilize CMake. 

This would be a problem for the nmake and autotools based build systems. I wrote a long apropos to this, but decided to post it as an issue, for discussion - cf. #622

aaronpuchert:
Alright, I understand. You want to be able to compile without any additional tools. What about this?

```cpp
#ifndef DISABLE_STATIC_ASSERT
    STATIC_ASSERT(condition);
#else
    if (!condition) {
        // Handle error.
    }
#endif
```

Users with the wrong size would first get a compilation error, and could then decide how to proceed. One possibility would be to add a flag like `-DDISABLE_STATIC_ASSERT`, which makes it compile but turns off some features.

I'm not sure what your position is, but I would prefer this to having features silently turned off and noticing only at run-time.

busstoptaktik:
> Alright, I understand. You want to be able to compile without any additional tools

It's true that I find it preferable to have software configuration tools as options, not hard dependencies, but that was not my point right here (although it was in the apropos, #622). My point was just that using CMAKE to detect `sizeof(int)` would not help us in the autotools case, and in a recent discussion with @kbevers, @hobu mentioned the importance of supporting a plain CMAKE-independent autotools build is important.

Back to #622, being able to build without a source configuration system is certainly nice as well, although for a different reason.

I prefer CMAKE to autotools, but that's just a personal preference, based on not-very-much experience.
But I also prefer to minimize **hard** dependencies, and since a source configuration system in many cases can be cast in the role of a **soft** dependency (at the cost of having to downgrade gracefully), I think it is worthwhile - especially in the case of PROJ.4, where the effort for doing this is minor.

busstoptaktik:
But I forgot to reply to the question:
> What about this?
    
    (code block left out)

That looks like a neat solution, which in many cases may even simplify to

```C
#ifndef DISABLE_STATIC_ASSERT
    STATIC_ASSERT(condition);
#else
    if (condition)
        return 0;
    /* now do the real stuff */
#endif
```

QuLogic:
Why do that instead of `#define STATIC_ASSERT` to return?

aaronpuchert:
I guess because error handling is not completely consistent. Some functions return 0 or `NULL` on error (those returning a pointer usually), other return 0 on success and an error code otherwise. And in `src/nad2bin.c` we call `exit(2)` on error.

aaronpuchert:
> My point was just that using CMAKE to detect `sizeof(int)` would not help us in the autotools case, and in a recent discussion with @kbevers, @hobu mentioned the importance of supporting a plain CMAKE-independent autotools build is important.

My understanding is that autotools can detect stuff like `sizeof(int)` as well, that is usually what the `configure` script does. But I don't know anything about `nmake`.

Also it's probably a pain in the ass to manage multiple build systems where each has its own way of detecting things, and all have to be supported. So adding more configuration doesn't make much sense right now.

aaronpuchert:
There is probably a 32-bit integer type on almost all platforms, the question is just how it is called. On most platforms, this is `int`, but on some it might be `long` (e.g. x86 in 16-bit mode).

For people not being familiar with the library I think it's useful if they see a compiler error when building the library rather than find out something is missing only after they've deployed the code. Of course you can disagree though, this is just my perspective.

Unfortunately the right way to do this is only available with C99, so I think this is as good as it gets.

src/projects.h
```diff
@@ -78,6 +79,14 @@ extern "C" {
 #  define ABS(x)        ((x<0) ? (-1*(x)) : x)
 #endif
 
+#if INT_MAX == 2147483647
+typedef int int32;
```
rouault: Can we use some 'namespace' ? So pj_int32. I know a number of other projects, which can potentially be mixed with proh, that also have typedef int32...

src/pj_gridinfo.c
```diff
@@ -431,23 +422,8 @@ static int pj_gridinfo_init_ntv2( projCtx ctx, PAFile fid, PJ_GRIDINFO *gilist )
...
-#ifdef _MSC_VER
-#pragma warning( pop )
-#endif
-    {
-        pj_log( ctx, PJ_LOG_ERROR,
-             "basic types of inappropraiate size in pj_gridinfo_init_ntv2()" );
-        pj_ctx_set_errno( ctx, -38 );
-        return 0;
-    }
+    STATIC_ASSERT( sizeof(int) == 4 );
```
rouault: Shouldn't STATIC_ASSERT( sizeof(int) == 4 ); be removed and (pj_)int32 used everywhere where a int is assumed to be 4 bytes ?

src/pj_gridinfo.c
```diff
@@ -669,23 +645,8 @@ static int pj_gridinfo_init_ntv1( projCtx ctx, PAFile fid, PJ_GRIDINFO *gi )
...
-#ifdef _MSC_VER
-#pragma warning( pop )
-#endif
-    {
-        pj_log( ctx, PJ_LOG_ERROR,
-                 "basic types of inappropraiate size in nad_load_ntv1()" );
-        pj_ctx_set_errno( ctx, -38 );
-        return 0;
-    }
+    STATIC_ASSERT( sizeof(int) == 4 );
```
rouault: Same here

src/pj_gridinfo.c
```diff
@@ -769,23 +730,8 @@ static int pj_gridinfo_init_gtx( projCtx ctx, PAFile fid, PJ_GRIDINFO *gi )
...
-#ifdef _MSC_VER
-#pragma warning( pop )
-#endif
-    {
-        pj_log( ctx, PJ_LOG_ERROR,
-                "basic types of inappropraiate size in nad_load_gtx()" );
-        pj_ctx_set_errno( ctx, -38 );
-        return 0;
-    }
+    STATIC_ASSERT( sizeof(int) == 4 );
```
rouault: Same here

src/projects.h
aaronpuchert: That makes sense. I will change it.

aaronpuchert:
Thank you for the review, I addressed your comments.

The Windows build failure seems to have been caused by network problems, unfortunately I can't restart the build.

rouault:
I've restarted the Windows build. Looks good to me. Other reviewers: ok to merge ?

kbevers (approved):

PR #619, opened 2017-10-20
Pro forma PR - do not merge

busstoptaktik:
What happens to tests if we get rid of proj_def.dat? See #201


kbevers:
You are not going to get away with removing WGS84 as the default ellipsoid :-)

kbevers:
I meant for this test... Try changing the default to GRS80 and see how big the impact on the test data is.

busstoptaktik:
> I meant for this test

That will be my next commit. Until now, I just want to see how bad it gets (how many non-express assumptions are, at least, in the code - apparently calcofi depends hard on WGS84, which would make sense if talking about the datum, but not the ellipsoid)

busstoptaktik:
Fails miserably, but only (?) due to the expected submillimetric differences in `testvarious` output file `tv_out`. Mostly a matter of creating a new `tv_out.dist` file.

kbevers:
It doesn't look *that* bad actually. It is at most 5 or 6 test cases that are affected. Should be easy enough to check if they are something other deviations on the 8th decimal, which is most likely the case. An updated `tv_out.dist` is definitely the way to go then.

busstoptaktik:
(Closing for now, having gained the insight, that solving the technical aspect of this is not a big problem. Deferring further work until this has been discussed on the mailing list - hopefully first after next release)

PR #625, opened 2017-10-24
Remove PJ_OBS from the API surface, rename pj_obs_api.c to pj_4D_api.c

### LINKED ISSUE #597: Rename OBS-API to 4D-API
To talk about something, it needs a name, so the "new geodetic API" got the name OBS-API, since it was based on the OBServation data type.

As most of the problems the OBS data type was intended to solve has now found other, and more elegant, solutions, the COORDinate data type is really the workhorse of the OBS API. So we may very well discourage the use of PJ_OBS, pj_fwdobs, pj_invobs, etc, even before it appears in an official release.

As it is probably better anyway to have an API name that relates to the core functionality exposed by the API, rather than the data type it happens to leverage, *I propose we change the name of the API to **the PROJ.4 4D-API**.*

We can leave the header name as is, or change it to proj4d.h if that is better (I think @rouault once mentioned a SGI 1990s system header with the proj.h-name, so changing to proj4d.h would make it easier for the SGI nostalgia crowd, to run PROJ.4 on their aging Octanes :-) ).

proj_obs, proj_trans_obs, proj_transform_obs, and PJ_OBS should go from the surface of the API. They cannot go (immediately) from the library, though, as they are used as workhorses for the COORD style functions.

Eventually ridding PROJ.4 of the PJ_OBS data type would, however, offer a slight speed advantage.

busstoptaktik:
Resolves #597

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -85,7 +85,7 @@ libproj_la_SOURCES = \
 	jniproj.c pj_mutex.c pj_initcache.c pj_apply_vgridshift.c geodesic.c \
 	pj_strtod.c \
 	\
-	pj_obs_api.c PJ_cart.c PJ_pipeline.c PJ_horner.c PJ_helmert.c \
+	proj_4D_api.c PJ_cart.c PJ_pipeline.c PJ_horner.c PJ_helmert.c \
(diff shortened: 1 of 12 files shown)


busstoptaktik:
Checks not fully successful as one of the Travis builds stalled, for reasons not related to this PR. Merging now.

PR #629, opened 2017-10-25
Allow nested pipelines.

kbevers:
The previous behaviour was to quit pipeline initialization when
encountering a nested pipeline definition. With this commit that
behaviour is changed so that nested pipelines are expanded into one big
pipeline. Example:

    +proj=pipeline +step +proj=pipeline +step +proj=merc

is turned into

    +proj=pipeline +step +proj=merc

This is useful in init-files where steps in complicated transformations
can be grouped in "sub-pipelines". These "sub-pipelines" can then be
used as individual steps in a larger and more complicated pipeline.

### DIFF
--- a/src/pj_init.c
+++ b/src/pj_init.c
@@ -441,6 +441,8 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
     int i;
     int found_def = 0;
     PJ *PIN = 0;
+    int n_pipelines = 0;
+    int n_inits = 0;
 
     if (0==ctx)
         ctx = pj_get_default_ctx ();
@@ -452,36 +454,45 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
         pj_ctx_set_errno (ctx, PJD_ERR_NO_ARGS);
         return 0;
     }
-    
+
+    /* count occurrences of pipelines and inits */
+    for (i = 0; i < argc; ++i) {
+        if (!strcmp (argv[i], "+proj=pipeline") || !strcmp(argv[i], "proj=pipeline") )
+                n_pipelines++;
+        if (!strncmp (argv[i], "+init=", 6) || !strncmp(argv[i], "init=", 5))
+            n_inits++;
+    }
+
+    /* can't have nested pipeline directly */
+    if (n_pipelines > 1) {
+        pj_ctx_set_errno (ctx, PJD_ERR_MALFORMED_PIPELINE);
+        return 0;
+    }
+
+    /* don't allow more than one +init in non-pipeline operations */
+    if (n_pipelines == 0 && n_inits > 1) {
+        pj_ctx_set_errno (ctx, PJD_ERR_TOO_MANY_INITS);
+        return 0;
+    }
+
     /* put arguments into internal linked list */
     start = curr = pj_mkparam(argv[0]);
     if (!curr)
         return pj_dealloc_params (ctx, start, ENOMEM);
 
-    /* build parameter list and expand +init's. Does not take care of a single +init. */
     for (i = 1; i < argc; ++i) {
         curr->next = pj_mkparam(argv[i]);
         if (!curr->next)
             return pj_dealloc_params (ctx, start, ENOMEM);
-
-        /* check if +init present */
-        if (pj_param(ctx, curr, "tinit").i) {
-            found_def = 0;
-            curr = get_init(ctx, &curr, curr->next, pj_param(ctx, curr, "sinit").s, &found_def);
-            if (!curr)
-                return pj_dealloc_params (ctx, start, PJD_ERR_NO_ARGS);
-
-            if (!found_def)
-                return pj_dealloc_params (ctx, start, PJD_ERR_NO_OPTION_IN_INIT_FILE);
-                
-        } else {
-            curr = curr->next;
-        }
+        curr = curr->next;
     }
 
-    /* in case the parameter list only consist of a +init parameter
-       it is expanded here (will not be handled in the above loop). */
-    if (pj_param(ctx, start, "tinit").i && argc == 1) {
+
+    /* Only expand +init's in non-pipeline operations. +init's in pipelines are  */
+    /* expanded in the individual pipeline steps during pipeline initialization. */
+    /* Potentially this leads to many nested pipelines, which shouldn't be a     */
+    /* problem when +inits are expanded as late as possible.                     */
+    if (pj_param(ctx, start, "tinit").i && n_pipelines == 0) {
         found_def = 0;
         curr = get_init(ctx, &start, curr, pj_param(ctx, start, "sinit").s, &found_def);
         if (!curr)
@@ -489,10 +500,10 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
         if (!found_def)
             return pj_dealloc_params (ctx, start, PJD_ERR_NO_OPTION_IN_INIT_FILE);
     }
-    
+
     if (ctx->last_errno)
         return pj_dealloc_params (ctx, start, ctx->last_errno);
-        
+
     /* find projection selection */
     if (!(name = pj_param(ctx, start, "sproj").s))
         return pj_default_destructor (PIN, PJD_ERR_PROJ_NOT_NAMED);
@@ -500,16 +511,18 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
 
     if (!s)
         return pj_dealloc_params (ctx, start, PJD_ERR_UNKNOWN_PROJECTION_ID);
-        
-    /* set defaults, unless inhibited */
-    if (!(pj_param(ctx, start, "bno_defs").i))
-        curr = get_defaults(ctx,&start, curr, name);
+
+    /* set defaults, unless inhibited or we are initializing a pipeline */
+    if (!(pj_param(ctx, start, "bno_defs").i) && n_pipelines == 0)
+            curr = get_defaults(ctx,&start, curr, name);
+
     proj = (PJ *(*)(PJ *)) pj_list[i].proj;
 
     /* allocate projection structure */
     PIN = proj(0);
     if (0==PIN)
         return pj_dealloc_params (ctx, start, ENOMEM);
+
 
     PIN->ctx = ctx;
     PIN->params = start;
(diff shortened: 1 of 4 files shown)


busstoptaktik:
I lost track, when trying to ensure myself that this will work properly, so I just want to suggest that you take a look at these lines around line 220:
```C
if (P->opaque->reverse_step[i])
    point = proj_trans_obs (P->opaque->pipeline[i], PJ_FWD, point);
else
    point = proj_trans_obs (P->opaque->pipeline[i], PJ_INV, point);
```
I think there will be a (resolvable) singularity in the case of:
```
+proj=pipeline +step +init=pipelines:pipeline2 +inv +step +init=pipelines:pipeline3
```
If the +inits just represent a single step this will work (as seen in e.g. the last line of [this](https://github.com/busstoptaktik/NordicTransformations/blob/master/resources/DK)), but in the case of a pipeline it is hard to see where the `+inv`ification should start.

Hence, the corresponding `reverse_step` entries must be inverted, the `+inv` removed, and the sub-pipeline steps reverted. This would probably become easier to implement if requiring that `+inv`s must precede `+init`s

(or am I wrong, and it is only the subpipeline steps that should be reverted?)

Perhaps it might be easier to just let proj=pipeline call itself recursively, so a subpipeline is just another PJ object, that happens to be a pipeline...

kbevers:
> I think there will be a (resolvable) singularity in the case of:
> +proj=pipeline +step +init=pipelines:pipeline2 +inv +step +init=pipelines:pipeline3

I think that is correct.

> Perhaps it might be easier to just let proj=pipeline call itself recursively, so a subpipeline is just another PJ object, that happens to be a pipeline...

It is worth looking into at least. Not sure it is easy, but probably the most robust solution.

busstoptaktik:
> It is worth looking into at least. Not sure it is easy, but probably the most robust solution.

Perhaps just by letting +init introduce a pipeline: It is probably only in the case of +inits it will be relevant, since handwriting something like
```
proj=pipeline step proj=pipeline...
```
will probably be easily avoidable in all practical cases :-)

So a rule along the lines of "when in a pipeline (or perhaps even always?), +init inroduces a new pipeline PJ object"

kbevers:
This was a spur of the moment thing, it obviously needs more thought. Closing for now.

busstoptaktik:
> This was a spur of the moment thing, it obviously needs more thought. Closing for now.

But certainly an extremely useful thing to have - so looking forward to see the result of some deep thinking :-)

busstoptaktik:
Perhaps introducing a `+pinit` ("pipeline init") thingy would be useful for a start: This would eliminate the premature expansion of `+init` and allow `proj=pipeline` to peacefully expand its dependency as a separate `PJ`. Later on, as things settle, it can be either continued as a plain synonym for `+init` or deprecated

kbevers:
> Perhaps introducing a +pinit ("pipeline init") thingy would be useful for a start: This would eliminate the premature expansion of +init and allow proj=pipeline to peacefully expand its dependency as a separate PJ. Later on, as things settle, it can be either continued as a plain synonym for +init or deprecated

I think I got it without adding another type of initialization. I need to add some proper tests but my playing around on the command line did not trigger any unexpected behaviour.

The rules for nested pipelines are now:

1. You can't have more than one literal `+proj=pipeline` in a proj-string
2. Pipelines can be nested if they are wrapped up in a `+init`
3. More than one `+init` is disallowed in non-pipeline proj-strings
4. `+inits` are expanded as late as possible, that is they will only be expanded in single operations (that can be a part of a pipeline)

What have I missed?

src/pj_init.c
```diff
@@ -452,55 +454,63 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
...
+        pj_ctx_set_errno (ctx, PJD_ERR_MALFORMED_PIPELINE);
+        return 0;
+    }
+
+    /* don't allow more than one +init in non-pipeline operations */
+    if (n_pipelines == 0 && n_inits > 1) {
+        pj_ctx_set_errno (ctx, PJD_ERR_TOO_MANY_INITS);
+        return 0;
+    }
+
```
busstoptaktik, line 476: Is it necessary to disallow more than 1 init in classic operations? Wouldn't it be sufficient to disallow more than one in each step of a pipeline?

kbevers: Well... each step boils down to a classic operation, so enforcing a one-init-rule solves the pipelines steps without further logic. It is also what has been done previously, before pipelines were a thing. Lastly, I can't think of a good reason to have several inits in a single operation.

busstoptaktik: But otherwise - If this works, I'll buy you beer: This is pipelines as I had intended them, but couldn't implement in an evidently safe way a year ago, and had to disallow nesting instead. This PR rocks!, @kbevers rocks!, pipelines rock!

And what about the `+step +init=foo +inv` case? works too? Then it **really** rocks!

busstoptaktik: ... and rolls :-)

kbevers: > And what about the +step +init=foo +inv case? works too? 

At least this abomination of a proj-string works without errors:

```
$ cat flaf
<boing> +proj=pipeline +step +proj=merc +step +proj=merc +inv +init=./flaf:puf

<puf> +proj=pipeline +step +proj=aea +step +proj=aea +inv

$ ./bin/cct +proj=pipeline +step +init=epsg:25833 +step +inv +init=epsg:25832 +step +inv +init=epsg:43
26 +step +init=./flaf:boing +step +init=epsg:25832
12 56 0 0
312928.560892657085788  6210141.326871711760759     0.000000000000000     0.000000000000000
```

It has both inverted inits and two sets of nested pipelines. I couldn't come up with something that didn't do what I expected, so maybe it's okay. I'll test it on a real world scenario tomorrow.

> But otherwise - If this works, I'll buy you beer

 Friday bar has Belgian beers tomorrow ;-)

busstoptaktik: Aaargh - proj -VC fails the pipeline test with exit 5001...

busstoptaktik: > Friday bar has Belgian beers tomorrow ;-)
  
Will practice my Flemish immediately!

busstoptaktik: > At least this abomination of a proj-string works without errors

Awesome - and totally crooked!

kbevers:
problem seems to be that in the last step the ellipsoid is somehow set to WGS84 instead of GRS80:

```
λ bin\cct.exe +proj=pipeline +step +init=epsg:25832 +inv +step +init=epsg:25833 +step +init=epsg:25833 +inv +step +init=epsg:25832
pj_open_lib(proj_def.dat): call fopen(C:\OSGeo4W64\\share\proj\proj_def.dat) - succeeded

Pipeline: Building arg list for step no. 0
Pipeline: init - init=epsg:25832, 2
    inv
pj_open_lib(epsg): call fopen(C:\OSGeo4W64\\share\proj\epsg) - succeeded

Pipeline: Step 0 at 0000000002401560
Pipeline:    step done
Pipeline: Building arg list for step no. 1
Pipeline: init - init=epsg:25833, 1
pj_open_lib(epsg): call fopen(C:\OSGeo4W64\\share\proj\epsg) - succeeded

Pipeline: Step 1 at 0000000002401DA0
Pipeline:    step done
Pipeline: Building arg list for step no. 2
Pipeline: init - init=epsg:25833, 2
    inv
Pipeline: Step 2 at 00000000024022B0
Pipeline:    step done
Pipeline: Building arg list for step no. 3
Pipeline: init - init=epsg:25832, 2
    ellps=WGS84
Pipeline: Step 3 at 00000000024027C0
Pipeline:    step done
Pipeline: 4 steps built. Determining i/o characteristics
Pipeline: Units - left: [linear], right: [linear]

700000 6000000
In[00]:               700000              6000000               1.#INF - Universal Transverse
In[01]:    0.210479616228122    0.944383808172103               1.#INF - Universal Transverse
In[02]:     307788.876117106     5999669.30360376               1.#INF - Universal Transverse
In[03]:    0.210479616228122    0.944383808172103               1.#INF - Universal Transverse
Out[ ]:      699999.99999784     6000000.00012413               1.#INF
699999.999997840150000  6000000.000124132300000     1.#INF00000000000     1.#INF00000000000
````

Gives correct results if I add `+ellps=GRS80`:

```
λ bin\cct.exe +proj=pipeline +step +init=epsg:25832 +inv +step +init=epsg:25833 +step +init=epsg:25833 +inv +step +init=epsg:25832 +ellps=GRS80
pj_open_lib(proj_def.dat): call fopen(C:\OSGeo4W64\\share\proj\proj_def.dat) - succeeded

Pipeline: Building arg list for step no. 0
Pipeline: init - init=epsg:25832, 2
    inv
pj_open_lib(epsg): call fopen(C:\OSGeo4W64\\share\proj\epsg) - succeeded

Pipeline: Step 0 at 00000000027E1560
Pipeline:    step done
Pipeline: Building arg list for step no. 1
Pipeline: init - init=epsg:25833, 1
pj_open_lib(epsg): call fopen(C:\OSGeo4W64\\share\proj\epsg) - succeeded

Pipeline: Step 1 at 00000000027E1DA0
Pipeline:    step done
Pipeline: Building arg list for step no. 2
Pipeline: init - init=epsg:25833, 2
    inv
Pipeline: Step 2 at 00000000027E22B0
Pipeline:    step done
Pipeline: Building arg list for step no. 3
Pipeline: init - init=epsg:25832, 3
    ellps=GRS80
    ellps=WGS84
Pipeline: Step 3 at 00000000027E27C0
Pipeline:    step done
Pipeline: 4 steps built. Determining i/o characteristics
Pipeline: Units - left: [linear], right: [linear]

700000 6000000
In[00]:               700000              6000000               1.#INF - Universal Transverse
In[01]:    0.210479616228122    0.944383808172103               1.#INF - Universal Transverse
In[02]:     307788.876117106     5999669.30360376               1.#INF - Universal Transverse
In[03]:    0.210479616228122    0.944383808172103               1.#INF - Universal Transverse
Out[ ]:               700000              6000000               1.#INF
700000.000000000120000  5999999.999999999100000     1.#INF00000000000     1.#INF00000000000
```

I'll look into it later.

busstoptaktik:
Hooray!!

busstoptaktik:
This is so awesome I can hardly express my excitement in socially acceptable terms :-)

PR #630, opened 2017-10-25
Remove superfluous internal selftests from projection files matching …

busstoptaktik:
…patterns PJ_f....c and PJ_g....c


busstoptaktik:
(One Travis build stalls for unrelated reasons - merging immediately)

rouault:
I may have miss something but why do we need those remaining stub tests all in pj_ files ?

busstoptaktik:
@rouault: The stubs are there because the calls used by `proj -VC` are generated by the PROJ_HEAD x-macro, so it's all or nothing, and I need the tests for regression testing, so until I have removed all of them, the stubs will have to remain for those that have been removed. After that, I'll remove all of the stubs, and the pj_selftest functions.

I'll get there, but there's so much to polish up :-)

rouault:
ok, thanks for the explanation

PR #631, opened 2017-10-26
Repair and improve broken cct output routine

busstoptaktik:
The cct output routine was badly broken. Output is more reasonable now.


busstoptaktik:
Travis OSX build stalled - everything else builds fine. Merging.

PR #634, opened 2017-10-29
change suspicious loop exit condition in nad_cvt

busstoptaktik:
This loop exit condition in nad_cvt.c looks suspicious:

    while (i-- && fabs(dif.lam) > TOL && fabs(dif.phi) > TOL)

This would mean that we exit the iteration if ***either*** `dif.lam` ***or*** `dif.phi` converged. Hence, we may exit as soon as one dimension is OK, while the other may still be kilometers off.

I have changed it to

    while (--i && hypot (dif.lam, dif.phi) > TOL)

This introduces a slightly stronger convergence criterium (we must be "inside the disk", rather than "inside the box"), and as the "inside the box" criterium was apparently misimplemented, the criterium change is even stronger.

This will probably introduce small regressions in all tests involving NADCON style datum shifts.

I discovered the suspicious loop condition while reviewing PR #588, where @kbevers uses essentially the same code.  I may, however, have misunderstood subtle aspects of this, so please try to find time to review this!

### DIFF
--- a/src/nad_cvt.c
+++ b/src/nad_cvt.c
@@ -1,70 +1,70 @@
 #define PJ_LIB__
-#include <projects.h>
-#define MAX_TRY 10
+#include "projects.h"
+#define MAX_ITERATIONS 10
 #define TOL 1e-12
-	LP
-nad_cvt(LP in, int inverse, struct CTABLE *ct) {
-	LP t, tb;
+
+LP nad_cvt(LP in, int inverse, struct CTABLE *ct) {
+	LP t, tb,del, dif;
+	int i = MAX_ITERATIONS;
+	const double toltol = TOL*TOL;
 
 	if (in.lam == HUGE_VAL)
 		return in;
+
 	/* normalize input to ll origin */
 	tb = in;
 	tb.lam -= ct->ll.lam;
 	tb.phi -= ct->ll.phi;
-	tb.lam = adjlon(tb.lam - M_PI) + M_PI;
-	t = nad_intr(tb, ct);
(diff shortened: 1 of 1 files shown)


rouault:
Looks reasonable to me, although I'm wondering about the computational cost of hypot(x,y). But this might be neglectable compared to nad_intr. A lightweight equivalent of hypot(x,y) > TOL would be x*x+y*y > TOL * TOL, avoiding the cost of the square root.

kbevers (approved):
Seems good to me. Probably not a problem in practice, given that the code has been working fine for many years. If I remember correctly there's a few reports of problems near grid boundaries - perhaps this PR fixes those cases?

kbevers (approved):

busstoptaktik:
@rouault ecrit:
> Looks reasonable to me, although I'm wondering about the computational cost of hypot(x,y). But this might be neglectable compared to nad_intr.

At least that was my reason for using it, even though in this use case, the extra digits of precision from `hypot`s internal refactoring of the arguments are superfluous, so I selected ease of comprehension rather than speed of computation.

But since I will take another trip through the code to normalize some spacing etc. (stuff I wouldn't do in the first commit to avoid drowning the subject matter in unrelated cleanups), I will introduce the lightweight test as well.

It seems that all tests went well, so I think the conclusion is that datum shift grids are so smooth, that the single iteration typically taken brings us all the way to the goal, even though the exit criterion was unintentionally optimistic.

Otherwise, the error had probably been detected sooner: This code is more than 20 years old.

PR #636, opened 2017-10-30
Add 'axisswap' operation - swap axes in pipelines

kbevers:
This operation let's you swap axes across all four dimensions of a coordinate. Brings similar functionality to `cs2cs`'s `+axis` parameter to transformation pipelines. This is useful on cases where coordinates systems are defined with axis order different that X, Y, Z.

A few examples:

```
# reverse the axis order
$ echo 1 2 3 4 | ./bin/cct +proj=axisswap +order=4,3,2,1
   4.000000000000000     3.000000000000000     2.000000000000000     1.000000000000000

# switch x and y axis and reverse direction of the new x axis
$ echo 1 2 3 4 | ./bin/cct +proj=axisswap +order=-2,1,3,4
  -2.000000000000000     1.000000000000000     3.000000000000000     4.00000000000000

# reverse direction on all axes
$ echo 1 2 3 4 | ./bin/cct +proj=axisswap +order=-1,-2,-3,-4
  -1.000000000000000    -2.000000000000000    -3.000000000000000    -4.000000000000000

# by default output units are meters...
$ echo 12 55 0 0 | ./bin/cct +proj=pipeline +step +proj=latlon +step +proj=axisswap +order=1,2,3,4
   0.209439510239320     0.959931088596881     0.000000000000000     0.000000000000000

# ... but can be changed to radians with +angularunits
$ echo 12 55 0 0 | ./bin/cct +proj=pipeline +step +proj=latlon +step +proj=axisswap +order=1,2,3,4 +angularunits
  12.000000000000000    55.000000000000000     0.000000000000000     0.000000000000000
```

Rules for axis ordering:

1. No repetitions
2. At least two axes in the list
3. Maximum four axes in the list
4. The highest axis number allowed is equal to the number of axes in the order list (e.g. "3,1" and "4,3,1" are not allowed)


rouault:
How hard would it be to require a comma separator instead of just concatenated numbers ? +order=4321 might be OK, but +order=-4-3-2-1 looks a bit strange to me. +order=-4,-3,-2,-1 would look better

kbevers:
I don't think it is that hard to do. I think the easiest is to make the comma optional, so that you can do both`+order=4321`, and `+order=4,3,2,1` (and of course `+order=-4,-3,-2,-1`.

Other ways to describe the axes are:

1. By using "named" axes such as `xyztXYZT` where capital letters describe a reversed axis:
      `+order=xyzt` (normal case), `+order=YXzt` (x and y swapped, both reversed).
2. Separate axis order and axis sign:
     `+order=yxzt +sign=nnpp` (x and y swapped, both reversed, "n" for negative, "p" for positive)

busstoptaktik:
> Other ways to describe the axes (...)

I would prefer to keep it at 1234/4321/2143 etc.: Historically all kinds of different names have been used for mapping axes, and as we are probably not striving for completeness, I think the abstract numeric naming is the overall least confusing option.

Overall, this PR is important because it takes the first step toward handling coordinate order according to the order used in the individual coordinate system definition. Although we would probably never try to handle all potential combinations natively, at least having a way of specifying what permutations/reflections must be done to arrive at an right-handed-ish system is a neat start.

kbevers:
I prefer the numbered axes as well. @rouault what do you think about the current implementation where it is possible to list the axes with and without comma-separation?

rouault:
I don't have more comments than my initial ones. Personnaly, I especially dislike the -1-2-3-4 syntax with negative numbers without separator, but I'm just -0 on it.

src/PJ_axisswap.c
```diff
@@ -0,0 +1,227 @@
...
+        sign = 1;
+    }
+
+    P->fwdobs = forward_obs;
+    P->invobs = reverse_obs;
+    P->fwd3d  = forward_3d;
+    P->inv3d  = reverse_3d;
+    P->fwd    = forward_2d;
+    P->inv    = reverse_2d;
+
```
busstoptaktik: Just before the code block above, it would make sense to add a check for whether the swap was valid for a given number of dimensions, and then using null pointers for the corresponding  fwd/inv combo if not.

Since pj_fwd, pj_inv, pj_fwd3d, pj_inv3d, pj_fwdobs and pj_invobs all check for nulls, and return appropriate error objects, you could eliminate the `if (Q->axis[i] < 3)` style checks, and get better run time performance, at the cost of a minimally longer setup time.

rouault:
Actually +order=-2134 is super confusing. It could be very much interpreted as 2134 and then negate all axis, instead of negating only the second axis.

src/proj_4D_api.c
```diff
@@ -342,16 +342,22 @@ PJ  *proj_create_crs_to_crs (PJ_CONTEXT *ctx, const char *srid_from, const char
...
+    if (!strchr(srid_from, ':') || !strchr(srid_to, ':')) {
+        proj_log_error(P, "input does not appear to be a valid SRID");
+        return 0;
+    }
+
     strcpy(buffer, "+proj=pipeline +step +init=");
     strncat(buffer, srid_from, 512-strlen(buffer));
     strncat(buffer, " +inv +step +init=", 512-strlen(buffer));
     strncat(buffer, srid_to, 512-strlen(buffer));
+    printf("%s\n", buffer);
```
busstoptaktik: Leftover debugging code?

kbevers: Yes. Should not have been included in this commit.

src/PJ_axisswap.c
kbevers: Good suggestion. I'll include that.

kbevers:
> Actually +order=-2134 is super confusing. It could be very much interpreted as 2134 and then negate all axis, instead of negating only the second axis.

I interpret that as you'd prefer always separating by commas. I agree that is is less confusing when having negative axes. For the sake of consistency I'll require comma-separation instead of leaving it optional.

rouault:
> I interpret that as you'd prefer always separating by commas

At least when there is a negative number

kbevers:
Okay, I changed the ordering input so that axes has to be separated by commas. This made it easier for me to also accept a list of only two or three axes. So now it is also possible to do 
`+proj=axisswap +order=2,1` and `+proj=axisswap +order=3,2,1` as well.

PR #637, opened 2017-10-31
Pipeline cleanup (WIP)

busstoptaktik:
Some debugging cruft was left over in the inner loop of the pipeline drivers `pipeline_forward_obs`  and `pipeline_reverse_obs`, which meant that for each step in the pipeline, an extra function call and some additional futile memory management, was incurred.

While the debugging info at high debug levels could be useful, really, debugging a pipeline should be done by adding probe steps to the pipeline, printing out intermediate results in between the payload steps.

Only problem is that such probes do not exist yet. Basically, they should work as the code removed in this PR, while having the IO type `PJ_IO_UNITS_ANY` which does not exist either, but is anticipated in line 166 of `PJ_pipeline.c`.

Hence, this PR is just a first step. It appears that a lot needs to be done to properly clean up the pipeline code. First and foremost, we should disallow the "one directional steps" (`omit_fwd` and `omit_inv`), since we have yet to see a use case.

Next we need to go over all the classic projections once more, and make them return values in meters, rather than in units of the semimajor axis. This will let us eliminate the `PJ_IO_UNITS_CLASSIC` flag, and make projections honest about what they return.

With this in place, we can let the `pj_init` functions handle the `+inv` flag, by swapping the right and left indicators, and the inverse and forward functions. The latter can, however,  be done for the 4D cases only, since `pj_fwdobs` and `pj_invobs` functions have identical prototypes, while fwd and inv differ for the two and three dimensional stuff.

All in all an enormous, but potentially quite rewarding task.

But for now, I will take the more pragmatic step of introducing *two* probe functions - one for printing linear coordinates and one for printing angular. This sidesteps the PJ_IO_UNITS_ANY problem.

Also, the one-way steps should be eliminated, and the entire "swap funcs and flags" thing can be handled by the pipeline code, since it uses proj_trans_obs, which in turn calls pj_fwdobs and pj_invobs to do the heavy lifting. All in all, this will dramatically reduce the complexity of the pipeline code, while keeping the changes localized. It will probably also result in a slight speed improvement.

### DIFF
--- a/src/proj.def
+++ b/src/proj.def
@@ -146,6 +146,3 @@ EXPORTS
     proj_list_ellps                @132
     proj_list_units                @133
     proj_list_prime_meridians      @134
-
-    pj_left                        @135
-    pj_right                       @136
--- a/src/proj.def
+++ b/src/proj.def
@@ -146,3 +146,6 @@ EXPORTS
     proj_list_ellps                @132
     proj_list_units                @133
     proj_list_prime_meridians      @134
+
+    pj_left                        @135
+    pj_right                       @136
--- a/src/PJ_pipeline.c
+++ b/src/PJ_pipeline.c
@@ -107,20 +107,18 @@ Thomas Knudsen, thokn@sdfe.dk, 2016-05-20
 PROJ_HEAD(pipeline,         "Transformation pipeline manager");
 
 /* Projection specific elements for the PJ object */
-#define PIPELINE_STACK_SIZE 100
 struct pj_opaque {
     int reversible;
     int steps;
-    int depth;
     int verbose;
     int *reverse_step;
     int *omit_forward;
     int *omit_inverse;
     char **argv;
     char **current_argv;
-    PJ_OBS stack[PIPELINE_STACK_SIZE];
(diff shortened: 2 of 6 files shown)


busstoptaktik:
Since this is primarily a cleanup, I think the next steps of this will become more straightforward, if I (rather than continuing development of this PR),  merge it now, then merge #639, and then go on with the introduction of probes. @kbevers and @rouault: Any objections to this?

src/proj.def
```diff
@@ -146,3 +146,6 @@ EXPORTS
     proj_list_ellps                @132
     proj_list_units                @133
     proj_list_prime_meridians      @134
+
+    pj_left                        @135
```
kbevers: These functions are private, right? If that's the case, they shouldn't be here.

kbevers (changes requested):
No objection from me, apart from the one in-lined comment I've made.

busstoptaktik:
> No objection from me, apart from the one in-lined comment I've made.

Done

kbevers (approved):

PR #639, opened 2017-11-04
Improve ISO 19000 alignment

busstoptaktik:
Predating the ISO 19000 series by more than a decade, it is not surprising that PROJ.4 concepts are not entirely aligned with ISO 19000., and hence are slightly out of tune with how the general geoprocessing software stack views the world.

There is, however, plenty of low hanging fruit, where we can significantly improve the alignment by small code adjustments. This is a first example, introducing explicit conceptual differentiation between conversions and transformations.

In ISO 19000 lingo, a ***conversion*** is of mathematical nature, and hence in principle *exact*, while a ***transformation*** is based on physical observations, and hence depending on our present knowledge of the environment.

So e.g. projections are conversions, while datum shifts are typically transformations. The union of conversions and transformations is called ***operations***.

This PR grew from a discussion with @kbevers about how to tell the PROJ.4 plumbing that a given operation did not need access to ellipsoid information. The problem is that PROJ.4 init functionality insists that a spheroid is specified, but in cases like the Helmert transformation, it is not needed, and you must specify a dummy.

So to introduce a better communication between the `PJ_***.c` individual operation implementation code, and the `pj_init.c` plumbing code,  let's introduce a new setup macro `OPERATION`, replacing `PROJECTION`, which now becomes a specialization (i.e. a *partial evaluation*) of `OPERATION`.

### DIFF
--- a/src/PJ_helmert.c
+++ b/src/PJ_helmert.c
@@ -461,7 +461,7 @@ static PJ_OBS helmert_reverse_obs (PJ_OBS point, PJ *P) {
 #define ARCSEC_TO_RAD (DEG_TO_RAD / 3600.0)
 
 /***********************************************************************/
-PJ *PROJECTION(helmert) {
+PJ *OPERATION(helmert, PJ_IO_UNITS_METERS, PJ_IO_UNITS_METERS, 0) {
 /***********************************************************************/
     struct pj_opaque_helmert *Q = pj_calloc (1, sizeof (struct pj_opaque_helmert));
     if (0==Q)
--- a/src/PJ_helmert.c
+++ b/src/PJ_helmert.c
@@ -461,7 +461,7 @@ static PJ_OBS helmert_reverse_obs (PJ_OBS point, PJ *P) {
 #define ARCSEC_TO_RAD (DEG_TO_RAD / 3600.0)
 
 /***********************************************************************/
-PJ *OPERATION(helmert, PJ_IO_UNITS_METERS, PJ_IO_UNITS_METERS, 0) {
+PJ *TRANSFORMATION(helmert, PJ_IO_UNITS_METERS, PJ_IO_UNITS_METERS, 0) {
 /***********************************************************************/
     struct pj_opaque_helmert *Q = pj_calloc (1, sizeof (struct pj_opaque_helmert));
     if (0==Q)
--- a/src/PJ_helmert.c
+++ b/src/PJ_helmert.c
@@ -461,7 +461,7 @@ static PJ_OBS helmert_reverse_obs (PJ_OBS point, PJ *P) {
 #define ARCSEC_TO_RAD (DEG_TO_RAD / 3600.0)
 
 /***********************************************************************/
-PJ *TRANSFORMATION(helmert, PJ_IO_UNITS_METERS, PJ_IO_UNITS_METERS, 0) {
+PJ *TRANSFORMATION(helmert, 0) {
 /***********************************************************************/
     struct pj_opaque_helmert *Q = pj_calloc (1, sizeof (struct pj_opaque_helmert));
     if (0==Q)
--- a/src/projects.h
+++ b/src/projects.h
@@ -327,6 +327,7 @@ struct PJconsts {
     int  geoc;                      /* Geocentric latitude flag */
     int  is_latlong;                /* proj=latlong ... not really a projection at all */
     int  is_geocent;                /* proj=geocent ... not really a projection at all */
+    int  need_ellps;                /* 0 for operations that are purely cartesian */
 
     enum pj_io_units left;          /* Flags for input/output coordinate types */
     enum pj_io_units right;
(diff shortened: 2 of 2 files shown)


src/PJ_helmert.c
```diff
@@ -461,7 +461,7 @@ static PJ_OBS helmert_reverse_obs (PJ_OBS point, PJ *P) {
 #define ARCSEC_TO_RAD (DEG_TO_RAD / 3600.0)
 
 /***********************************************************************/
-PJ *PROJECTION(helmert) {
+PJ *OPERATION(helmert, PJ_IO_UNITS_METERS, PJ_IO_UNITS_METERS, 0) {
```
kbevers: Surely this is a `TRANSFORMATION`

src/projects.h
```diff
@@ -327,6 +327,7 @@ struct PJconsts {
     int  geoc;                      /* Geocentric latitude flag */
     int  is_latlong;                /* proj=latlong ... not really a projection at all */
     int  is_geocent;                /* proj=geocent ... not really a projection at all */
+    int  need_ellps;                /* 0 for operations that are purely cartesian */
```
kbevers, line 330: Does `need_ellps` also imply that the output has to be scaled by the semimajor axis?

src/projects.h
```diff
@@ -597,14 +600,27 @@ C_NAMESPACE PJ *pj_##name (PJ *P) {                          \
...
+    P->need_ellps = NEED_ELLPS;                              \
+    P->left  = LEFT;                                         \
+    P->right = RIGHT;                                        \
     return P;                                                \
 }                                                            \
+                                                             \
 PJ *pj_projection_specific_setup_##name (PJ *P)
+
+/* In ISO19000 lingo, an operation is either a conversion or a transformation */
+#define CONVERSION(name, left, right, need_ellps)      OPERATION (name, left, right, need_ellps)
```
kbevers: Wouldn't it be simpler to assume that the units are meters on both sides for `CONVERSION`'s and `TRANSFORMATIONS`'s? In the rare case that something different is needed it can be changed in the operation setup.

kbevers (commented):
It looks good to me. When we discussed this earlier I imagined setting the type more explicitly on the PJ object. Perhaps by having a `type` member, that would tell calling functions if this is a projection, conversion or transformation. With the way you've set it up now that is probably not necessary, but it could make it easier to compile a list of the various kinds of operations. This could be useful for users of the library.

src/PJ_helmert.c
busstoptaktik: Yes (but also an operation, since operation is the union of transformations and conversions). Will, however, update, since there is no reason not using the most tight concept.

src/projects.h
busstoptaktik: No - and it doesn't even mean that it is evident from the outside that an ellipsoid is needed:

Consider e.g. a hypothetical  case where you do a correction of cartesian coordinates.

Input is cartesian, output is cartesian, so no need for ellipsoid is expected.

But the correction grid is in geographical coordinates, so you need the ellipsoid to convert to geographical, in order to fetch the correction to apply to the cartesian.

`need_ellps`  is really just a flag to allow the plumbing  in `pj_init.c` to relax the initial check for whether a spheroid is specified.

In general, I would like to get rid of the scaled output from projections, since i complicates lots of other stuff, but it takes another tour through the entire code base, so I would prefer to have more to do than just adding `xy.y *= P->a` here and there.

src/projects.h
busstoptaktik: For most conversions it would be angular in, linear out, since most conversions are projections, and all projections are conversions, so in most cases the left and right arguments will be partially evaluated through the definition of the `PROJECTION` macro.

I actually think that direct calls to the `CONVERSION` macro will be rare, but it plays a role in making the code architecture reflect the conceptual model of ISO19000.

The transformations from the geodetic toolbox may be mostly linear in both ends, although e.g. Molodensky is angular in both ends, so I find it useful to have it comparatively clearly stated in the macro, (which amounts to a "kind of function header"), what kinds of I/O goes on.

src/projects.h
kbevers: Good, just wanted to check that we had the same thing in mind. Your case is not very hypothetical since that is exactly what the `deformation` transformation does :)

> In general, I would like to get rid of the scaled output from projections, since i complicates lots of other stuff

Would that also be the case if there was information on the type of operation attached to each `PJ`? That would enable you to skip the scaling logic when not processing a projection.

src/projects.h
kbevers: > I actually think that direct calls to the CONVERSION macro will be rare, but it plays a role in making the code architecture reflect the conceptual model of ISO19000.

`unitconvert` and `axisswap`, but yes, they are likely to be the only ones.

> The transformations from the geodetic toolbox may be mostly linear in both ends, although e.g. Molodensky is angular in both ends, so I find it useful to have it comparatively clearly stated in the macro, (which amounts to a "kind of function header"), what kinds of I/O goes on.

I don't feel too strongly about this either way, but I do prefer having a simple setup for most cases and paying a bit of extra attention in the few special cases like the Molodensky. 
Also, I think that the macro-part of the `PJ` initialisation is best reserved for things that are needed right away in `pj_init` such as `need_ellps` and the name of the operation. While important, I don't think the units fall in that category.

busstoptaktik:
> With the way you've set it up now that is probably not necessary, but it could make it easier to compile a list of the various kinds of operations. This could be useful for users of the library.

Yes pretty useful. Should probably go in another PR?

PR #641, opened 2017-11-05
Make tol a parameter in pj phi2 (lcc, merc, omerc, gstmerc)

busstoptaktik:
In a recent mailing list discussion, Dmitry Mefed revealed that the roundtrip precision was insufficient for Lambert Conformal Conic (lcc).

The solution was to change `TOL` in `pj_phi2.c` from 1e-10 to 1e-13 and recompiling.

This makes the iteration in `pj_phi2` take a few extra steps, which affects run time - not only for lcc, but also for merc, omerc, gstmerc, and calcofi.

To avoid that, this PR introduces a separate setup parameter `tol=`, indicating the desired precision level., for lcc, merc, omerc and gstmerc.

The tol parameter defaults to 1e-10 (the original value from `pj_phi2.c`), so not specifying it will result in status quo. It has been left out from the CalCOFI case, due to its rather special, and probably not precision intensive, use case.

It would probably be better to let the `tol=` parameter somehow indicate the target precision, rather than a more abstact (although related) convergence target, but I'm not sure how to do this in a sufficient general way, since that will probably mean that `tol=` should become a general PROJ.4 parameter. So I have left this aspect out for now.

### DIFF
--- a/src/PJ_calcofi.c
+++ b/src/PJ_calcofi.c
@@ -57,7 +57,7 @@ static XY e_forward (LP lp, PJ *P) {          /* Ellipsoidal, forward */
     l1 = (xy.y - oy) * tan(ROTATION_ANGLE);
     l2 = -xy.x - l1 + PT_O_LAMBDA;
     ry = l2 * cos(ROTATION_ANGLE) * sin(ROTATION_ANGLE) + xy.y;
-    ry = pj_phi2(P->ctx, exp(-ry), P->e); /*inverse Mercator*/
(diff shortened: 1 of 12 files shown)


kbevers:
You seem to have based this branch on the branch behind #637. Should be cleaned up before merging. Otherwise I think it is nice non-intrusive addition.

busstoptaktik:
> You seem to have based this branch on the branch behind #637. Should be cleaned up before merging. Otherwise I think it is nice non-intrusive addition.

Oops - yes, will have to look into this...

busstoptaktik:
This was an experiment which I will not have time to continue. Closing immediately.

PR #642, opened 2017-11-05
Pipeline initialisation improvements

kbevers:
Move pipeline initialisation logic to PJ_pipeline.c and decrease the number of special cases to handle in pj_init.c.

With the changes made in #639 it is now easier to take control of the pipeline initialisation as part of the setup of that operation instead of doing it as special cases in pj_init. With these changes ellipsoid setup is handled better for pipelines than previously.

### DIFF
--- a/src/PJ_axisswap.c
+++ b/src/PJ_axisswap.c
@@ -127,36 +127,31 @@ static LPZ reverse_3d(XYZ xyz, PJ *P) {
 }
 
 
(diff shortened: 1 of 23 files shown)


busstoptaktik:
Oops - could you wait a bit merging this?  I have been working for some hours on an extremely complex PR that cleans up the last traces of PJ_OBS. This also touches most of the files you have up in the air (also pipeline), but perhaps even more.

kbevers:
Sure. There's a few more tests I'd like to do before I merge it anyway.

PR #643, opened 2017-11-05
Eliminate the last traces of PJ_OBS

### LINKED ISSUE #627: Eliminate PJ_OBS entirely
In continuation of #597, which eliminated PJ_OBS from the API level, the next step is to eliminate PJ_OBS from the deeper waters of the code base.

@kbevers did a quick check this morning, and it appears that most of the work will be search-and-replace style. But there are also a few devils in the details, making it not entirely routine work.

busstoptaktik:
PJ_OBS eliminated, API adjusted to reflect that we now have only one 4D data type, so we do not have to decorate API calls with "_obs" and "_coord"

Fixes #627 and simplifies a lot of stuff.

### DIFF
--- a/src/PJ_axisswap.c
+++ b/src/PJ_axisswap.c
@@ -127,36 +127,31 @@ static LPZ reverse_3d(XYZ xyz, PJ *P) {
 }
 
 
-static PJ_OBS forward_obs(PJ_OBS obs, PJ *P) {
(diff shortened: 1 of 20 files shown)


busstoptaktik:
OK - this builds cleanly and passes all tests. I think it is ready to merge. @kbevers would you want to check how hard this PR stomps on your #642 before I press the merge button?

PR #644, opened 2017-11-06
Use enumerations where appropriate

aaronpuchert:
Enumerations have the following advantages over #define:
- they clearly connect a variable and the allowed constants,
- the meaning of code is not obfuscated by integer values,
- they are visible to the compiler, which can warn about (possibly) incorrect usage.

There should be no functional change.


kbevers (approved):
I approve of these changes 👍 

Thanks.

busstoptaktik (approved):
+1

PR #646, opened 2017-11-07
Make proj_roundtrip do a real TrLib style Poder dual autochecking dance.

busstoptaktik:
The backward-forward combo implemented in `proj_roundtrip` originates from a 1980s idea by Knud Poder, which he implemented in an enigmatic, yet elegant way in [TrLib](https://bitbucket.org/KMS/trlib).

Until now, the PROJ.4 implementation has provided means for evaluating algorithmic stability only: It returns a roundtrip deviation, without returning the result of the transformation.

This PR remedies that, by changing the `PJ_COORD`argument to `PJ_COORD *`, and writing the first half-step result back to the input argument.

Hence, `proj_roundtrip` now provides a transformation result, with an associated stability check.

### DIFF
--- a/src/PJ_cart.c
+++ b/src/PJ_cart.c
@@ -318,8 +318,8 @@ int pj_cart_selftest (void) {
     b = proj_trans (P, PJ_FWD, a);
 
     /* Check roundtrip precision for 10000 iterations each way */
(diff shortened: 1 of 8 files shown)


busstoptaktik:
Closing this for now - will reimplement after #648 has been merged

PR #652, opened 2017-11-10
Poder autochecking again (WIP)

busstoptaktik:
A re-run of #646:

**Make proj_roundtrip do a real TrLib style *Poder dual autochecking* dance.**

The backward-forward combo implemented in proj_roundtrip originates from a 1980s idea by [Knud Poder](http://lists.maptools.org/pipermail/proj/2015-October/007288.html), which he implemented in an enigmatic, yet elegant way in [TrLib](https://bitbucket.org/KMS/trlib).

Until now, the PROJ.4 implementation has provided means for evaluating algorithmic stability only: It returns a roundtrip deviation, without returning the result of the transformation.

This PR remedies that, by changing the `PJ_COORD` argument to `PJ_COORD *`, and writing the first half-step result back to the input argument.

Hence, proj_roundtrip now provides a transformation result, with an associated stability check.

The material is marked as WIP, because there is a known bug, when roundtripping on grid operations. So do not merge until that has been resolved.


test/gie/deformation.gie
```diff
@@ -64,10 +64,35 @@ ROUNDTRIP   1000
...
+TOLERANCE   1 um
 ACCEPT      -3004295.5882503074 -1093474.1690603832 5500477.1338251457 2000.0
 EXPECT      -3004295.5888766116 -1093474.1688513425 5500234.008855661  2000.0
 ROUNDTRIP   1000
+
+
+
+
+------------------------------------------------------------------------------
+operation +proj=vgridshift +grids=egm96_15.gtx +ellps=GRS80
```
kbevers: I'd prefer to put this and the one below in their own seperate files. I am working on this already.

busstoptaktik: Fine with me - I only put them in temporally, to get an extra check because I get  errors with the roundtrips of the deformation stuff in this PR, and I'm quite stuck in the debugging of the problem.

kbevers: Alright, carry on then :)

PR #659, opened 2017-11-12
Remove selftests from projection files starting with r,s,t,u

busstoptaktik:


busstoptaktik:
The Travis CI builds succeed, but this does not get reflected to github. Merging anyway

PR #664, opened 2017-11-13
Support numerical factors only

busstoptaktik:
Support for the "special" entry of PJ, which enabled PROJ.4 to compute analytically derived, rather than numerically approximated, values for a. o. scale factors and convergence, was limited to the `PJ_lcc.c` and `PJ_eqdc.c` projection implementations.

And for the past 20 years, the need for that functionality has not been so huge as to animate anyone to implement it for more than those two projections - so the need is apparently limited.

It could even be argued that it is more relevant to get the numerical approximation for the *actually implemented algorithm*, than the analytically derived values for the *mathematical abstract* version.

Removing the feature simplifies the code a bit, and also reduces the documentation needed to be written a comparative bit.

**Note:** This PR does not touch the availability of the convergence, etc. computations - it just results in treating all projections equal, by using numerical approximation even in those two cases that formerly provided analytical versions.

### DIFF
--- a/src/projects.h
+++ b/src/projects.h
@@ -257,9 +257,6 @@ struct PJconsts {
     PJ_COORD (*fwd4d)(PJ_COORD, PJ *);
     PJ_COORD (*inv4d)(PJ_COORD, PJ *);
 
-
-    void (*spc)(LP, PJ *, struct FACTORS *);
-
     void *(*destructor)(PJ *, int);
 
 
@@ -460,14 +457,7 @@ struct FACTORS {
     double conv;           /* convergence */
     double s;              /* areal scale factor */
     double a, b;           /* max-min scale error */
-    int code;              /* info as to analytics, see following */
 };
-
-
-#define IS_ANAL_XL_YL 01    /* derivatives of lon analytic */
-#define IS_ANAL_XP_YP 02    /* derivatives of lat analytic */
-#define IS_ANAL_HK    04    /* h and k analytic */
-#define IS_ANAL_CONV 010    /* convergence analytic */
 
 /* datum_type values */
 #define PJD_UNKNOWN   0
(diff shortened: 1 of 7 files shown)


kbevers:
Have you tested how the analytical and numerical calculations compare?

busstoptaktik:
@kbevers : they are indiscernible to within the number of decimals provided by `proj.exe`:
                                                                                        
    $ echo 12 55 | bin\proj -S  +proj=lcc +ellps=GRS80
    794682.65       6469768.20      <1.03785 1.03785 1.07714 0.000285335 1.03785 1.03785>   
    
    $ echo 12 55 | proj -S  +proj=lcc +ellps=GRS80
    794682.65       6469768.20      <1.03785 1.03785 1.07714 0.000285335 1.03785 1.03785>   

    $ echo 12 55 | proj -S  +proj=eqdc +lat_1=45 +ellps=GRS80                       
    858340.13       6130795.52      <1 1.11887 1.11887 6.43214 1.11887 1>                   
                                                                                        
    $ echo 12 55 | bin\proj -S  +proj=eqdc +lat_1=45 +ellps=GRS80                  
    858340.13       6130795.52      <1 1.11887 1.11887 6.43214 1.11887 1>                   

(bin\proj is the numerical version, proj is the older supporting analytical)

src/projects.h
```diff
@@ -460,14 +457,17 @@ struct FACTORS {
     double conv;           /* convergence */
     double s;              /* areal scale factor */
     double a, b;           /* max-min scale error */
-    int code;              /* info as to analytics, see following */
+    int    code;           /* always 0 */
+};
+
+enum deprecated_constants_for_now_dropped_analytical_factors {
```
kbevers, line 463: I may have missed something, but why are these necesary still?

busstoptaktik:
@kbevers, they are necessary still because user programs may be checking them. By ensuring that `fac->code==0`, these programs will still get the right answer, that "these factors were computed numerically"

PR #669, opened 2017-11-16
Introduce geodetic-geocentric conversions ...

busstoptaktik:
as PJ_xxx style conversion steps and as API entry points


src/PJ_geoc.c
```diff
@@ -0,0 +1,56 @@
+/******************************************************************************
+ * Project:  PROJ.4
+ * Purpose:  Transfornation from geographic to geocentric latitude and back.
```
kbevers: Transfor_m_ation

src/proj.h
```diff
@@ -381,7 +380,9 @@ size_t proj_trans_generic (
     double *t, size_t st, size_t nt
 );
 
-int proj_trans_array (PJ *P, PJ_DIRECTION direction, size_t n, PJ_COORD *coord);
+
+/* Geographical to geocentric latitude */
+PJ_COORD proj_geoc_lat (PJ *P, PJ_DIRECTION direction, PJ_COORD coo);
```
kbevers: Is this really meant to go in the public API?

src/PJ_geoc.c
busstoptaktik: Well - actually *Conversion*, but yes - wrong, and not even "just a speeling missteak"

src/proj.h
busstoptaktik: Yes, that was my intention: We want to make it possible for an app to massage its input data into the form PROJ.4 expects, and we do not want the app to need to know which specific ellipsoid was selected (which can possibly have happened externally or through user input at a different access level).

Will update proj.def

busstoptaktik:
Coverage reported by coveralls dropped 6% - odd, and probably not directly related to this PR.

UPDATE: yes, it actually had - I accidentally deleted the call to execute the temporarily builtin "non-gie-formatted" tests from more_builtins.gie, when introducing the tests for geocentric/geodetic latitude conversion. Will update in a moment

PR #673, opened 2017-11-17
Increment version number to 5.0.0

kbevers:
By popular demand, I have updated the version number to 5.0.0.

Unfortunately it is spread across several files, so I hope I updated it everywhere. Also, the format of the version number in `proj_api.h` is a bit convoluted - have I got right?

### DIFF
--- a/src/proj_api.h
+++ b/src/proj_api.h
@@ -28,19 +28,14 @@
 
 
  /*
-  * This version number should be updated with every release!  The format of
-  * PJ_VERSION is
+  * This version number should be updated with every release!
   *
-  * * Before version 4.10.0: PJ_VERSION=MNP where M, N, and P are the major,
-  *   minor, and patch numbers; e.g., PJ_VERSION=493 for version 4.9.3.
+  * This file is expected to be removed from the PROJ distribution
+  * when a few minor-version releases has been made.
   *
-  * * Version 4.10.0 and later: PJ_VERSION=MMMNNNPP later where MMM, NNN, PP
-  *   are the major, minor, and patch numbers (the minor and patch numbers
-  *   are padded with leading zeros if necessary); e.g., PJ_VERSION=401000
-  *   for version 4.10.0.
   */
  #ifndef PJ_VERSION
- #define PJ_VERSION 493
+ #define PJ_VERSION 500
  #endif
 
 
--- a/src/pj_release.c
+++ b/src/pj_release.c
@@ -2,7 +2,7 @@
 
 #include <projects.h>
 
-char const pj_release[]="Rel. 4.9.3, 15 August 2016";
+char const pj_release[]="Rel. 5.0.0, development version";
 
 const char *pj_get_release()
 
--- a/src/proj.h
+++ b/src/proj.h
@@ -135,31 +135,10 @@
 extern "C" {
 #endif
 
-/************************************************************************
-    These version numbers should be updated with every release!
-    The format of PJ_VERSION is
-
-    * Before version 4.10.0: PJ_VERSION=MNP
-      where M, N, and P are the major, minor, and patch numbers;
-      e.g., PJ_VERSION=493 for version 4.9.3.
-
-    * Version 4.10.0 and later: PJ_VERSION=MMMNNNPP later
-      where MMM, NNN, PP are the major, minor, and patch numbers.
-      The minor and patch numbers are padded with leading zeros if
-      necessary);
-      e.g., PJ_VERSION=401000 for version 4.10.0.
-************************************************************************/
-#define PROJ_VERSION_MAJOR 10
+/* The version numbers should be updated with every release! **/
+#define PROJ_VERSION_MAJOR 5
 #define PROJ_VERSION_MINOR 0
 #define PROJ_VERSION_PATCH 0
-
-#ifndef PJ_VERSION
-#define PJ_VERSION 10##000##00
-#endif
-#ifndef PROJ_VERSION
-#define PROJ_VERSION PJ_VERSION
-#endif
-
 
 extern char const pj_release[]; /* global release id string */
 


rsbivand:
Thanks!

rsbivand:
Will the acid test be that < 5, there is no config-proj --version at all (on systems supporting config-*) but that proj_api.h declares the version? For ./configure, confirmation that config-proj --version returns the authoritative version number if it exists is sufficient.

kbevers:
As far as I know there's not a `config-proj` in the coming release. Maybe there should be one? Anyway, I guess you check the version number specified in whatever include file you are using (either `proj_api.h` or `proj.h`). On the command line you can run `proj` without any arguments and it will return the version number.

rsbivand:
There is proj.pc.in in the root directory ... commented [here](https://github.com/OSGeo/proj.4/issues/3) I think. 

When running ./configure, there may be no installed proj. In rgdal we use this [configure.ac](https://r-forge.r-project.org/scm/viewvc.php/pkg/configure.ac?view=markup&revision=681&root=rgdal), with this program stub to find the version number used for conditioning other checks (for files in /share, mostly, and I'm not sure that all the version checking logic holds either):

```
# PROJ.4

AC_ARG_WITH([proj-include],
    AC_HELP_STRING([--with-proj-include=DIR],
                   [location of proj header files]),
    [proj_include_path=$withval])
if test [ -n "$proj_include_path" ] ; then
   AC_SUBST([PKG_CPPFLAGS],["-I${proj_include_path} ${INPKG_CPPFLAGS}"])
fi

# honor PKG_xx overrides
# for CPPFLAGS we will superfluously double R's flags
# since we'll set PKG_CPPFLAGS with this, but that shouldn't hurt
CPPFLAGS="${INCPPFLAGS} ${PKG_CPPFLAGS}"


          
# dnl ditto for a library path
AC_ARG_WITH([proj-lib],
    AC_HELP_STRING([--with-proj-lib=LIB_PATH],
               [the location of proj libraries]),
               [proj_lib_path=$withval])
if test [ -n "$proj_lib_path" ] ; then
    AC_SUBST([PKG_LIBS], ["-L${proj_lib_path} ${INPKG_LIBS}"])
fi


LIBS="${INLIBS} ${PKG_LIBS}"
PKG_LIBS="${PKG_LIBS} -lproj"
AC_MSG_NOTICE([${CFLAGS} ${PKG_CPPFLAGS}])
AC_MSG_CHECKING(proj_api.h presence and usability)
# FIXME re-using names
[cat > proj_conf_test1.c <<_EOCONF
#include <stdio.h>
#include <proj_api.h>

int main() {
    printf("%d\n", PJ_VERSION);
    exit(0);
}
_EOCONF]

if test `${CC} ${CFLAGS} ${PKG_CPPFLAGS} -o proj_conf_test1 proj_conf_test1.c ${PKG_LIBS} 2> /dev/null; echo $?` = 0; then
  AC_MSG_RESULT(yes)
else
  AC_MSG_RESULT(no)
  AC_MSG_ERROR([proj_api.h not found in standard or given locations.])
fi

proj_version=`./proj_conf_test1`

AC_MSG_CHECKING([PROJ.4 version: ])
if test ${proj_version} -eq 480; then
  AC_MSG_RESULT([4.8.0])
  AC_SUBST([PKG_CPPFLAGS],["-DP4CTX=1 ${PKG_CPPFLAGS}"])
else if test ${proj_version} -ge 490; then
  AC_MSG_RESULT([> 4.8.0])
  AC_SUBST([PKG_CPPFLAGS],["-DP4CTX=1 ${PKG_CPPFLAGS}"])
else
  AC_MSG_RESULT([< 4.8.0])
fi
fi
```

rsbivand:
I wrote in an issue on r-spatial/sf:

>  I've found an installed {--prefix}/lib/pkgconfig/proj.pc file, but no further build of that to proj-config. Does anyone know how to create a stand-alone script?
> 
> ```
> $ pkg-config --exists proj
> $ echo $?
> 1
> $ pkg-config --with-path=/usr/local/lib/pkgconfig --exists proj
> $ echo $?
> 0
> $ pkg-config --with-path=/usr/local/lib/pkgconfig --modversion proj
> 4.9.3
> $ pkg-config --with-path=/usr/local/lib/pkgconfig --libs proj 
> -L/usr/local/lib -lproj 
> $ pkg-config --with-path=/usr/local/lib/pkgconfig --cflags-only-I proj
> -I/usr/local/include 
> ```
> 
> work, but aren't obvious unless /usr/local/lib/pkgconfig is known. Can we work with this and a configure-arg for --with-path, falling back on the include and libs configure args if it fails?

micahcochran:
`proj_api.h` is the legacy PJ_VERSION format. It should be changed to the new `proj.h` format, which is a [semantic version number](http://semver.org/).

That changed was made in anticipation of a 4.10.0 version.

The comments above the PJ_VERSION number format should be updated to read 5.0.0 instead of 4.10.0.  Since, there will not be a version 4.10.0.

rsbivand:
Re. config: please also see comments on [this issue](https://github.com/r-spatial/sf/issues/545) in r-spatial/sf, and a [GEOS](https://github.com/OSGeo/geos/blob/master/tools/geos-config.in) example.

kbevers:
@micahcochran This is what is says in `proj_api.h`:

```
  /*
   * This version number should be updated with every release!  The format of
   * PJ_VERSION is
   *
   * * Before version 4.10.0: PJ_VERSION=MNP where M, N, and P are the major,
   *   minor, and patch numbers; e.g., PJ_VERSION=493 for version 4.9.3.
   *
   * * Version 4.10.0 and later: PJ_VERSION=MMMNNNPP later where MMM, NNN, PP
   *   are the major, minor, and patch numbers (the minor and patch numbers
   *   are padded with leading zeros if necessary); e.g., PJ_VERSION=401000
    *   for version 4.10.0.
    */
````
I do believe the `PJ_VERSION = 500` fits this description since it is not necessary to pad with zeros in this case. I also think that this is a shit way of doing it, but if this is what we've told people to check against we have to respect that. It is also why is different in `proj.h` (I will make sure to update the comment, thanks for pointing it out).

**Comment by rouault** [comment]
> I do believe the PJ_VERSION = 500 fits this description

I believe @micahcochran point is that there should be no mention of 4.10.0 as this will never exist. Just replace 4.10.0 by 5.0.0 in the text (and PJ_VERSION=500000) and that will do

By the way I find the pattern MMMNNNPP not really logical. Either use 2 digits for all number, or 3. But the mix is really hard to grasp.

**Comment by rsbivand** [comment]
Which version number should be exposed in proj.pc.in (and/or proj-config.in)? Maybe --version should be the semantic version number (as it does now)? Going from this to the old-style string is feasible.

**Comment by kbevers** [comment]
> I believe @micahcochran point is that there should be no mention of 4.10.0 as this will never exist. Just replace 4.10.0 by 5.0.0 in the text (and PJ_VERSION=500000) and that will do

Ah, alright. I've updated it now.

> By the way I find the pattern MMMNNNPP not really logical. Either use 2 digits for all number, or 3. But the mix is really hard to grasp.

Yes, it is horrible. Eventually we'll be able to get rid of it though. We could remove it completely from `proj.h` entirely. I don't think that will cause any harm, and we won't have to deal with removing it in the future.

**Comment by rsbivand** [comment]
@rouault : what would you see as easiest for GDAL, a free-standing {prefix}/bin/proj-config, a {prefix}/lib/pkgconfig/proc.pc, or both? GDAL does both, right? Once we know that, the version numbers should fall into place, shouldn't they? I guess that since 4.10.0 will never appear, so the comment could be that <= 4.9.3, the 493 model was also used, > 4.9.3, only MAJOR.MINOR.PATCH.

**Comment by micahcochran** [comment]
> Yes, it is horrible. Eventually we'll be able to get rid of it though. We could remove it completely from `proj.h` entirely. I don't think that will cause any harm, and we won't have to deal with removing it in the future.

This is where the comment was committed.  https://github.com/OSGeo/proj.4/pull/364    The comment made it into the 4.9.3 release.  Since you are doing API change in PROJ.5, I don't see an issue with changing the version numbering scheme.  

IMHO If you are going to fix it, go ahead and make it a "sensible" version numbering scheme for 5.0.0.

**Comment by rouault** [comment]
> what would you see as easiest for GDAL, a free-standing {prefix}/bin/proj-config, a {prefix}/lib/pkgconfig/proc.pc, or both?

Either approach would be ok, although the pkgconfig approach is perhaps easier to test the version than a proj-config. PKG_CHECK_MODULES([PROJ], [libproj >= 5.1.0], ...)

**Comment by rsbivand** [comment]
Did proj 4.8.0 and earlier support pkgconfig? Yes, 4.8.0 has a populated proj.pc! < 4.8.0 do not have, checked on the [old download site](http://download.osgeo.org/proj/). So if the pkgconfig path is right, >= 4.8.0 we have MAJOR.MINOR.PATCH returned by pkg-config and testable. If the pkgconfig path is right with respect to {prefix}, but pkg-config fails, we know the version is < 4.8.0, so will not use contexts either (legacy checking for metadata files in /share).

**Comment by kbevers** [comment]
> IMHO If you are going to fix it, go ahead and make it a "sensible" version numbering scheme for 5.0.0.

I'll leave it "as is" in `proj_api.h` (we with numbers though) and remove the `PJ_VERSION` stuff from `proj.h`. I think that's best compromise between backwards compatibility and not creating a big future mess.

File: src/proj_api.h [hunk 1]
```diff
@@ -31,16 +31,16 @@
   * This version number should be updated with every release!  The format of
   * PJ_VERSION is
   *
-  * * Before version 4.10.0: PJ_VERSION=MNP where M, N, and P are the major,
+  * * Before version 5.0.0: PJ_VERSION=MNP where M, N, and P are the major,
   *   minor, and patch numbers; e.g., PJ_VERSION=493 for version 4.9.3.
   *
-  * * Version 4.10.0 and later: PJ_VERSION=MMMNNNPP later where MMM, NNN, PP
+  * * Version 5.0.0 and later: PJ_VERSION=MMMNNNPP later where MMM, NNN, PP
```
micahcochran: If 5.0.0 is to be PJ_VERSION=50000,  then please change this to read as so:

`* * Version 5.0.0 and later: PJ_VERSION=MMNNPP later where MM, NN, PP`

kbevers: Although not very obvious, and probably incorrect, I was trying to follow the recipe in the comment. It says that minor and patch numbers should be padded with zeros. So what I've got here is 5 *padding zero* 0 *padding zero* 0. I'm happy to change it if it's too obscure.

rsbivand:
Please only use 500 as 5.0.0. The 4.10.0 was never used, never will be, and creates havoc. Simply retire the comment about zero padding. I assume that proj.h will access the version number from proj_api.h internally - a correct version number must be exposed also for those trying to implement the new API.

kbevers:
> Please only use 500 as 5.0.0. 

What's the advantage of this versus 50000 (or whatever). Both are larger numbers than 493 which is the important bit. Mind you, I don't care either way.

> I assume that proj.h will access the version number from proj_api.h internally - a correct version number must be exposed also for those trying to implement the new API.

`proj.h` is completely orthogonal to `proj_api.h` so the version number will have to be maintained in both files.

rsbivand:
Yes, but anyone writing a configure.ac needs to be able to discover whether proj.h is available (> 4.9.3), or if not which version of proj_api.h is presented. Once this is known, compile flags can be defined and passed on down (until know this has been a 3 character string by default in my usage). If the x.y.z format had been adopted previously, testing major, minor or patch would have been easier, but it wasn't. Once this lands, I'll do a PR to the migration guide for configure.ac and similar, after testing, and following your merging the PR.

PR #674, opened 2017-11-17
Declare non-local variables as const where possible

aaronpuchert:
Having non-const variables of static lifetime or even global scope is usually a bad idea. These variables are inherently constants, and this should be enforced.

This required marking some functions as not modifying input parameters and marking some pointers as pointers to const.

One advantage is that the compiler usually puts const static variables in a read-only code segment, so they can't be modified physically. This can be verified with `nm` (on POSIX systems).

### DIFF
--- a/src/PJ_healpix.c
+++ b/src/PJ_healpix.c
@@ -64,7 +64,7 @@ typedef struct {
     enum Region {north, south, equatorial} region;
 } CapMap;
 
-double rot[7][2][2] = ROT;
+static const double rot[7][2][2] = ROT;
 
 /**
  * Returns the sign of the double.
@@ -310,7 +310,7 @@ static LP healpix_sphere_inverse(XY xy) {
  * Return the vector sum a + b, where a and b are 2-dimensional vectors.
  * @param ret holds a + b.
  **/
-static void vector_add(double a[2], double b[2], double *ret) {
+static void vector_add(const double a[2], const double b[2], double *ret) {
     int i;
     for(i = 0; i < 2; i++) {
         ret[i] = a[i] + b[i];
@@ -322,7 +322,7 @@ static void vector_add(double a[2], double b[2], double *ret) {
  * Return the vector difference a - b, where a and b are 2-dimensional vectors.
  * @param ret holds a - b.
  **/
-static void vector_sub(double a[2], double b[2], double*ret) {
+static void vector_sub(const double a[2], const double b[2], double*ret) {
     int i;
     for(i = 0; i < 2; i++) {
         ret[i] = a[i] - b[i];
@@ -335,7 +335,7 @@ static void vector_sub(double a[2], double b[2], double*ret) {
  * b is a 2 x 1 matrix.
  * @param ret holds a*b.
  **/
-static void dot_product(double a[2][2], double b[2], double *ret) {
+static void dot_product(const double a[2][2], const double b[2], double *ret) {
     int i, j;
     int length = 2;
     for(i = 0; i < length; i++) {
@@ -453,7 +453,7 @@ static XY combine_caps(double x, double y, int north_square, int south_square,
     double vector[2];
     double v_min_c[2];
     double ret_dot[2];
-    double (*tmpRot)[2];
+    const double (*tmpRot)[2];
     int pole = 0;
 
     CapMap capmap = get_cap(x, y, north_square, south_square, inverse);
--- a/src/PJ_isea.c
+++ b/src/PJ_isea.c
@@ -159,7 +159,7 @@ struct snyder_constants {
 };
 
 /* TODO put these in radians to avoid a later conversion */
-ISEA_STATIC
+ISEA_STATIC const
 struct snyder_constants constants[] = {
     {23.80018260, 62.15458023, 60.0, 3.75, 1.033, 0.968, 5.09, 1.195, 1.0},
     {20.07675127, 55.69063953, 54.0, 2.65, 1.030, 0.983, 3.59, 1.141, 1.027},
@@ -217,7 +217,7 @@ static int      tri_v1[] = {0, 0, 0, 0, 0, 0, 6, 7, 8, 9, 10, 2, 3, 4, 5, 1, 11,
 #define F_RAD 0.18871053072122403508
 
 /* triangle Centers */
-struct isea_geo icostriangles[] = {
+static const struct isea_geo icostriangles[] = {
     {0.0, 0.0},
     {-DEG144, E_RAD},
     {-DEG72, E_RAD},
--- a/src/PJ_mod_ster.c
+++ b/src/PJ_mod_ster.c
@@ -12,7 +12,7 @@ PROJ_HEAD(gs50, "Mod. Stereographic of 50 U.S.") "\n\tAzi(mod)";
 #define EPSLN 1e-12
 
 struct pj_opaque {
-    COMPLEX *zcoeff; \
+    const COMPLEX *zcoeff; \
     double  cchio, schio; \
     int     n;
 };
@@ -120,7 +120,7 @@ static PJ *setup(PJ *P) { /* general initialization */
 
 /* Miller Oblated Stereographic */
 PJ *PROJECTION(mil_os) {
-    static COMPLEX AB[] = {
+    static const COMPLEX AB[] = {
         {0.924500, 0.},
         {0.,       0.},
         {0.019430, 0.}
@@ -143,7 +143,7 @@ PJ *PROJECTION(mil_os) {
 
 /* Lee Oblated Stereographic */
 PJ *PROJECTION(lee_os) {
-    static COMPLEX AB[] = {
+    static const COMPLEX AB[] = {
         {0.721316,    0.},
         {0.,          0.},
         {-0.0088162, -0.00617325}
@@ -165,7 +165,7 @@ PJ *PROJECTION(lee_os) {
 
 
 PJ *PROJECTION(gs48) {
-    static COMPLEX /* 48 United States */
+    static const COMPLEX /* 48 United States */
     AB[] = {
         {0.98879,   0.},
         {0.,        0.},
@@ -191,7 +191,7 @@ PJ *PROJECTION(gs48) {
 
 
 PJ *PROJECTION(alsk) {
-    static COMPLEX  ABe[] = { /* Alaska ellipsoid */
+    static const COMPLEX  ABe[] = { /* Alaska ellipsoid */
         { .9945303, 0.},
         { .0052083, -.0027404},
         { .0072721,  .0048181},
@@ -200,7 +200,7 @@ PJ *PROJECTION(alsk) {
         { .3582802, -.2884586},
     };
 
-    static COMPLEX ABs[] = { /* Alaska sphere */
+    static const COMPLEX ABs[] = { /* Alaska sphere */
         { .9972523, 0.},
         { .0052513, -.0041175},
         { .0074606,  .0048125},
@@ -231,7 +231,7 @@ PJ *PROJECTION(alsk) {
 
 
 PJ *PROJECTION(gs50) {
-    static COMPLEX  ABe[] = { /* GS50 ellipsoid */
+    static const COMPLEX  ABe[] = { /* GS50 ellipsoid */
         { .9827497, 0.},
         { .0210669,  .0053804},
         {-.1031415, -.0571664},
@@ -244,7 +244,7 @@ PJ *PROJECTION(gs50) {
         {-.0210072,  .0834037}
     };
 
-    static COMPLEX ABs[] = { /* GS50 sphere */
+    static const COMPLEX ABs[] = { /* GS50 sphere */
         { .9842990, 0.},
         { .0211642,  .0037608},
         {-.1036018, -.0575102},
--- a/src/PJ_nzmg.c
+++ b/src/PJ_nzmg.c
@@ -34,7 +34,7 @@ PROJ_HEAD(nzmg, "New Zealand Map Grid") "\n\tfixed Earth";
 #define SEC5_TO_RAD 0.4848136811095359935899141023
 #define RAD_TO_SEC5 2.062648062470963551564733573
 
-static COMPLEX bf[] = {
+static const COMPLEX bf[] = {
     { .7557853228, 0.0},
     { .249204646,  0.003371507},
     {-.001541739,  0.041058560},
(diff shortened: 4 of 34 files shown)


aaronpuchert:
There is a weird crash in `src/gie ./test/gie/more_builtins.gie`. I'll have a look.

aaronpuchert:
Could someone have a look at this please? I'm working on reducing the amount of shared modifiable data to make sure there are really no problems in a multi-threaded environment.

rouault:
Looks generally good to me, but could you revert the changes in proj_api.h and projects.h, as they could potentially break external code ?

aaronpuchert:
The problem is that for some reason `pj_list`, `pj_ellps`, `pj_units`, `pj_datums`, `pj_prime_meridians` are forward declared in `projects.h`, although there are also functions providing access to the same data.

I could leave `pj_strerrno` and the `pj_get_*_ref` functions as they are, but the forward declarations must match with the definition. Would that be OK?

aaronpuchert:
The changes in `pj_zpoly1` and `pj_zpolyd1` are backwards compatible. I would leave the forward declarations with `const`, but return a non-const pointer in `pj_strerrno` and the `pj_get_*_ref` functions:

```patch
diff --git a/src/proj_api.h b/src/proj_api.h
index abc10795..3b817d2d 100644
--- a/src/proj_api.h
+++ b/src/proj_api.h
@@ -163,7 +163,7 @@ void pj_dalloc(void *);
 void *pj_calloc (size_t n, size_t size);
 void *pj_dealloc (void *ptr);
 char *pj_strdup(const char *str);
-const char *pj_strerrno(int);
+char *pj_strerrno(int);
 int *pj_get_errno_ref(void);
 const char *pj_get_release(void);
 void pj_acquire_lock(void);
diff --git a/src/projects.h b/src/projects.h
index ebc410e5..fc4d1c0e 100644
--- a/src/projects.h
+++ b/src/projects.h
@@ -778,11 +778,11 @@ LP     pj_inv_gauss(projCtx, LP, const void *);
 
 extern char const pj_release[];
 
-const struct PJ_ELLPS            *pj_get_ellps_ref( void );
-const struct PJ_DATUMS           *pj_get_datums_ref( void );
-const struct PJ_UNITS            *pj_get_units_ref( void );
-const struct PJ_LIST             *pj_get_list_ref( void );
-const struct PJ_PRIME_MERIDIANS  *pj_get_prime_meridians_ref( void );
+struct PJ_ELLPS            *pj_get_ellps_ref( void );
+struct PJ_DATUMS           *pj_get_datums_ref( void );
+struct PJ_UNITS            *pj_get_units_ref( void );
+struct PJ_LIST             *pj_get_list_ref( void );
+struct PJ_PRIME_MERIDIANS  *pj_get_prime_meridians_ref( void );
 
 void *pj_default_destructor (PJ *P, int errlev);
 
```

We could even, if people actually use the forward-declared arrays and assume they are non-const, make them macros:

```patch
diff --git a/src/projects.h b/src/projects.h
index 569141e3..e346884d 100644
--- a/src/projects.h
+++ b/src/projects.h
@@ -561,20 +561,20 @@ struct PJ_LIST {
 
 
 #ifndef USE_PJ_LIST_H
-extern const struct PJ_LIST pj_list[];
+#define pj_list pj_get_list_ref()
 #endif
 
 #ifndef PJ_ELLPS__
-extern const struct PJ_ELLPS pj_ellps[];
+#define pj_ellps pj_get_ellps_ref()
 #endif
 
 #ifndef PJ_UNITS__
-extern const struct PJ_UNITS pj_units[];
+#define pj_units pj_get_units_ref()
 #endif
 
 #ifndef PJ_DATUMS__
-extern const struct PJ_DATUMS pj_datums[];
-extern const struct PJ_PRIME_MERIDIANS pj_prime_meridians[];
+#define pj_datums pj_get_datums_ref()
+#define pj_prime_meridians pj_get_prime_meridians_ref()
 #endif
 
 
```

However, this isn't very beautiful.

rouault:
Honestly I would prefer defering any potential breaking changes in the old API after we have taken the decision to not install anymore (so in 5.1 or 6.0), so that proj 5 can be installed as easily as possible with existing software.

aaronpuchert:
Well, I see your point. Neglecting `const` is only a warning in C, but could cause errors in C++ code, or when people compile with `-Wall -Werror`.

Does the API also include that users _can actually modify_ `pj_err_list`, `pj_list`, `pj_ellps`, `pj_units`, `pj_datums`, and `pj_prime_meridians`? Of course they can't add or remove entries, but they could change some. I'm not sure why someone would do this, but people have [reasons](https://xkcd.com/1172/). However, the error messages returned by `pj_strerrno` are indeed [unmodifiable](http://en.cppreference.com/w/c/language/string_literal).

If this is considered part of the interface, then I have to (reluctantly) remove the `const` specifiers from those tables again, because with these changes they can't be modified.

rouault:
All of this is of course intended to be non modifiable. It is just the API is too loose. But as it is ancient API, fixing it would probably do more harm than good at that point. IMHO

hobu:
> But as it is ancient API, fixing it would probably do more harm than good at that point.

As a 30 year old API, we can't really go and "fix" anything that people are using TODAY. We can go fix a new on that they should use TOMORROW and wait the 5+ years until people start to do that. API migration for PROJ is *glacial*, and we should plan as such.

kbevers:
@hobu Slightly off topic, sorry. 

> API migration for PROJ is glacial, and we should plan as such.

Do you mind outlining such a plan? At the moment we are all working in sort of the same direction, but my perception is that there is no common understanding of what exactly that direction is. It would be nice to have that settled once and for all. Coming from you, the project maintainer, it can be regarded as the "authorative plan" I guess.

hobu:
> Do you mind outlining such a plan? At the moment we are all working in sort of the same direction, but my perception is that there is no common understanding of what exactly that direction is. It would be nice to have that settled once and for all. Coming from you, the project maintainer, it can be regarded as the "authorative plan" I guess.

Well, I'm the self-appointed maintainer, and I'm happy to hand off the hat to people actually doing all the work 😄  I've been trying to stay back and let you guys evolve things without worrying too much about the project maintenance implications. 

In my opinion, the 5.0.0 release cannot break projects that are using the existing API, except to force them to recompile. I think it is just fine to annoyingly warn them the API they're using is deprecated, and in one or two releases, require them to `#define I_KNOW_THIS_API_IS_OLD` to continue to use the old API. We can't force a migration by breaking compilation of existing usage that's out there in one release though, or we'll end up splitting the project into two, and the packagers will choose the old one and start patching because of all of the existing software.

image

Yes I know this ends up being a little bit like XKCD, except for APIs instead of standards, but we don't have much choice. The library is 30 years old, and we don't get to break people for our own satisfaction -- no matter how right, correct, fast, or convenient our new reality might be.

kbevers:
I don't want to shift the focus of this thread. Continuing off topic discussion in https://github.com/OSGeo/proj.4/issues/686

aaronpuchert:
Now that we have this discussion, let's clear up a misunderstanding of mine. In the world I grew up in, version numbers have a [meaning](https://semver.org/). Specifically, the major version is supposed to be incremented "when you make incompatible API changes". That's also what I would expect as user. This has nothing to do with how old the library is and for how long people have been using it.

Let's be clear about this: nothing breaks here. People can just use their PROJ.4 as they used to. They might not be able to update to version 5.0.0, but I'm not sure why people would assume that a major version increment goes by without the slightest API changes. If updating turns out too much trouble (and to be honest, I don't see why, but more about that later), I'd stay on the 4.9.x branch and hope that developers port back the occasional bug fix, or do it myself.

That's usually no problem. Linux has a [mechanism](https://en.wikipedia.org/wiki/Soname) that allows having multiple versions of the same library. In fact `libproj.so` comes now in shared object version 12, which means there have been 12 incompatible changes to the ABI already. On Windows and Mac OS, applications usually ship with their own libraries, so there is even less of a problem here.

Now let's get back to the changes at hand, because it's not like I'm redesigning the entire API:
* The values returned by `pj_strerrno` are string literals from `pj_err_list`. String literals are [unmodifiable](http://en.cppreference.com/w/c/language/string_literal), most compilers put them in the `.rodata` (meaning read-only data) segment, meaning that attempts at modification result in a segmentation fault. So this change is only *clarifying* the API.
* The tables with ellipses, datums, units, and so on, are supposed to be unmodifiable as well. I don't see any reason to modify these tables, it pretty much defeats their purpose.
* Not withstanding the other two points, discarding qualifiers like `const` is only a warning in C.
* Const correctness is usually easy to achieve by adding `const` in a few places, so I don't see a lot of migration hassle.

Not withstanding all these points, I suggested to leave the header files as they are, and only add the `const` qualifier in the source files, effectively making them unmodifiable without telling the user. That would let user's applications compile but is a bit sneaky.

Lastly, I can of course leave the tables as they are. I'm not a big fan of this, but if I can't make them `const`, I want at least the other changes.

aaronpuchert:
It's not the intention of this change to change the API, but to make internal data unmodifiable. I thought it's just fair to note that in the API. But as mentioned, there are three options:

1. Make data unmodifiable, and denote that in the API.
2. Make data unmodifiable, but don't change the headers.
3. Leave data modifiable that can currently be modified.

Of course we can go with options 2 or 3 instead of 1.

The only problem with option 3 is that we can then only make the tables unmodifiable starting with version 6.0 as of the current plan in #686, meaning about 15 months down the road. That's why it isn't my favorite.

aaronpuchert:
@rouault Could you have a look again? I reverted the changes in `proj_api.h` and made the changes in `projects.h` backwards-compatible. To have const-correctness within the library, I decided to use the `proj_list_*` functions from the new API internally instead of the `pj_get_*_ref` functions.

The build failure on Mac OS seems unrelated, maybe you can restart the build.

@busstoptaktik Since there are changes to `proj_4D_api.c` you should probably have a look, too. I moved the functions to where the data is, so that we avoid conversions `const` → non-`const` → `const`. It would also allow us to make the data arrays `static` once the old API is no longer exposed.

Also, do you intend to expose something like `pj_strerrno` in the new API? If yes, please consider making it return `const char*` instead `char*`, string literals are unmodifiable.

rouault:
Why do you need to add in projects.h
```
+#ifndef PROJ_H
+const struct PJ_LIST             *proj_list_operations(void);
+const struct PJ_ELLPS            *proj_list_ellps(void);
+const struct PJ_UNITS            *proj_list_units(void);
+const struct PJ_PRIME_MERIDIANS  *proj_list_prime_meridians(void);
+#endif
```

Couldn't users of those functions in the library include proj.h instead ?

aaronpuchert:
I tried that, but there are some conflicts. When I include `proj.h` in `proj.c` like this:
``` patch
diff --git a/src/proj.c b/src/proj.c
index dc4db466..b18dc2e8 100644
--- a/src/proj.c
+++ b/src/proj.c
@@ -1,4 +1,5 @@
 /* <<<< Cartographic projection filter program >>>> */
+#include "proj.h"
 #include "projects.h"
 #include <stdio.h>
 #include <stdlib.h>
```

I get the following errors:
```
../src/proj.c:122:39: error: passing 'projUV' to parameter of incompatible type 'LP'
                facs_bad = pj_factors(data, Proj, 0., &facs);
                                      ^~~~
../src/projects.h:716:18: note: passing argument to parameter here
int pj_factors(LP, const PJ *, double, struct FACTORS *);
                 ^
../src/proj.c:126:39: error: passing 'projUV' to parameter of incompatible type 'LP'
                facs_bad = pj_factors(data, Proj, 0., &facs);
                                      ^~~~
../src/projects.h:716:18: note: passing argument to parameter here
int pj_factors(LP, const PJ *, double, struct FACTORS *);
                 ^
../src/proj.c:241:29: error: passing 'projUV' to parameter of incompatible type 'XY'
            dat_ll = pj_inv(dat_xy, Proj);
                            ^~~~~~
../src/proj_api.h:123:21: note: passing argument to parameter here
projLP pj_inv(projXY, projPJ);
                    ^
../src/proj.c:255:29: error: passing 'projUV' to parameter of incompatible type 'LP'
            dat_xy = pj_fwd(dat_ll, Proj);
                            ^~~~~~
../src/proj_api.h:122:21: note: passing argument to parameter here
projXY pj_fwd(projLP, projPJ);
                    ^
../src/proj.c:273:24: error: passing 'projUV' to parameter of incompatible type 'LP'
        if (pj_factors(dat_ll, Proj, 0., &facs)) {
                       ^~~~~~
../src/projects.h:716:18: note: passing argument to parameter here
int pj_factors(LP, const PJ *, double, struct FACTORS *);
                 ^
../src/proj.c:493:14: error: incompatible pointer types assigning to 'projUV (*)(projUV, PJ *)' (aka 'projUV (*)(projUV, struct PJconsts *)') from 'LP (XY, projPJ)' (aka 'LP (XY, struct PJconsts *)') [-Werror,-Wincompatible-pointer-types]
        proj = pj_inv;
             ^ ~~~~~~
../src/proj.c:495:14: error: incompatible pointer types assigning to 'projUV (*)(projUV, PJ *)' (aka 'projUV (*)(projUV, struct PJconsts *)') from 'XY (LP, projPJ)' (aka 'XY (LP, struct PJconsts *)') [-Werror,-Wincompatible-pointer-types]
        proj = pj_fwd;
             ^ ~~~~~~
```

I guess the reason is that the definitions of some basic types are incompatible.

Specifically, we don't have these `#define`s any more:
```cpp
#define XY projUV
#define LP projUV
#define XYZ projUVW
#define LPZ projUVW
```

Now if I change
```patch
diff --git a/src/projects.h b/src/projects.h
index c9a56c6e..52dfea66 100644
--- a/src/projects.h
+++ b/src/projects.h
@@ -713,7 +713,7 @@ COMPLEX pj_zpoly1(COMPLEX, const COMPLEX *, int);
 COMPLEX pj_zpolyd1(COMPLEX, const COMPLEX *, int, COMPLEX *);
 
 int pj_deriv(LP, double, const PJ *, struct DERIVS *);
-int pj_factors(LP, const PJ *, double, struct FACTORS *);
+int pj_factors(projUV, const PJ *, double, struct FACTORS *);
 
 struct PW_COEF {    /* row coefficient structure */
     int m;          /* number of c coefficients (=0 for none) */
```

then the implementation does not work because it relies on the internal definition
```cpp
typedef struct { double lam, phi; }    LP;
```
which is active if `PJ_LIB__` is defined.

aaronpuchert:
I think we can include `proj.h` in `proj.c` after some changes. I'll open a separate pull request for that.

busstoptaktik:
I suppose the problem is with the deffing XY and LP to UV in projects.h
ifndef PJ_LIB__ - probably it would work just #define'ing PJ_LIB__ prior to
including projects.h

2017-12-11 17:51 GMT+01:00 Aaron Puchert <notifications@github.com>:

aaronpuchert:
If I include `proj.h`, the defines are already excluded, because all that is surrounded by `#ifndef PROJ_H`. But right now the code in `proj.c` relies on these defines, specifically that `XY` = `LP`. I addressed this issue in #704, though only to the extent it was necessary.

aaronpuchert:
Now that #704 is merged, I removed the additional `proj_list_*` declarations from `projects.h` again.

aaronpuchert:
@rouault Could you have a look again?

rouault:
Looked good to me. Merged. Thanks

PR #687, opened 2017-11-27
Fix bad typedef that surfaces when including proj.h and proj_api.h in the same file

kbevers:
This error appears when trying to include `proj.h` and `proj_api.h` in the same file. Fixed with this PR.

```
$ gcc -L../../build/proj4/lib -I../src -lproj projapitest.c -o test                                 
In file included from projapitest.c:3:                                                                
../src/proj_api.h:87:13: error: must use 'struct' tag to refer to type 'projCtx_t'
    typedef projCtx_t *projCtx;  /* projCtx is a pointer to projCtx_t */ 
            ^                                                                                         
            struct                                                                                    
1 error generated.
```

### DIFF
--- a/src/proj_api.h
+++ b/src/proj_api.h
@@ -84,6 +84,7 @@ extern int pj_errno;    /* global error return code */
 #ifdef PROJ_API_H_NOT_INVOKED_AS_PRIMARY_API
     /* These make the function declarations below conform with classic proj */
     typedef PJ *projPJ;          /* projPJ is a pointer to PJ */
+    typedef struct projCtx_t projCtx_t;
     typedef projCtx_t *projCtx;  /* projCtx is a pointer to projCtx_t */
 #   define projXY        XY
 #   define projLP       LP
--- a/src/proj_api.h
+++ b/src/proj_api.h
@@ -84,8 +84,7 @@ extern int pj_errno;    /* global error return code */
 #ifdef PROJ_API_H_NOT_INVOKED_AS_PRIMARY_API
     /* These make the function declarations below conform with classic proj */
     typedef PJ *projPJ;          /* projPJ is a pointer to PJ */
-    typedef struct projCtx_t projCtx_t;
-    typedef projCtx_t *projCtx;  /* projCtx is a pointer to projCtx_t */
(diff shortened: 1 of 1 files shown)


busstoptaktik:
The two old API headers are a minefield to step around in, and it cost me significant amounts of frustration to typedef my way around the traditions.

There is a `typedef struct projCtx_t projCtx_t;` in `projects.h`, and `projects.h` includes `proj_api.h`.

So a workaround, if you need to access `proj_api.h` in a file also needing `proj.h`, could be to include `projects.h` instead.

Repeated typedefs are, however, not an error (as long as they are identical), so I suppose that this PR does no harm (although it reminds me why I look so much forward to make these old header files strictly internal, so we can skip backwards compatibility, and begin to clean them up properly).

busstoptaktik:
Oops... taking another look at your patch: These typedefs are *not* identical, but it shouldn't matter either, because they are typedeffing different things.

    projects.h:  typedef struct projCtx_t projCtx_t;
    this patch:  typedef struct projCtx_t *projCtx;
    replacing:   typedef projCtx_t *projCtx;  /* projCtx is a pointer to projCtx_t */

As I mentioned: Cleaning up, in a backwards compatible way, after the "information-hiding-through-typdedeffing-of-everything-to-void" culture that used to be, is like stepping through a minefield.
So for consistency, I think you should try whether this two line version fits:

    typedef struct projCtx_t projCtx_t;
    typedef projCtx_t *projCtx;  /* projCtx is a pointer to projCtx_t */

kbevers:
Both versions seem to work fine. I'll just go with the one you suggested.

kbevers:
> Both versions seem to work fine. I'll just go with the one you suggested.

Turns out the tests fail with the following 

```
/Users/travis/build/OSGeo/proj.4/src/proj_api.h:86:30: error: redefinition of typedef 'projCtx_t' is a C11 feature [-Werror,-Wtypedef-redefinition]
    typedef struct projCtx_t projCtx_t;
                             ^
/Users/travis/build/OSGeo/proj.4/src/projects.h:220:26: note: previous definition is here
typedef struct projCtx_t projCtx_t;
                         ^
```
when doing the proposed two-line version.

Issue #678, opened 2017-11-20
Namespacing for library internal vs. file static functions


busstoptaktik:
While discussing #675 with @kbevers, we touched the subject of a library internal name space.

As things are, the pj_-namespace is used for too many things. It would be useful to go for an architecture, where

* the API lives in the proj_-namespace, and is declared in the `proj.h` header (this is already the case).
* library internal material lives in the pj_-namespace, and is declared in the `proj_internals.h` header
* file scoped static functions are moved **out** of the pj_-namespace, and does not appear in any header file.

This also suggests some elements of a long term roadmap toward gradual retirement of `projects.h` and `proj_api.h`, by (at whatever pace is realistic) moving elements from these headers to `proj_internals.h` (and perhaps even `proj.h`), potentially under different names, and implementing temporary workarounds in the old headers.

PR #693, opened 2017-12-06
Free format everywhere

busstoptaktik:
This PR (hopefully) finalizes the free formatting work started in PR #691: Free formatting can now be used everywhere - in gie test files, in definition files (such as epsg), and on the `cct` command line (for backwards compatibility, this has not been enabled in `proj` and `cs2cs`).

In `cct`, free format on the command line is indicated by using the `--` operator to separate operation definition and input files (this has historically been used in the Unix world whenever options and input files could be mistaken for each other, e.g. when one input file was named `-starts-with-minus`).

The free format, and the orthogonality of the init and gie syntaxes make it possible to interleave system definition and system validation blocks in the same file. This is demonstrated in the file `nad/val_def.demo`.

Touching the input format is a major operation, so this PR is complex, and touches code here, there and everywhere - sometimes in surprising places. The latter mostly, because some detours during development, for various reasons were only partially rolled back (i.e. "keep improvements, roll back mistakes").

All in all, however, I'm quite confident that this PR is an improvement.


src/pj_init.c
```diff
@@ -33,314 +33,295 @@
...
-    char_requested = sizeof(state->buffer) - state->buffer_filled - 1;
-    bytes_read = pj_ctx_fread( state->ctx, state->buffer + state->buffer_filled,
-                               1, char_requested, state->fid );
-    if (bytes_read < char_requested)
-    {
-        state->at_eof = 1;
-        state->buffer[state->buffer_filled + bytes_read] = '\0';
-    }
+/**************************************************************************************/
+static paralist *string_to_list (PJ_CONTEXT *ctx, char *definition) {
```
kbevers: I think `string_to_paralist` is a better name for this.

src/pj_initcache.c
```diff
@@ -51,7 +51,7 @@ paralist *pj_clone_paralist( const paralist *list)
       newitem->used = 0;
       newitem->next = 0;
       strcpy( newitem->param, list->param );
-      
+
```
kbevers: Only non-functional changes in this file. While trailing whitespace is annoying as hell, removing it in this context only contributes clutter to the commit log of this file.

src/pj_internal.c
```diff
@@ -114,14 +115,15 @@ void proj_context_inherit (PJ *parent, PJ *child) {
 
 
 
-size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
-/*******************************************************************
+/**************************************************************************************/
+size_t pj_strlcpy(char *dst, const char *src, size_t dsize) {
```
kbevers, line 119: Perhaps it's time to make `pj_string.c`? There's a bunch of string manipulation functions spread out across the code base. It would be nice to gather them all in the same space.

src/pj_param.c
```diff
@@ -1,20 +1,78 @@
...
+    strncpy(newitem->param, str, len);
+
+    newitem->used = 0;
+    newitem->next = 0;
+
+    return newitem;
+}
+
+/**************************************************************************************/
+paralist *pj_param_exists (paralist *list, const char *parameter) {
```
kbevers, line 54: I like this a lot!

src/projects.h
```diff
@@ -685,7 +688,7 @@ int pj_angular_units_set(paralist *, PJ *);
 paralist *pj_clone_paralist( const paralist* );
 paralist *pj_search_initcache( const char *filekey );
 void      pj_insert_initcache( const char *filekey, const paralist *list);
-paralist *pj_get_init(projCtx ctx, paralist **start, paralist *next, char *name, int *found_def);
+paralist *pj_get_init(projCtx ctx, paralist *start, char *key);
```
kbevers: This breaks public "API" :(

It's an undocumented part of the `projects.h` "API", so at least we need to discuss if this should be allowed to change or not. Since `projects.h` is a somewhat loose definitions of an API I think changes to functions that are mostly there for internal use should be allowed.

src/readblock.c
```diff
@@ -0,0 +1,419 @@
...
+
+****************************************************************************************/
+    free (F->args);
+    free (F->next_args);
+    free (F);
+    return 0;
+}
+
+#if 0
+/***************************************************************************************/
```
kbevers: Don't include this out-commented code (it has apparently found a new home in `proj_internal.c`).

src/pj_initcache.c
busstoptaktik: You're obviously and evidently right - and I'm quite surprised that all functional changes that were entered during the work, ended up being eradicated again (but as mentioned in the PR intro - a lot of stuff, spanning 3 continents, has been going on during this work).

Will back this out.

src/pj_internal.c
busstoptaktik: ... and while at it, move `pj_free` to `pj_malloc.c`, and perhaps renaming to `pj_memory.c`

kbevers: +1. Seperate PR though. For both I think.

src/projects.h
busstoptaktik: At least, I believe that anyone outside the PROJ developer group, who would be able to make use of this in an application program, really should step forward and join the fun as PROJ developer :-)

As mentioned in #678, the `pj_`-namespace is heavily overloaded, and deciding what part can be considered API, and what cannot, is probably a matter of balanced judgment.

In the  present case, I believe the risk of stepping on someone's code is minimal - and at least, I believe that anyone making sense of `pj_get_init` in an application context, would probably have no problems migrating.

src/pj_init.c
busstoptaktik: +1

src/pj_internal.c
busstoptaktik: Agree

src/pj_param.c
busstoptaktik: Thx

schwehr (commented):
Partial review.  first 1/3.   Ran out of time

src/PJ_pipeline.c
```diff
@@ -427,7 +427,7 @@ PJ *OPERATION(pipeline,0) {
 
         err = proj_errno_reset (P);
 
-        next_step = proj_create_argv (P->ctx, current_argc, current_argv);
+        next_step = proj_create_argv (P->ctx, current_argc, (const char **) current_argv);
```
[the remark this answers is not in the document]
busstoptaktik: Yes, that was also what I believed, but apparently at high enough warning levels, gcc insists these are incompatible types - and as we treat warnings as errors, I needed the cast to get things to build. But I agree that it is both stupid and unexpected.

src/optargpm.h
```diff
@@ -219,6 +219,7 @@ struct OPTARGS {
     FILE *input;
     int   input_index;
     int   record_index;
+    int   free_format;       /* plus-style specs replaced by free format */
```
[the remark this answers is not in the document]
busstoptaktik: Yes - it's really mostly a flag, but doubles as index. But it is actually superfluous as struct member, so rather than trying to explain that in a comment, I think the easy option is just to eliminate it...

src/optargpm.h
```diff
@@ -516,7 +525,10 @@ OPTARGS *opt_parse (int argc, char **argv, const char *flags, const char *keys,
                 char *equals;
                 crepr = argv[i] + 2;
 
-                /* need to maniplulate a bit to support gnu style --pap=pop syntax */
+                /* We need to maniplulate a bit to support gnu style --foo=bar syntax.  */
```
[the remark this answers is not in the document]
busstoptaktik: Oops - will elimiplulate that (and all other speeling (and typng) missteaks) :-)

src/pj_factors.c
```diff
@@ -17,12 +17,18 @@ int pj_factors(LP lp, const PJ *P, double h, struct FACTORS *fac) {
     PJ_COORD coo = {{0, 0, 0, 0}};
     coo.lp = lp;
 
-    err = proj_errno_reset (P);
+    /* These are probably due to earlier errors, so we leave errno alone */
+    if (0==fac)
```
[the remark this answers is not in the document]
busstoptaktik: But this is C89, not C++, and a pretty common idiom. I have been using it for "clarity through brevity" for many years. It was actually my understanding that **by stdc definition** a `0` interpreted as a pointer  would have to expand to the null pointer, even though the actual hardware representation of that might be different from integer `0`. 

But `NULL` is a macro that expands to `(void *) 0`, and I have recently become in doubt whether it is actually strictly true that a plain `0`, interpreted as a pointer, is identical to a plain `0` cast as `void *`.  I will have to do some standards studying.

But if strictly `NULL` (i.e. `(void *) 0`) is needed here, it will break a common idiom (and hence a lot of code).

Nevertheless, it is valuable to know that this idiom is confusing, so I will retract to my monastery to contemplate, and may return enlightened...

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
 
-  Source: http://www.i-pi.com/Training/EthicalHacking/Solutions/strlcpy.c
+  Source: http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libc/string/strlcpy.c
 
-********************************************************************/
-    register char *d = dst;
-    register const char *s = src;
-    register size_t n = siz;
+***************************************************************************************/
+    const char *osrc = src;
```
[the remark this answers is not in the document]
busstoptaktik: Yes - certainly the compiler is in a better position to determine what should go in a register and what not

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+    /* Not enough room in dst, add NUL and traverse rest of src. */
+    if (nleft == 0) {
+        if (dsize != 0)
+            *dst = '\0';		/* NUL-terminate dst */
+        while (*src++)
             ;
     }
 
-    return(s - src - 1);    /* count does not include NUL */
+    return(src - osrc - 1);	/* count does not include NUL */
```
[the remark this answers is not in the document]
busstoptaktik: so would mine - making return look like a function call seems utterly strange to me. But this is code lifted directly from the OpenBSD repo, and I think I will just reimplement using strncpy and friends,  also in order to avoid the inline license and attribution stuff.

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+}
+
+
+
+/***************************************************************************************/
+char *pj_chomp (char *c) {
+/****************************************************************************************
+    Strip pre- and postfix whitespace. Note: Inline comments (indicated by '#')
```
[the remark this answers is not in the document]
busstoptaktik: Yes - that is definitely more clear - thanks!

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+****************************************************************************************/
+    size_t i, j, n, ws;
+
+    if (0==c)
+       return 0;
+
+    pj_chomp (c);
+    n = strlen (c);
+
+    /* First clash repeated whitespace (including +/;) */
```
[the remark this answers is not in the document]
busstoptaktik: Will collapse that and replace with collapse :-)

src/pj_internal.c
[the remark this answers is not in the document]
schwehr: C++17 removes register as (at least for C++) modern compilers will ignore the register keyword and do what they think is best.

https://stackoverflow.com/questions/20618008/replacement-for-deprecated-register-keyword-c-11

src/pj_param.c
```diff
@@ -1,20 +1,78 @@
 /* put parameters in linked list and retrieve */
-#include <projects.h>
 #include <stdio.h>
+#include <ctype.h>
```
schwehr: Unless there is a specific reason, please alphabetize the includes.  If there is a reason for the order, please add a comment.

src/pj_param.c
```diff
@@ -1,20 +1,78 @@
...
+        newitem->next = 0;
+        if (*str == '+')
+            ++str;
+        (void)strcpy(newitem->param, str);
+    }
+    return newitem;
+}
+
+
+/* As pj_mkparam, but payload is terminated by whitespace, rather than 0 */
```
schwehr: `0` -> `\0` or `NUL`

src/pj_param.c
```diff
@@ -1,20 +1,78 @@
...
+paralist *pj_mkparam_ws (char *str) {
+    paralist *newitem;
+    size_t len = 0;
+
+    if (0==str)
+        return 0;
```
schwehr, line 29: Here and everywhere else (I'll not comment on them more), please use `NULL` rather than `0` for nullptrs.

src/pj_param.c
```diff
@@ -1,20 +1,78 @@
...
+    if (*str == '+') {
+        str++;
+        len--;
+    }
+
+    /* Use calloc to automagically 0-terminate the strncpy */
+    newitem = (paralist *) pj_calloc (1, sizeof(paralist) + len);
+    if (0==newitem)
+        return 0;
+    strncpy(newitem->param, str, len);
```
schwehr: So there is pj_strlcpy and yet here you directly use strncpy?  What is the proj4 policy on the strcpy family of functions?

src/pj_param.c
```diff
@@ -1,20 +1,78 @@
...
+paralist *pj_param_exists (paralist *list, const char *parameter) {
+/***************************************************************************************
+    Determine whether a given parameter exists in a paralist. If it does, return
+    a pointer to the corresponding list element - otherwise return 0.
+
+    This function is equivalent to the pj_param (...) call with the "opt" argument
+    set to the parameter name preceeeded by a 't'. But by using this one, one avoids
+    writing the code allocating memory for a new copy of parameter name, and prepending
+    the t (for compile time known names, this is obviously not an issue).
+***************************************************************************************/
```
schwehr, line 63: Make an explicit statement about memory ownership and that there isn't a NUL terminator at the end of the param.

src/pj_param.c
```diff
@@ -1,20 +1,78 @@
...
+    newitem->next = 0;
+
+    return newitem;
+}
+
+/**************************************************************************************/
+paralist *pj_param_exists (paralist *list, const char *parameter) {
+/***************************************************************************************
+    Determine whether a given parameter exists in a paralist. If it does, return
+    a pointer to the corresponding list element - otherwise return 0.
```
schwehr, line 57: `0` -> NULL.

src/pj_param.c
```diff
@@ -35,76 +93,76 @@ paralist *pj_mkparam(char *str) {
 /*                                                                      */
 /************************************************************************/
 
-	PROJVALUE /* test for presence or get parameter value */
+    PROJVALUE /* test for presence or get parameter value */
```
schwehr, line 97: If the union is char *s, who owns the string?

src/pj_param.c
```diff
@@ -35,76 +93,76 @@ paralist *pj_mkparam(char *str) {
...
-			value.i = atoi(opt);
-			break;
-		case 'd':	/* simple real input */
-			value.f = pj_atof(opt);
-			break;
-		case 'r':	/* degrees input */
-			value.f = dmstor_ctx(ctx, opt, 0);
-			break;
-		case 's':	/* char string */
+    int type;
```
schwehr, line 100: If this is all just formatting, can you commit this separately from this PR to reduce the noise in this pr?

src/proj.def
```diff
@@ -149,3 +149,5 @@ EXPORTS
 
     pj_ellipsoid                   @134
     pj_calc_ellipsoid_params       @135
+    pj_chomp                       @136
+    pj_shrink                      @137
```
schwehr: Add new line at the end.

src/proj_4D_api.c
```diff
@@ -795,7 +758,7 @@ PJ_INIT_INFO proj_init_info(const char *initname){
     metadata is missing, the value is set to "Unknown".
 
 ******************************************************************************/
-    int file_found, def_found=0;
+    int file_found;
     char param[80], key[74];
```
schwehr, line 768: It would be better to have 80 and 74 be symbolic constants.  Out of scope for this PR, but just want to mention this.

src/proj_4D_api.c
```diff
@@ -858,6 +823,9 @@ PJ_FACTORS proj_factors(PJ *P, LP lp) {
     PJ_FACTORS factors = {0,0,0, 0,0,0, 0,0};
     struct FACTORS f;
 
+    if (0==P)
+        return factors;
```
schwehr, line 831: Return type is  `PJ_INIT_INFO` but factors is `PJ_FACTORS`.  Confusing.

src/proj_internal.h
```diff
@@ -107,6 +107,13 @@ void pj_inherit_ellipsoid_def (const PJ *src, PJ *dst);
 void pj_erase_ellipsoid_def (PJ *P);
 int pj_calc_ellipsoid_params (PJ *P, double a, double es);
 
+
```
schwehr: Why 2 blank lines?

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+    return(src - osrc - 1);	/* count does not include NUL */
+}
+
+
+
+/***************************************************************************************/
+char *pj_chomp (char *c) {
+/****************************************************************************************
+    Strip pre- and postfix whitespace. Note: Inline comments (indicated by '#')
+    are considered whitespace.
```
schwehr: Add a comment that the return should not be freed and that the caller's `c` still owns the memory.

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>
+
+#include <string.h>
+#include <ctype.h>
+
+#include <math.h>
+#include <errno.h>
+
```
schwehr: Remove the extra white space.

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>
+
+#include <string.h>
+#include <ctype.h>
+
+#include <math.h>
+#include <errno.h>
```
schwehr: I don't follow the order of headers here.  Is there some reason for the 3 groups and unusual order?  Why not just alphabetize them and remove the groupings

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+
+
+
+static const char *gie_tags[] = {
+    "<gie>", "operation", "accept", "expect", "roundtrip", "banner", "verbose",
+    "direction", "tolerance", "builtins", "echo", "</gie>"
+};
+
+static const size_t n_gie_tags = sizeof gie_tags / sizeof gie_tags[0];
```
schwehr: Is there no array size macro for Proj?

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+
+/***************************************************************************************/
+static ffio *ffio_destroy (ffio *F) {
+/****************************************************************************************
+
+****************************************************************************************/
+    free (F->args);
+    free (F->next_args);
+    free (F);
+    return 0;
```
schwehr: Why does this return anything?

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+    F->n_tags = n_tags;
+    return F;
+}
+
+
+/***************************************************************************************/
+static ffio *ffio_destroy (ffio *F) {
+/****************************************************************************************
+
+****************************************************************************************/
```
schwehr: What about `f`, `tags`, and `tag`?  If you don't need to handle them, add a comment as to why.

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+    if (at_tag (F))
+        return 1;
+    return 0;
+}
+
+
+/***************************************************************************************/
+static int nextline (ffio *F) {
+/****************************************************************************************
+    Read next line of input file
```
schwehr: Describe that the return is a boolean and what true and false represent

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+        return 0;
+    pj_chomp (F->next_args);
+    F->next_lineno++;
+    return 1;
+}
+
+/***************************************************************************************/
+static int locate_tag (ffio *F, const char *tag) {
+/****************************************************************************************
+    find start-of-line tag
```
schwehr: So I think this is consuming lines until the tag is reached and returns true.  And if it runs out of file without finding it, it returns a false.  Yes?  Please add what it does to the comment

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+        if (0==nextline (F))
+            return 0;
+    return 1;
+}
+
+
+/***************************************************************************************/
+static int step_into_gie_block (ffio *F) {
+/****************************************************************************************
+    Make sure we're inside a <gie>-block
```
schwehr: Describe the boolean return.

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+
+    /* We're ready at the start - now step into the block */
+    return nextline (F);
+}
+
+
+/***************************************************************************************/
+static int skip_to_next_tag (ffio *F) {
+/****************************************************************************************
+    Skip forward to the next command tag
```
schwehr: Describe return

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+        F->args[0] = 0;
+        return skip_to_next_tag (F);
+    }
+    F->lineno = F->next_lineno;
+
+    return 1;
+}
+
+
+static int append_args (ffio *F) {
```
schwehr: Needs comments describing the function

src/readblock.c
```diff
@@ -0,0 +1,299 @@
...
+}
+
+
+
+
+
+/***************************************************************************************/
+static int get_inp (ffio *F) {
+/****************************************************************************************
+    The primary command reader for gie. May be useful in the init-file reader as well.
```
schwehr: Sorry, what does this do?  What state changes to F does this cause?  Why would it be use in the init-file reader?

schwehr (commented):
I hope these comments are useful.  I spend most of my time with https://google.github.io/styleguide/cppguide.html and a fair bit of time in GDAL's C++ code, so apologies for comments that might not make sense in the context of Proj and C (version ?).

src/proj_4D_api.c
busstoptaktik: No - there is a new function, `proj_factors` starting in between

src/proj_internal.h
busstoptaktik: Not for any good reason - will eat one of them

src/readblock.c
busstoptaktik: I have never met the idiom of alphabetic inclusion order before, but it's a good idea for cases where the order otherwise doesn't matter, and I will reorganize here, and try to remember following the rule in new code

src/readblock.c
busstoptaktik: No - and I think this is the only place where one would be needed

src/readblock.c
busstoptaktik: for this use case:
    
    f = calloc (42, sizeof (f));
    ....
    if (some error occurred)
        return  ffio_destroy (f);

A coding style focusing on eliminating the easy errors first

src/pj_param.c
busstoptaktik: This is classic 1980s PROJ code, which I consider eliminating (the `pj_param_exists` function is the first step towards this). I see your point in being clear about memory deallocation duty, and I have spent a lot of time adding comments to older PROJ code, but in this case, I will really prefer to work toward better solution, rather than documenting code with more complex use patterns than need be.

src/readblock.c
kbevers: I usually follow the Python recommendations for this. Standard library headers first,  external headers and finally local headers. Since we don't have any external dependencies it's just stdlib then local headers, preferably separated by a space. But since the PROJ.4 codebase is a big pile of disorder with regards to code style it hardly matters much. There's no such thing as consistency here :)

busstoptaktik: ... but at least we could strive for *orderly* disorder

src/PJ_pipeline.c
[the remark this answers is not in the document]
aaronpuchert: These are [incompatible types](https://isocpp.org/wiki/faq/const-correctness#constptrptr-conversion), although that isn't obvious. However, converting `char**` to `const char* const*` should be fine.

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+
+
+
+/***************************************************************************************/
+char *pj_free_format_from_argc_argv (size_t argc, const char **argv) {
+/****************************************************************************************
+    Convert free format command line input to something proj_create can eat.
+****************************************************************************************/
+    size_t i, n;
+    char *p;
```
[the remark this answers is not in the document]
aaronpuchert: See `CONTRIBUTING.md`, section "Programming language":

> PROJ.4 is developed strictly in ANSI C 89.

Which means variables can't be declared within loop headers.

Compilers are also able to detect if variables are used uninitialized, so I think this can go either way.

src/readblock.c
schwehr: Gotchya.  That does seem reasonable for C.  This is where my mind being C++ has me off on the wrong track... I forget that there are no std::unique_ptr or gtl::MakeCleanup kinds of things available.

src/pj_factors.c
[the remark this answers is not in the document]
schwehr: Ok, that answers my question about which version of C.  The C++ community pushed for 0 over NULL or a long time and then when running for the typed nullptr.   nullptr helped catch quite a few long term lurking bugs.

I suggesting NULL to make it clear that the return is a pointer and not a boolean false or something else.  I think that if the 0 is applied to a pointer value, it should always be safely  converted to a `NULL`.  But I think you are correct that 0 assigned to a ptr always is the same as NULL.

Some quickly found related readings:

C:
- http://en.cppreference.com/w/c/types/NULL
- https://stackoverflow.com/questions/1296843/what-is-the-difference-between-null-0-and-0

C++:
- https://stackoverflow.com/questions/36484473/is-c-null-equal-to-c11-nullptr
- https://stackoverflow.com/questions/20509734/null-vs-nullptr-why-was-it-replaced

src/pj_param.c
busstoptaktik: Another case of classic 1980s proj code, where probably no harm is done by reordering.

src/pj_param.c
busstoptaktik: NUL does not exist in C, \0 should be '\0', but really just indicates an octal character constant (which in context is promoted to an int) - and that is expressed with less clutter as a plain decimal integer constant.

src/pj_param.c
busstoptaktik: Actually there is, so your comment is an indication that this is not described clearly enough. Will try to think up a better description.

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
  * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
  * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
 
-  Source: http://www.i-pi.com/Training/EthicalHacking/Solutions/strlcpy.c
+  Source: http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libc/string/strlcpy.c
```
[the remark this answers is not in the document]
aaronpuchert: As long as we stick with C89, we can't assume having `strncpy`. It's only available [since C99](http://en.cppreference.com/w/c/string/byte/strncpy) and the compiler will not make it available when we pass `-std=c89`, regardless of whether it is supported.

src/pj_param.c
aaronpuchert: That is just slang, `\0` is called [NUL](https://en.wikipedia.org/wiki/Null_character) in ASCII. People don't talk about 0-terminated strings, but [null-terminated strings](https://en.wikipedia.org/wiki/Null-terminated_string).

src/pj_param.c
busstoptaktik: I could, and I am actually surprised to see this amount of formatting changes show up - this is definitlely a case of editor setup hiding some changes.

I do, however, think (although I may misremember, and @kbevers  a.o., please correct me if I'm wOrNg)  that about a year ago, after @kbevers  did a number of commits unifying whitespace (to the canonical Guido van Rossum compatible) "no tabs, 4 spaces per indent"), we agreeed that the remaining work on this would have to be "drive by" corrections. Unintendedly, this is one of them...

src/proj.def
busstoptaktik: done

src/pj_param.c
busstoptaktik: (or "strings in ASCIIZ format", if they're old enough)

schwehr: Since this is a comment, it's not super critical.  It was just that someone could (when overtired) interpret `0` as `'0'` (`\x30`).  NUL and \0 in a comment should be less likely to be read incorrectly by a human.

src/optargpm.h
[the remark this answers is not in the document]
busstoptaktik: Will handle this in a separate PR.

aaronpuchert:
I think instead of replacing `const char **` by `char **`, you should change it to `const char * const *`, which is probably what is intended anyway. String literals should be treated as `const char *`, since they are [unmodifiable](http://en.cppreference.com/w/c/language/string_literal).

When compilers warn about casting `char **` to `const char **` they're [right](https://isocpp.org/wiki/faq/const-correctness#constptrptr-conversion), this is not a mistake.

busstoptaktik:
I am aware of the semantics, but I think it is strictly C++?

busstoptaktik:
I am aware of the semantics, but I think the syntax is strictly C++?

aaronpuchert:
You're right. The syntax `const char * const *` is allowed in C and has the same meaning as in C++, but it's not allowed to (implicitly) cast `char **` to `const char * const *`. Apparently C uses simpler (arguably too strict) rules, as explained in the [C language FAQ](http://c-faq.com/ansi/constmismatch.html).

> C++ has more complicated rules for assigning const-qualified pointers which let you make more kinds of assignments without incurring warnings, but still protect against inadvertent attempts to modify const values. C++ would still not allow assigning a `char **` to a `const char **`, but it would let you get away with assigning a `char **` to a `const char * const *`.

I would still argue that using `const char * const *` is right when no modification is intended, but it's not going to make compiler warnings disappear.

src/gie.c
```diff
@@ -1975,6 +1975,10 @@ static int pj_unitconvert_selftest (void) {
     char args6[] = "+proj=unitconvert +xy_in=m +xy_out=m +z_in=m +z_out=m";
     PJ_COORD in6 = {{12.3, 45.6, 7.89, 0}};
 puts ("pj_unitconvert_selftest - start");
```
kbevers: This needs to move down a few lines

kbevers: This needs to go down a few lines

src/pj_factors.c
```diff
@@ -17,12 +17,18 @@ int pj_factors(LP lp, const PJ *P, double h, struct FACTORS *fac) {
     PJ_COORD coo = {{0, 0, 0, 0}};
     coo.lp = lp;
 
-    err = proj_errno_reset (P);
+    /* These are probably due to earlier errors, so we leave errno alone */
```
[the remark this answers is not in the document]
busstoptaktik: "These checks" (implicitly: *Failing* these checks). Replaced by

    /* Failing the 3 initial checks will most likely be due to */
    /* earlier errors, so we leave errno alone */

src/pj_internal.c
busstoptaktik: Refactored, and added the comment `It is the duty of the caller to free this array` at the relevant place

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+        c[i] = c[i + n];
+    c[i] = 0;
+    return c;
+}
+
+
+/***************************************************************************************/
+char *pj_shrink (char *c) {
+/****************************************************************************************
+    Clash repeated whitespace, remove '+' and ';', make ',' and '=' greedy, eating
```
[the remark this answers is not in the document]
busstoptaktik: Thanks!

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+    if (0==c)
+       return 0;
+
+    pj_chomp (c);
+    n = strlen (c);
+
+    /* First clash repeated whitespace (including +/;) */
+    for (i = j = ws = 0;  j < n;  j++) {
+
+        /* Blank out prefix '+', only if preceeded by whitespace (i.e. keep it in 1.23e+08 */
```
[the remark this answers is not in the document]
busstoptaktik: Replaced by "Eliminate"

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
 }
 
 
 
-/* stuff below is *not* considered API, and will be moved to an "internal plumbing toolset" */
+
+/***************************************************************************************/
+size_t pj_trim_args (char *args) {
+/****************************************************************************************
+    Trim an argument string and count its number of elements.
```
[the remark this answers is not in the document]
busstoptaktik: Refactored and added clarifying comments

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+            args[i] = 0;
+            m++;
+        }
+    }
+    return m + 1;
+}
+
+
+/***************************************************************************************/
+char **pj_trimmed_args_to_argv (size_t argc, char *args) {
```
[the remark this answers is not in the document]
busstoptaktik: Done

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+/****************************************************************************************
+    Create an argv-style array from elements placed in args.
+
+    args is a trimmed string as returned by pj_trim_args(), and argv is the number of
+    trimmed strings found (i.e. the return value of pj_trim_args()).
```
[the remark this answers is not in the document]
busstoptaktik: Clarifying comment added

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+****************************************************************************************/
+    size_t i, j;
+    char **argv;
+    /* turn the massaged input into an array of strings */
```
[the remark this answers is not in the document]
busstoptaktik: By following your previous advice, I think this has become more clear from the (improved) context. To avoid any remaining possibilities of confusion, the comment now reads *turn the input string into an array of strings*

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+    }
+    return argv;
+}
+
+
+
+/***************************************************************************************/
+char *pj_free_format_from_argc_argv (size_t argc, const char **argv) {
+/****************************************************************************************
+    Convert free format command line input to something proj_create can eat.
```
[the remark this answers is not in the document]
busstoptaktik: It was intended as a joke, referring to the operators being "greedy". But I think you're right that consume is less surprising, and may aid in the consumption of the code, so replaced by *consume*

src/pj_internal.c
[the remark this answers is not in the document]
busstoptaktik: @schwehr I agree that uninitialized variables are in general horrible. Zeroing them at definition time (in cases where a proper value can first be computed later) is, however, not always the proper thing to do, since the compiler may warn about unused assignments, and we treat warnings as errors in all builds.

src/pj_internal.c
```diff
@@ -134,36 +136,193 @@ size_t pj_strlcpy(char *dst, const char *src, size_t siz) {
...
+****************************************************************************************/
+    size_t i, n;
+    char *p;
+
+    for (i = n = 0;  i < argc;  i++)
```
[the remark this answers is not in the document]
busstoptaktik: Clearly a better choice - thanks!

src/pj_param.c
busstoptaktik: Changed to *As pj_mkparam, but payload ends at first whitespace, rather than at end of **str***

src/pj_param.c
busstoptaktik: I don't think there is any policy, but the comment by @aaronpuchert above, we should probably eliminate all calls to `strncpy` in order to be compliant with the explicit policy of being straight C89.

src/projects.h
busstoptaktik: For extra safety, and because it better reflects the actual effect of the function, I have changed the name to `pj_expand_init` and added an extensive comment describing the action of the function.

This will make it *fail early and noisily* for anyone who may have used pj_get_init, and provide the necessary details for migrating.

src/readblock.c
busstoptaktik: Done

src/readblock.c
busstoptaktik: Done

src/readblock.c
busstoptaktik: Done

kbevers:
Good job on weeding out the bugs. Are you're a happy with this now? Time for a new round of (quick) reviews?

src/readblock.c
busstoptaktik: Done

src/readblock.c
busstoptaktik: Done

src/readblock.c
busstoptaktik: Improved and increased comment to address this

busstoptaktik:
@kbevers -  yes, I think (with the last round of responses to the review by @schwehr) things are getting ready for merging. I intend to merge later today if no objections pop up.

kbevers:
I'll give it a good this afternoon then. Can you rebase the commits before that? It is quite hard to follow at the moment with 50+ debugging commits.

busstoptaktik:
@kbevers rebasing this is apparently beyond my git-fu: I tried, but run into problems with some of the merges-of-master into the PR (two have been needed to eliminate conflicts).

Although I see the annoyance in following this huge, fragmented  set of commits, I can also see that github claims to be able to merge automatically, and had hoped to go ahead by just squashing everything at merge time.

There really is not much story to be told by keeping more than one final commit: This PR is just plain messy, because it messes around in already messy parts of the code, attempting (with some success) to partially un-messify it, although taking some temporarily messifying debugging detours on the way.

rouault:
You could potentially grab https://github.com/OSGeo/proj.4/pull/693.diff and apply it as a patch (patch -p0 < 693.diff) to current master and commit that (make sure to git add / git rm files if needed)

rouault:
Ah I see there's a "Squash and merge" option in the Merge pull request combobox.

busstoptaktik:
@rouault - yes, that was what I figured I would want to use, to make Github make up for my meagre git-fu.

Your diff/patch trick looks neat, but I have even lost track of adds and removes along the way (some files have had temporary lives during development), so I would not feel entirely confident doing that.

busstoptaktik:
OK - all compile checks have passed, all regression tests pass. Time to merge this monstrous PR, which cleans a bit up in a complex part of the code (pj_init.c) and much improves the use case of PROJ as an authoring/registration/validation tool for geodetic transformations.

Huge thank yous to everyone helping and commenting - especially @aaronpuchert, @rouault, @kbevers, and @schwehr.

appveyor.yml
```diff
@@ -58,11 +58,13 @@ test_script:
...
   - echo "Contents of current directory:"
   - dir
+  - echo "Contents of ..\test\gie:"
+  - dir ..\test\gie
   - gie.exe ..\test\gie\builtins.gie
-  - gie.exe ..\test\gie\more_builtins.gie
   - gie.exe ..\test\gie\deformation.gie
   - gie.exe ..\test\gie\axisswap.gie
   - gie.exe ..\test\gie\ellipsoid.gie
+  - gie.exe -vvvvv ..\test\gie\more_builtins.gie
```
kbevers, line 67: Is there a reason why this is super verbose compared to the other tests?

nad/val_def.demo
```diff
@@ -0,0 +1,83 @@
+
+-----------------------------------------------------------------------
+
+      INTEGRATING DEFINITION AND VALIDATION OF GEODETIC SYSTEMS
```
kbevers, line 4: Perhaps this should be placed in `examples/` instead?

busstoptaktik: Sorry, @kbevers - two comments from you popped up when I pressed "squash and merge". Will handle these in a later PR.

src/gie.c
```diff
@@ -1706,10 +1890,10 @@ static int pj_unitconvert_selftest (void) {
     ret = test_time(args3, 1e-6, in3, in3);   if (ret) return ret + 30;
     ret = test_time(args4, 1e-6, in4, exp4);  if (ret) return ret + 40;
     ret = test_xyz (args5, 1e-10, in5, exp5); if (ret) return ret + 50;
+    ret = test_xyz (args6, 1e-10, in6, in6);  if (ret) return ret + 50;
```
kbevers, line 1893: Delete this line. It seems to be a mishap in solving a merge conflict. The same thing is tested twice here, with the first returning a unfortunate code (this is test 6, but it looks like it is test 5 if it fails). The next line is correct.

nad/val_def.demo
kbevers: And there's a few more coming. Would have been nice to be able to do a final review before merging.

src/pj_init.c
```diff
@@ -27,320 +27,352 @@
...
-    while (*next_char)
-    {
-        next_char = fill_buffer(state, next_char);
+    fname = pj_malloc (MAX_PATH_FILENAME+ID_TAG_MAX+3);
+    if (0==fname) {
+        pj_dealloc (line);
+        return 0;
+    }
+
+    /* Support "init=file:section", "+init=file:section", and "file:section" format */
```
kbevers, line 102: > ... "file:section" format

You haven't really advertised this last part in the PR. Effectively this introduces a special new type of key/value description only valid for init-files. I am not entirely sure it is a smart move to allow such a special case control structure.

src/pj_init.c
```diff
@@ -27,320 +27,352 @@
...
-    init_items = pj_search_initcache( name );
-    if( init_items != NULL )
-    {
-        next->next = init_items;
-        while( next->next != NULL )
-            next = next->next;
-        *found_def = 1;
-        return next;
-    }
+           'init=foo:bar ellps=intl proj=utm zone=32 ellps=GRS80',
```
kbevers, line 341: This is new behaviour (previously the overridden parameter would be left out) and is potentially very confusing. Not sure what I think about it. At the very least it should be documented somewhere.

test/gie/pip.gie
```diff
@@ -0,0 +1,41 @@
+These two tests have occasionally been causing segfaults from gie, but
```
kbevers, line 1: This looks like leftovers from the debugging session. If not, it should have a more descriptive name than "pip".

travis/install.sh
```diff
@@ -81,7 +81,7 @@ if [ $TRAVIS_OS_NAME == "osx" ]; then
 make -j3
 make check
 PROJ_LIB=$GRIDDIR ./src/gie ./test/gie/builtins.gie
-PROJ_LIB=$GRIDDIR ./src/gie ./test/gie/more_builtins.gie
+PROJ_LIB=$GRIDDIR ./src/gie -vvvvv ./test/gie/more_builtins.gie
```
kbevers, line 84: Same comment as in the appveyor file. Why the extra need for verbosity here?

appveyor.yml
busstoptaktik: No reason (anymore) - cleaned up as part of #714

src/pj_init.c
busstoptaktik: I considered changing init=foo to xnit=foo when expanding, and adding beginit=foo/endinit=foo sentinels around the expanded expression. This would ease the comprehension of list dumps. But since it is not necessary currently, I just left them out.

Leaving the init=foo thing in the list made it possible to handle lists with the init at the first position of the list, without having to juggle around with a pointer-to-pointer-to-list, which was confusing and error prone (perhaps partially because it - like everything else in that corner of the code - was undocumented).

So it made for a much cleaner call sequence, by giving up the possibility of dropping the list element. We can, however, still mark it as expanded, by changing init= to e.g. exit=, xnit=, tini=, or any other 4 letter tag, should it show necessary.

Since the overall design of the previous version of the code was highly confusing, and came with "plenty of rope to shoot oneself in the foot with", I went for a call convention that was as tightly pruned as at all possible.

test/gie/pip.gie
busstoptaktik: Exactly - will remove

travis/install.sh
busstoptaktik: Same reply :-)

src/gie.c
busstoptaktik: done

src/pj_init.c
busstoptaktik: I do not think this is the case: The function is entirely internal, called only from pj_init.c. The "skip init=" part is needed because it serves both calls for init expansion (syntax init=foo:bar) and default expansion (syntax foo:bar)

src/pj_init.c
busstoptaktik: Actually now introduced the one line patch changing an expanded init to an xnit

kbevers: Hmm.. Now that I read it again, I wasn't very clear in my comment. I have no problem with init=xxx staying in there - that's the way it has been previously. My issue was with having two occurences of `ellps` which makes it difficult to comprehend which one is used. But having though it over I don't think it is a problem since only the first occurence will get marked as "used".

I don't like the `xnit` variant of `init`.

src/pj_internal.c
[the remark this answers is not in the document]
busstoptaktik: @aaronpuchert, I think your statement about strncpy if wrong: if I read the page at the end of the link you provide correctly, the prototype simply changed to include the 'restrict' qualifier with the publication of the C99 standard.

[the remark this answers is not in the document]
schwehr: strncpy is in C89/90.

https://stackoverflow.com/questions/17014835/where-can-i-find-the-c89-c90-standards-in-pdf-format
http://port70.net/~nsz/c/c89/c89-draft.txt :

    4.11.2.4 The strncpy function

    Synopsis

         #include <string.h>
         char *strncpy(char *s1, const char *s2, size_t n);

http://pubs.opengroup.org/onlinepubs/9699919799/functions/strncpy.html

    CHANGE HISTORY
        First released in Issue 1. Derived from Issue 1 of the SVID.

    Issue 6
        The strncpy() prototype is updated for alignment with the ISO/IEC 9899:1999 standard.

    Issue 7
        The stpncpy() function is added from The Open Group Technical Standard, 2006, Extended API Set Part 1.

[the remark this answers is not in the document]
busstoptaktik: @schwehr thanks for a link to an excellent resource - diving into it, it took med 3 hours to get back to the surface :-)

I will drop my planned effort of removing all calls to `strncpy`.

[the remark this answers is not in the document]
aaronpuchert: You are right, I was a bit quick to judge.

PR #694, opened 2017-12-06
Use t_epoch and t_obs instead of epoch and tobs.

kbevers:
Prefixing with t_ is more consistent with the existing parameters
in PROJ, such as x_0 and friends. t_epoch and t_obs is already used
in PJ_deformation. Now users can expect consistency across
time-varying transformations.

Default behaviour in the Helmert transform is changed to use the approximate equations instead of the exact equations, since in most cases transformations are expressed in terms of the approximate formulas.


busstoptaktik:
Awesome. Both the improved consistency for parameter names, and the improved adherence to the "principle of least astonishment" by defaulting to linear approximation of the rotation matrix.

kbevers:
Documented here: https://kbevers.github.io/usage/operations/transformations/helmert.html

Issue #696, opened 2017-12-06
Fix Visual Studio cmake compiler options


pstieber:
Using the current master branch (cbb497636ec29f5350864b128e7d09d603f6a15d), line 20 of `CMakeLists.txt` https://github.com/OSGeo/proj.4/blob/master/CMakeLists.txt#L20 uses the Visual Studio C/C++ compiler option to makes warnings equivalent to errors (`/WX`).  Unfortunately, the compiler considers certain functions security threats and is flagging them as C4996 warnings (https://msdn.microsoft.com/en-us/library/ttcz0bys.aspx).  Adding `/wd4996` to the compiler options in line 20 of `CMakeLists.txt` fixes the problem, but I see other attempts to deal with MSC specific code in headers (https://github.com/OSGeo/proj.4/blob/master/src/projects.h#L33) so I'm not sure what this projects policy is for dealing with such issues.

PR #698, opened 2017-12-07
Squelch Visual Studio's obnoxious warnings about sprintf being unsafe.

### LINKED ISSUE #695: Incorrect declaration of hypot using Visual Studio 2017
I'm attempting to build the master branch (86b96ccb8c926f831451c3e6d364319a0d833522) using the 64-bit version of Visual Studio 2017 version 15.5.0.  The default cmake setup indicates `Performing Test C99_MATH - Failed`. This leads to a declaration of `hypot` via https://github.com/OSGeo/proj.4/blob/master/src/projects.h#L97, but this declaration causes the following compiler warning:

```
2>C:\OutsideSource\proj.4\src\projects.h(97): error C2220: warning treated as error - no 'object' file generated
2>C:\OutsideSource\proj.4\src\projects.h(97): warning C4273: 'hypot': inconsistent dll linkage
2>C:\Program Files (x86)\Windows Kits\10\Include\10.0.16299.0\ucrt\corecrt_math.h(509): note: see previous definition of 'hypot'
```

This indicates `hypot` is declared in `corecrt_math.h`, which is included via `math.h`.

The preprocessor test to include `hypot` was changed by @kbevers 29 days ago.

https://github.com/OSGeo/proj.4/commit/e09999127da8dea56d9af52bc346cf7cc6613547

Is the `C99_MATH` test is flawed?

cffk:
Also set policy CMP0054 to stop a cmake warning.
This fixes #695

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,8 @@ if (NOT CMAKE_VERSION VERSION_LESS 3.1)
 endif ()
 # Set warnings
 if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
-  set(CMAKE_C_FLAGS "/WX ${CMAKE_C_FLAGS}")
+  # Suppress warning 4996 about sprintf, etc., being unsafe
+  set(CMAKE_C_FLAGS "/wd4996 /WX ${CMAKE_C_FLAGS}")
 elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
   set(CMAKE_C_FLAGS "-std=c89 -Wall -Wextra -Wswitch -Werror \
     -Wunused-parameter -Wmissing-prototypes -Wmissing-declarations -Wformat \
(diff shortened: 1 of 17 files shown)


cffk:
Whoops, this is to fix #696 (not #695).

pstieber:
@cffk Wouldn't the addition of `/wd4996` on line 20 of `CMakeLists.txt` be a simpler solution that wouldn't pollute the code base with Microsoft specific `pragma` lines?

The following works for me:

```
if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
  set(CMAKE_C_FLAGS "/WX /wd4996 ${CMAKE_C_FLAGS}")
elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
```

cffk:
I can go either way on this... The benefits of my approach:
* targets just the files in question;
* handles the case where the medicine needs to be applied to installed header files.

kbevers:
After thinking about this for a while, I lean towards the "simple" solution of adding `/wd4996` to the CMake flags. I prefer to have the code base as neutral as possible. Keeping the platform specifics (mostly) to header files and build systems achieve that goal. It makes things a little easier for for one-time contributors as well as allow us to govern the rules for warning- and error-handling centrally. Or at least as centrally as possible when maintaining three orthogonal build systems...

kbevers (approved):

PR #704, opened 2017-12-11
Allow including the new API header in proj.c

aaronpuchert:
This didn't work before because the elaborate type punning scheme was deactivated by including proj.h. Now we use the type punning from the new API header and build our own.

This change is required for #674.

### DIFF
--- a/src/proj.c
+++ b/src/proj.c
@@ -1,4 +1,5 @@
 /* <<<< Cartographic projection filter program >>>> */
+#include "proj.h"
 #include "projects.h"
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


aaronpuchert:
@rouault This change would allow me to address your [comment](https://github.com/OSGeo/proj.4/pull/674#issuecomment-350254095), please have a look.

rouault:
Looks good to me. @kbevers Seems github suggests you as reviewer

kbevers:
Seems good to me as well. If this makes things easier elsewhere I am fine with merging.

Of course the gold star solution would be to reimplement `proj`  using only the new API, but that is a task for another day :-)

PR #383, opened 2016-05-23
Add BETA2007 grid and change potsdam datum

rouault:
Functionnaly equivalent to https://github.com/OSGeo/proj.4/pull/371 by Jürgen Fischer

See
http://crs.bkg.bund.de/crseu/crs/descrtrans/BeTA/de_dhdn2etrs_beta.php

Confirmed with Uwe Schmitz uwe.schmitz@bezreg-koeln.nrw.de that free
redistribution is allowed and welcome.

PR #708, opened 2017-12-17
Add Australian ICSM transformation grids

nyalldawson:
These files are available under a BSD-3 Clause license from https://github.com/icsm-au/transformation_grids

I can see the file size for these grids may be an issue. But having them included by default with proj would certainly be a big win for open source geo software in Australia...

### DIFF
--- a/nad/A66_National_13_09_01.LICENSE.md
+++ b/nad/A66_National_13_09_01.LICENSE.md
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2017, ICSM Australia
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of the copyright holder nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/nad/GDA94_GDA2020_conformal.LICENSE.md
+++ b/nad/GDA94_GDA2020_conformal.LICENSE.md
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2017, ICSM Australia
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of the copyright holder nor the names of its
+  contributors may be used to endorse or promote products derived from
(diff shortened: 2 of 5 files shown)


busstoptaktik:
Do you have any idea why a BSD software license is used for data? A creative commons license (and preferably CC0, although I think CC-BY is more in line with the 3 clause version of BSD) would make more sense.

PROJ is distributed under the MIT license, and although similar, MIT and BSD are not identical - I'm not sure what license the final product should be distributed under.

nyalldawson:
@alexgleith

Any insights into this license choice?

alexgleith:
Hey folks

To be honest, the license choice was fairly arbitrary. We were aiming to be as permissive as possible.

If this license is too restrictive, let me know what is the preferred license and I can put a decision up to change it.

nyalldawson:
Any reason why not to just go public domain and be done with it?

QuLogic:
Because public domain is not really a valid license everywhere.

Is Australia still not [part of the Open Government Partnership](https://www.finance.gov.au/blog/2013/05/22/australia-joins-open-government-partnership/)? IOW, default to CC-BY.

rouault:
Rather than putting those quite heavy grids in git, why not including them in a new version of proj-datumgrid-XXXX.zip available in http://download.osgeo.org/proj/ ? Binary distributions of proj normally ship it as a recommended package with libproj (proj-nad on Debian/Ubuntu, proj-datumgrid  on osgeo4w)

We might perhaps also consider setting up a dedicated proj-grid repo if we still want to benefit from git, without being too annoying for most contributors that don't want to clone a too large repo just from libproj itself

kbevers:
> We might perhaps also consider setting up a dedicated proj-grid repo if we still want to benefit from git, without being too annoying for most contributors that don't want to clone a too large repo just from libproj itself

+1! This is in my opinion the correct way forwards. It will make it more obvious how to add new grids to the grid distribution, keep track of history (the egm96_15 grid is a good example of why that's beneficial) and we don't have to drag along unnecessary MB's in the main repo.

alexgleith:
Regarding a separate grid repo, I think that's a great idea! But I don't know how they integrate into the package, so I defer to those that do.

Regarding licenses, does CC-BY have an advantage over BSD? I thought BSD was more permissive. Regardless, I think that the license that is here means that they can be used in the project and redistributed...

busstoptaktik:
@QuLogic can you provide a reference indicating that joining the Open Government Partnership implies defaulting to a CC-BY license? I cannot find it on the OGP web site, and being able to point people to the relevant place would be a huge help in arguing for the "use of useful licenses".

The EU commission actually recommends CC0 for public sector data, but it is a recommendation, not a requirement, and not a default. If OGP membership means defaulting to CC-BY, many roundabouts could probably be avoided.

busstoptaktik:
@alexgleith - BSD is for (source) code. CC-BY is for any product, although most useful for material that is human readable (reports, artwork, poetry, raw data, whatever).

The BSD license explicitly mentions "source code" in its license text, which is why I find it "not entirely suitable" for data distribution (although any data set could be wrapped in `double *data = {/* data set goes here */};` and play the role of source code).

If you want to be as interoperable as possible, use CC0. If attribution is important, use CC-BY.

andrewharvey:
> Because public domain is not really a valid license everywhere.

CC0 a copyright license granting for all intensive purposes the same rights as public domain.

I'm in favour of CC0.

alexgleith:
Ok, thanks @busstoptaktik, that's convincing enough.

I'll have a go at getting the ICSM Transformation File repo changed to CC0 or CC-BY.

QuLogic:
@busstoptaktik Joining the OGP does not imply a specific license; only that it's open. The link to the Australian Government's press release that I posted does however state in the 5th bullet point that one of its goals is to default to CC-BY. For example, everything published on [data.gov.au is CC-BY](https://data.gov.au/about).

I'm not Australian so I can't really point to the specific legislation or commitments made since then.

alexgleith:
Can I please confirm that licensing the grids as CC-BY-4.0, as promoted by AusGOAL, is appropriate for incorporation into software projects? There was a [bit of discussion in the past](http://osgeo-org.1560.x6.nabble.com/ICSM-Australia-transformation-file-licensing-td5303797.html) that sort of suggested that it wasn't appropriate.

alexgleith:
I have approval to change the license to [CC-BY-4.0 International](https://creativecommons.org/licenses/by/4.0/).

Can I ask a couple of things:

 * Is this definitely compatible with this project and better than BSD 3 clause?
 * Can we agree on the new repo for transform grids now, and if so, I'll set up a PR on there?

alexgleith:
Hey @nyalldawson, license has been changed to CC-BY-4.0, see: https://github.com/icsm-au/transformation_grids

sebastic:
This issue should be moved to the https://github.com/OSGeo/proj-datumgrid repo which was created to build proj-datumgrid archives.

busstoptaktik:
> This issue should be moved to the https://github.com/OSGeo/proj-datumgrid repo which was created to build proj-datumgrid archives.

... and probably the same is the case for #383

sebastic:
> ... and probably the same is the case for #383

That PR also changes `src/pj_datums.c` which is appropriate for this repository.

The `nad/BETA2007.gsb` could be included in proj-datumgrid, but proj already includes the French NTv2 grid shift file (`ntf_r93.gsb`) so the German NTv2 grid shift file can also be included in proj itself.

kbevers:
> That PR also changes src/pj_datums.c which is appropriate for this repository.
>
> The nad/BETA2007.gsb could be included in proj-datumgrid, but proj already includes the French NTv2 grid shift file (ntf_r93.gsb) so the German NTv2 grid shift file can also be included in proj itself.

You could also argue that the `+datum` option is a legacy feature that should not be expanded on. That is my personal opinion. I'd rather go in the direction of separating grids and init-files completely from the library and having that as external packages that you can download to suit your needs. At the moment we are giving some users special treatment, I am not in favour of that. 

I haven't thought this all the way through yet but at some point I will come with a proposal the subject of how we treat grids and init files in the future.

busstoptaktik:
Having re-read the discussion over at #383, I think that PR *should* be merged as part of the 5.0.0.RC2 work. - I have written a few words about why over at #383, and (well knowing and admitting that I was the one mixing that case into this) suggest to move the part of this discussion relating to the BeTA grid back to its home PR.

nyalldawson:
PR opened for proj-datumgrid at https://github.com/OSGeo/proj-datumgrid/pull/3

PR #729, opened 2018-01-04
Bar step-internal params from influencing pipeline driver init

busstoptaktik:
In cases such as:

    proj=pipeline step proj=utm zone=32 ellps=GRS80 ...

The pj_init code would pick up the `ellps` information (and other material such as false eastings and northings) from the utm step and place it into the `PJ` object of the pipeline driver.

This is not the intention, and is eliminated in this PR by terminating parameter searching (done by `pj_param`) once a `step` parameter is reached.

### DIFF
--- a/src/pj_param.c
+++ b/src/pj_param.c
@@ -56,6 +56,11 @@ paralist *pj_param_exists (paralist *list, const char *parameter) {
     Determine whether a given parameter exists in a paralist. If it does, return
     a pointer to the corresponding list element - otherwise return 0.
 
+    In support of the pipeline syntax, the search is terminated once a "+step" list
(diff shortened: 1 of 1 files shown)


src/pj_param.c
```diff
@@ -92,78 +100,82 @@ paralist *pj_param_exists (paralist *list, const char *parameter) {
...
 
     type = *opt++;
-    /* simple linear lookup */
-    l = (int)strlen(opt);
-    while (pl && !(!strncmp(pl->param, opt, l) &&
-      (!pl->param[l] || pl->param[l] == '=')))
-        pl = pl->next;
-    if (type == 't')
+
+    if (0==strchr ("tbirds", type)) {
```
kbevers: Birds?

Issue #700, opened 2017-12-10
4D API does not process cs2cs-style proj-strings


kbevers:
# The problem

Parameters such as `towgs84`, `nadgrids`, `geoidgrids`, `axis`, `units` and `vunits` are only parsed and used by the `pj_transform` function. For the new 4D API to be truly useful it should be able to convert `cs2cs`-style proj-strings to pipelines. This is especially important as long as we still distribute init-files, such as the `epsg`-file, that uses the `cs2cs` parameters.

It should be possible to mimic the behaviour of `cs2cs` with the many new primitives that can be used in pipelines. The order of the steps in a pipeline is crucial. As far as I can tell from `pj_transform` the order when going from "WGS84"  to e.g. `epsg:xxxx` is as described below. I've used plusses in front of the `cs2cs`-style to make it clearer which is which.

1. Apply datum shift of horizontal coordinates:
First look for a `+datum` parameters and expand it if it exist. After that we look for either `+nadgrids` or `+towgs84`. If both exists (they shouldn't, but they can), `+nadgrids` takes precedence. If we find one of the two parameters it expands to either:
 `+towgs84=???` -> `step proj=helmert x=? y=? z=? s=? rx=? ry=? rz=?`
or
`+nadgrids=???`->  `step proj=hgridshift grids=???`

2. Apply datum shift of vertical coordinates:
`+geoidgrids=????` ->  `step proj=vgridshift grids=????`

3. Projection:
`+proj=????` -> `proj=??? arg1 arg2 arg3 ... argn`

4. Units:
`+units=??? +vunits=???` -> `proj=unitconvert xy_in=m xy_out=??? z_in=m z_out=???`

5. Axis order:
`+axis=???` -> `step proj=axisswap +order=?,?,?,?`

Which in the end becomes pipeline that looks like:

```
proj=pipeline 
step proj=helmert x=? y=? z=? s=? rx=? ry=? rz=?
step proj=vgridshift grids=????
proj=??? arg1 arg2 arg3 ... argn
proj=unitconvert xy_in=m xy_out=??? z_in=m z_out=???
step proj=axisswap +order=?,?,?,?
```

In case the transfomation is reversed, that is `init=epsg:xxxx` to "WGS84", the above is also reversed.

# Implementation

Implementation of the above is not entirely straight forward. We very quickly run into nesting pipelines which we only allow in case they are wrapped in init-files. This is a catch-22 for the epsg init-file and friends.

[...]

PR #705, opened 2017-12-11
Run gie tests with CMake and autotools builds

kbevers:
This adds the gie tests to the CMake and autotest testing facilities. GIGS tests have been converted from Python/json to gie and added as well. Note that some GIGS tests do not pass. The reason for this is described in #700. The failing tests are commented out for now.

Resolves #702.

PR #731, opened 2018-01-06
[WIP] 4D API cs2cs style

busstoptaktik:
This PR is in response to issue #700. It is WIP and currently untested, but I push the PR in order to check (or actually re-establish) orthogonality to pj_transform and cs2cs.

UPDATE: Now very much tested and very little WIP. Getting ready for merging.


busstoptaktik:
Neat - we now have (to the extent test coverage allows) confirmed the orthogonality of this PR to the cs2cs and pj_transform sub-systems.

Next step will be to construct test data, to confirm (or probably rather reject) the hypothesis that the code does what it's supposed to. Despite just having to follow the recipe by @kbevers from #700, I doubt that I have managed to get everyting right at first shot.

src/pj_fwd.c
```diff
@@ -54,18 +54,46 @@ PJ_COORD pj_fwd_prepare (PJ *P, PJ_COORD coo) {
...
+        }
+        if (P->vgridshift)
+            coo = proj_trans (P->vgridshift, 1, coo);
+        return coo;
+    }
 
-        coo.lp = lp;
+
+    if (P->left==PJ_IO_UNITS_CARTESIAN) {
+        /* Do we really support gridshifts on cartesian input? */
```
kbevers: > Do we really support gridshifts on cartesian input?

Not in "classic PROJ" mode. This can only occur if when using the deformation operation which is not compatible with classic PROJ.

src/PJ_helmert.c
```diff
@@ -486,6 +486,22 @@ PJ *TRANSFORMATION(helmert, 0) {
         P->right = PJ_IO_UNITS_PROJECTED;
     }
 
+    /* Support the classic PROJ towgs84 parameter, but allow later overrides.*/
+    /* Note that if towgs84 is specified, the datum_params array is set up   */
+    /* for us automagically by the pj_datum_set call in pj_init_ctx */
+    if (pj_param (P->ctx, P->params, "ttowgs84").i) {
```
kbevers: pj_param_exists?

src/pj_fwd.c
```diff
@@ -81,9 +109,42 @@ PJ_COORD pj_fwd_finalize (PJ *P, PJ_COORD coo) {
     }
 
     /* Handle false eastings/northings and non-metric linear units */
-    coo.xyz.x = P->fr_meter  * (coo.xyz.x + P->x0);
-    coo.xyz.y = P->fr_meter  * (coo.xyz.y + P->y0);
-    coo.xyz.z = P->vfr_meter * (coo.xyz.z + P->z0);
+    if (P->right==PJ_IO_UNITS_CARTESIAN) {
```
kbevers: in `pj_inv` you define `INPUT_UNITS` and `OUTPUT_UNITS`. Doing that here as well will make it easier to read the code

src/pj_fwd.c
```diff
@@ -81,9 +109,42 @@ PJ_COORD pj_fwd_finalize (PJ *P, PJ_COORD coo) {
...
+    if (P->right==PJ_IO_UNITS_RADIANS && P->left!=PJ_IO_UNITS_RADIANS) {
+        /* Distance from central meridian, taking system zero meridian into account */
+        coo.lp.lam = coo.lp.lam + P->from_greenwich + P->lam0;
+
+         /* adjust longitude to central meridian */
+        if (0==P->over)
+            coo.lpz.lam = adjlon(coo.lpz.lam);
+
+        if (P->vgridshift)
+            coo = proj_trans (P->vgridshift, -1, coo);
```
kbevers: Use `PJ_INV` here. And many other places in the code below.

src/pj_fwd.c
```diff
@@ -81,9 +109,42 @@ PJ_COORD pj_fwd_finalize (PJ *P, PJ_COORD coo) {
...
+        if (P->hgridshift)
+            coo = proj_trans (P->hgridshift, 1, coo);
```
kbevers: `PJ_FWD` here and below.

src/proj_4D_api.c
```diff
@@ -381,6 +381,111 @@ char*  proj_rtodms(char *s, double r, int pos, int neg) {
...
+
+    /* Don't recurse when calling proj_create (which calls us back) */
+    if (pj_param_exists (P->params, "break_cs2cs_recursion"))
+        return 1;
+
+    /* Swap axes? */
+    p = pj_param_exists (P->params, "axis");
+    /* p->param is a *string array*, not a *pointer to char*, */
+    /* so we do not check here and below whether it is non-null */
+    if (p && strcmp ("enu", p->param)) {
```
kbevers: At this point I believe `P->axis` can be expected to contain valid data if it is set. So no need for the `strcmp`.

test/gie/4D-API_cs2cs-style.gie
```diff
@@ -0,0 +1,89 @@
+===============================================================================
+
+Test the 4D API handling of cs2cs style transformation options.
+
+These tests are mostly based on the same material as those in
+more_builtins.gie, since we are testing the same kinds of things,
+but provided through a different interface.
+
```
kbevers, line 8: This should also contain a bunch of tests using the epsg init-file. Test data is easy to prepare with cs2cs. I can come up with some suggestions if needed.

src/pj_fwd.c
```diff
@@ -54,18 +54,46 @@ PJ_COORD pj_fwd_prepare (PJ *P, PJ_COORD coo) {
         if (lp.phi < -M_HALFPI)
             lp.phi = -M_HALFPI;
 
+
+        coo.lp = lp;
```
kbevers: May I suggest calling it `coord` instead. I think it reads better.

kbevers (commented):
Just a few comments. Seems like it's going in the right direction.

src/PJ_helmert.c
busstoptaktik: Yes, it makes sense here, and we could redesign the entire parameter handling in this function to  use `pj_param_exists`, which I think we should do, but not now. Changed this instance, because it is different in having its work already done by a friend up in `pj_datum_set`

src/pj_fwd.c
busstoptaktik: `coo` has been conventionally used here, there, and everywhere:

    grep coo *.c *.h | grep -v coord | wc -l

gives 74 hits.

`coord` is used in methods and type names (`PJ_COORD`, `proj_coord ()`, etc.) - `coo` is used for an otherwise unspecific instance, intentionally modelled after the equivalent unspecificness of the versatile variable `foo` of `foo, bar, baz` fame.

`coord` is used as an instance name a few places in `proj_4D_api.c` and `gie.c` - I would be less reluctant renaming these instances to `coo`, than changing the already widespread `coo` idiom.

src/pj_fwd.c
busstoptaktik: So this will first really be an issue to decide on when we introduce global `+inv` handling and/or a `proj=cartesian` (or `proj=geocent` in deprecated PROJ terms).

I'll leave the code in, but we will have to decide what we mean with this. I can imagine situations where it makes sense, but mostly for convenience (but then again, there is nothing in this PR which cannot be handled by properly constructed pipelines - it is by and large a matter of improving backward compatibility by providing some convenience functionality, for interoperability between new code and old parameter file systems)

src/pj_fwd.c
busstoptaktik: Yes, and I will do that. I managed to do the mental math necessary in the forward case, without having this translation/reminder, whereas when having to do the same thing *backwards and in high heels*, in the inverse case, I really needed these persistent pings

src/pj_fwd.c
busstoptaktik: I don't agree. `PJ_INV/PJ_FWD` were introduced for your convenience. I actually find it quite a lot more readable to have negative mean backward, and positive mean forward.

And since there is more than enough of layered negations in this code, I really prefer the concise to the verbose here.

src/pj_fwd.c
busstoptaktik: (ibid.)

src/pj_fwd.c
kbevers: > coo has been conventionally used here, there, and everywhere

I know, but I don't like the convention. It somehow made sense previously when `coo` was also used in the now gone `PJ_OBS` where a short name was justified. Since that no longer exists I think it is perfectly reasonable to use a more descriptive convention.

I would also object to using non-descriptive names like `foo` and `bar` in anything other than example code so I am not buying that argument.

Anyway, my point is just that `coord` makes more sense to readers unfamiliar with the code-base as a whole. I am not suggesting that the existing occurences of `coo` be replaced now, I just wanted to raise a flag here since it has been bothering me for a while. Do what you feel is best - ultimately it is in the nitpicking department!

src/pj_fwd.c
kbevers: > (but then again, there is nothing in this PR which cannot be handled by properly constructed pipelines - it is by and large a matter of improving backward compatibility by providing some convenience functionality, for interoperability between new code and old parameter file systems)

That's my point. Keep it as simple as possible and don't open the door too much for new ideas based on the classic PROJ.4 idioms.

src/pj_fwd.c
kbevers: I guess we will just have to agree to disagree here :-)

busstoptaktik: Sounds very agreeable :-)

src/pj_fwd.c
busstoptaktik: OK, I get your point now, and I fully agree, so I will remove that chunk of code

src/proj_4D_api.c
busstoptaktik: The strcmp returns true (i.e. non-null) if  *(p->param) != "enu", so it's a way to avoid calling axisswap if the data are already in the expected order.

I agree that the data must be valid at this point, so this is not a validation, just a cheap optimization, which may deserve a comment for explanation.

test/gie/4D-API_cs2cs-style.gie
busstoptaktik: That would be nice. What is really needed (and will probably cause hiccups) is cases where more than one of the cs2cs style options are used.

src/proj_4D_api.c
kbevers: Ah of course. Makes sense.

test/gie/4D-API_cs2cs-style.gie
kbevers: I'll merge #705 s soon as all the tests pass. Sync your branch with master and you'll have a bunch of test material in `test/gigs/`. I think that should be enough for this purpose.

src/pj_fwd.c
busstoptaktik: @kbevers: Since you dislike the convention, I think we should consider reconsidering the convention(s).

From classic PROJ/Gerald, we have **`P`** for the generic `PJ` object. From 3D-PROJ/Frank, we have **`ctx`** for the generic `PJ_CONTEXT` object. So apparently, the smaller the object, the longer the name.

With that in mind, it actually makes quite a lot of sense to switch **`coo`** (same length as `ctx`, but smaller) to your preferred **`coord`** (longer and smaller than `ctx`).

But if that's to be done I suggest we do it consequently, and reserve a rainy summer day, in time for 5.1.0, for a sweep-over of the entire code base, switching `coo` to `coord` - and collecting a list of other *paper cuts*/minor annoyances to be remedied at the same occasion.

I'll buy the beer, and open the issue for discussion - you bring the stamina, and I'll try to follow your typing pace :-)

src/pj_fwd.c
busstoptaktik: That being said, if you find it unreadable, I would not mind switching to `PJ_FWD/PJ_INV` once we have established that the material works as intended (last step before merging).

It's just that I find the `PJ_FWD/PJ_INV` convention less readable than the mathematical convention of +1/-1, and this stuff is pretty obnoxious already.

test/gie/4D-API_cs2cs-style.gie
busstoptaktik: Have sync'ed, but currently drowning in all the "almost OK" tests. I think there is a bug in the interpretation of the GIGS tolerance parameter.

If I understand @micahcochran's comments correctly, they are considered to be max absolute coordinatewise deviations (i.e. in decimal degrees for latitude/longitude pairs), whereas in the `.gie` version, they are translated into "same numerical value but interpreted as having a *meter* unit".

This makes a.o. all NADCON tests fail, because 1e-7 m (0.0001 mm) is much less than 1e-7 degree (1 cm at the Equator), at least if you are more than a few minutes of arc away from the poles.

kbevers: Which tests fail? There's a few of them that doesn't work in the current python setup as well. It might be that it is only those tests that fail, in which case you can disregard them, since the problems are not related to the work you are doing here.

busstoptaktik: I can see from the notes of @micahcochran that a large number is supposed to fail slightly and a much smaller number is supposed to fail miserably.

But getting the general view of more than 600 failing tests takes some time, so I'm starting out in a horrible way by introducing the infinity norm for any tests where `gigs` is part of the path to the test file.

Whatever remains, I'll check case-by-case - or, if many remain, by whatever works.

kbevers: Here https://github.com/OSGeo/proj.4/blob/master/travis/install.sh#L98-L99 you can see which tests are working. Tests not in those two calls will fail. Maybe that helps you a bit?

busstoptaktik: Certainly - down to 223 failures :-) I'll go on debugging tomorrow

kbevers: Happy hunting :)

busstoptaktik: > Happy hunting :)

Already down to 199...

kbevers:
Regarding test 5208 (and possibly others). The problem is due to prime meridian not being adjusted correctly in the inverse operation, as examplified here:

```
λ echo 0 0 | bin\cct.exe +proj=latlong +pm=paris
 -2.3372291667    0.0000000000        1.#INF        1.#INF

λ echo 0 0 | bin\cct.exe -I +proj=latlong +pm=paris
  0.0000000000    0.0000000000        1.#INF        1.#INF # this should be  2.3372291667    0.0000000000
```

The forward operation is handled correctly because this

```
    case PJ_IO_UNITS_RADIANS:
        if (INPUT_UNITS==PJ_IO_UNITS_RADIANS)
            break;
```

happens in `pj_fwd_finalize` (line 131 and below), which basically omits the adjusted the longitude back to the greenwhich meridian. A similar construct is missing in `pj_inv_prepare` (remember we are going backwards now), so the longigude is adjusted to the desired prime meridan and then back to greenwich. The last bit breaks the test.

test/gie/4D-API_cs2cs-style.gie
busstoptaktik: Still some problems regarding the handling of torad/todeg in gie, when `axis != enu`.

The same problem is, as far as I can see, present in cs2cs: expecting enu, converting first two fields to radians on input, not taking care of different axis order on neither input, nor output.

busstoptaktik: but repaired the double superfluous datum shifts in 5103.1

kbevers: > but repaired the double superfluous datum shifts in 5103.1

It seems to me that you've just changed the test. Is that correct?

kbevers: > Still some problems regarding the handling of torad/todeg in gie, when axis != enu.

> The same problem is, as far as I can see, present in cs2cs: expecting enu, converting first two fields to radians on input, not taking care of different axis order on neither input, nor output.

When I tried to fix the axis ordering business, I missed your `4D-API_cs2cs-style.gie` test since it wasn't added to the CMake test-suite. I see now that there's a few errors. I'm looking into that now.

test/gie/4D-API_cs2cs-style.gie
```diff
@@ -0,0 +1,132 @@
...
+-------------------------------------------------------------------------------
+
+
+
+-------------------------------------------------------------------------------
+operation  proj=latlong geoidgrids=egm96_15.gtx  ellps=GRS80
+-------------------------------------------------------------------------------
+tolerance 15 cm      # lax tolerance due to widespread bad egm96 file
+accept    12.5 55.5   0
+expect    12.5 55.5 -36.0213
```
kbevers, line 91: With `cs2cs` I get the following output:

```
λ echo 55.5 12.5  0|bin\cs2cs.exe -f %.2f +proj=latlong +to +proj=latlong +geoidgrids=egm96_15.gtx  +axis=neu +ellps=GRS80
12.50   55.50 42.37

C:\dev\build\ninjaproj
λ echo 55.5 12.5  0|bin\cs2cs.exe -f %.2f +proj=latlong +geoidgrids=egm96_15.gtx  +axis=neu +ellps=GRS80 +to +proj=latlong
12.50   55.50 36.02
```
gie returns:

```
     -----
     FAILURE in 4D-API_cs2cs-style.gie(94):
     expected: 12.5 55.5 -36.0213
     got:      12.500000000   55.500000000   42.372592926
     deviation:  78393.893 mm,  expected:  150.000 mm
``` 

which corresponds to the first call to `cs2cs` above. I guess this all comes down to what we mean by `+axis`. Does `+axis` describe the ordering of the input or the output coordinates?

test/gie/4D-API_cs2cs-style.gie
busstoptaktik: > It seems to me that you've just changed the test. Is that correct?

I overwrite the `towgs84=...` elements from the epsg file with a null operation which is, in turn, ignored by the prepare/finalize code.

This makes pj_fwd/pj_inv work the same way as pj_transform: omitting dual identical helmert shifts, so it doesn't really change the test - it rewrites the test in order to act identically to the json case.

The alternative to declaring this *an idiom for signalling "no datum shift here please"*, is to write a large amount of quite intrusive, and probably quite brittle, code in `pj_pipeline.c` doing hand held manipulation across otherwise individual pipeline steps.

I intend to do further clean up of the gigs handling in gie in time for 5.1.0, so this is a simple, non-intrusive stop gap for now.

busstoptaktik: > I'm looking into that now.

In that context, please note that I have merged your previous PR over at busstoptaktik/PROJ.4.

kbevers: > This makes pj_fwd/pj_inv work the same way as pj_transform: omitting dual identical helmert shifts, so it doesn't really change the test - it rewrites the test in order to act identically to the json case.

I understand why you've done it, I just don't think it is going to work. The end goal of this PR is, amongst other things, to be able to put to `epsg` CRS definitions into `proj_create_crs_to_crs()` and get correct results when using the resulting `PJ` in transformations. While your solution fixes the test, it doesn't fix the transformation in the test when used in the wild (in the 4D API).

kbevers: > In that context, please note that I have merged your previous PR over at busstoptaktik/PROJ.4.

Duly noted.

busstoptaktik: > The end goal of this PR is, amongst other things, to be able to put to epsg CRS definitions into `proj_create_crs_to_crs()` and get correct results when using the resulting PJ in transformations

`proj_create_crs_to_crs()` is a much different situation:

**First**, we do not necessarily need this, if we can live with the time and slight precision penalty:  For typical crs_to_crs uses we do just one half-roundtrip, and do not accumulate helmert imprecision.

**Second**, we can implement the `pj_transform` trick as an optimization in a much simpler way than in the generic `pipeline` case: Here we know we have a one input, one output situation, and if both are from epsg definitions we can check whether the `towgs84=...` incantations are identical and bypass Helmert by setting the `towgs84=0.0.0` thingy at the `proj=pipeline` level.

This can be economically implemented by first changing

        strcpy(buffer, "+proj=pipeline +step +init=");

to

    strcpy(buffer, "+proj=pipeline +towgs84=0,0,0 +step +init=");

Then going on with the setup, but adding some 15 lines of code, checking the case of identical Helmert parameters (instantiate, check), and then finally:

    if (helmerts_not_identical)
        return P;
    proj_destroy (P);
    t = strstr (buffer, "towgs84");
    strcpy (t, "tonight"); /* or xxxxxxx or whatever fits */
    P = proj_create(ctx, buffer);
    return P;

Simpler and more localized than solving the general problem in `pj_pipeline`

So, by and large, see the handheld introduction of towgs84=0,0,0 in the `*.gie` files as a temporary exercise, helping pinpoint some cases where we can do *somewhat* mechanically introduced optimizations in a comming (version 6-ish) early version of a *late binding* EPSG implementation.

kbevers: Okay, that works for me. Do you want to make that change to `proj_create_crs_to_crs`? It fits nicely within this PR.

busstoptaktik: Sure, but I'm offline until Monday, so not until then.

kbevers:
Seems like this is getting close to the finish line. I still get errors when running the 5201.gie test. I guess it is not run on CI yet. The problem is related to +proj=geocent, which at the moment is not taken care of in the cs2cs emulation. I don't think it is too difficult to take care of.

PR #734, opened 2018-01-09
Remove or use unused macros

aaronpuchert:
Some macros seemed to be leftover from earlier code, so I removed them. Others seemed like they should have been used, but weren't.

There should be no functional change, except the following: in floating-point arithmetic, x / y is not the same as x * (1.0 / y). It can be argued that using the multiplication is significantly faster, and the algorithm is approximative anyway. Otherwise, the constants are obviously not required.

Also fixes one location in PJ_aitoff.c, where an enumeration value should have been used.


kbevers:
Seems good to me. Thanks.

PR #577, opened 2017-09-25
Docs for the upcoming release

kbevers:
Essentially a duplicate of https://github.com/OSGeo/proj.4/pull/563, this time on a branch in the project repository. This way it is easier for others to contribute as well. The overall structure of the docs are in place now, so feel free to correct existing text and add additional text.

Closes https://github.com/OSGeo/proj.4/issues/442
Closes https://github.com/OSGeo/proj.4/issues/454

## Want to contribute to the version 5.0.0 docs?

Awesome! It's quite simple:

* Clone the PROJ.4 repo
* Check out the branch `docs-release-4.10.0` (it was created before we decided on v. 5)
* Document something
* If  you have commit access to the main repo, you can push your changes directly to this branch, otherwise you will have to make a pull request targeting the `docs-release-4.10.0` branch. Please add `[skip ci]` to your commit messages (otherwise proj4.org will be updated with the v. 5.0.0 docs...).
* Feel good about yourself for giving some much needed love to the PROJ.4 docs!

The reason the docs for 5.0.0 are maintained in a separate branch is that changes to the docs in the master branch get's pushed to proj4.org as soon as they land in master. Un-released features should not be documented on the website.

## Readable version of the new docs: https://kbevers.github.io/ 

I try to keep this updated when new changes are made.

Issue #735, opened 2018-01-10
PROJ version 5 vs. PROJ.4


busstoptaktik:
**TL;DR: Some problems associated with the PROJ.4-to-PROJ name change can be worked around by adopting a few *additional* naming conventions.**

Rationale
---------

About a month ago, @kbevers and @hobu conducted a discussion on the mailing list, ending up in adopting that the next version of PROJ.4 will not be PROJ.4 version 5.0.0, but PROJ version 5.0.0.

I didn’t take part in the discussion, as I was quite busy actually developing what is now about to become version 5.0.0, so I'm sorry for being late to the party, but I have some additional input which I think is relevant.

I also think this input first and foremost calls for developer consensus, which is why I post it as an Github issue, rather than a mailing list discussion item (obviously, the mailing list is where it will go, once I have seen developer consensus).

The Problem
-----------

The problem with the name change is that, while decoupling the name of PROJ with the version number buys us a communication advantage, it also buys us a lot of trouble which, I believe, can be worked around by adopting a few more naming conventions.

The trouble is that the name change invalidates the connection between PROJ and the URLs associated with PROJ. That is (mostly) http://proj4.org and https://github.com/OSGeo/proj.4, but also (probably) a number of OSGeo pages, and e.g. the Wikipedia page https://en.wikipedia.org/wiki/PROJ.4.

Also, both the PROJ.4 and PROJ names are, in various combinations, baked into build- and packaging systems, where it will be a hassle (and largely out of our hands) to change - cf. the mailing list discussion http://osgeo-org.1560.x6.nabble.com/Time-for-a-new-release-td5341895.html.

Solution Proposed
-----------------

The solution I propose is to **keep PROJ.4** - not as the name of the *product*, but as the name of the *project*. With a few more straightforward naming conventions, the message we will communicate is this:

* We are the PROJ.4 project
* We provide the product **PROJ**, which is now at version 5.0.0
* The foundational component of PROJ is the library `libproj`
* Other components include the application `proj`, which provide a command line interface to `libproj`
* You can contribute to the project by e.g. providing test data, writing documentation, suggesting improvements, helping fellow users on the mailing list, and by submitting code improvements and extensions

By **switching** the name of the product from PROJ.4 to PROJ, we gain the advantage of a product name that does not need an explanation to go with it.

By **keeping** PROJ.4 as the name of the project, rather than the product, we avoid a lot of unneeded trouble with web presence etc.

[...]

Issue #686, opened 2017-11-27
API migration timeline. Continuation of side-discussion in #674


kbevers:
I asked an off-topic question in https://github.com/OSGeo/proj.4/pull/674#issuecomment-347221630 regarding the plan ahed API wise. Continuing the discussion here as to not clutter the existing thread.

My question:

> Do you mind outlining such a plan? At the moment we are all working in sort of the same direction, but my perception is that there is no common understanding of what exactly that direction is. It would be nice to have that settled once and for all. Coming from you, the project maintainer, it can be regarded as the "authorative plan" I guess.

to which @hobu answered:

> Well, I'm the self-appointed maintainer, and I'm happy to hand off the hat to people actually doing all the work 😄 I've been trying to stay back and let you guys evolve things without worrying too much about the project maintenance implications.
> 
> In my opinion, the 5.0.0 release cannot break projects that are using the existing API, except to force them to recompile. I think it is just fine to annoyingly warn them the API they're using is deprecated, and in one or two releases, require them to #define I_KNOW_THIS_API_IS_OLD to continue to use the old API. We can't force a migration by breaking compilation of existing usage that's out there in one release though, or we'll end up splitting the project into two, and the packagers will choose the old one and start patching because of all of the existing software.
> 
> ![image](https://user-images.githubusercontent.com/87751/33276224-8b2d0fb2-d35a-11e7-8235-f578adb189e3.png)
> 
> Yes I know this ends up being a little bit like XKCD, except for APIs instead of standards, but we don't have much choice. The library is 30 years old, and we don't get to break people for our own satisfaction -- no matter how right, correct, fast, or convenient our new reality might be.

kbevers:
> Well, I'm the self-appointed maintainer, and I'm happy to hand off the hat to people actually doing all the work 😄 

If you want to step down I am sure we can find someone to take over the reins (if no one else volunteers I'll consider it). If that is the case, Ideally that would be after 5.0.0 is out the door. 

> I've been trying to stay back and let you guys evolve things without worrying too much about the project maintenance implications.

And that's been great so far. As we are getting close to a new release, I think now is the time to worry about those implications.

[...]

PR #736, opened 2018-01-10
Preparing for 5.0.0 release

kbevers:
We are getting close to a new release. There's only a few things left to do, most notably resolving #700 which is well on its way. I'd say we should be able to get the first release candidate out the door some time next week.

I have set the release date to 2018-02-15. It is arbitrarily chosen, but I think the final release will be ready around that time. If the date turns out to be different we need to remember to change those.

This PR takes care of much of the bookkeeping related to the release. I have bumped the remaining version numbers, updated the README and NEWS files and updated the HOWTO-RELEASE file since some of the info in it was outdated or missing. Obviously I haven't created new tarballs and other stuff relelated to distribution of the package.

I have not changed the version number in the Sphinx docs since that is already taken care of in #577 which should be merged just before releasing (it updates the website, so merging needs to be coordinated with the release).

I think most of the changes are straight forward but more eyes on this is needed. **One thing in particular that I am unsure about is the SO-version number. I have changed it from "12.0.0" to "12.1.0". This can be debated since I believe there's a few minor changes in `projects.h` that can be interpreted as breaking backwards compatibility (this basically depends on what we define as the API in previous versions). I need your inpout on this.**

Otherwise most of the work has gone into writing up the release notes. I have included everything I could find in PR's and issues that affects users of the package. I might have missed a few minor things, let me know if you stumple on something.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,9 +49,9 @@ colormsg(_HIBLUE_ "Configuring PROJ:")
 #PROJ version information
 #################################################################################
 include(Proj4Version)
-proj_version(MAJOR 4 MINOR 9 PATCH 3)
-set(PROJ_API_VERSION "12")
-set(PROJ_BUILD_VERSION "12.0.0")
+proj_version(MAJOR 5 MINOR 0 PATCH 0)
+set(PROJ_API_VERSION "13")
+set(PROJ_BUILD_VERSION "13.0.1")
 
 #################################################################################
 # Build features and variants
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,298 @@
+5.0.0 Release Notes
+-------------------
+
+This version of PROJ introduces some significant extensions and
+improvements to (primarily) the geodetic functionality of the system.
+
+The main driver for introducing the new features is the emergence of
+dynamic reference frames, the increasing use of high accuracy GNNS,
+and the related growing demand for accurate coordinate
+transformations.  While older versions of PROJ included some geodetic
+functionality, the new framework lays the foundation for turning PROJ
+into a generic geospatial coordinate transformation engine.
+
+The core of the library is still the well established projection code.
+The new functionality is primarily exposed in a new programming
+interface and a new command line utility, "cct" (for "Coordinate
+Conversion and Transformation").  The old programming interface is
+still available and can - to some extent - use the new geodetic
+transformation features.
+
+The internal architecture has also seen many changes and much
+improvement.  So far, these improvements respect the existing
+programming interface. But the process has revealed a need to simplify
+and reduce the code base, in order to support sustained active
+development.
+
+!!!
+!!! Therefore we have scheduled regular releases over the coming years
+!!! which will gradually remove the old programming interface.
+!!!
+!!! This will cause breaking changes with the next two major version
+!!! releases, which will affect all projects that depend on PROJ
+!!! (cf. section "deprecations" below).
+!!!
+
+The decision to break the existing API has not been easy, but has
+ultimately been deemed necessary to ensure the long term survival of
+the project. Not only by improving the maintainability immensely, but
+also by extending the potential user (and hence developer) community.
+
+The end goal is to deliver a generic coordinate transformation
+software package with a clean and concise code base appealing to
+both users and developers.
+
+
+VERSIONING AND NAMING
+---------------------
+
+For the first time in more than 25 years the major version number of
+the software is changed. The decision to do this is based on the many
+new features and new API. While backwards compatibility remains -
+except in a few rare corner cases - the addition of a new and improved
+programming interface warrants a new major release.
+
+The new major version number unfortunately leaves the project in a bit
+of a conundrum regarding the name. For the majority of the life-time
+of the product it has been known as PROJ.4, but since we have now
+reached version 5 the name is no longer aligned with the version
+number.
+
+Hence we have decided to decouple the name from the version number and
+from this version and onwards the product will simply be called PROJ.
+
+In recognition of the history of the software we are keeping PROJ.4 as
+the *name of the organizing project*. The same project team also
+produces the datum-grid package.
+
+In summary:
+
+o The PROJ.4 project provides the product PROJ, which is now at
+  version 5.0.0.
+
+o The foundational component of PROJ is the library libproj.
+
+o Other PROJ components include the application proj, which provides
+  a command line interface to libproj.
+
+o The PROJ.4 project also distributes the datum-grid package,
+  which at the time of writing is at version 1.6.0.
+
+
+ UPDATES
+ -------
+
+ o Introduced new API in proj.h.
+   - The new API is orthogonal to the existing proj_api.h API and the
+     internally used projects.h API.
+   - The new API adds the ability to transform spatiotemporal (4D)
+     coordinates.
+   - Functions in the new API use the "proj_" namespace.
+   - Data types in the new API use the "PJ_" namespace, with a few
+     historic exceptions such as XY, XYZ, LP and LPZ.
+
+ o Introduced the concept of "transformation pipelines" that makes it
+   possible to do complex geodetic transformations of spatiotemporal
+   coordinates by daisy chaining simple coordinate operations.
+
+ o Introduced cct, the Coordinate Conversion and Transformation
+   application.
+
+ o Introduced gie, the Geospatial Integrity Investigation Environment.
+   - Selftest invoked by -C flag in proj has been removed
+   - Ported approx. 1300 built-in selftests to gie format
+   - Ported approx. 1000 tests from the gigs test framework
+   - Added approx. 200 new tests
+
+ o Adopted terminology from the OGC/ISO-19100 geospatial standards
+   series. Key definitions are:
+   - At the most generic level, a *coordinate operation* is a change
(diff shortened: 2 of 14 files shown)


NEWS
```diff
@@ -1,3 +1,210 @@
...
+
+ o Adopted terminology from the OGC/ISO-19100 geospatial standards
+   series. Key definitions are:
+   - A coordinate operation is a change of coordinates, based on a
+     one-to-one relationship, from one coordinate reference system
+     to another
+   - A transformation is a coordinate operation in which the two
+     coordinate reference systems are based on different datums, e.g.
+     a change from a global reference frame to a regional frame.
+   - A coordinate operation is an operation in which both coordinate
```
rouault: a bit confusing: coordinate operation is defined in first and third items

NEWS
```diff
@@ -1,3 +1,210 @@
...
+
+ o Added metadata to init-files that can be read with the
+   proj_init_info() function in the new proj.h API.
+
+ o Added ITRF2000, ITRF2008 and ITRF2014 init-files with ITRF
+   transformation parameters, including plate motion model
+   parameters.
+
+ o Added parameters for GSK2011, PZ90 and danish (similar to existing
+   andrae which has an incorrect semimajor axis) ellipsoids.
```
rouault: andrae ?

NEWS
```diff
@@ -1,3 +1,210 @@
...
+
+ o Support for analytical partial derivatives has been removed.
+
+ o Improved performance in Winkel Tripel and Aitoff. #406.
+
+
+ BUG FIXES
+ ------------
+
+ o Fixed incorrect convergence calculation in Lamber Conformal
```
rouault: Lambert

NEWS
```diff
@@ -1,3 +1,210 @@
...
+ o Return correct result at origin in inverse Lee Oblated
+   Stereographic. #164.
+
+ o Return correct result at origin in inverse Miller Oblated
+   Stereographic. #164.
+
+ o Fixed scaling and wrap-around issues in Oblique Cylindrical
+   Equal Area. #166.
+
+ o Corrected a coefficient error in inverse Transform Mercator. #174.
```
rouault: Transform->Transverse

src/Makefile.am
```diff
@@ -40,7 +40,7 @@ geodtest_LDADD = libproj.la
 
 lib_LTLIBRARIES = libproj.la
 
-libproj_la_LDFLAGS = -no-undefined -version-info 12:0:0
+libproj_la_LDFLAGS = -no-undefined -version-info 12:1:0
```
rouault: This should normally be 13:0:1 I think. If I follow correctly the rule in GDAL's HOWTO-RELEASE, in the "adding interfaces" case
```
4) Update LIBGDAL_CURRENT/REVISION/AGE macros in GDALmake.opt.in.
   - For a release with no interface changes just bump REVISION.
   - Adding interfaces, bump CURRENT/AGE, set REVISION to 0.
   - Deleting interfaces / compatibility issues - bump CURRENT, others to zero.
```

CMakeLists.txt
```diff
@@ -49,9 +49,9 @@ colormsg(_HIBLUE_ "Configuring PROJ:")
 #PROJ version information
 #################################################################################
 include(Proj4Version)
-proj_version(MAJOR 4 MINOR 9 PATCH 3)
+proj_version(MAJOR 5 MINOR 0 PATCH 0)
 set(PROJ_API_VERSION "12")
-set(PROJ_BUILD_VERSION "12.0.0")
+set(PROJ_BUILD_VERSION "12.1.0")
```
rouault: I'd say 13.0.1 (see other related comment)

NEWS
kbevers: Oh, I meant to write "conversion", ont "coordinate operation". Good catch.

NEWS
kbevers: Yes, that's a thing:

```
 proj -le | grep andrae
   andrae a=6377104.43     rf=300.0         Andrae 1876 (Den., Iclnd.)
```

Named after Danish geodesist Andræ, but "æ" doesn't play well non-unicode character sets. I'll change it to "andrae ellipsoid" which hopefully is less confusing.

src/Makefile.am
kbevers: Yes, this is a bit easier to understand. After reading the rules in the PROJ HOWTO-RELEASE i interpret it the same way. 13.0.1 it is.

kbevers:
Updated according to @rouault's comments. ABI version is now set to `13.0.1`.

kbevers:
I few other things I am thinking about adding to the NEWS file:

* Something about naming conventions re: #735. If some sort of consensus is reached at least.
* Some sort of description of the release schedule as described in #686. I think it is only fair to be up front about the expected time schedule for the announced removal of deprecated modules.
* Should we also announce the removal of the COM-modules? I've enquried the mailing list about it's status but so far haven't gotten any feedback. If no one replies I take it as it is unused. I'd remove it with the 5.1.0 release similar to the WinCE module.

busstoptaktik:
A few relevant mailing list discussions:

[Current master...](http://osgeo-org.1560.x6.nabble.com/Current-master-still-at-released-4-9-3-td5336300.html)

[Coming releases...](http://osgeo-org.1560.x6.nabble.com/Coming-releases-of-PROJ-4-td5325216.html)

Also, someone somewhere had a request for some specific numbering scheme for release candidates It might have been @rsbivand or @sebastic but I cannot find the actual comment.

In general, I think we need input on this subject, from people packaging PROJ: We developers are  probably unaware of a number of ways we can mess up things for the people doing packaging, testing and distribution.

Also note [this message](http://lists.maptools.org/pipermail/proj/2017-June/007710.html) from @rsbivand, mentioning the rproj use of a "detect invertibility" function, that will no longer work as of 5.0.0. May be worth mentioning somewhere - or perhaps it should just be filed as an issue on rproj?

kbevers:
> Also note this message from @rsbivand, mentioning the rproj use of a "detect invertibility" function, that will no longer work as of 5.0.0. May be worth mentioning somewhere - or perhaps it should just be filed as an issue on rproj?

Are you talking about this test:

```
#include <projects.h>

int inversetest(PJ *P){
  return (P->inv ? 1: 0);

}
```

If so, it is still possible in 5.0.0. From 6.0.0 `projects.h` will be removed from the public API though.

busstoptaktik:
No, I'm talking about the PJ->inverted being set

Den 25. jan. 2018 4.02 PM skrev "Kristian Evers" <notifications@github.com>:

kbevers:
Ah okay. How likely is that to happen if you are using the old API(s)? I suspect not very

busstoptaktik:
The old APIs can still be initialized through EPSG style definition files, where the actual definition text is opaque to the user, so I still think it worth to warn, and provide an API workaround.

Actually, it seems, we also have the problem in pj_transform.c around line 158.

kbevers:
Sure, we can mention it in the release notes, no problem. We could also make a pj_has_inverse function and expose it in proj_api.h. Should be quite simple to do.

rsbivand:
This is not the only problem in rgdal - I am not the maintainer of the proj4 package which has not been revised since 2012. Once we've checked, I'll post separate issues on breaking changes. @edzer please note sf consequences.

edzer:
Thanks, @rsbivand - I stumbled in a much earlier phase, see #745

kbevers:
> I did note that "make test" fails on testIGNF. I didn't get a chance to figure out why. If something needs to be done before this test can be successfully run, perhaps this could be checked when cmake is configured?

They seem to work on CI so I am not too worried about that. 

> Finally, the presence of -std=c89 with gnu compiles causes the C99_MATH test to fail. I'm OK with this. But why not support c99 if it's available? The C99_MATH usage is supposed only to affect how proj is compiled. Client packages shouldn't need to know about this.

I think it is unrelated to this PR, but please make the changes you feel are necessary. I have no strong opinions about this subject.

cffk:
> > I did note that "make test" fails on testIGNF. I didn't get a chance to figure out why. If something needs to be done before this test can be successfully run, perhaps this could be checked when cmake is configured?

> They seem to work on CI so I am not too worried about that.

Surely, the tests should work for someone trying to build the code?  I'll try to figure out why this one isn't working.

In fact, there's another glaring problem with most of the tests -- all, save geodesic-test, require write access to the source directory.  This is a bad practice.  I'll see about fixing this too.

> > Finally, the presence of -std=c89 with gnu compiles causes the C99_MATH test to fail. I'm OK with this. But why not support c99 if it's available? The C99_MATH usage is supposed only to affect how proj is compiled. Client packages shouldn't need to know about this.

> I think it is unrelated to this PR, but please make the changes you feel are necessary. I have no strong opinions about this subject.

I'm happy to drop this (too close to the release date to be mucking with compiler flags).

kbevers:
> Surely, the tests should work for someone trying to build the code? I'll try to figure out why this one isn't working.

Sorry I should have elaborated on that. What I meant was that usually when the tests fail locally it is because you are missing grids or some other configuration. If that isn't case then indeed it should be fixed.

> In fact, there's another glaring problem with most of the tests -- all, save geodesic-test, require write access to the source directory. This is a bad practice. I'll see about fixing this too.

Good point. I guess it should be possible to write to a temporary location instead. Just for the record, the gie tests doesn't need write access either.

PR #743, opened 2018-01-22
Handle ellipsoid parameters correctly when using +nadgrids=@null. 

kbevers:
Attempt at fixing #22.

Some of the IGNF tests are failing slightly. I can't tell if the test data is correct or not. It is likely that the test data was constructed simply as regression tests and not based on different source.


kbevers:
The affected IGNF tests are for transformations between the following systems:

LAMBE -> LAMB93
LAMBE -> GEOPORTALFX
RGR92 -> REUN47

As far as I can tell they are, at best, defined at a meter level accuracy. The tests that fail are roughly 1 mm off on the y-coordinate. I dont think it is unreasonable to update the regression tests to reflect this slight change caused by the bug fix.

busstoptaktik (commented):
What does the double strcmp test do? I think a comment here would be useful.

kbevers:
Just checking for both scenarios where the null grid might be applied by it self: `+nadgrids=@null` and `+nadgrids=null`. If either one of them exists the strcmp-expressions will evalutate to false. Is the current comment not adequate?

kbevers:
I have now updated the comments to be more specific about what the double strcmp test does. I'll merge once the tests pass.

Issue #759, opened 2018-02-02
5.0.0-rc1 fails to build (./src/gen_cheb.c:59: undefined reference to `pow')


sebastic:
proj-5.0.0-rc1 fails to build. Looks like missing `-lm` somewhere:
```
ibtool: link: ranlib .libs/libproj.a
libtool: link: ( cd ".libs" && rm -f "libproj.la" && ln -s "../libproj.la" "libproj.la" )
/bin/bash ../libtool  --tag=CC   --mode=link gcc -Wall -Wdeclaration-after-statement -g -O2 -fdebug-prefix-map=/build/proj-5.0.0~rc1=. -fstack-protector-strong -Wformat -Werror=format-security -I/usr/lib/jvm/default-java/include/linux -DHAVE_C99_MATH=1  -Wl,-z,relro -Wl,-z,now -o proj proj.o gen_cheb.o p_series.o libproj.la -lpthread 
/bin/bash ../libtool  --tag=CC   --mode=link gcc -Wall -Wdeclaration-after-statement -g -O2 -fdebug-prefix-map=/build/proj-5.0.0~rc1=. -fstack-protector-strong -Wformat -Werror=format-security -I/usr/lib/jvm/default-java/include/linux -DHAVE_C99_MATH=1  -Wl,-z,relro -Wl,-z,now -o geod geod.o geod_set.o geod_interface.o libproj.la -lpthread 
/bin/bash ../libtool  --tag=CC   --mode=link gcc -Wall -Wdeclaration-after-statement -g -O2 -fdebug-prefix-map=/build/proj-5.0.0~rc1=. -fstack-protector-strong -Wformat -Werror=format-security -I/usr/lib/jvm/default-java/include/linux -DHAVE_C99_MATH=1  -Wl,-z,relro -Wl,-z,now -o nad2bin nad2bin.o libproj.la -lpthread 
libtool: link: gcc -Wall -Wdeclaration-after-statement -g -O2 "-fdebug-prefix-map=/build/proj-5.0.0~rc1=." -fstack-protector-strong -Wformat -Werror=format-security -I/usr/lib/jvm/default-java/include/linux -DHAVE_C99_MATH=1 -Wl,-z -Wl,relro -Wl,-z -Wl,now -o .libs/proj proj.o gen_cheb.o p_series.o  ./.libs/libproj.so -lpthread
[...]

PR #762, opened 2018-02-02
Add AC_SEARCH_LIBS for libm to configure.ac.

sebastic:
Some systems need `-lm` to use sqrt and related functions.

Fixes: #759


kbevers:
Thanks for fixing this!

PR #771, opened 2018-02-04
Add --version option to gie and cct.

### LINKED ISSUE #764: Add --version to gie and cct
It should be possible to check what version you are using (also, it is needed for `html2man` re: #760)

busstoptaktik:
Additionally correct a spelling error in optargpm.h and
remove two lines of leftover cruft from gie.c

Resolves #764

### DIFF
--- a/src/cct.c
+++ b/src/cct.c
@@ -157,7 +157,7 @@ int main(int argc, char **argv) {
     int nfields = 4, direction = 1, verbose;
     double fixed_z = HUGE_VAL, fixed_time = HUGE_VAL;
     int columns_xyzt[] = {1, 2, 3, 4};
-    const char *longflags[]  = {"v=verbose", "h=help", "I=inverse", 0};
+    const char *longflags[]  = {"v=verbose", "h=help", "I=inverse", "version", 0};
     const char *longkeys[]   = {"o=output",  "c=columns", "z=height", "t=time", 0};
 
     o = opt_parse (argc, argv, "hvI", "cozt", longflags, longkeys);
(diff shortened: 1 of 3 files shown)


src/cct.c
```diff
@@ -157,7 +157,7 @@ int main(int argc, char **argv) {
     int nfields = 4, direction = 1, verbose;
     double fixed_z = HUGE_VAL, fixed_time = HUGE_VAL;
     int columns_xyzt[] = {1, 2, 3, 4};
-    const char *longflags[]  = {"v=verbose", "h=help", "I=inverse", 0};
+    const char *longflags[]  = {"v=verbose", "h=help", "I=inverse", "version", 0};
```
kbevers: It would be nice to use `proj_info().release` or `proj_info().version` here.

src/cct.c
```diff
@@ -173,6 +173,11 @@ int main(int argc, char **argv) {
     direction = opt_given (o, "I")? -1: 1;
     verbose   = opt_given (o, "v");
 
+    if (opt_given (o, "version")) {
+        fprintf (stdout, "%s: %s\n", o->progname, pj_get_release ());
```
kbevers, line 177: It would be nice to use `proj_info().release` or `proj_info().version` here.

src/gie.c
```diff
@@ -250,6 +247,13 @@ int main (int argc, char **argv) {
         return 0;
     }
 
+
+    if (opt_given (o, "version")) {
+        fprintf (stdout, "%s: %s\n", o->progname, pj_get_release ());
```
kbevers, line 252: It would be nice to use `proj_info().release` or `proj_info().version` here.

src/cct.c
busstoptaktik: Yes, I agree, but cf. PR #775 to see why I didn't do it that way. I suggest, for now, to
1. merge this PR (which is tidy in the sense of implementing --version in an identical way to cs2cs and proj),
2. Then wait for #775 to get merged,
3. And finally update cct & gie in a new (small) PR.

src/gie.c
busstoptaktik: (ibid.)

busstoptaktik:
@kbevers please just merge this if you agree with the procedure outlined in the comment inside your review. #775 simplifies the INFO code a bit and dramatically reduces the size of the very large structs being passed/allocated on the stack, as suggested by @schwehr recently

PR #772, opened 2018-02-04
Make a README from README.md when running 'make dist'

kbevers:
Fixing insufficient makefile update in 9e412249e5a7c6eaab1e73a11ca09e2bc7ca5df9

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ CTestTestfile.cmake
 cmake_install.cmake
 install_manifest.txt
 cmake/project-config*.cmake
+README
 
 /.vs*
 /*.manifest
--- a/Makefile.am
+++ b/Makefile.am
@@ -10,7 +10,7 @@ ACLOCAL_AMFLAGS =	-I m4
 
 
 README: README.md
-	cp $< $@
+	fgrep -v "[![" $< > $@
 
 clean-local:
 	$(RM) README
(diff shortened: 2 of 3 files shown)


mloskot:
Looks like the markdown conversion PR made you busy :(

`make[1]: *** No rule to make target `README.md', needed by `README'.  Stop.` is puzzling, the former does exist, doesn't it.

kbevers:
Yes, I am quite aware of my limited knowledge of autotools right now. But I am learning a lot at least. I believe the problem now is files referenced in the Makefile is not correct when making out-of-source builds.

mloskot:
I have no experienced with out of source builds using Autotools.

Might be easier to keep it in Markdown but rename `README.md` back to `README` until Autotools are dropped.

kbevers:
Autotools is not likely to get dropped just yet though. I believe there's several "easy fixes" here that's close to perfect, for instance one option would be to add `README.md` to `EXTRA_DIST` and accept that two READMEs will be included in the source distribution. But I am not ready for that compromise just yet :-)

kbevers:
The biggest problem for me is that all the various things I am trying works locally. Can't get to the bottom of the actual problem.

kbevers:
The problem was in the travis setup. Running `make dist-all` from the source-distribution created from `make dist-all` is not going to work from here on. I would say that is reasonable.

mloskot:
@kbevers Thank you for cleaning up the mess I've started :-)

PR #775, opened 2018-02-05
Shrink PJ_XXX_INFO structs, but keep same syntax.

busstoptaktik:
A number of the fixed length strings in the INFO structs are simply
reflections of material that already exists as static or preallocated strings at a
number of places in the library (or in the case of PJ_INFO, really
*should* exist, and is now implemented).

This PR replaces these cases of constant length strings with const
char pointers. The usage syntax is unchanged, and so is the nice
property of having the return value allocated on the stack, and
hence not requiring explicit memory management by the caller.

proj_info now only does setup once - and the searchpath entry of
PJ_INFO is not arbitrarily truncated at 512 bytes. Repeated calls
simply returns a copy of already prepared material.

The id, description and definition entries of PJ_PROJ_INFO are now
also guaranteed to hold the entire text of the corresponding static or preallocated
string, by being represented by a const char pointer to that actual
string.

PJ_GRID_INFO and PJ_INIT_INFO (i.e. the two smallest INFO structs)
are unchanged.

As part of the work, the remaining calls to pj_strlcpy were also eliminated, so pj_strlcpy is removed.

### DIFF
--- a/src/proj_4D_api.c
+++ b/src/proj_4D_api.c
@@ -758,7 +758,7 @@ static char *path_append (char *buf, const char *app, size_t *buf_size) {
             pj_dealloc (buf);
             return 0;
         }
-        *buf_size = 2 * len;
+        *buf_size = 2 * len + 2;
         if (buf != 0)
             strcpy (p, buf);
         pj_dealloc (buf);
--- a/src/proj_4D_api.c
+++ b/src/proj_4D_api.c
@@ -753,7 +753,7 @@ static char *path_append (char *buf, const char *app, size_t *buf_size) {
 
     /* "pj_realloc", so to speak */
     if (*buf_size < len) {
-        p = pj_calloc (2 * len, sizeof (char));
+        p = pj_calloc (2 * len + 2, sizeof (char));
         if (0==p) {
             pj_dealloc (buf);
             return 0;
@@ -806,6 +806,7 @@ PJ_INFO proj_info (void) {
         info.searchpath = empty;
         info.version    = version;
         info.release    = pj_get_release ();
+printf ("HOME=[%s],  PROJ_LIB=[%s]\n", getenv("HOME"), getenv("PROJ_LIB"));
 
         /* build search path string */
         buf = path_append (buf, getenv ("HOME"), &buf_size);
@@ -821,11 +822,13 @@ PJ_INFO proj_info (void) {
 
         for (i = 0;  i < n;  i++) {
             buf = path_append (buf, paths[i], &buf_size);
+printf ("loop: buf_size=%3.3d,  info.searchpath=[%s]\n", (int) buf_size, info.searchpath);
             if (0==buf)
                 break;
         }
+        if (0==buf)
+            break;
         info.searchpath = buf;
-printf ("HOME=[%s],  PROJ_LIB=[%s]\n", getenv("HOME"), getenv("PROJ_LIB"));
 printf ("buf_size=%3.3d,  info.searchpath=[%s]\n", (int) buf_size, info.searchpath);
 
         info.paths = paths;
--- a/src/proj_4D_api.c
+++ b/src/proj_4D_api.c
@@ -792,40 +792,45 @@ PJ_INFO proj_info (void) {
     char   *buf = 0;
 
     pj_acquire_lock ();
-    while (0==info_initialized) {
-        info_initialized = 1;
-        info.major = PROJ_VERSION_MAJOR;
-        info.minor = PROJ_VERSION_MINOR;
-        info.patch = PROJ_VERSION_PATCH;
 
-        /* This is a controlled environment, so no risk of sprintf buffer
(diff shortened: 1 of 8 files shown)


src/proj_4D_api.c
```diff
@@ -724,22 +724,73 @@ PJ_CONTEXT *proj_context_destroy (PJ_CONTEXT *ctx) {
...
+    if (*buf_size < len) {
+        p = pj_calloc (2 * len, sizeof (char));
+        if (0==p) {
+            pj_dealloc (buf);
+            return 0;
+        }
+        *buf_size = 2 * len;
+        if (buf != 0)
+            strcpy (p, buf);
+        buf = p;
```
kbevers: I think this is your memory leak. You don't deallocate buf before `buf = p;` which causes the original content of buf to leak.

src/proj_4D_api.c
```diff
@@ -724,22 +724,74 @@ PJ_CONTEXT *proj_context_destroy (PJ_CONTEXT *ctx) {
...
+    applen = strlen (app);
+    if (0 == applen)
+        return buf;
+
+    /* Start checking whether buf is long enough */
+    if (0 != buf)
+        buflen = strlen (buf);
+    len = buflen+applen+2;
+
+    /* "pj_realloc", so to speak */
```
rouault: indeed, why not using realloc() ?

src/proj_4D_api.c
```diff
@@ -724,22 +724,74 @@ PJ_CONTEXT *proj_context_destroy (PJ_CONTEXT *ctx) {
...
     const char * const *paths;
-    char *tmpstr;
     int i, n;
-    size_t len = 0;
 
-    memset(&info, 0, sizeof(PJ_INFO));
+    size_t  buf_size = 0;
+    char   *buf = 0;
+
+    if (0 != info.version)
```
rouault: Be careful: this is not thread-safe. Ideally this should be protected by a mutex

src/proj_4D_api.c
busstoptaktik: For two reasons: First (and least), to get the null-padding from calloc.

Second: Because (for reasons beyond my comprehension) we use wrappers for memory allocation.

I would really like to get rid of pj_malloc, pj_calloc, pj_dalloc, and just use malloc, calloc and free. I would equally like to see the fileapi go, and just use direct stdio calls where needed. But I cannot convince myself that this will not break something for someone. In the case of removing fileapi, stuff will probably break for Google (@schwehr?), since I think Frank @warmerdam worked for Google when it was introduced.

For pj_malloc and friends: Yes - getting rid would be fantastic. Could be a thing to consider for v. 5.1.0 or 6.0.0?

src/proj_4D_api.c
busstoptaktik: Wow - that was really a blunder. Thanks for spotting it!

src/proj_4D_api.c
```diff
@@ -793,6 +793,8 @@ PJ_INFO proj_info (void) {
     if (0 != info.version)
         return info;
 
+    pj_acquire_lock();
```
rouault: the pj_acquire_lock() should be placed before the if(0 != info.version).
And there are missing pj_release_lock() in the early return code paths

src/proj_4D_api.c
schwehr: I think going with the standard malloc, calloc and free calls should be fine.  The one thing to watch out is that malicious input can't allocate huge amounts of memory.aq

The one thing that I need is a way to have the external data files not be required for a binary.  If there is some alternative way to do this, that would be great. 

My constraints, which are different from the normal linux install: containers have no local storage.  Each container runs one binary.  That binary can be "large."  There is a bit of local temp ramdisk, but it is so small that data files can't go there.

The same issue happens with GDAL and I solved that with a library that sadly isn't part of the public ABSL library release.  But then I stash them in /vsimem/ and set the GDAL_DATA env var.   In a perfect world, proj, libgeotiff, and gdal would all have the same way to stash support data in the binary that was an optional configure flag.

FYI, The other thing we do is hook in the google logger so that proj messages end up in the standard logs.

kbevers: > The same issue happens with GDAL and I solved that with a library that sadly isn't part of the public ABSL library release. But then I stash them in /vsimem/ and set the GDAL_DATA env var. In a perfect world, proj, libgeotiff, and gdal would all have the same way to stash support data in the binary that was an optional configure flag.

What sort of data are we talking about here? Grids and init-files?

> FYI, The other thing we do is hook in the google logger so that proj messages end up in the standard logs.

Are your needs sufficiently covered in the new API?

schwehr: The embedded data was indeed setup by Warmerdam.

This is indeed the init (epsg, proj_def.dat) and grid shift files.  BTW, looking in proj_def.dat, I was amused to see "SCCSID"

Logging is all set as long as there is pj_ctx_set_logger().

PR #781, opened 2018-02-09
Avoid XY, LP and UV datatype clashes with other libraries.

kbevers:
Reaction to @sebastic's report of on the mailing list: http://lists.maptools.org/pipermail/proj/2018-February/007976.html

Remove unnecessary definitions of UV and UVW from project.h that collides with
external libraries. To prevent similar problems in the future the
datatypes XY, LP, UV, XYZ, LPZ and UVW has been prefixed by PJ_ in
proj.h and proj_internal.h

### DIFF
--- a/src/proj.h
+++ b/src/proj.h
@@ -212,13 +212,13 @@ typedef struct { double o, p, k; }          PJ_OPK;  /* Rotations: omega, phi, k
 typedef struct { double e, n, u; }          PJ_ENU;  /* East, North, Up */
 
 /* Classic proj.4 pair/triplet types */
-typedef struct { double   u,   v; }  UV;
-typedef struct { double   x,   y; }  XY;
-typedef struct { double lam, phi; }  LP;
+typedef struct { double   u,   v; }  PJ_UV;
+typedef struct { double   x,   y; }  PJ_XY;
+typedef struct { double lam, phi; }  PJ_LP;
 
-typedef struct { double   x,   y,  z; }  XYZ;
-typedef struct { double   u,   v,  w; }  UVW;
-typedef struct { double lam, phi,  z; }  LPZ;
+typedef struct { double   x,   y,  z; }  PJ_XYZ;
+typedef struct { double   u,   v,  w; }  PJ_UVW;
+typedef struct { double lam, phi,  z; }  PJ_LPZ;
 
 
 /* Avoid preprocessor renaming and implicit type-punning: Use a union to make it explicit */
@@ -229,12 +229,12 @@ union PJ_COORD {
     PJ_LPZT lpzt;
      PJ_OPK opk;
      PJ_ENU enu;
-        XYZ xyz;
-        UVW uvw;
-        LPZ lpz;
-         XY xy;
-         UV uv;
-         LP lp;
+     PJ_XYZ xyz;
+     PJ_UVW uvw;
+     PJ_LPZ lpz;
+      PJ_XY xy;
+      PJ_UV uv;
+      PJ_LP lp;
 };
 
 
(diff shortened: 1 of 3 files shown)


busstoptaktik (approved):
I have no substantial comments, really: It seems to work, it moves some additional data types into the PJ_ namespace. I'm surprised it could be done so compactly, but it evidently is due to the types almost never are referrred directly to in the new API - rather they are "called by nickname" through the union accessor syntax for the PJ_COORD union data type

c0nk:
I am getting a compile error on master now. I think it is related to me including <proj.h> and <proj_api.h> at the same time. I am getting this error:
```
proj_api.h:122:1: error: ‘XY’ does not name a type
 projXY pj_fwd(projLP, projPJ);
 ^
proj_api.h:123:1: error: ‘LP’ does not name a type
 projLP pj_inv(projXY, projPJ);
 ^
proj_api.h:125:1: error: ‘XYZ’ does not name a type
 projXYZ pj_fwd3d(projLPZ, projPJ);
 ^
proj_api.h:126:1: error: ‘LPZ’ does not name a type
 projLPZ pj_inv3d(projXYZ, projPJ);
 ^
```
I am using the new API. The reason I am including <proj_api.h> is to be able to use `pj_ctx_get_errno()` to get the error when `proj_create()` fails. For example:
```
pj = proj_create(ctx, definition));
  if (!pj) {
    errno = pj_ctx_get_errno(ctx);
  }
```
Maybe I shouldn't be using `pj_ctx_get_errno` for this? Either way including <proj_api.h> shouldn't be a compile error.

kbevers:
Use `proj_errno()` to read the error number.

It is not a good idea to use `proj.h` and `proj_api.h` together. This should probably be prevented for library users (internally we need to be able to do it). I'll look into that.

busstoptaktik:
@c0nk - This is truly  a bug. The workaround seems to be to include proj.h, projects.h, and proj_api.h, in that sequence.

I mentioned somewhere that undoing the projXXX defines was a trip through a hell of tripwires, and that I advised great care to be taken when touching that stuff.

Apparently @kbevers managed to do it anyway - but then again, according to your experience, not entirely successful.

We will have to adjust the API to handle this. Basically the problem you experience is the result of me trying to talk everyone into getting rid of contexts: Their single purpose is to provide a thread local errno - *but errno is thread local already:* errno is a macro evaluating to a dereferencing of a pointer to something (a function) returning something that is thread local.

Since I did not see any reason at all for PJ_CONTEXT, I was probably not careful enough to design an API that was entirely bulletproof.

But if you are running in a single threaded environment, you can work around this by 1) setting your context to 0 and 2) using proj_errno(0) to access the context errno in case of PJ=0.

busstoptaktik:
> Use proj_errno() to read the error number.

@kbevers this will not work if the problem is that proj_create returns 0.

We should probably change the signature from `proj_errno(PJ *)` to `proj_errno(PJ_CONTEXT *)`

Apparently, we also need a few test cases checking different header include strategies. The current problem is that we have no cases left in the command line tool department which solely includes `proj_api.h`

kbevers:
I see the problem now. It is probably the best solution to go with the context as input instead of a `PJ`.

> Apparently @kbevers managed to do it anyway - but then again, according to your experience, not entirely successful.

No, apparently not. I think the solution is to disallow combining `proj.h` with  `proj_api.h` and/or `projects.h`. They are supposed to be orthogonal afterall. I'll try a few things tomorrow.

busstoptaktik:
> I see the problem now. It is probably the best solution to go with the context as input instead of a PJ.

@kbevers, @c0nk: Actually, after having given it some extra thought, i think we should (despite the risk of API sprawl) probably have both versions, i.e. a `proj_errno`and a `proj_context_errno`.

The thing is that if I introduce a new context, `ctx` in (say) the `main` function, then **un**succesfully attempt to do `P = proj_create (ctx, "foo bar baz")`, I sit in a context where the value of `ctx` is readily available, and I can call `proj_context_errno`.

If the call to `proj_create` is successfull, I go on and pass `P` to the function `do_something_with (P)`. If something goes wrong down here, I do not have access to `ctx`, but `P!=0` and I can call `proj_errno`.

But really - just using `errno` had been much easier, and I really wish I had been more determined and unreasonable in my fight against contexts last spring. Contexts are ***EVIL*** :-)

kbevers:
The include-problem is fixed here: https://github.com/OSGeo/proj.4/pull/790

>  Actually, after having given it some extra thought, i think we should (despite the risk of API sprawl) probably have both versions, i.e. a `proj_errno` and a `proj_context_errno`.

I am okay with that.

> But really - just using errno had been much easier, and I really wish I had been more determined and unreasonable in my fight against contexts last spring. Contexts are EVIL :-)
 
You are allowed to fight that battle again for v. 7.0.0 :-)

c0nk:
I can confirm that the compile error related to including <proj_api.h> has been fixed.

PR #786, opened 2018-02-10
Interdependent wrap ups

busstoptaktik:
This PR introduces a number of repairs and improvements which, unfortunately are interdependent, and would break CI if introduced as separate pull requests.

Instead, they are introduced as individual commits here:

* gie: repair handling of angular distances

* update gie tests to agree with improved gie distance handling

* pj_init: set ellipsoid if given, even if not needed

* Fix numerous bugs in the cs2cs emulation


test/gie/DHDN_ETRS89.gie
```diff
@@ -0,0 +1,373 @@
...
+accept       7.466666666667  50.500000000000  # DE_DHDN_Lat-Lon
+expect       7.465834308888  50.498791390585  # ETRS89_Lat-Lon
+accept       7.533333333333  49.333333333333  # DE_DHDN_Lat-Lon
+expect       7.532503616986  49.332250779407  # ETRS89_Lat-Lon
+accept       7.250000000000  49.333333333333  # DE_DHDN_Lat-Lon
+expect       7.249209260581  49.332249456364  # ETRS89_Lat-Lon
+accept       7.533333333333  47.666666666667  # DE_DHDN_Lat-Lon
+expect       7.532530252396  47.665765608135  # ETRS89_Lat-Lon
+-------------------------------------------------------------------------------
+</gie>
```
kbevers: Are you terminating the test on purpose here? Or is it a blunder?

test/gie/4D-API_cs2cs-style.gie
```diff
@@ -12,74 +12,62 @@ but provided through a different interface.
...
+# operation  proj=latlong nadgrids=nzgd2kgrid0005.gsb  ellps=GRS80
 -------------------------------------------------------------------------------
-tolerance 1 nm
-ignore    pjd_err_failed_to_load_grid
-accept    173                 -45                  0 0
-expect    172.999892181021551 -45.001620431954613  0 0
-direction inverse
-accept    172.999892181021551 -45.001620431954613  0 0
-expect    173                 -45                  0 0
+Note: This test appears to use wrong coordinate data, and is commented out.
```
kbevers: Why are the coordinates wrong? I agree that it would be good to have some authoritative test data, but these should still function fine as a regression test.

test/gie/GDA.gie
```diff
@@ -50,6 +50,7 @@ operation proj = pipeline ellps=GRS80;
                z  = -0.04019;   rz = -0.0328979;  s = -0.009994;
           step proj = cart inv;
 -----------------------------------------------------------------------------------
+tolerance  1.1 mm
```
kbevers: That's an oddly specific tolerance. Is there a reason for that?

test/gie/more_builtins.gie
```diff
@@ -208,12 +208,26 @@ expect    failure errno failed_to_load_grid
...
 
 
-
 -------------------------------------------------------------------------------
 Some tests from PJ_hgridshift.c
 -------------------------------------------------------------------------------
-operation  proj=hgridshift +grids=nzgd2kgrid0005.gsb  ellps=GRS80
+# operation  proj=hgridshift +grids=nzgd2kgrid0005.gsb  ellps=GRS80
+-------------------------------------------------------------------------------
+Note: This test appears to use wrong coordinate data, and is commented out.
```
kbevers: Same as above

src/gie.c
```diff
@@ -842,10 +842,17 @@ Tell GIE what to expect, when transforming the ACCEPTed input
     if (T.verbosity > 3)
         printf ("GOT      %.12f  %.12f  %.12f  %.12f\n", co.v[0],co.v[1],co.v[2],co.v[3]);
 
+#if 0
+    /* We need to handle unusual axis orders - that'll be an item for version 5.1 */
```
kbevers: Please add a ticket for that and add it to the 5.1.0 milestone

src/gie.c
```diff
@@ -829,42 +829,35 @@ Tell GIE what to expect, when transforming the ACCEPTed input
...
-        if (e < d)
-            d = e;
-
-        /* or the tolerance may be based on euclidean distance */
-        if (d > T.tolerance)
-            e = proj_xyz_dist (T.b.xyz, T.e.xyz);
-        if (e < d)
-            d = e;
+#if 0
+    /* We need to handle unusual axis orders - that'll be an item for version 5.1 */
```
kbevers, line 846: Please add a ticket for this and add it to the 5.1.0 milestone

kbevers (approved):
Seems fine to me. A few comments on minor details.

test/gie/4D-API_cs2cs-style.gie
busstoptaktik: They are actually not wrong - they are just swapped. Repaired now. Sign conventions are hard to get right, especially when mixing the imperative `grids` and doing the same test elsewhere with the declarative `nadgrids`...

I have added a comment indicating that these data are for regression testing only.

src/gie.c
busstoptaktik: Will do

test/gie/GDA.gie
busstoptaktik: Given the number of decimals given in the original publication you can hardly get closer, except by chance, but 2 mm seems a bit sloppy in a geodetic context. I will, however change to 2 mm for good (or rather: bad :-)) measure.

PR #789, opened 2018-02-10
Re-do pull request #451 export CMake targets with PROJ4:: namespace

cffk:
This pretty much duplicates pull request #451 of @dg0yt.

This version takes to add the include path to the target definition for
cmake 2.8.11 and later.  Also the documentation sticks to the existing
convention of using cmake variables ${PROJ4_LIBRARIES} and
${PROJ4_INCLUDE_DIRS}.  However, the namespace variables are still being
included.

Here's the roll-out plan

(0) Version 4.9.x: The target is proj and PROJ4_LIBRARIES is set to
this.

(1) Version 5.0.x: Two targets, proj and PROJ4::proj, are defined;
PROJ4_LIBRARIES = proj.

(2) In a year or two: Two targets, proj and PROJ4::proj, are defined;
PROJ4_LIBRARIES = PROJ4::proj.

(3) With a change in the library which breaks backwards compatibility:
The target is PROJ4::proj and PROJ4_LIBRARIES = PROJ4::proj.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -21,3 +21,7 @@ install (FILES
 install (EXPORT targets
   FILE ${PROJECT_NAME_LOWER}-targets.cmake
   DESTINATION "${CMAKECONFIGDIR}")
+install (EXPORT targets
+  NAMESPACE ${PROJECT_NAME}::
+  FILE ${PROJECT_NAME_LOWER}-namespace-targets.cmake
+  DESTINATION "${CMAKECONFIGDIR}")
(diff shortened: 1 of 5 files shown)


kbevers:
Awesome! Please create issues for your roll-out plan and set the milestones to the appropriate release. Just so we don't forget to do it :-). I am thinking 2) == v. 6.0.0 and 3) == 7.0.0. They are scheduled for 2019 and 2020 respectively.

dg0yt:
Good to see this being picked up. I would still recommend to include a test for the legacy CMake usage, and another one for the namespaced target (including transitive usage requirements).

kbevers:
@cffk  Can this be merged? I am planning on pushing RC3 today or tomorrow and I would like to include this so it can be tested in the wild.

PR #791, opened 2018-02-10
Add proj_context_errno to API and missing proj_geod to proj.def

busstoptaktik:
`proj_errno` takes a `PJ *` argument, so it is not enough if the problem is that `proj_create` returns `0`


c0nk:
I think changing the signature of `proj_errno(PJ *)` to `proj_errno(PJ_CONTEXT *)` is a much better solution. Why not add a function `proj_get_context(PJ*)` instead so that if you only have access to `P` you can get a pointer to its owning context which can be used to get the errno?

busstoptaktik:
> Why not ...

Because it leads to somewhat more clumsy code and a large number of corrections all over the code base.

And hopefully (as @kbevers said), I get another chance to start the fight for the elimination of the ***EVIL*** contexts soon, so I'm not much into making them more accessible.

I really think contexts are either an utterly stupid idea (likely), or a brilliant idea I just fail to see the justification for (also likely).

If the latter is the case, I hope some more enlightened persons will help me see the light :-)

c0nk:
As the API stands now, I also don't really see the point of the context. Is there any reason that every `P` couldn't be its own context?

busstoptaktik:
The PJ_CONTEXT makes it possible to assign a separate logger to each thread, and provides room for storage of an opaque object for the logger.

And then there's the homegrown errno and the fileapi. The latter has probably never been used anywhere.

So we could have a proj_create that instead of a context took a logger function as optional argument, then use the already thread local errno, and eliminate an, in my eyes, not very useful subsystem ...

c0nk:
Ahh. I didn't realize that PJ_CONTEXT was also used for logging and for the fileAPI. I just got the impression from your previous discussion that it was mainly for errno. Given this fact, I think I understand the point of contexts and they probably make sense. If you have a thread context, you probably just want to set up the logging / fileAPI once and not have to pass in that information every time you are creating a `PJ`. I was also just imagining the fileAPI could also be used to work around systems that don't have a filesystem and/or want to embed proj data files in the executable itself.

kbevers:
> The PJ_CONTEXT makes it possible to assign a separate logger to each thread, and provides room for storage of an opaque object for the logger.

`proj_log_func()` is hidden in `proj_internal.h` at the moment. We should move that to the public API.

>  I was also just imagining the fileAPI could also be used to work around systems that don't have a filesystem and/or want to embed proj data files in the executable itself.

Apparently that is what they do at Google. Probably also the only users of that feature. It is so far not supported in the new API. I think we should hold back on adding it to the new API until someone specifically requests it.

busstoptaktik:
I appreciate the possibility of linking in stuff for embedded systems without any file system, but I fail to see why the fileapi needs to be thread local.

I assume that one globally accessible pointer would be enough? Since it provides access to material that is compiled into the binary, nothing really is saved by having individual accessors for each thread?

The opaque object on the other hand, I would understand if needed to be PJ local, but I cannot imagine a case where it needed to be thread local

Please enlighten me if anyone can?!

c0nk:
Having a library have global state is generally a bad idea. Consider what happens if application A itself depends on proj directly and, at the same time, another library L that internally uses proj. So we have the following dependency graph:
A -> proj
A -> L -> proj.
Now consider that L (for whatever reason) relies on some custom fileAPI to load special internal +init files. This could conflict with application A use of this feature. The same can be said for logging: Application A may want to include its use of proj in its debug log, but it may be undesirable to see library L's private use of the library in the debug log. Or perhaps library L intercepts and parses its internal proj log to provide better error diagnostics to application A in another format.

PR #793, opened 2018-02-12
A few repairs in and around pj_init.c

busstoptaktik:
* The defaulting-to-WGS84-ellipsoid handling introduced in 82cbab19db5ec8fd7b2fec7f6faf75993381f105 didn't work for `need_ellps=0` in connection with `no_defs` set.

* Some recurring function types (`PJ_CONSTRUCTOR`, `PJ_DESTRUCTOR` and `PJ_OPERATOR`) were typedeffed for clarity.

### DIFF
--- a/src/gie.c
+++ b/src/gie.c
@@ -1836,12 +1836,12 @@ static int pj_cart_selftest (void) {
 
     /* linear in and out */
     P = proj_create(PJ_DEFAULT_CTX,
-        " +proj=helmert +ellps=GRS80"
+        " +proj=helmert"
         " +x=0.0127 +y=0.0065 +z=-0.0209 +s=0.00195"
         " +rx=-0.00039 +ry=0.00080 +rz=-0.00114"
         " +dx=-0.0029 +dy=-0.0002 +dz=-0.0006 +ds=0.00001"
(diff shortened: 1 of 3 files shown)


kbevers:
The debug output from cppcheck is not very useful. I suggest you change

```
grep "error," /tmp/cppcheck.txt
```
to
```
grep -A 25 -B 25 "error," /tmp/cppcheck.txt
```

in `travis/linux_gcc/before_install.sh`. It should give you some more output from `cppcheck.txt`

busstoptaktik:
> I suggest you change...

@kbevers - I agree it would make sense to provide some context when cppcheck makes noise, but that will probably be for another PR.

For now, I have repaired the mess by typedeffing away the pointer semantics of the three function pointer typedefs. We're in a somewhat dark corner of the C standard here: The only thing you can do with a function is to take its address, so functions and pointers to functions are easily mistaken and/or intermixed - and I'm not entirely sure whether my original definitions were correct or not: All compilers accepted them, though, and only cppcheck complained.

After changing the semantics to include the pointer aspect into the typedef, and removing it from the declarations using the typedefs, things still work as expected, and cppcheck is happy.

PR #794, opened 2018-02-12
Avoid invalid ellps error messages from pj_init

busstoptaktik:
Reset error type PJD_ERR_MAJOR_AXIS_NOT_GIVEN for operations
that do not need an ellipsoid.

### DIFF
--- a/src/pj_init.c
+++ b/src/pj_init.c
@@ -600,6 +600,8 @@ pj_init_ctx(projCtx ctx, int argc, char **argv) {
             return pj_default_destructor (PIN, proj_errno(PIN));
         }
         else {
+            if (PJD_ERR_MAJOR_AXIS_NOT_GIVEN==proj_errno (PIN))
+                proj_errno_reset (PIN);
             PIN->f = 1.0/298.257223563;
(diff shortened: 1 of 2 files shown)


busstoptaktik:
@kbevers I suspect that when this is merged, we can actually drop the "set_ellipsoid" code in PJ_pipeline.c - but I do not remember the details of the problem leading to its original inclusion, so I'm not sure we have a test case testing for that.

kbevers:
> I suspect that when this is merged, we can actually drop the "set_ellipsoid" code in PJ_pipeline.c - but I do not remember the details of the problem leading to its original inclusion, so I'm not sure we have a test case testing for that.

The problem was, I think, came up when trying to calculate distances between two points with `proj_lp_dist()`. You need to put a `PJ`-object into the function.  It would fail if the pipeline was not set up with `+ellps=...` in the global parameters. E.g. `+proj=pipeline +proj=merc +ellps=bessel` would fail. The fix to that was to ensure that if no ellipsoid was specified WGS84 would be set as a best guess that at least would allow you to calculate an approximate distance between points.

I can't remember if there's a specific test case.

PR #805, opened 2018-02-14
Fix case of proj-datumgrid filename.

sebastic:
The files use lowercase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ The `makefile.vc` builds `proj.exe`, `proj.dll` and `proj.lib`.
 ### Distribution files and format
 
 Sources are distributed in one or more files.  The principle elements
-of the system are in a compress tar file named `PROJ-x.y.z.tar.gz` where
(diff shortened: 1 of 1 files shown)


kbevers:
Actually I prefer the uppercase "PROJ" after the lengthy discussions about names. But I haven't been consistent in that regard myself with the release candidates so it is probably better to stick with the lowercase "proj". In line 92 PROJ-x.y.z.tar.gz should also be changed to lowercase.

sebastic:
The filenames of the release tarballs should not be changed.

kbevers:
> The filenames of the release tarballs should not be changed.

Out of curriosity, is it common practice to always have source tarballs in lowercase letters only?

sebastic:
> Out of curriosity, is it common practice to always have source tarballs in lowercase letters only?

Yes, uppercase or mixed case release tarballs are quite uncommon (in the Linux world).

kbevers:
Alright, no need to break that tradition. Thanks for pointing it out.

Issue #803, opened 2018-02-14
Remove duplicate files from nad directory which are included in proj-datumgrid


sebastic:
As mentioned in #383 none of the .gsb files included in the `nad/` directory are included in the proj tarballs, these files are included in the proj-datumgrid archives.

We should remove the files from the `nad/` directory that are included in proj-datumgrid, and in the future only accept new binary grids in proj-datumgrid.

kbevers:
I agree that we should do this. Let's wait until the dust has settled on 5.0.0. I am not completely sure of the ramifications. At least we have to improve docs so it clearly explains that you need both packages in order to have a working setup. Some changes to the test setup will probably also be needed.

sebastic:
`travis/install.sh` already unpacks the proj-datumgrid archive into the `nad/` directory, only the proj-datumgrid version will need to be updated when the new release will be the only one to include the .gsb files removed from the proj repository.

The `README.md` already mentions the proj-datumgrid archives (although the case is wrong), the `INSTALL` file does not.

Both files should probably recommend to unpack the latest proj-datumgrid archive into the `nad/` directory before configuring & building PROJ because the grids are used by the tests. This is what distribution packages have done for many years now.

kbevers:
Updates on the website will also be necessary.

sebastic:
The documentation for the `.lla` files in `nad/README` should probably also be moved to the proj-datumgrid repository.

PR #806, opened 2018-02-14
Remove .gsb files included in proj-datumgrid.

sebastic:
Fixes: #803


kbevers:
I didn't want to merge this when it still looked like we had a solid release candidate, but now that's beed disproven we  might as well get this merged as well.

I think the failing Windows tests will be fixed by adding "`ignore pjd_err_failed_to_load_grid`" to the tests in `test/gie/DHDN_ETRS89.gie`. After the `tolerance`-lines.

sebastic:
`ntv1_can.dat` should also be removed as it is included in proj-datumgrid too.

kbevers:
> I think the failing Windows tests will be fixed by adding "ignore pjd_err_failed_to_load_grid" to the tests in test/gie/DHDN_ETRS89.gie. After the tolerance-lines.

Scratch that. I'll fix the tests myself in a follow-up PR.

PR #809, opened 2018-02-19
Minor Mistype GNNS -> GNSS

nicholas-l:
Cant wait for the new release!


kbevers:
Nice catch. I make this mistake an alarmingly high percentage of the times I want to write GNSS.

PR #800, opened 2018-02-13
Four minor corrections, mostly improving testing

busstoptaktik:
* Enhanced gie stats summary

* Add a few range tests

* Repair: Geocent needs an ellipsoid

* pj_fwd/pj_inv: Move checks related to Helmert inside Helmert block

PR #811, opened 2018-02-22
Repair some failures due to missing features in horner & gie

busstoptaktik:
This PR repairs a few bugs revealed during the work on [these](https://github.com/NordicGeodesy/NordicTransformations/pull/4)  parameter files for handling current and legacy Danish systems.

The changes, especially the ones touching the library core, are straightforward, and repair actual bugs (material in `PJ_horner,c`  forgotten during the original porting effort). It is crucial for support of legacy Danish systems, so I find it important to get this into the 5.0.0 cycle, and hence call for yet another RC.

### **Horner:**

**Needs support for westings/southings.**

Introduced for the complex domain. May also be useful for polynomials on the real domain, but I have no test cases available for that. So this will have to wait until an actual use case (and hence test case) shows up.

For the complex case, we have ample test material in legacy Danish systems.

### **gie:**

**Ignore unspecified dimensions, improve reporting and builtin tests.**

For pipelines including a Helmert shift (or any other 3D/4D steps), we need to run `gie` tests through the 4D transformation interface, even though the input coordinate system may be 2D.
    
This can be enforced by appending "0 0" to the 2D coordinate pair in the `accept` command, which is sufficiently recognizable to be considered an idiom for selecting 4D.
    
On return, however, (i.e. in the `expect` command), the last dimensions will contain garbage, and this garbage will be compared with "0 0" when computing the deviation.

This obviously leads to nonsensical results, which this PR repairs by zeroing all dimensions *not given* in `expect`, before computing the deviation.

Additionally, for the built in testing of geo/cartesian roundtrip precision, the test tolerance has been relaxed from picometer to nanometer level.

These tests have shown to blow up intermittently. And as the pm level tolerance is probably a leftover from when deviation was computed in degrees, not meter (and hence a factor of 111000 tighter than intended at its introduction), relaxing it by a factor of 1000 makes ample sense.

Also, two new features, reintroduced from abandoned PR #800, while debugging this case have been left in the code:

- improved reporting, for verbosity levels higher than 2
    
- a "skip" instruction, forcing all remaining work to be skipped (i.e. run until something strange happens - then stop to handle debugging, while avoiding additional garbage)

### DIFF
--- a/src/PJ_horner.c
+++ b/src/PJ_horner.c
@@ -96,6 +96,8 @@ static HORNER *horner_alloc (size_t order, int complex_polynomia);
 static void    horner_free (HORNER *h);
 
 struct horner {
+    int    uneg;     /* u axis negated? */
+    int    vneg;     /* v axis negated? */
     int    order;    /* maximum degree of polynomium */
     int    coefs;    /* number of coefficients for each polynomium  */
     double range;    /* radius of the region of validity */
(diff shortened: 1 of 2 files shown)


kbevers:
I'll let this one slide, but you are now officially banned from looking at PROJ code until 5.0.0 is out. We can't push new RC's forever.

busstoptaktik:
> you are now officially banned from looking at PROJ code until 5.0.0 is out

What a nice relief :-)

Issue #812, opened 2018-02-22
Pole encircling logic in geod_polygon_addedge is bad


cffk:
This test (soon to be added to `geodtest.c`)

    static int AddEdge1() {
      /* Check fix to transitdirect vs transit zero handling inconsistency */
      struct geod_geodesic g;
      struct geod_polygon p;
      double area;
      int result = 0;
      geod_init(&g, wgs84_a, wgs84_f);
      geod_polygon_init(&p, 0);
      geod_polygon_addpoint(&g, &p, 0.0, 0.0);
      geod_polygon_addedge(&g, &p, 90.0, 1000.0);
      geod_polygon_addedge(&g, &p,  0.0, 1000.0);
      geod_polygon_addedge(&g, &p,-90.0, 1000.0);
      geod_polygon_compute(&g, &p, 0, 1, &area, 0);
      result += assertEquals(area, 1000000.0, 0.01);
      return result;
    }

should return 1000000.00 m^2 as the area.  Instead it gets -2.550328e+14
m^2.  The problem is in inconsistent treatment of longitude = 0 by
geod_polygon_addedge and geod_polygon_addpoint.  I have this fixed in my
code base and will be making a pull request this weekend.  (I'm wanting
also to add additional tests to improve the code coverage.)

This is an outright *BUG* and I'm hoping it can make it into
release 5.0.0.  Nothing in the rest of the proj code base calls
`geod_polygon_addedge`, so incorporating this last-minute fix should be
low-risk.

kbevers:
Sounds like a reasonable thing fix to include in 5.0.0, although very last minute.

I was going to put out RC5 this afternoon.  Can you give a rough estimate on when you expect the fix to be ready? I know you said the weekend, but when in the weekend? I don't want to prolong the release process too much if it can be avoided. March 1st is still the release date and I want to give people some time for testing as well.

cffk:
I'll make the pull request Saturday morning (my time) -- around midday
in Western Europe.  Maybe I'll do the bug fix and the single test before
midnight today (Thurs), with the additional coverage tests (in geodtest.c) on
Sat morning.

PR #816, opened 2018-02-23
Fix issue #812

cffk:
Implement Polygon AddEdge fix in C library (will be version 1.49.2 of C
library for proj 5.0.0).  This fixes issue #812.

Still to do: add tests to expand code coverage.  This will only affect
geodtest.c which is not part of the compiled library.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -162,7 +162,7 @@ o The PROJ.4 project also distributes the datum-grid package,
 
  o Updated EPSG database to version 9.2.0.
 
- o Geodesic library updated to version 1.49.1-c.
+ o Geodesic library updated to version 1.49.2-c.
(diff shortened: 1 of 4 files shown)


kbevers:
Do you mind updating `NEWS` with the proper version number of the geodesic library as well?

cffk:
geodesic.c is down to 3 uncovered lines now.  I think that's as far as I
can take it.  In my second commit message you will see that I've
uncovered another problem.  However this only affects polygons which are
not "simple" so I'll defer the fix to this to another day.

kbevers:
Sloppy testing - what about that last line of live code? :-)

Thanks. I'll get started on RC5.

Issue #819, opened 2018-02-23
gie test failures (on amd64)


sebastic:
proj-5.0.0RC5 fails to build due to gie test failures:
```
-------------------------------------------------------------------------------
Reading file '4D-API_cs2cs-style.gie'
-------------------------------------------------------------------------------
proj=latlong geoidgrids=egm96_15.gtx ellps=GRS80                      
-------------------------------------------------------------------------------
Invalid operation definition in line no. 71:
       (null) (errno=PJD_ERR_UNKNOWN/0)
-------------------------------------------------------------------------------
proj=merc geoidgrids=egm96_15.gtx ellps=GRS80                         
-------------------------------------------------------------------------------
Invalid operation definition in line no. 84:
       (null) (errno=PJD_ERR_UNKNOWN/0)
-------------------------------------------------------------------------------
proj=latlong geoidgrids=egm96_15.gtx axis=neu ellps=GRS80             
-------------------------------------------------------------------------------
Invalid operation definition in line no. 104:
       (null) (errno=PJD_ERR_UNKNOWN/0)
-------------------------------------------------------------------------------
proj=merc geoidgrids=egm96_15.gtx ellps=GRS80                         
-------------------------------------------------------------------------------
Invalid operation definition in line no. 124:
       (null) (errno=PJD_ERR_UNKNOWN/0)
-------------------------------------------------------------------------------
total:  6 tests succeeded,  6 tests skipped,  4 tests FAILED!
-------------------------------------------------------------------------------
```

kbevers:
God dammit, I am getting tired of those frickin' grids.

The error must've been introduced in https://github.com/OSGeo/proj.4/pull/811. I'll see if I can find the problem, otherwise @busstoptaktik  will have to look at it. These tests where supposed to be skipped if the grid isn't found (which I suspect is the problem).

kbevers:
@sebastic Can you run `gie -vvvv 4D-API_cs2cs-style.gie` and post the output? And the same with the `egm96_15.gtx` grid in `PROJ_LIB`. I am 99% percent certain the problem is that the grid is missing, but I can't understand why it doesn't just skip those tests. I can't reproduce it locally.

sebastic:
**Without `egm96_15.gtx`**
```
PROJ_LIB=../../nad/ gie -vvvv 4D-API_cs2cs-style.gie 
-------------------------------------------------------------------------------
Reading file '4D-API_cs2cs-style.gie'
[...]

PR #820, opened 2018-02-23
Don't reset errno when skipping tests.

kbevers:
The error number should not be reset until a new instance of "operation
..." is reached. The ignore-feature initially worked by accident since
pj_errno was not being reset when calling proj_errno_reset. This was
fixed in #808, which subsequently caused ignored tests to fail.

Closes #819

PR #821, opened 2018-02-23
Rerun tests without grids not included in proj-datumgrid.

sebastic:
To catch regressions like #819 the tests should also be run with only the grids included in proj-datumgrid.


kbevers:
Good point. After the dust has settled on 5.0 I'll try to come up with a more robust solution for the CI tests. They are rather fragile at the moment.

kbevers:
Fails because the fix in #820 isn't included in this branch. Please merge that in and push the branch again so we can have a proper run of the tests.

PR #822, opened 2018-02-23
Update travis to use proj-datumgrid-1.7.

sebastic:
No need to download `BETA2007.gsb` separately any more.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -59,8 +59,8 @@ test_script:
   - if "%BUILD_TYPE%" == "cmake" set PROJ_LIB=C:\projects\proj-4\nad
   - if "%BUILD_TYPE%" == "nmake" set PROJ_LIB=C:\PROJ\SHARE
   - cd %PROJ_LIB%
-  - curl -O http://download.osgeo.org/proj/proj-datumgrid-1.7RC2.zip
-  - 7z e -aoa -y proj-datumgrid-1.7RC2.zip
+  - curl -O http://download.osgeo.org/proj/proj-datumgrid-1.7.zip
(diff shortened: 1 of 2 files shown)


kbevers:
I disagree. The package is 100+ MB. It is just a waste of time and bandwith to download that. That is one of the reasons why I want to create subdivisions of the datumgrids package.

sebastic:
Then the grids that are considered "too big" should be removed from proj-datumgrid.

kbevers:
I understand where you are coming from and in principle I agree. The thing is though, there's a difference between users that download the package once and an automated system downloading the package several of a hundred times a day (there's 15 test targets that all need the package, that's ~1.5GB pr commit). It takes extra time to download the package, but more importantly it wastes bandwith for both OSGeo and the CI providers. We get all of this for free and I don't want to abuse that privilige unless it is necessary. It isn't in this case. So for now i prefer to download what's needed for the tests to run and not much else.

I believe it should be possible to do caching of frequently downloaded files. When I have some more time on my hands I will look into that as a part of a more thorough overhaul of the test-suite.

sebastic:
proj-datumgrid-1.7RC2 is of much more reasonable size (only slightly more than double the 1.6 size).

Both the Travis & AppVeyor configuration were updated to use it.

kbevers:
@sebastic Do you mind updating this to use the final 1.7 package?

sebastic:
> Do you mind updating this to use the final 1.7 package?

Done in 973886758abc62e762e1fb75c838599bf8de9552

kbevers:
Thanks!

Issue #826, opened 2018-02-27
Intel compiled test failures with geodtest


mwtoews:
Building proj-5.0.0RC6 using autotools with Intel compilers, I see some test failures. (These are non-critical, and can be addressed after PROJ v 5 is out.)

Here is the workflow
```
$ CC=icc ./configure
...
$ make
...
$ make check
Making check in src
make[1]: Entering directory `/tmp/proj-5.0.0/src'
make  geodtest
make[2]: Entering directory `/tmp/proj-5.0.0/src'
make[2]: Leaving directory `/tmp/proj-5.0.0/src'
make  check-TESTS
make[2]: Entering directory `/tmp/proj-5.0.0/src'
make[3]: Entering directory `/tmp/proj-5.0.0/src'
FAIL: geodtest
============================================================================
Testsuite summary for PROJ.4 Projections 5.0.0
============================================================================
# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See src/test-suite.log
Please report to warmerdam@pobox.com
============================================================================
$ cat src/test-suite.log
==================================================
   PROJ.4 Projections 5.0.0: src/test-suite.log
==================================================

# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: geodtest
==============

assertEquals fails: 0 != 3.5e-14 +/- 1.5e-14
assertEquals fails: 180 != 180 +/- 1.5e-14
GeodSolve59 fail: 2
assertEquals fails: 180 != -180 +/- 5e-06
GeodSolve67 fail: 1
FAIL geodtest (exit status: 2)
```
I've started an open discussion with @cffk regarding some of these with GeographicLib, where they may perhaps be resolved.

cffk:
Regarding the geodtest failures...  This isn't a case where relaxing the tests is the right solution.  It's possible that icc is optimizing too aggressively.  Is it possible to compile the code without optimization (-O0)?  I'll also see if I can lay my hands on a system with icc installed.

mwtoews:
@cffk your guess is correct; `make clean && CC=icc CFLAGS=-O0 ./configure && make && make check` is clean.
I'm not sure how to get around Intel's aggressive optimizations here...

mwtoews:
[...]

PR #827, opened 2018-02-27
Modify project bugreport (autoconf)

mwtoews:
See autoconf documentation for [AC_INIT](http://www.gnu.org/software/autoconf/manual/html_node/Initializing-configure.html) where the third parameter `AC_PACKAGE_BUGREPORT` is "Typically an email address, or URL to a bug management web page."

This parameter is visible at the end of a `make check` failure (via autotools build method), e.g. see the "report to" line the output in #826

With this PR (and re-`autogen.sh`) this is modified to:
```
...
=======================================================
1 of 1 test failed
Please report to https://github.com/OSGeo/proj.4/issues
=======================================================
...
```
which is the preferred place to submit bug reports.


kbevers:
Thanks!

PR #832, opened 2018-03-02
Improve consistency of parameters

awulkiew:
no_rot in omerc projection (boolean)
tilt, azi in nsper projection (angle in dms format)


kbevers:
Please add tests cases for omerc that uses the `+no_rot` and for nsper that uses `+tilt` og `+azi`. Preferable in a commit before the proposed changes so we can be sure that no regressions are introduced.

You can add the tests to `test\gie\builtins.gie` next to the existing tests of the same projections.

awulkiew:
@kbevers Done. `+azi` was already tested.

Btw, is it expected that `12% tests  passed, 22 tests failed out of 25`?

kbevers:
> Btw, is it expected that 12% tests passed, 22 tests failed out of 25?

If you haven't installed the grids from [proj-datumgrid](https://github.com/OSGeo/proj-datumgrid), probably yes.

PR #815, opened 2018-02-23
Tidy a few typos

mwtoews:
Mostly found in documentation or other documents.

PR #840, opened 2018-03-05
Fix a few typos (again)

mwtoews:
A few typos from #815 crept back in, but I think I've identified them. And found several more typos.


kbevers:
Thanks!

PR #849, opened 2018-03-09
Remove angle brackets for proj headers

### LINKED ISSUE #841: Switch #include <proj(...).h> to #include "proj(...).h"
... to avoid including an old, already installed `proj.h`, `proj_api.h` and `projects.h` when building a new version of the library.

Not realizing the extent of this (due to misconstructed grep pattern, when searching for additional cases), I introduced a single correction in 20e5610911a6d6231121eb3c0d3c38330f8d708e

Given the number of actual cases, this is probably only a minor problem, which will only show during in-tree builds. So I believe repairing it can wait until we have a number of tasks requiring a full pass over the source (I can think of a few, and will add issues for those)

busstoptaktik:
Using angle brackets in include directives increases the risk of accidentally picking up an older system installed header file while building an updated version of PROJ.

This PR changes the angle brackets to "" for proj.h, projects.h, proj_api.h and geodesic.h

Resolves #841.

### DIFF
--- a/src/PJ_aea.c
+++ b/src/PJ_aea.c
@@ -28,7 +28,7 @@
  *****************************************************************************/
 
 #define PJ_LIB__
-#include <proj.h>
+#include "proj.h"
 #include <errno.h>
 #include "projects.h"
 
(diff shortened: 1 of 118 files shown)


kbevers:
Cherry-picked into 5.0 branch

PR #851, opened 2018-03-11
Add Web Mercator test case

kbevers:


test/gie/4D-API_cs2cs-style.gie
```diff
@@ -186,4 +186,15 @@ tolerance 20 cm
...
 -------------------------------------------------------------------------------
+
+-------------------------------------------------------------------------------
+Test that Google's Web Mercator works as intended (see #834 for details).
+-------------------------------------------------------------------------------
+operation   proj=pipeline step init=epsg:26915 inv step init=epsg:3857
+-------------------------------------------------------------------------------
+tolerance   20 cm
+accept      487147.594520173 4934316.46263998
+expect      -10370728.80    5552839.74
```
busstoptaktik: Shouldn't the test be 3D, to check that heights are referred to the ellipsoid, while latitudes are  referred to the circumscribed sphere?

kbevers: I dont know really. According to [this](http://epsg.io/3857) it is a 2D system. Maybe both a 2D and 3D test should be added?

PR #853, opened 2018-03-11
Typo fixes

rouault:
Question: should we keep the mispelled PJD_ERR_UNKNOW_UNIT_ID in projects.h ?


rouault:
OK, mispelled #defined removed per 290834e

kbevers:
> OK, mispelled #defined removed per 290834e

I probably wasn't very clear before. What I meant was that keeping both the misspelled and the correctly spelled constant would be best until we remove `projects.h`. We'll release 5.1 and 5.2 before 6.0 so theoretically we're are changing public API without a bump of major version number. It is extremely unlikely to cause problems but you never know what's going on out there :-)

rouault:
@kbevers OK, I just cancelled the last commit that removed the misspelled constant.

kbevers:
Thanks!

PR #854, opened 2018-03-11
Allow mingw32 travis tests to fail

kbevers:
The previous attempt at allowing mingw32 tests to fail was not
sucessfull. It seems that all environment variables need to be matched
completely before the allowed failure setting is picket up. This commit
attempts to do that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,9 +38,8 @@ matrix:
       env:
         - BUILD_NAME=mingw32
         - DETAILS="mingw32"
-
   allow_failures:
-      - env: BUILD_NAME=mingw32
+      - env: BUILD_NAME=mingw32 DETAILS="mingw32"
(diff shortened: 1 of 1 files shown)


kbevers:
Hmmm... not having much luck allowing this test to fail. It seems that the problem is multistresstest hanging maybe half the time the test is run. I don't know if there's a fix to that apart from removing the call to multristresstest.exe.

@rouault  Do you mind if I remove multistresstest from the mingw32 travies test?

rouault:
@kbevers Alternate strategy attempted at https://github.com/OSGeo/proj.4/pull/855

kbevers:
Nice. Seems like a better solution. Let's see how that works out.

kbevers:
Better fix in #855

PR #855, opened 2018-03-11
mingw32 travis: reduce number of iterations for multistresstest

rouault:


rouault:
Runs now in 4 min vs 12 before. Merging

PR #857, opened 2018-03-11
Make sure that transient errors are returned correctly

kbevers:
The prepare/finalize steps in pj_fwd*/pj_inv* now checks the error level
of the PJ object before returning the coordinate. If an error is
detected proj_coord_error() is returned.

Prompted by https://github.com/pytroll/mpop/issues/49#issuecomment-372142012


src/pj_fwd.c
```diff
@@ -174,11 +174,23 @@ static PJ_COORD fwd_finalize (PJ *P, PJ_COORD coo) {
 }
 
 
+static PJ_COORD error_or_coord(PJ *P, PJ_COORD coord, int last_errno) {
+    if (proj_errno(P))
+        coord = proj_coord_error();
+    else
```
busstoptaktik: Could handle the non-common case by  `if (...) return proj_coord_error()` here.

This would raise the common case to top level, rather than partially depressed behind an else clause

src/pj_fwd.c
```diff
@@ -174,11 +174,23 @@ static PJ_COORD fwd_finalize (PJ *P, PJ_COORD coo) {
...
+        proj_errno_restore(P, last_errno);
+
+    return coord;
+}
+
 
 XY pj_fwd(LP lp, PJ *P) {
+    int last_errno;
     PJ_COORD coo = {{0,0,0,0}};
     coo.lp = lp;
```
busstoptaktik, line 189: It is tempting, at this stage, to consider whether we want to play ball *at all* if there's an unhandled exception - i.e. just do

    if (proj_errno(P))
        return proj_coord_error();

It will, however, require some additional error handling in the `pj_transform()`, and the `proj_trans...()`family, to avoid an error for a single point (e.g. out of area) to result in all subsequent points in that call failing.

(this comment also applies to `pj_fwd3d`, `pj_fwd4d`, and their `inv`counterparts).

kbevers: That's a problem for another day :-)

busstoptaktik: :-)

kbevers:
Cherry-picked into 5.0 branch

PR #860, opened 2018-03-12
Fix various issues spotted by cppcheck and clang static analyzer

rouault:
- Fixes all cppcheck and clang static analyzer warnings
- Add stricter cppcheck checks (errors  + warnings) in Travis-CI
- Add a clang static analyzer target in Travis-CI

### DIFF
--- a/src/PJ_isea.c
+++ b/src/PJ_isea.c
@@ -139,11 +139,13 @@ struct isea_geo {
     double lon, lat;
 };
 
+#ifdef unused
 struct isea_address {
     int type; /* enum isea_address_form */
     int number;
     double  x,y; /* or i,j or lon,lat depending on type */
 };
+#endif
 
 /* ENDINC */
 
--- a/src/PJ_isea.c
+++ b/src/PJ_isea.c
@@ -149,7 +149,9 @@ enum snyder_polyhedron {
 };
 
 struct snyder_constants {
-    double          g, G, theta, ea_w, ea_a, ea_b, g_w, g_a, g_b;
+    double          g, G, theta;
+    /* cppcheck-suppress unusedStructMember */
+    double ea_w, ea_a, ea_b, g_w, g_a, g_b;
 };
 
 /* TODO put these in radians to avoid a later conversion */
--- a/src/PJ_isea.c
+++ b/src/PJ_isea.c
@@ -139,14 +139,6 @@ struct isea_geo {
     double lon, lat;
 };
 
-#ifdef unused
-struct isea_address {
-    int type; /* enum isea_address_form */
-    int number;
-    double  x,y; /* or i,j or lon,lat depending on type */
-};
-#endif
-
 /* ENDINC */
 
 enum snyder_polyhedron {
--- a/src/PJ_lsat.c
+++ b/src/PJ_lsat.c
@@ -61,8 +61,6 @@ static XY e_forward (LP lp, PJ *P) {          /* Ellipsoidal, forward */
         sav = lampp;
         lamtp = lp.lam + Q->p22 * lampp;
         cl = cos(lamtp);
-        if (fabs(cl) < TOL)
-            lamtp -= TOL;
         if( cl < 0 )
             fac = lampp + sin(lampp) * M_HALFPI;
         else
--- a/src/PJ_misrsom.c
+++ b/src/PJ_misrsom.c
@@ -78,8 +78,6 @@ static XY e_forward (LP lp, PJ *P) {          /* Ellipsoidal, forward */
             sav = lampp;
             lamtp = lp.lam + Q->p22 * lampp;
             cl = cos(lamtp);
-            if (fabs(cl) < TOL)
-                lamtp -= TOL;
             if( cl < 0 )
                 fac = lampp + sin(lampp) * M_HALFPI;
             else
--- a/src/pj_param.c
+++ b/src/pj_param.c
@@ -131,17 +131,20 @@ PROJVALUE pj_param (projCtx ctx, paralist *pl, const char *opt) {
 
     /* Not found */
     if (0==pl) {
+        /* Return value after the switch, so that the return path is */
+        /* taken in all cases */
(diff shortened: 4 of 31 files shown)


src/proj_4D_api.c
```diff
@@ -1004,7 +1008,7 @@ PJ_INIT_INFO proj_init_info(const char *initname){
     key[64] = 0;
     strncat(key, ":metadata", 9);
     strcpy(param, "+init=");
-    strncat(param, key, 73);
+    strncat(param + strlen(param), key, sizeof(param)-1-strlen(param));
```
kbevers: I don't understand the `param + strlen(param)` part of this. What's going on here?

rouault: This helps cppcheck (probably false positive otherwise). strncat(foo, bar, baz) is equivalent of strncat(foo + strlen(foo), baz, baz)
The real fix is to make sure to use sizeof(foo) - 1 - strlen(foo) to make sure it doesn't go beyond end of buffer

kbevers: Ah, okay. I think a comment here is warranted. At least I'll forget this explanation quickly and wonder what's going on here the next time I stop by.

rouault: Comment added

kbevers: Thanks

rouault:
Any additional comments / opposition to merging this ?

cffk:
I'm fine with this...

kbevers:
I only looked it over briefly yesterday. There's a thing or two that I'd like to check again. I'll get back tomorrow.

src/pj_param.c
```diff
@@ -134,14 +134,15 @@ PROJVALUE pj_param (projCtx ctx, paralist *pl, const char *opt) {
         switch (type) {
         case 'b': case 'i':
             value.i = 0;
-            return value;
+            break;
```
kbevers, line 139: This has been changed back and forth a few times now. I don't care much which is used, but a comment here would probably be a good idea to avoid this getting changed back again at a later stage.

src/PJ_isea.c
```diff
@@ -139,11 +139,13 @@ struct isea_geo {
     double lon, lat;
 };
 
+#ifdef unused
```
kbevers: Why not remove this entirely? It's still there in the log if needed in the future.

src/PJ_lsat.c
```diff
@@ -61,8 +61,11 @@ static XY e_forward (LP lp, PJ *P) {          /* Ellipsoidal, forward */
         sav = lampp;
         lamtp = lp.lam + Q->p22 * lampp;
         cl = cos(lamtp);
+#ifdef fixme
+        /* lamtp is not read any more */
```
kbevers: Has `lamtp` been used previously or is this just something that has been left in by mistake when originally implementing this projection?

src/PJ_misrsom.c
```diff
@@ -78,8 +78,11 @@ static XY e_forward (LP lp, PJ *P) {          /* Ellipsoidal, forward */
             sav = lampp;
             lamtp = lp.lam + Q->p22 * lampp;
             cl = cos(lamtp);
+#ifdef fixme
```
kbevers: Same as above. Has `lamtp` been used previously or is this just something that has been left in by mistake when originally implementing this projection?

kbevers (commented):
Generally this looks very good. I have a few minor comments in-lined.

The Clang static analyzer is awesome. So is the cppcheck stuff. I think it would be good to write something for the docs about which settings are used etc. So it is easier to make the same checks locally before pushing the code.

src/PJ_lsat.c
rouault: History ( https://github.com/OSGeo/proj.4/blame/4.8.0/src/PJ_lsat.c ) shows that this issues dates back to the initial commit 19 years ago...

src/PJ_misrsom.c
rouault: Similarly as PJ_lsat.c, the issues dates back to the initial commit ( https://github.com/OSGeo/proj.4/commit/991ed3603ecd46284b91bdbaaa39a01cfa6c0c70#diff-0dca5024bc8819892dd0d81ca8be2b90 ) , but this a "recent" file, probably a copy&paste from PJ_lsat

src/PJ_isea.c
rouault: OK I'll remove it

src/pj_param.c
rouault: OK I'll add a comment

rouault:
> So is the cppcheck stuff. I think it would be good to write something for the docs about which settings are used etc. So it is easier to make the same checks locally before pushing the code.

Where would you document this ?
- cppcheck: run scripts/cppcheck.sh
- clang static analyzer: basically run travis/csa/before_install.sh and travis/csa/install.sh

kbevers:
> Where would you document this ?

I was thinking somewhere under docs/source/development actually. I have been toying with the idea of adding a section called "For PROJ developers" or "Developing PROJ". What you've done in this PR is the kind of stuff that I think would be useful in a section like that. Both as a reminder for current devs but also to make it easier to get started for new contributors. It doesn't need to be added as part of this PR, it just hit close to an idea that has been slowly growing in my head so I wanted to mention it.

src/PJ_lsat.c
kbevers: I think it is safe to remove those lines then :-) Same for the one below.

rouault:
> It doesn't need to be added as part of this PR, it just hit close to an idea that has been slowly growing in my head so I wanted to mention it.

I've created https://github.com/OSGeo/proj.4/issues/862 about that

kbevers:
Thanks.

src/PJ_misrsom.c
rouault: Lines removed

kbevers (approved):

PR #863, opened 2018-03-14
Fix Debian package name

sebastic:
On Debian/Ubuntu the `proj-bin` package provides the commandline utilities, and `libproj-dev` the library headers and `.so` symlinks, the `-dev` required to build software that links to libproj.


kbevers:
Thanks!

PR #867, opened 2018-03-17
Removal of assignment statements that do not have any functionality

hwiesmann:
The assigned variable are not used anymore after the assignment.


rouault:
This has already been committed a few days ago per https://github.com/OSGeo/proj.4/commit/98258282af54a8ec315aaa216d48f890dd09bbdf . Not sure why github doesn't detect a conflict here

busstoptaktik:
> Not sure why github doesn't detect a conflict here.

@rouault: It probably shouldn't - it's hardly a conflict removing something that isn't there. You wouldn't want it to yell if merging a PR removing two lines already removed in another PR, then adding one new line. Same for this case, except for adding one line less :-)

Issue #831, opened 2018-03-01
mac osx geodtest fails


mikehagerty:
I seem to have the same issue as #826 (geodtest fail on intel) on mac OS X 10.10 (gcc-4.9):

> more src/test-suite.log
==================================================
   PROJ.4 Projections 5.0.0: src/test-suite.log
==================================================
#TOTAL: 1
#PASS:  0
#SKIP:  0
#XFAIL: 0
#FAIL:  1
#XPASS: 0
#ERROR: 0
FAIL: geodtest
==============

assertEquals fails: 180 != -180 +/- 5e-06
GeodSolve61 fail: 1
FAIL geodtest (exit status: 1)

gcc does not appear to like the -fp-model precise CFLAG.
This appears to be an intel flag that doesn't work with mac (?)

Also, from what I can tell, 'precise' is suppose to be the default setting
for GCC/CLANG.

Any suggestions would be appreciated.

Thanks!

cffk:
There are two tests of lon2 == -180 in GeodSolve61 and only one of them is failing.  They are on lines 523 and 528 of geodtest.c.  Could you please figure out which one is failing?  Also could you try the test with all optimization turned off (-O0)?

Also can you check whether this problem happens with the previous version of proj?

Thanks.

cffk:
The test failure you're seeing is consistent with `sin(-0.0)` returning `0.0` instead of `-0.0`.  Perhaps you can check this on your system.  I already have a workaround for this bug (that's what it is) for old versions of Visual Studio; see line 248 in geodesic.c.  If this is the problem, then could you figure out how to modify the test at line 248 to include your system?  (While you're about it, could you also determine whether `fmod(-0.0, 360.0)` returns `-0.0`?)

mikehagerty:
Hi Charles,

Sorry, I just saw your mails.

1. It's the first test (line 523) that fails.

2. sine(-0.0) looks right:

  MTH line 523: sin(0.0)=0.000000 sin(-0.0)=-0.000000

3. -O0 (optimization off) didn't fix it.

It's late here, I'll try to look at it tomorrow.

Cheers,
-Mike

mikehagerty:
Correction:

3. Turning off optimization (-O0) *does* fix the test fail.  I must not
have been done a clean compile last night.

Cheers,
-Mike

cffk:
Please apply these patches (to print out some debuggining information)

[...]

PR #868, opened 2018-03-17
Patch 1.49.3 for geodesic package.

cffk:
Set flags for Intel compiler to prevent incorrect optimization of
arithmetic expressions #826.  Guard against nans in sincosdx #843.
Issue #831 is not addressed here (need more information...).

@kbevers and @rouault I flagged some of the changes made to
satisfy cppcheck, e.g.,

        if ((outmask & GEOD_LATITUDE) && plat2) /* plat2 check redundant */
          *plat2 = lat2;

The outmask test makes the plat2 test unnecessary.  So, if possible, I'd
like to back out of this change and revert to

        if (outmask & GEOD_LATITUDE) 
      
Sorry not to have caught this earlier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,18 @@ elseif("${CMAKE_C_COMPILER_ID}" MATCHES "Clang")
     -Wfloat-conversion ${CMAKE_C_FLAGS}")
 endif()
 
+# Tell Intel compiler to do arithmetic accurately.  This is needed to
+# stop the compiler from ignoring parentheses in expressions like
+# (a + b) + c and from simplifying 0.0 + x to x (which is wrong if
+# x = -0.0).
+if (CMAKE_C_COMPILER_ID STREQUAL "Intel")
+  if (MSVC)
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /fp:precise")
+  else ()
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fprotect-parens -fsigned-zeros")
+  endif ()
+endif ()
+
 #################################################################################
 # PROJ4 CMake modules
 #################################################################################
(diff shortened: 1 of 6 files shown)


cffk:
The setting of the flags for the Intel compiler has been tested on Linux only (cmake + autoconf).  I've specified the right flag for the Intel compiler on Windows (with cmake), but I haven't tested this.

rouault:
> The outmask test makes the plat2 test unnecessary.

This is actually not for cppcheck but for clang static analyzer, for which I don't know how to put suppressions as annotations in the code. This was so immediately obvious to me that the test is unnecessary. CSA is probably confused by the line "outmask &= l->caps & OUT_ALL;" which modifies the flags, but as and masking can only remove a bit, indeed the current code is correct, but it needs a (too) clever brain to realize it. I'd vote to let the additional check. If not, some exception should be added in the last line of travis/csa/install.sh

cffk:
@rouault OK, I'll let it be.  Maybe I'll change the comment to put the blame on CSA.  (You'll see that I did the same with the initial assignments.  Perhaps you can check that I blamed the right party!)

cffk:
I can't easily debug the configuration failures on the CI machines.  `autogen.sh` and `configure` work OK for me (Intel and gcc).  I suspect something needs to be done to bring `AX_CHECK_COMPILE_FLAG` into scope.  Worst comes to worst, I remove the changes to `configure.ac` and Intel is only supported via cmake.

QuLogic:
You need to add the file defining the `AX_CHECK_COMPILE_FLAG` macro into the `m4` directory (the autoconf archive is all copylibs).

cffk:
@QuLogic Thanks that did it.
@mwtoews Perhaps you could verify that I've got the Intel flags right on as many platforms as possible (+ configure + cmake).  Thanks.

kbevers:
cherry-picked into 5.0 branch

PR #869, opened 2018-03-18
Auto-generate man pages from Sphinx docs

kbevers:
This PR reformats the existing application doc pages on the webpage to better use the Sphinx syntax. A new page for the gie application has been added as well. Using the proper Sphinx syntax makes it possible to generate man pages from the Sphinx docs making sure that the web page and man pages are aligned and up to date.

Because the way options are described in Sphinx is not as flexible as I would like it has been necessary to describe them a little differently to the old man-pages. This is especially with regards to flags like `-w` and `-W` that are described as `-[wIW]n` in the current man pages. This is hard to do in Sphinx so I have split those into two different flags `-w` and `-W` with each it's own description. I might have interpreted the original text wrong so please take a good look on those and check that the description matches reality.

I have added rule to the autoconf makefiles that takes care of building the man pages. I am not entirely happy with the way I have manage to set it up since the man-pages are generated every time `make` is called. Even if nothing in the docs has changed. Ideally I would like the man-pages only be generated when running `make install` and `make dist` but haven't been able to figure out how to do this. Ideas how to change that are welcome. 

I haven't looked at CMake yet and have most likely mocked that up by changing the folder structure in `man/`. Also, Sphinx is now a dependency when building with automake. I don't think this is too big a deal since we are already using it, but with this it is a hard dependency instead of a soft dependency. If it is possible to reconfigure the makefiles as described above it will only have to be a soft dependency.

I expect this needs some changes before merging but I wanted it out there to show what I am working on and hopefully get some tips on how to configure the makefiles in a better way.


kbevers:
@busstoptaktik @cffk I've added you as reviewers for this because I need your feedback on the updated application doc pages. Charles I need you to check that the flags for `geod` are described correctly. Thomas I need your input on the `cct` page and especially the `gie` page.

For easier reading run `make man` and `make html` in `doc/` and find the relevant files in `docs/build/`.

cffk:
The man pages look fine.  Potential problems are:

* cmake build+install fails
* configure build+install creates the documentation twice
* creating the documentation requires sphinx-build

Would it make sense to create the documentation as part of autogen.sh?

kbevers:
It might. Or handle it somewhat manually as part of the release process at least. I've been trying to get CMake sorted for the last hour or so, and I think it might end up being more complicated than is good.

It could be as simple as adding a step to `HOWTO-RELEASE` with something like:

```
Generate man pages:

1. cd docs
2. make man
3. cp -f build/man/*.1 ../man/man1
4. cp -f build/man/*.3 ../man/man3 # eventually, maybe
5. git add /man/man1/* man/man3/*
6. git commit -m "update man-pages."
```

Although the whole reason for this reworking of the docs was to simplify the release process. Oh well :-)

kbevers:
I have abandoned the idea with the build systems and added a script for updating man pages semi-automatically. New versions of the man pages have been committed as a test of the script and to show the resulting Sphinx-build man pages.

PR #873, opened 2018-03-20
Add 'exact' flag to Helmert operation when doing cs2cs emulation.

### LINKED ISSUE #871: Pipeline with +towgs84 doesn't work as expected
Trying to port GDAL to use new API, I'm doing some unexpected results when doing supposedly no-op transformations involving towgs84

The output of the following code
```
#include "proj.h"
#include <math.h>
#include <stdio.h>

{
    PJ* pj = proj_create(0, "+proj=pipeline +step "
                            "+proj=longlat +ellps=intl +towgs84=-104.1,-49.1,-9.9,0.971,-2.917,0.714,-11.68 "
                            "+inv +step "
                            "+proj=longlat +ellps=intl +towgs84=-104.1,-49.1,-9.9,0.971,-2.917,0.714,-11.68");
    double x[] = { 2. / 180 * M_PI };
    double y[] = { 49. /180 * M_PI };
    double z[] = { 0 };
    printf("%.18g %.18g %.18g\n", x[0], y[0], z[0]);
    proj_trans_generic (pj, PJ_FWD,
                        x, sizeof(double), 1,
                        y, sizeof(double), 1,
                        z, sizeof(double), 1,
                        0, 0, 0);
    printf("%.18g %.18g %.18g\n", x[0], y[0], z[0]);
    return 0;
}
```
is
```
0.0349065850398865909 0.855211333477221336 0
0.0349065851566477472 0.855211333481314284 0.00130653195083141327
```
Whereas the second line corresponding to the output should be identical to the first one
Doing the same using the proj_api.h with pj_transform() results in the expected result

kbevers:
fixes #871.

### DIFF
--- a/src/proj_4D_api.c
+++ b/src/proj_4D_api.c
@@ -490,7 +490,7 @@ Returns 1 on success, 0 on failure
         def = malloc (100+n);
         if (0==def)
             return 0;
-        sprintf (def, "break_cs2cs_recursion     proj=helmert %s transpose", s);
+        sprintf (def, "break_cs2cs_recursion     proj=helmert exact %s transpose", s);
         Q = proj_create (P->ctx, def);
         pj_inherit_ellipsoid_def (P, Q);
         free (def);


kbevers:
cherry-picked into 5.0 branch

PR #880, opened 2018-03-21
Avoid shadowed declaration errors with old gcc

mwtoews:
On an old CentOS with gcc 4.4.7, building with CMake has errors compiling a few files due to the `-Wshadow` flag. (The build with autotools does not complain since it doesn't enable this flag, but that's possibly a different issue with whatever `AX_CFLAGS_WARN_ALL(C_WFLAGS)` is doing).

I suspect `-Wshadow` is more strict with older gcc releases, which is why this not the normal behaviour.

The modifications rename some function variable names, so that they are different than other function names used globally (e.g. `log`) or elsewhere (e.g. `geodetic`). If other variable names are preferred, I can swap these.

### DIFF
--- a/src/PJ_cart.c
+++ b/src/PJ_cart.c
@@ -131,39 +131,39 @@ static double geocentric_radius (double a, double b, double phi) {
 
 
 /*********************************************************************/
-static XYZ cartesian (LPZ geodetic,  PJ *P) {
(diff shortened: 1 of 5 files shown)


kbevers:
Seems good to me. Thanks for fixing this, I would never have stumbled on this.

kbevers:
Cherry-picked into 5.0 branch

PR #882, opened 2018-03-21
Take into account +towgs84=0,0,0 in pipeline to still imply geodetic->cartesian->geodetic (fixes #881)

### LINKED ISSUE #881: +towgs84=0,0,0 ignored in pipelines
With proj master
```
echo "440720 3751320 0" |  src/cct +proj=pipeline +step +proj=utm +zone=11 +ellps=clrk66 +towgs84=0,0,0 +inv +step +proj=utm +zone=11 +datum=WGS84
   440721.3314   3751514.1319        0.0000           inf
```
whereas the equivalent cs2cs gives 440719.958709357 3751294.2109841 -4.44340920541435

If you slightly modify the towgs84 values to be almost 0 like
```
echo "440720 3751320 0" |  src/cct +proj=pipeline +step +proj=utm +zone=11 +ellps=clrk66 +towgs84=0.000000000001,0,0 +inv +step +proj=utm +zone=11 +datum=WGS84
   440721.3314   3751514.1319        0.0000           inf
```
you get the expected result. This strong discontinuity in outputs regarding very slight input parameter changes is not expected from my point of view.

The reason is that the completely 0 towgs84 is ignored in cs2cs_emulation_setup(), and so no cartesian/geodetic transformations are done to change the ellipsoids.

Suggested fix to come in a pull request

rouault:
…


src/proj_4D_api.c
```diff
@@ -480,9 +481,18 @@ Returns 1 on success, 0 on failure
         double *d = P->datum_params;
         size_t n = strlen (s);
 
+
```
schwehr: nit: I've been seeing a number of 2 blank lines inside of functions.  Is there a reason for the extra white space?

kbevers: I think in some cases it is just by mistake. In others who ever wrote the code felt that it improves clarity/readability. Hard to tell really. Since we don't have any strict formatting rules this thing happen from time to time.

src/proj_4D_api.c
```diff
@@ -480,9 +481,18 @@ Returns 1 on success, 0 on failure
...
         /* We ignore null helmert shifts (common in auto-translated resource files, e.g. epsg) */
         if (0==d[0] && 0==d[1] && 0==d[2] && 0==d[3] && 0==d[4] && 0==d[5] && 0==d[6])
+        {
+            /* If the current ellipsoid is not WGS84, then make sure frame */
```
kbevers: I suggest you change the comment to:

"If the current ellipsoid is not WGS84, then make sure the change in ellipsoid is still done". Since the Helmert transform is a no-op it is a little confusing to me to talk about a change in frame.

src/proj_4D_api.c
```diff
@@ -480,9 +481,18 @@ Returns 1 on success, 0 on failure
         double *d = P->datum_params;
         size_t n = strlen (s);
 
+
         /* We ignore null helmert shifts (common in auto-translated resource files, e.g. epsg) */
         if (0==d[0] && 0==d[1] && 0==d[2] && 0==d[3] && 0==d[4] && 0==d[5] && 0==d[6])
+        {
```
kbevers: Since the Pandoras box of code style discussion already has been opened, please follow the few [rules](http://proj4.org/development/for_proj_contributors.html#coding-style) we do have :-)
(in the rest of this file starting curly braces are on the same line as `if (...)` etc.

test/gigs/5111.1.gie
```diff
@@ -21,7 +21,7 @@ tolerance  0.05 m
 accept    100.0876483 77.6534822
 expect    2800000.0 15000000.0
 
-tolerance  0.05 m
+tolerance  0.055 m
```
kbevers, line 24: Why is the extra 5 mm needed here?

kbevers (changes requested):
Looks good to me. Only a few minor details to sort out.

It is good to have this part of the good properly tested so all these bugs can be weeded out. Looking forward to seeing GDAL ported to the new API!

test/gigs/5111.1.gie
rouault: To make the test pass... I can't make sense about the towgs84=0,0,0 default in the pipeline definition

src/proj_4D_api.c
rouault: Extra line added by mistake removed

src/proj_4D_api.c
rouault: Style fixed

src/proj_4D_api.c
rouault: Suggested comment adopted

test/gigs/5111.1.gie
kbevers: I think the idea is to override the `+towgs84` parameters in the epsg init file, probably in an effort to work around `towgs84=0,0,0` not being handled correctly before.

It makes sense that you have to change the tolerance since `towgs84` is now handled differently.

rouault:
Note: with this pull request and the previous fixes, this GDAL branch https://github.com/rouault/gdal2/tree/proj5 passes its autotest suite using proj.5 API :-)

kbevers (approved):

kbevers:
That is friggin' amazing!!

kbevers:
I guess that means you are dont expect to find more bugs today? I'll start work on a 5.0.1 release candiate later today if that is the case.

kbevers:
Cherry-picked into 5.0 branch

rouault:
Just to be sure, I've re-run my GDAL proj5 branch against proj 5.0 branch, and things work fine. So +1 from my side to start on 5.0.1 !

schwehr:
So far so good with GDAL and the proj 5 API in my very different bazel world.

PR #883, opened 2018-03-21
Handle the case where nad_ctable2_init returns NULL.

schwehr:
Make CTABLE V2 handling the same as CTABLE.

The pj_log call hit a NULL pointer during fuzzing.

Found with autofuzz.


kbevers:
cherry-picked into 5.0 branch

PR #886, opened 2018-03-21
3 -> PJ_LOG_TRACE for pj_log level in pj_init.c

schwehr:
Bare literals are hard to follow and inconsistent with the rest of proj.


kbevers:
Thanks. I thought the extend of this was worse actually.

PR #893, opened 2018-03-23
horner: Fail if the order is unreasonably large.

schwehr:
Overflow in horner_alloc with "2*(int)order"

Found with autofuzz with UndefinedBehaviorSanitizer: signed-integer-overflow (b/73260430)


kbevers:
I am sure the limit could be lowered by a order of magnitude or two but let's just go with this for now as it fixes the problem. For the record, 15 is the highest degree I have seen in the wild so far.

Issue #894, opened 2018-03-23
Convert x == x to !pj_is_nan(x) and x != x to pj_is_nan(x)


schwehr:
Not quite the correct grep, but close enough... only a few false positives.

```
egrep '([a-zA-Z0-9_]+) [!=]= \1' *.[ch]

PJ_geos.c:    if (P->phi0 != 0.0)
PJ_healpix.c:        if (capmap.region == north) {
PJ_healpix.c:        if (capmap.region == north) {
PJ_healpix.c:        if (capmap.region == north) {
PJ_helmert.c:        if (pj_param (P->ctx, P->params, "ttheta").i && Q->scale_0 == 0.0)
PJ_lcca.c:    if (P->phi0 == 0.) {
PJ_unitconvert.c:    return ((year % 4 == 0 && year % 100 != 0) || year % 400 ==0);
gen_cheb.c:	if (low.u == upp.u || low.v >= upp.v)
geodesic.c:  q = r == r ? (int)(floor(r / 90 + (real)(0.5))) : 0;
geodesic.c:    if (l->calp0 == 0 || l->salp0 == 0) {
geodesic.c:    if (calp0 != 0 && salp0 != 0) {
geodtest.c:  result += azi1 == azi1 ? 1 : 0;
geodtest.c:  result += azi2 == azi2 ? 1 : 0;
geodtest.c:  result += s12 == s12 ? 1 : 0;
geodtest.c:  result += azi1 == azi1 ? 1 : 0;
geodtest.c:  result += azi2 == azi2 ? 1 : 0;
geodtest.c:  result += s12 == s12 ? 1 : 0;
geodtest.c:  result += azi1 == azi1 ? 1 : 0;
geodtest.c:  result += azi2 == azi2 ? 1 : 0;
geodtest.c:  result += s12 == s12 ? 1 : 0;
geodtest.c:  result += a12 == a12 ? 1 : 0;
geodtest.c:  result += area == area ? 1 : 0;
geodtest.c:  result += perim == perim ? 1 : 0;
geodtest.c:  result += perim == perim ? 1 : 0;
geodtest.c:  result += area == area ? 1 : 0;
geodtest.c:  result += perim == perim ? 1 : 0;
geodtest.c:  result += perim == perim ? 1 : 0;
multistresstest.c:                    if ( skipTest != test->skip )
pj_apply_vgridshift.c:    if( input.phi != input.phi || input.lam != input.lam )
pj_gridinfo.c:            if( gi != gilist )
```

PR #895, opened 2018-03-23
Use pj_is_nan rather than x == x or x != x (#894)

schwehr:


cffk:
Since `geodesic.c` is a "standalone" component of proj.4 (I do initial
testing of this component within GeographicLib), I'd prefer that it not
use `pj_is_nan` (the issue only arises in one place); possibly a comment
to that effect is warranted in the code.

The same goes for `geodtest.c`; here, for clarity, it should use its own
internal definition of `isnan.`

schwehr:
Removed geodesic.c and geodtest.c from the pr.

PR #896, opened 2018-03-23
geodtest: introduce assertNaN.  geodesic: comment on non-use of pj_is_nan.

cffk:
@schwehr: An alternative to #895 for geodesic.c and geodtest.c


cffk:
Fair enough, I'll change "assert" to "check".  BTW, we're only talking about `geodtest.c` which *isn't* part of the library (it's only used for testing).  Still, better not to cause confusion for the next person to look at the code.

schwehr:
Thanks!

PR #898, opened 2018-03-24
Don't set -std=c89 in CMakeLists.txt

### LINKED ISSUE #892: CMakeLists shouldn't unconditionally set -std=c89
@aaronpuchert added `-std=c89` to `CMAKE_C_FLAGS` in `CMakeLists.txt` for
the laudable reason to force compliance with the c89 standard.  However
standard compiles should not include this flag -- it makes the
`HAVE_C99_MATH` check fail.  Instead this flag should be added in one of
builds on the CI machines.  I'll note that `configure.ac` doesn't include
this flag so, at present, the cmake and autoconf builds don't use the
same compiler flags.

aaronpuchert:
This allows us to take advantage of newer features when they are available. However, builds on Travis still use `-std=c89` to ensure C89 compatibility. Locally this can be achieved with either

* cmake -DCMAKE_C_STANDARD=90 ..
* cmake -DCMAKE_C_FLAGS='-std=c89' ..
* C_FLAGS='-std=c89' cmake ..

We also reorder the warning flags: they are all part of the standard build now, but `-Werror` is only applied on Travis.

Fixes #892.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,19 +18,21 @@ set(PROJECT_INTERN_NAME PROJ)
 if (NOT CMAKE_VERSION VERSION_LESS 3.1)
   cmake_policy(SET CMP0054 NEW)
 endif ()
+
 # Set warnings
 if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
+  # Suppress warning 4706 about assignment within conditional expression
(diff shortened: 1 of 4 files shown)


cffk:
@aaronpuchert Thanks for taking care of this.

PR #899, opened 2018-03-25
Catch over range lam0 in proj_etmerc.c

schwehr:
lam0 of inf caused a nan in the cast to int.
Picked +/-1000 for lam0 as a guess for what constitutes reduculously large values.

proj_etmerc.c:361:16: runtime error: nan is outside the range of representable values of type 'int'

Found with autofuzz: UndefinedBehaviorSanitizer: float-cast-overflow

### DIFF
--- a/src/proj_etmerc.c
+++ b/src/proj_etmerc.c
@@ -345,6 +345,10 @@ PJ *PROJECTION(utm) {
         proj_errno_set(P, PJD_ERR_ELLIPSOID_USE_REQUIRED);
         return pj_default_destructor(P, ENOMEM);
     }
+    if (P->lam0 < -1000.0 || P->lam0 > 1000.0) {
+        return pj_default_destructor(P, PJD_ERR_INVALID_UTM_ZONE);
+    }
+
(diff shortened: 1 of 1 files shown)


schwehr:
poc for gdal/autotest2/cpp/ogr:ogr_srs_proj4_fuzzer:

```
+proj=pipeline	++proj=utm +lon_0=7e902
+stex	+step	+ste
```

kbevers:
The fix seems find. Thanks for catching the problem. May I ask though that you move the `P->lam0` check up ~15 lines, below the `P->es` check, so that the error checking is done before entering the main logic of the code? We[0] try to follow that principle in new code. The reason being that you might as well fail as early as possible and save some CPU cycles and that it removes unnecessary mental obstructions when reading the code since the algorithm is written in it's pure form without any detours.

[0] at least @busstoptaktik and me

schwehr:
I'm of the same mindset... range check up front as much as possible.

Hopefully this redo of the patch doesn't have any surprises.  I don't know proj well enough to have a sense of when something might have a surprise side effect.

kbevers:
This should be fine. Generally speaking, when inside a `PJ_*.c` file you will only be able to cause side effects in code within that file.

PR #900, opened 2018-03-25
Rename PACKAGE_NAME: PROJ.4 Projections -> PROJ

mwtoews:
Both autoconf and CMake defined PACKAGE_NAME as "PROJ.4 Projections", which I've modified to simply "PROJ". I'm not sure if there are any possible consequences for this change. The ones that I've noted are some defines in src/proj_config.h, and a few improved messages from "make check". I suspect the file name src/org_proj4_Projections.h is derived from the former PACKAGE_NAME.

I've left PACKAGE_NAME in proj.spec as "proj", as it was previously already "PROJ" in 2004-05-17 (see ChangeLog). Is file even used? It's a metadata file for "FreeGIS CD".

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2,7 +2,7 @@ dnl Process this file with autoconf to produce a configure script.
 
 AC_PREREQ(2.59)
 AC_INIT([PROJ], [5.0.0],
-        [https://github.com/OSGeo/proj.4/issues], proj)
+        [https://github.com/OSGeo/proj.4/issues], proj, [http://proj4.org])
 AC_CONFIG_MACRO_DIR([m4])
 AC_LANG(C)
 
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,8 @@
 dnl Process this file with autoconf to produce a configure script.
 
 AC_PREREQ(2.59)
-AC_INIT([PROJ.4 Projections], 5.0.0, [https://github.com/OSGeo/proj.4/issues], proj)
+AC_INIT([PROJ], [5.0.0],
+        [https://github.com/OSGeo/proj.4/issues], proj)
 AC_CONFIG_MACRO_DIR([m4])
 AC_LANG(C)
 
--- a/cmake/Proj4Config.cmake
+++ b/cmake/Proj4Config.cmake
@@ -31,8 +31,8 @@ check_library_exists(m ceil "" HAVE_LIBM)
 
 set(PACKAGE "proj")
 set(PACKAGE_BUGREPORT "https://github.com/OSGeo/proj.4/issues")
-set(PACKAGE_NAME "PROJ.4 Projections")
-set(PACKAGE_STRING "PROJ.4 Projections ${${PROJECT_INTERN_NAME}_VERSION}")
+set(PACKAGE_NAME "PROJ")
+set(PACKAGE_STRING "PROJ ${${PROJECT_INTERN_NAME}_VERSION}")
 set(PACKAGE_TARNAME "proj")
 set(PACKAGE_VERSION "${${PROJECT_INTERN_NAME}_VERSION}")
 
--- a/cmake/Proj4Config.cmake
+++ b/cmake/Proj4Config.cmake
@@ -34,6 +34,7 @@ set(PACKAGE_BUGREPORT "https://github.com/OSGeo/proj.4/issues")
 set(PACKAGE_NAME "PROJ")
 set(PACKAGE_STRING "PROJ ${${PROJECT_INTERN_NAME}_VERSION}")
 set(PACKAGE_TARNAME "proj")
+set(PACKAGE_URL "http://proj4.org")
 set(PACKAGE_VERSION "${${PROJECT_INTERN_NAME}_VERSION}")
 
 configure_file(cmake/proj_config.cmake.in src/proj_config.h)


kbevers:
I don't think this has any consequences either. But this is not my area of expertise.
@cffk @sebastic, what do you guys say to this? Is it safe to make these changes?

I believe `src/org_proj4_Projections.h` is somehow used in the Java bindings, although I don't understand how that all works. It hasn't been touched in 14 years. `src/org_proj4_PJ.h` seems to be related to JNI as well.

`proj.spec` is used for generating RPM files for RedHat Linux. I have no idea if it is used by anyone. I don't think we have a contact with the RedHat community so I don't know how to find out apart from just removing the file. I'll ask on the mailing list and see if anyone needs it.

sebastic:
@devrimgunduz maintains the proj package for Fedora, they use their own [proj.spec](https://src.fedoraproject.org/rpms/proj/blob/master/f/proj.spec) file.

kbevers:
Thanks, Bas. I figured something like that would be the case. It is probably the same for other distros using RPM's. I'll add it to my list of things to get rid of.

@devrimgunduz Thanks for making the PROJ packages for Fedora! In case you don't know, we are releasing version 5.0.1 in roughly a week. I don't know how your strategy regarding package updates is, but in case you are aiming for the most up to date packages a new version will soon be ready. Just a heads up :-)

configure.ac
```diff
@@ -1,7 +1,8 @@
 dnl Process this file with autoconf to produce a configure script.
 
 AC_PREREQ(2.59)
-AC_INIT([PROJ.4 Projections], 5.0.0, [https://github.com/OSGeo/proj.4/issues], proj)
+AC_INIT([PROJ], [5.0.0],
+        [https://github.com/OSGeo/proj.4/issues], proj)
```
kbevers: As per [this](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Initializing-configure.html) you could also add the project website at the end of the `AC_INIT` call, so it becomes:

```
AC_INIT([PROJ], [5.0.0],
        [https://github.com/OSGeo/proj.4/issues], proj, http://proj4.org)
```

mwtoews:
I've added the corresponding `PACKAGE_URL` variable for CMake too. It was already correctly set for `proj.spec`.

sebastic:
Regarding the JNI wrapper, @desruisseaux is the person to talk to, its origin was to support https://github.com/opengeospatial/geoapi/tree/master/geoapi-proj4 according to the comments in the source.

The Debian package builds the Java wrapper as it was requested for JGRASS in [Debian Bug #293803](https://bugs.debian.org/293803), but according to the [popcon data](https://qa.debian.org/popcon.php?package=proj) it doesn't have any actual users.

[`ant`](http://ant.apache.org/) generates the headers and the jar using the `build.xml`, I don't see any actual references to `org_proj4_Projections.h` in any code, though. Unless the Java wrapper will be updated to support the new API, it should probably be removed in PROJ 6.0.0.

The initial changes to `AC_INIT` in `configure.ac` didn't break the JNI build, so it should be fine to merge these changes.

kbevers:
> Unless the Java wrapper will be updated to support the new API, it should probably be removed in PROJ 6.0.0

I should look closer into this stuff, really. I've been thinking that it would be able to survive until 7.0.0, but now I am not so sure anymore. In any case, it will be removed with either version 6 or 7 if we can't find someone to maintain it. I have no idea about what impact that might have. I would hope that Java applications that needs a transformation library has already ditched PROJ JNI for the Apache SIS library, which seem to be a much better in that environment.

> The initial changes to AC_INIT in configure.ac didn't break the JNI build, so it should be fine to merge these changes.

And following that I guess it doesn't break anything else?

sebastic:
> And following that I guess it doesn't break anything else?

The `AC_INIT` changes just set some `m4` macros, that should be fairly safe. The only visible change was in the `gie` output:
```diff
 ============================================================================
-Testsuite summary for PROJ.4 Projections 5.0.1
+Testsuite summary for PROJ 5.0.1
 ============================================================================
```

kbevers:
Awesome. Thanks for testing it. I am way too under-educated on both our build systems I must admit so I don't really trust my gut feeling on anything.

devrimgunduz:
Hi @kbevers.

I am maintaining PROJ in 2 repos:

1. Fedora / EPEL repo: We can push Proj 5 to only Fedora rawhide, if PROJ 5 is released early next week.

2. PostgreSQL RPM repo (for RH, CentOS, Fedora and SuSE): I already created proj49 (means 4.9) package there, and we are more flexible in that repo. Once 5.0 is out, I can add proj50 package, and rebuild all packages using this one.

Regards, Devrim

kbevers:
> if PROJ 5 is released early next week.

Actually PROJ 5 was released March 1st. We are pushing the first patch release April 1st (or there about) which will be version 5.0.1. It fixes a bunch of bugs found after testing 5.0.0 in the wild. A [release candidate](http://download.osgeo.org/proj/proj-5.0.1RC1.tar.gz) is out at the moment if you want to test it before we release.

Also note that the [proj-datumgrid](http://download.osgeo.org/proj/proj-datumgrid-1.7.tar.gz) packages has been updated. See the [release announcement](http://lists.maptools.org/pipermail/proj/2018-March/008098.html) for more info.

In case it is useful to you, [here](https://github.com/OSGeo/proj.4/milestones?direction=asc&sort=due_date&state=open) is an overview of the scheduled releases for the next couple of years.

devrimgunduz:
Apologies, I missed that release -- a bit behind my emails.

I'll work on packaging 5.0 as soon as I can (early next week)

Regards, Devrim

kbevers:
Don't worry, it's hard to keep up with everything. I suggest you wait until 5.0.1 is out. That would save you some time in the end. You can use the buggyness of 5.0.0 as the reason for not packaging that in the first place :)

desruisseaux:
I can volunteer for updating the JNI binding to the new API. Should we create a separated issue for that? I'm not sure it would be ready in time for the 5.0.1 release in April 1st however.

kbevers:
@desruisseaux, awesome. Thanks!

A new issue covering this would be good, yes. 5.0.1 is strictly a bug fixing release and a release candidate is already out so in any case it is not going in there. It would be a nice addition for 5.1.0 though. There's still a few months until that is released so no need to rush it.

Issue #802, opened 2018-02-13
Remove legacy CMake target name "proj"


kbevers:
In #789 @cffk lays out the plan to move PROJ into a namespace that follows current best practice. This ticket schedules part 3 of the plan to the PROJ 7.0.0 release where only one CMake target will be available onwards. This breaks backwards compatibility. The target is PROJ4::proj and PROJ4_LIBRARIES = PROJ4::proj.

kbevers:
@cffk I have assigned this ticket to you in case you want to change the wording to something more accurate. Feel free to unassign yourself if you are happy with the description above.

cffk:
@kbevers This is fine.  Feel free to remind me about this again when 7.0.0 is imminent.

kbevers:
You will be duly notified in the beginning of 2020 :)

mwtoews:
> You will be duly notified in the beginning of 2020 :)

Oh look, the future has finally arrived! @cffk are you OK to proceed with this?

cffk:
Sure, I'll undertake this.  It might also be a good time to revisit the
decision made a couple of years ago to stick with "PROJ4" as the
cmake project name.  Perhaps, we can make it that there's a
graceful transition from PROJ4 to PROJ.

kbevers:
Yes, it would be nice to smoothly transition into just "PROJ". I'll let you guys come with a suggestion for how to best to do that (you understand CMake a lot better than I do!).

Issue #801, opened 2018-02-13
Add namespace to CMake target proj


kbevers:
In https://github.com/OSGeo/proj.4/pull/789 @cffk lays out the plan to move PROJ into a namespace that follows current best practice. This ticket schedules part 2 of the plan to the PROJ 6.0.0 release where two targets, proj and PROJ4::proj, are defined; PROJ4_LIBRARIES = PROJ4::proj.

kbevers:
@cffk I have assigned this ticket to you in case you want to change the wording to something more accurate. Feel free to unassign yourself if you are happy with the description above.

kbevers:
@cffk Release time is coming up, can I persuade you to taking care of this issue within the next couple of days?

PR #901, opened 2018-03-25
Rename CMake infrastructure to remove 4

mwtoews:
A few CMake files were renamed and edited to remove the former major version number from them. Also a few old "[SOCI](https://github.com/SOCI/soci)" references were updated.


kbevers:
Does this in any way interfere with #801 and #802?

mwtoews:
Good point, I didn't consider these open issues. @cffk ?

As for `PROJ4_LIBRARIES` and `PROJ4_INCLUDE_DIRS` (which are not touched in this PR), they are only used in docs/source/development/cmake.rst. I only see `PROJ_LIBRARIES` used in various *.cmake files.

kbevers:
I am sure most of this is fine. But I am worried that this `project(PROJ4 C)` -> `project(PROJ C)` could break something.

cffk:
This PR breaks backward compatibility for cmake users.  The discussion lead by @kbevers a couple of months ago ended up deciding the although "proj" would be the official name of this project, proj4 would be retained in some contexts, e.g., proj4.org.  In particular, I liked the idea that

    find_project(PROJ4)
    target_link_libraries(${PROJ4_LIBRARIES})

 still "worked".  The change to CMakeLists.txt in this PR

    project(PROJ C)

breaks this.

mwtoews:
The reference to the previous discussion where PROJ4 would be retained for CMake is #735

Question from the roll out plain in #789 : does the CMake `PROJECT-NAME` ever change to "PROJ"? And if so, at which release number?

cffk:
In the plan described in #789, it was envisioned that the `PROJECT-NAME` stay as `PROJ4` indefinitely.  You could devise a separate plan to migrate from `PROJ4` to `PROJ` where for some period both names were supported.  I'm not sure this is necessary?

mwtoews:
@cffk I've restored the previous `PROJECT-NAME` as `PROJ4` with a note for future readers. As far as I can see the only sensitive CMake [variable](https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html#id2) that needs attention is [`<PROJECT-NAME>_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html) as `PROJ4_SOURCE_DIR`. I'm not sure if there are other CMake variables that provide provide information, and need to reference `PROJECT-NAME`. Is the CMake namespace restored with the previous commit? Any other concerns?

Also removed `${PROJ4_SOURCE_DIR}/cmake/modules`, which was another SOCI leftover (from `soci/cmake/modules/`) not used in this project.

PR #912, opened 2018-03-30
Tests and bugfix for the Orthographic projection

kbevers:
Test cases from Snyder added to gie test suite. While doing this I discovered that erroneous points where returned as ``(0,0)`` instead of the expected ``(HUGE_VAL, HUGE_VAL)``. This is also fixed in this PR. A error message is also added to the trace log when such transient errors occur.

### DIFF
--- a/src/PJ_ortho.c
+++ b/src/PJ_ortho.c
@@ -1,6 +1,7 @@
 #define PJ_LIB__
 #include <errno.h>
 #include "proj.h"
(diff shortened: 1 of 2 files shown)


kbevers:
98.67% test coverage. I guess that's as good as it gets.

@busstoptaktik I need an option in gie that lets me hit those ``return pj_default_destructor(P, ENOMEM);`` lines :-)

PR #914, opened 2018-04-01
Add a CITATION file (fixes #309)

### LINKED ISSUE #309: How to reference the proj.4 library
I'm writing a scientific journal paper at the moment in which I use [Iris](http://scitools.org.uk/iris/) to re-calculate atmospheric wind data in a rotated coordinate system (i.e. one where the north pole has been moved). In the methods section of the paper I list the major software packages I used for the research, including any references to academic papers written about that software (e.g. many authors of scientific software are publishing with journals like the _Journal of Open Research Software_ so they can get academic credit for their work). Since Iris uses proj.4 for grid rotation, I'd like to reference proj.4 if possible. Is there an appropriate reference that I could cite?

rouault:


kbevers:
That Zenodo doi thing looks cool too. Should we set that up as well?

There is also the [paper on transformation pipelines](http://www.fig.net/resources/proceedings/fig_proceedings/fig2017/papers/iss6b/ISS6B_evers_knudsen_9156.pdf) that Thomas and I wrote. It is probably the closest we get to a proper academic reference at the moment, although that isn't saying much...

kbevers:
Also, shouldn't `CITATION` be added to at least the autoconf makefiles since it is referenced by `README`?

rouault:
> There is also the paper on transformation pipelines that Thomas and I wrote

It is cited among other relevant ones (such as G. Evenden ones or @cffk ones) in http://proj4.org/references.html 

>  shouldn't CITATION be added to at least the autoconf makefiles since it is referenced by README?

Done

kbevers:
> it is cited among other relevant ones (such as G. Evenden ones or @cffk ones) in http://proj4.org/references.html

Yes, I was just thinking if it made sense as a the paper to reference. Probably not.

> Done

Thanks

rouault:
I'm completely a stranger to academics habits and customs, so I won't take any strong position on this. It seems to me that the purpose was to credit the software and its numerous contributors in a short way. Of course there are a number of works and papers that have lead to it but we cannot quote all the centenial long publications on geodesy and cartography ;-)  (or could we ? like https://physics.aps.org/featured-article-pdf/10.1103/PhysRevLett.116.061102 and its 6 page long autorship :-)) I guess someone writing a paper where PROJ is used could/should cite PROJ itself with the above citation, and if he uses a particular aspect of it, cite the appropriate paper that implements the details he use (so perhaps we could add a link in CITATION to http://proj4.org/references.html so they are aware of those publications related to PROJ)
We could extend the authorship a bit with the most significant contributors: "G. Evenden, F. Warmerdam, C. Karney, K. Evers, T. Knudsen et al. PROJ contributors"

mwtoews:
I'd consider R's [CITATION](http://www.blopig.com/blog/2013/07/citing-r-packages-in-your-thesispaperassignments/) the gold standard, as I see it used [frequently in literature](https://scholar.google.com/scholar?q=R), which is impressive since it's a one letter software. Following their lead (and removing `organization`/`address`, modifying `url`):
```
To cite PROJ in publications use:

  PROJ contributors (2018). The PROJ coordinate transformation software
  library. URL http://proj4.org/.

A BibTeX entry for LaTeX users is

  @Manual{,
    title = {The {PROJ} coordinate transformation software library},
    author = {{PROJ contributors}},
    year = {2018},
    url = {http://proj4.org/},
  }
```

kbevers:
I like the way R does it. Great example.

I think OSGeo should be used as the organization so that it instead becomes:

```
To cite PROJ in publications use:

  PROJ contributors (2018). The PROJ coordinate transformation software
  library. The Open Source Geospatial Foundation. URL http://proj4.org/.

A BibTeX entry for LaTeX users is

  @Manual{,
    title = {The {PROJ} coordinate transformation software library},
    author = {{PROJ contributors}},
    organization = {The Open Source Geospatial Foundation},
    year = {2018},
    url = {http://proj4.org/},
  }
```

The address seems to be related to the organization but it is not entirely clear to me if that is the case. If it is, there's a postal address for OSGeo [here](https://www.osgeo.org/about/contact/) that could used.

How is version numbers dealt with in this? There's a hell of a difference if you are using version 4.9.3 and some future version 7 for instance. 

Is the year static? Or is that to be updated with every major release?

rouault:
CITATION file updated with latest @kbevers 's content

CITATION
```diff
@@ -0,0 +1,14 @@
...
+
+A BibTeX entry for LaTeX users is
+
+  @Manual{,
+    title = {The {PROJ} coordinate transformation software library},
+    author = {{PROJ contributors}},
+    organization = {The Open Source Geospatial Foundation},
+    year = {2018},
+    url = {http://proj4.org/},
+  }
```
kbevers: There's a nasty special character at the end of the file here.

kbevers:
@cffk @busstoptaktik does this look reasonable to you?

CITATION
rouault: Ah was rather the lack of newline character. Fixed.

cffk:
Yes, this is fine.  I prefer "corporate authorship" to trying to keep a list of contributors up-to-date.

One possible addition: say how a specific version of the library should be cited, with a "version" field and with a more specific URL?

kbevers:
> Yes, this is fine. I prefer "corporate authorship" to trying to keep a list of contributors up-to-date.

Good point. Referring to the citation in the text would with the above be "... something PROJ-related (PROJ Contributors, 2018) ...". It is a bit unusual, but I guess it works.

> One possible addition: say how a specific version of the library should be cited, with a "version" field and with a more specific URL?

[Here](https://superuser.com/a/900254) is an example of using the "software" BibTex category instead of the "manual" category as above. For PROJ this would be something like:

```
@software{PROJ,
  author = {{The Open Source Geospatial Foundation}},
  title = {PROJ},
  url = {https://proj4.org},
  version = {5.0.1},
  date = {2018-04-01},
}
```

I think actually this is preferred since we still haven't got a stable manual that can be tied to each release (I hope we will have that with version 6, but let's see).

mwtoews:
For these BiBTeX  entries, `@manual` is the normal [entry type](https://en.wikipedia.org/wiki/BibTeX#Entry_types) for this sort of thing. `@software` is not common, and is effectively an alias for `@misc` where most of the fields like `version` are ignored when typesetting. Also `date` is not recognized. If you want to see the differences between the entry types on a PDF, take a look at a quick example I made on [Overleaf](https://www.overleaf.com/read/wzbjxrfxfkwd).

Similar to what R does, only `year` should be updated based on the year of the release, and the version shouldn't matter in the reference. Also, `HOWTO-RELEASE` should be amended to check the `year` entries in `CITATION`.

Another nitpick, remove the leading "The" from `author` (they are not called "The OSGeo") and also `title` (it's not called "The PROJ"). I wasn't 100% sure about choosing OSGeo as either author or organization, since PROJ is still not an OSGeo Project; it's still a community project.

I'm not convinced a DOI will add any value that a URL already offers.

kbevers:
> For these BiBTeX entries, @manual is the normal entry type for this sort of thing. @software is not common, and is effectively an alias for @misc where most of the fields like version are ignored when typesetting. Also date is not recognized. If you want to see the differences between the entry types on a PDF, take a look at a quick example I made on Overleaf.

Alright. My knowledge on this stuff is quite limited. The "software" option just seemed like the right thing to do, but if it isn't really used it is better to go with the "manual" option.

So we update the year when new releases are issued and then it is up to whoever uses the reference to note exactly which version of the software was used? Maybe that should be explained in `CITATION` as well.

> I wasn't 100% sure about choosing OSGeo as either author or organization, since PROJ is still not an OSGeo Project; it's still a community project

That is a good point. To me, and likely many others, the lines are a bit blurred here.  It should be checked wether it is okay or not to brand PROJ with the OSGeo name in this situation. What is the correct forum to ask that sort of thing? The discuss list? The OSGeo board list? This might also be a good occasion for the first small steps towards becoming a proper OSGeo project instead of the confusing setup we have today with MetaCRS.

rouault:
Can we merge this as it ? This can be later tuned...

PR #924, opened 2018-04-05
Logging with new api

kbevers:
This moves `proj_log_level` and `proj_log_func` to the public API in `proj.h`. 

I have also added the use of the PROJ logging functionality to cct. Mostly because I wanted to make sure that the logging functions worked as I expected but also to have a demonstration of how to use these API-functions. It may not make much sense for a simple app as cct. Let me know if you think this is stupid or if it should be done smarter. This is just a quick thing I did while spending a few hours on an airplane...

~~I still need to add documentation for the `proj_log_*` functions.~~


src/proj.h
```diff
@@ -268,6 +268,17 @@ struct PJ_INIT_INFO {
     char        lastupdate[16];     /* Date of last update in YYYY-MM-DD format */
 };
 
+typedef enum PJ_LOG_LEVEL {
+    PJ_LOG_NONE  = 0,
+    PJ_LOG_ERROR = 1,
+    PJ_LOG_DEBUG = 2,
+    PJ_LOG_TRACE = 3,
+    PJ_LOG_TELL  = 4,
+    PJ_LOG_DEBUG_MAJOR = 2, /* for proj_api.h compatibility */
```
kbevers, line 277: Would it perhaps be better to `#define` these in `proj_api.h` instead?

PR #925, opened 2018-04-07
Add a dedicated proj=webmerc operation

rouault:
Directly based on the auxiliary sphere type parameter of
http://desktop.arcgis.com/en/arcmap/10.3/guide-books/map-projections/mercator.htm

Implemented:
- 0: use semimajor axis or radius of the geographic coordinate system (WebMercator style)
- 1: use semiminor axis or radius
- 2: calculate and use authalic radius

Unimplemented:
- 3: use authalic radius and convert geodetic latitudes to authalic latitudes


rouault:
Was also wondering if it would be worth to have a +proj=gmerc that would be an alias for +proj=merc +aux_sphere_type=0 ?

kbevers:
This is nice and definitely a better solution to what we have now! This was on my list of things to do in the future so thanks for taking care of it.

> Was also wondering if it would be worth to have a +proj=gmerc that would be an alias for +proj=merc +aux_sphere_type=0 ?

That would have been my approach. I think it is a cleaner solution although it is a little more work to add a new projection. 
I would suggest not making it an alias but instead a "proper" projection in it's own right. As far as I can tell it would be fairly simple and mostly just requires an added  `PROJECTION` setup function. I would add the auxiliary sphere stuff to that and keep it out of the Mercator setup. `+lon_0=0`, `+lat_ts=0` and `+k_0=1` could then also be fixed, simplifying the user interface. Maybe what you meant was along the lines of that solution - if so, please disregard the previous paragraph.

I think calling it `webmerc` or `pseudomerc` is more describing. Also comes with the benefit of being less tied to the name of a large company.

src/PJ_merc.c
```diff
@@ -48,11 +49,36 @@ static LP s_inverse (XY xy, PJ *P) {           /* Spheroidal, inverse */
     return lp;
 }
 
+/* Cf http://desktop.arcgis.com/en/arcmap/10.3/guide-books/map-projections/mercator.htm */
+#define AUX_SPHERE_TYPE_SEMI_MAJOR 0
```
kbevers: Why not use a enum here? You could use a switch below instead of the else if's. I think that is cleaner.

rouault:
> I would add the auxiliary sphere stuff to that and keep it out of the Mercator setup

I really tried to implement ESRI stuff, which as far as I can see (from the doc ! not tested against Arc), is general mercator with all the the bells and whistles, and the ability to define how to compute the spherical radius

src/PJ_merc.c
```diff
@@ -48,11 +49,36 @@ static LP s_inverse (XY xy, PJ *P) {           /* Spheroidal, inverse */
...
+#define AUX_SPHERE_TYPE_SEMI_MINOR 1
+#define AUX_SPHERE_TYPE_AUTHALIC_RADIUS 2
+/* Unimplemented #define AUX_SPHERE_TYPE_AUTHALIC_LATITUDE 3 */
```
kbevers: healpix gives some hints as to how to implement this: https://github.com/OSGeo/proj.4/blob/master/src/PJ_healpix.c#L225-L246

src/PJ_merc.c
rouault: Done

rouault:
> Do you know of any other case, apart from the webmercator, where this is used?

No... Perhaps @melitakennedy could comment on the use cases for auxiliary_sphere_type=1/2/3 for the Mercator projection ?

src/PJ_merc.c
```diff
@@ -76,3 +119,22 @@ PJ *PROJECTION(merc) {
     return P;
 }
 
+PJ *PROJECTION(webmerc) {
+
+    if( pj_param(P->ctx, P->params, "tk_0").i ) {
```
kbevers: This behaviour is different to the rest of PROJ. Usually you are allowed to set a parameter even if it is not used. If it is not used by the projection it will just get ignored. `proj` will tell you if that is the case:

```
$ proj -v +proj=merc +lat_1=23
#Mercator
#       Cyl, Sph&Ell
#       lat_ts=
# +proj=merc +ellps=WGS84
#--- following specified but NOT used
# +lat_1=23
```

I think it would be better to keep this "tradition". Just make sure to override them in the setup with `P->k0=1.0; P->lam0=0.0`.

test/gie/4D-API_cs2cs-style.gie
```diff
@@ -215,6 +215,32 @@ accept      487147.594520173    4934316.46263998    0
...
+operation   proj=pipeline step init=epsg:26915 inv step proj=webmerc datum=WGS84
+-------------------------------------------------------------------------------
+tolerance   20 cm
+accept      487147.594520173    4934316.46263998
+expect      -10370728.80        5552839.74
+
+accept      487147.594520173    4934316.46263998    0
+expect      -10370728.80        5552839.74          0
+-------------------------------------------------------------------------------
+
```
kbevers, line 230: How about some tests for `aux_sphere_type` == 1 and 2?

kbevers:
Another thing, if the `epsg` file is updated to use `+proj=webmerc` we can re-instate the fix to https://github.com/OSGeo/proj.4/issues/22. Would it cause problems downstream if we change the `epsg`-file in PROJ? Wil that have an effect on GDAL, QGIS, etc?

src/PJ_merc.c
rouault: I kept the ones for k_0 and lat_0 since those are done in generic parts. I remove the one for lat_ts (which was wrongly spelled in my original commit)

test/gie/4D-API_cs2cs-style.gie
rouault: Added, by the way while adding them I had originally issues with the z component being modified, which I think was wrong. I fixed that issue with commit https://github.com/OSGeo/proj.4/pull/925/commits/ac78787bf3bfa15dabbbad0042ca762020542365 which I believe is a bugfix that should probably be applied to the 5.0 branch

rouault:
> Would it cause problems downstream if we change the epsg-file in PROJ? Wil that have an effect on GDAL, QGIS, etc?

I'm not completely sure. GDAL should probably not parse itself the file, and QGIS probably not too. But the epsg file is generated by GDAL that for now will still generate the old hacky definition of web mercator. And we should expect that old hacky definition to be still passed for years !

rouault:
By the way: this had apparently no impact on spherical Mercator but pj_calc_ellipsoid_params() behaviour is quite surprising: it doesn't recompute e, f and b is they are originally non zero ! I see that in ellps_spherification() there is a 'workaround' for that which I copied, but this feels messy

src/PJ_merc.c
kbevers: I understand why you would want to do that but I still don't think it is the correct way to handle it. I think it will cause more confusion than is necessary. This would be the only place in the code base where using one of the default projection parameters would cause an error. There are plenty [examples](https://github.com/OSGeo/proj.4/blob/master/src/PJ_stere.c#L295-L314) of [overriding](https://github.com/OSGeo/proj.4/blob/master/src/PJ_nzmg.c#L108-L113) default [parameters](https://github.com/OSGeo/proj.4/blob/master/src/PJ_krovak.c#L185-L186) in the setup without raising errors.

I think it would be awesome if we had a system that caused errors when using parameters that can't be used with a given operation but that is a problem for another day (and major version release :-) ).

rouault: OK, checks reverted and parameters overriden

test/gie/4D-API_cs2cs-style.gie
kbevers: > Added, by the way while adding them I had originally issues with the z component being modified, which I think was wrong. 

Do you mind writing that up in an issue so it can be referred to in the release notes later on? 

> I fixed that issue with commit 1a575b6 which I believe is a bugfix that should probably be applied to the 5.0 branch

I don't plan on doing more bug fix release for the 5.0 branch. 5.1 will be out in two months. I think that will be fine.

kbevers:
> I'm not completely sure. GDAL should probably not parse itself the file, and QGIS probably not too. But the epsg file is generated by GDAL that for now will still generate the old hacky definition of web mercator. And we should expect that old hacky definition to be still passed for years !

Alright. Do you plan on updating GDAL so it generates the epsg file using `webmerc`? 
Well yes, but we can't keep supporting that hack forever! It would be nice if we could issue a warning when the hack is used. Not sure if we have a smart way of doing that at the moment though...

> By the way: this had apparently no impact on spherical Mercator but pj_calc_ellipsoid_params() behaviour is quite surprising: it doesn't recompute e, f and b is they are originally non zero ! I see that in ellps_spherification() there is a 'workaround' for that which I copied, but this feels messy

It does indeed. I wonder what the reason for that is. Can you answer that @busstoptaktik ? It doesn't affect the tests (apart from one minor detail in the gie `pj_cart_selftest` function) if changed to not respect the original values.

rouault:
> Do you plan on updating GDAL so it generates the epsg file using webmerc?

Potentially, the exportToProj4() code could, once proj 5.1 is released and detected at runtime, export EPSG:3857 as webmerc.

> we can't keep supporting that hack forever!

The hack is only in the pj_transform() legacy interface where WGS84 has always (and will remain until we remove it) the pivot datum, so I don't think that's a big issue.

kbevers:
> The hack is only in the pj_transform() legacy interface where WGS84 has always (and will remain until we remove it) the pivot datum, so I don't think that's a big issue.

That's true. I was thinking about the `+nadgrids=@null` part of the hack but we seem to have that covered in the cs2cs emulation functions so that is not a problem.

melitakennedy:
Although I believe "web Mercator" was the instigator, Esri added several "Auxiliary Sphere" versions of existing projections. At the time, we were trying to match results from NGA's GEOTRANS and ERDAS. They both had several projections that were implemented as sphere-only, but Esri had ellipsoidal versions. One or the other (I don't remember which) would internally calculate the authalic radius and use that. Esri's default is to use the semimajor axis. 

The developer, David Burrows, pointed out that for an equal-area sphere-only projection, calculating the authalic radius is not quite enough. The geodetic latitudes should be converted to authalic latitudes as well. As far as I know, no one using any of this except Mercator aux sphere.

kbevers:
@melitakennedy Thanks for your insight. 

It sounds to me the "Auxiliary Sphere" doesn't bring anything new to PROJ. We already have both ellipsoidal and spherical versions of the Mercator projection and with the introduction of `webmerc` we can also handle the Webmercator satisfactorily.

@rouault Based on the above I suggest removing the aux. sphere stuff. Do you agree with that sentiment?

rouault:
> Based on the above I suggest removing the aux. sphere stuff. Do you agree with that sentiment?

OK, I've just kept the new webmerc and dropped the aux_sphere_type option of merc

kbevers:
> OK, I've just kept the new webmerc and dropped the aux_sphere_type option of merc

Thanks!

PR #928, opened 2018-04-11
Enhance the precision of Spherical Mercator projection near the equator

jgoizueta:
This uses a linear approximation of tan(x+pi/4) for better precision at small latitudes.

Before this patch:

```
cs2cs +init=epsg:4326 +to +init=epsg:3857  -v -f "%.9e" <<EOF
0 0
0 1E-14
0 1E-13
0 1E-12
0 1E-11
0 1E-10
0 1E-9
0 1
EOF
0.000000000e+00	-7.081154552e-10 0.000000000e+00
0.000000000e+00	1.416230910e-09 0.000000000e+00
0.000000000e+00	9.913616372e-09 0.000000000e+00
0.000000000e+00	1.104660110e-07 0.000000000e+00
0.000000000e+00	1.113157496e-06 0.000000000e+00
0.000000000e+00	1.113015872e-05 0.000000000e+00
0.000000000e+00	1.113199982e-04 0.000000000e+00
0.000000000e+00	1.113251429e+05 0.000000000e+00
```

After: 

```
./src/cs2cs +init=epsg:4326 +to +init=epsg:3857  -v -f "%.9e" <<EOF
0 0
0 1E-14
0 1E-13
0 1E-12
0 1E-11
0 1E-10
0 1E-9
0 1
EOF
0.000000000e+00	0.000000000e+00 0.000000000e+00
0.000000000e+00	1.416230910e-09 0.000000000e+00
0.000000000e+00	1.132984728e-08 0.000000000e+00
0.000000000e+00	1.118822419e-07 0.000000000e+00
0.000000000e+00	1.113157496e-06 0.000000000e+00
0.000000000e+00	1.113157496e-05 0.000000000e+00
0.000000000e+00	1.113199982e-04 0.000000000e+00
0.000000000e+00	1.113251429e+05 0.000000000e+00
```

All values are now more accurate (I checked using arbitrary precision arithmetic), and we have the nice property of preserving the (0,0) coordinates of the origin point.

### DIFF
--- a/src/PJ_merc.c
+++ b/src/PJ_merc.c
@@ -9,11 +9,31 @@ PROJ_HEAD(webmerc, "Web Mercator / Pseudo Mercator") "\n\tCyl, Sph\n\t";
 
 #define EPS10 1.e-10
 
-static double _tan_near_fort_pi(double x) {
+#if !defined(HAVE_C99_MATH)
+#define HAVE_C99_MATH 0
+#endif
+
+#if HAVE_C99_MATH
+#define log1px log1p
+#else
+static double log1px(double x) {
+  volatile double
+    y = 1 + x,
+    z = y - 1;
+  /* Here's the explanation for this magic: y = 1 + z, exactly, and z
+   * approx x, thus log(y)/z (which is nearly constant near z = 0) returns
+   * a good approximation to the true log(1 + x)/x.  The multiplication x *
+   * (log(y)/z) introduces little additional error. */
+  return z == 0 ? x : x * log(y) / z;
+}
+#endif
+
+static double logtanpfpim1(double x) {       /* log(tan(x/2 + M_FORTPI)) */
     if (fabs(x) <= DBL_EPSILON) {
-        return 2*x + 1.0;
+        /* tan(M_FORTPI + .5 * x) can be approximated by  1.0 + x */
+        return log1px(x);
     }
-    return tan(M_FORTPI + x);
+    return log(tan(M_FORTPI + .5 * x));
 }
 
 static XY e_forward (LP lp, PJ *P) {          /* Ellipsoidal, forward */
@@ -35,7 +55,7 @@ static XY s_forward (LP lp, PJ *P) {           /* Spheroidal, forward */
         return xy;
 }
     xy.x = P->k0 * lp.lam;
-    xy.y = P->k0 * log(_tan_near_fort_pi(.5 * lp.phi));
+    xy.y = P->k0 * logtanpfpim1(lp.phi);
     return xy;
 }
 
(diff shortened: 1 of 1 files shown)


kbevers:
> Perhaps now is a good time to revisit the Mercator formulas.

Now is as good a time as any. What about the inverse formulas? There's an `atan(exp(x))` thing going on. Presumably that can also be done better numerically.

jgoizueta:
I won't get into larger changes like using `atanh` at the moment (I think it would required further analysis) but I think using `logp1` is worth and straightforward to include here.

Looking at the inverse formula would also be interesting, but similarly I won't get
into analyzing `tan(exp(x))` now, but will look instead at the cancellation near Y=0 for consistency with the other changes.

Thanks @cffk & @kbevers!

kbevers:
Thanks for giving this a go @jgoizueta. We're at the limit of my numerical math capabilities here so I'll defer to @cffk's judgement on this. Are you happy with this approach, Charles?

kbevers:
Looking a bit more in to this it turns out that we have a `log1px` defined already in the extended transverse mercator:
https://github.com/OSGeo/proj.4/blob/master/src/proj_etmerc.c#L69-L78

There's even a `asinh` function as well:
https://github.com/OSGeo/proj.4/blob/master/src/proj_etmerc.c#L84-L88

We are getting close to the point were a dedicated non-C99-math function collection is needed.

cffk:
I agree...  Perhaps we should let this modification go through but queue of a bigger survey of these hygene + accuracy issues.

Note well that several conformal projections use similar formulas (converting between geographic latitude to conformal latitude so that a spherical conformal projection can be used).  Off the top of my head, I suspect that

  * Mercator
  * transverse Mercator
  * polar stereographic
  * Lambert conformal conic

are all candidates for a uniform treatment.

Resources are [my paper on transverse Mercator](https://arxiv.org/abs/1002.1417), eqs (3), (4).  Here the treatment works for arbitrary eccentricity.  [Engsager and Poder](https://icaci.org/files/documents/ICC_proceedings/ICC2007/documents/doc/THEME%202/oral%201/2.1.2%20A%20HIGHLY%20ACCURATE%20WORLD%20WIDE%20ALGORITHM%20FOR%20THE%20TRANSVE.doc) (and etmerc) use a trigonometric series for the transformation.

In addition to a C99 math library collection, perhaps it would be good to consider a centralized location for Clenshaw and Horner sums, and the basic ellipsoidal Mercator formulas (for use by etmerc etc.)

The various equal-area projections should also use the same machinery (if they don't already) to do the transformation between geographic latitude and authalic latitude.  See also my table of [trigonometric series](https://geographiclib.sourceforge.io/html/auxlat.html) for transformations between all the various latitudes (E+K's series appear here &phi; &minus; &chi; and &chi; &minus; &phi;).

kbevers:
> I agree... Perhaps we should let this modification go through but queue of a bigger survey of these hygene + accuracy issues.

I have merged this (thanks @jgoizueta) and opened for #930 for further discussion on the generic solution to the math functions problem.

PR #931, opened 2018-04-12
Enhance precision of inverse Spherical Mercator projection near the equator

jgoizueta:
This is for consistency with #928


src/PJ_merc.c
```diff
@@ -36,6 +44,14 @@ static double logtanpfpim1(double x) {       /* log(tan(x/2 + M_FORTPI)) */
     return log(tan(M_FORTPI + .5 * x));
 }
 
+static double hpip2atanexp(double x) {       /* M_HALFPI - 2. * atan(exp(x)) */
+    if (fabs(x) <= DBL_EPSILON) {
+        /* atan(exp(x))  can be approximated by (exp(x)-1)/2 + pi/4 */
+        return -expm1x(x));
```
kbevers: The extra `)` at the end here causes the build failure.

kbevers:
I wouldn't call ~200 m small deviations :-) It seems to be a matter of getter the sign correct on the northing. Probably just a missed minus somewhere.

kbevers:
Superseded by #1105

Issue #930, opened 2018-04-12
Create a collection of math function for use non-C99 with compilation


kbevers:
We are relying more and more on numerical math functions that are only available in C99 and newer C standards. So far we have added functions when needed, mostly local to the code that needed the function, although some functions such as`hypot` are defined across all of the PROJ code base. In 
In https://github.com/OSGeo/proj.4/pull/928 `log1px` was added locally. The same function also exists in `PJ_etmerc.c`. There are likely more examples of this but these were the first that came to mind.

Instead of having these function spread across the code base we should gather them in the same file. Potentially with it's own header file as well, making it obvious that we are using C99 math functions. For example `pj_math.c`/`pj_math.h`. Ideally all the `HAVE_C99_MATH` logic can take place in `pj_math.h` which should then be included when using functions like `hypot`, `log1px`, etc.

So far this is just some loose ideas based on the discussion in #928. More input is welcome!

PR #937, opened 2018-04-14
Collect custom C99 math functions in proj_math.h

kbevers:
A first go at #930.

This PR gathers C99 functions in the same file and introduces the
`proj_math.h` header. The build system checks for C99 math functions. If
not found the `proj_math.h` header make sure that C99 functions are
defined as their pj_ equivalent.

Ideally `proj_math.h` is included instead of math.h. This removes the need
for any checks against `HAVE_C99_MATH` in the code making it easier to
read. It also makes life easier for contributors not accustomed to C89 quirks.

For this commit the functions `hypot`, `log1p` and `asinh` has been taken care
of. Other potential candidates for the `proj_math.h` header are likely to be found in the code base.


src/pj_math.c
```diff
@@ -0,0 +1,72 @@
...
+        return (x);
+    if ( x < y ) {
+        x /= y;
+        return ( y * sqrt( 1. + x * x ) );
+    } else {
+        y /= x;
+        return ( x * sqrt( 1. + y * y ) );
+    }
+}
+
```
cffk: Surely, it's better to let the cases of zero `x` or `y` to be handled by the general case.  Also the function might return `-0` which would lead to unexpected results, e.g., with `atan2(hypot(x,y),z)`.  So I suggest

    x = fabs(x);
    y = fabs(y);
    if (x < y) {
      ...

src/pj_math.c
```diff
@@ -0,0 +1,72 @@
...
+     * a good approximation to the true log(1 + x)/x.  The multiplication x *
+     * (log(y)/z) introduces little additional error. */
+    return z == 0 ? x : x * log(y) / z;
+}
+
+/* Compute asinh(x) accurately */
+double pj_asinh(double x) {
+    double y = fabs(x); /* Enforce odd parity */
+    y = log1p(y * (1 + y/(hypot(1.0, y) + 1)));
+    return x < 0 ? -y : y;
```
cffk: Perhaps

    return x > 0 ? y : (x < 0 ? -y : x);

to ensure `asinh(-0) = -0`

kbevers:
Thanks @cffk. I've updated the code to reflect your comments. What do you think about the `proj_math.h` header?  My idea was that it would be simpler this way where you can basically just include `proj_math.h` instead of `math.h`.

cffk:
Re: hypot.  Whoops, sorry.  Change

    y /= x;

to

    y /= (x ? x : 1.0);

kbevers:
That makes sense, yes. Let's see if that makes the tests pass.

cffk:
I assume that proj_math.h isn't installed?  (If it is, the worry is that
it gets included with a different setting for HAVE_C99_MATH.).

Other C99 math functions which are candidates for inclusion:

    isnan
    isfinite
    copysign
    atanh
    cbrt
    (expm1?)
    
Note that `geodesic.c` has its private version of some these functions.
This is necessary so that `geodesic.[ch]` make an independent package.
However, I suggest that `geodesic.c` detect when it's being compiled as
part of `proj` and then include `proj_math.h`.

kbevers:
> I assume that proj_math.h isn't installed?

No, it is not installed. This is for internal use only.

> Other C99 math functions which are candidates for inclusion

I think we should just add them as needed. The expm1 function might soon be necessary for the inverse spherical Mercator so it will be added soon I guess.

> However, I suggest that geodesic.c detect when it's being compiled as
part of proj and then include proj_math.h.

Good idea. I'll take a look at implementing that.

cffk:
Inverse Mercator could avoid needing `expm1` by using `sinh`.

Issue #423, opened 2016-09-30
Invalid return 1; in PJ_isea.c:isea_hex


schwehr:
https://github.com/OSGeo/proj.4/blob/6a921265db9ff12a263ff1a88118a69a65b7a2df/src/PJ_isea.c#L919

From discussion with Nathan Wagner nw hydaspes if.org:

> I'm looking into it.  ... the code after the return is the right code,
> but maybe there's a bug or something that I missed.  My tests don't
> exercise that code path.

Just removing that code in #418 is not the correct fix.

direvus:
I'm speculating here, but the comments in the unreachable code talk about aperture 3 and 4:

```cpp
    /* Aperture 3 odd resolutions */
    if (g->aperture == 3 && g->resolution % 2 != 0) {
...
    }

    /* aperture 3 even resolutions and aperture 4 */
```

However the code below the comment "aperture 3 even resolutions and aperture 4" doesn't test the aperture, it just assumes that the aperture will meet those conditions if it reaches this line of code.

All of these clues lead me to think that the early `return` was originally meant to occur if the aperture is neither 3 nor 4, for example:

```cpp
if (g->aperture < 3 || g->aperture > 4)
    return;
```

I'm happy to raise a PR if this solution seems appropriate.

jerstlouis:
FYI @direvus @schweh I'm proposing to remove the `aperture`, `resolution` and `mode` parameters from PROJ as this has nothing to do with the projection itself, but rather with grid constructions which I would think are out of scope

See https://github.com/OSGeo/PROJ/issues/4389#issuecomment-3624139667

PR #943, opened 2018-04-20
Fix float-cast-overflow in PJ_isea.c:hexbin2

schwehr:
src/PJ_isea.c:79:10: runtime error: nan is outside the range of representable values of type 'int'

Found with autofuzz UndefinedBehaviorSanitizer

Input:
src: "++mode=di ++proj=isea +  ++step +stepp +"
dst: "++init=epsg:3249 +f=2src: \"++proj=naearth + +R_hj=nae=naea  +R_a +R_hj=natea0-src\\\" +axis=wde  +pr+R_h +oR_aejea"
x: 0
y: 3
z: 3

### DIFF
--- a/src/PJ_isea.c
+++ b/src/PJ_isea.c
@@ -66,13 +66,6 @@ void hexbin2(double width, double x, double y, int *i, int *j) {
     int ix, iy, iz, s;
     struct hex h;
 
-    if (pj_is_nan(x) || pj_is_nan(y) || pj_is_nan(width)) {
-        fprintf(stderr, "hexbin2 cannot handle NaN inputs\n");
-        *i = 0;
-        *j = 0;
-        return;
-    }
-
     x = x / cos(30 * M_PI / 180.0); /* rotated X coord */
     y = y - x / 2.0; /* adjustment for rotated X */
 
@@ -1044,6 +1037,12 @@ static XY s_forward (LP lp, PJ *P) {           /* Spheroidal, forward */
     struct pj_opaque *Q = P->opaque;
     struct isea_pt out;
     struct isea_geo in;
+
+    if (pj_is_nan(lp.lam) || pj_is_nan(lp.phi)) {
+        proj_log_error(P, "isea lp input has a NaN: lam: %lf phi: %lf",
+                       lp.lam, lp.phi);
+        return xy;
+    }
 
     in.lon = lp.lam;
     in.lat = lp.phi;
--- a/src/PJ_isea.c
+++ b/src/PJ_isea.c
@@ -7,6 +7,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <float.h>
+#include "proj_internal.h"
 
 #ifndef M_PI
 #  define M_PI 3.14159265358979323846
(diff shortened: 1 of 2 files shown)


src/PJ_isea.c
```diff
@@ -59,13 +60,19 @@ int hex_iso(struct hex *h) {
...
 ISEA_STATIC
-int hexbin2(double width, double x, double y,
-                int *i, int *j) {
+void hexbin2(double width, double x, double y, int *i, int *j) {
     double z, rx, ry, rz;
     double abs_dx, abs_dy, abs_dz;
     int ix, iy, iz, s;
     struct hex h;
 
+    if (pj_is_nan(x) || pj_is_nan(y) || pj_is_nan(width)) {
```
kbevers: The NaN-input can't be avoided earlier in the code? Without being particularly familiar with this code it seems like a fairly low-level function. What happens when `*i` and `*j` are set to 0? Does this still produce useful coordinates in the end?

src/PJ_isea.c
```diff
@@ -59,13 +60,19 @@ int hex_iso(struct hex *h) {
...
+        fprintf(stderr, "hexbin2 cannot handle NaN inputs\n");
```
kbevers: `proj_log_error` or `proj_log_debug` depending on what happens with the final coordinate

src/PJ_isea.c
schwehr: I will take another go at fixing this.  I'm sure it can be caught earlier.  

I always get nervous the higher up the stack I go.  I'd bet the results are garbage in the case of incoming nans.  Zeroing the result and complaining are better than the SIGILL I was getting.

src/PJ_isea.c
schwehr: Will do.  Lower down in this file, it uses fprintf stderr.  I'll convert that too.

kbevers: Thanks. Looking at this code I think there are several things that can be improved in the future. Especially keeping #423 in mind. This will be a nice start.

src/PJ_isea.c
kbevers: I can relate. But if it can be caught it is easier to fail gracefully I think. I wonder if this should really fail at the initialisation stage? The input doesn't seem to be unreasonable. Especially compared to the proj-strings :)

src/PJ_isea.c
```diff
@@ -1039,6 +1038,12 @@ static XY s_forward (LP lp, PJ *P) {           /* Spheroidal, forward */
     struct isea_pt out;
     struct isea_geo in;
 
+    if (pj_is_nan(lp.lam) || pj_is_nan(lp.phi)) {
+        proj_log_error(P, "isea lp input has a NaN: lam: %lf phi: %lf",
+                       lp.lam, lp.phi);
+        return xy;
```
schwehr: Should I be calling this?
```C
proj_errno_set(P, PJD_ERR_INVALID_X_OR_Y);
```

schwehr: Definitely better to do the NaN check in pj_fwd.  I very much doubt that any project is expecting a NaN.

cffk: I disagree...

I can plot a coastline in MATLAB using NaNs to cause the breaks which
are necessary to skip from one island to the next.  Reprojecting the
data should silently preserve NaNs.

cffk: Yes, this is fine.

However, as a matter of programming policy, I prefer to avoid special
handling for NaNs -- just let them percolate through the code.  This
keeps the code simple.  NaNs are (should be) rare so there's not much
benefit to an early exit from a function.

Unfortunately this policy hits a glitch with the double to int
conversion (which is why I wish that the C standard said that the
*result* of the conversion of NaN to int is undefined rather than that
the *behavior* is undefined).

Finally, note that the treatment of NaNs requires attention to
conditionals.  E.g., the convergence criterion should not be
`abs(d) <= eps` but `!(abs(d) > eps)`.

kbevers: > Yes, this is fine.

Great. Question is then, @schwehr, is that enough to avoid the problem your fuzzer uncovered? From you description the input seem to be actual numbers.

kbevers:
I've created https://github.com/OSGeo/proj.4/pull/949 to address this in a more generic way.

PR #945, opened 2018-04-22
Update documentation title, author, and other details

mwtoews:
There are a few commits here that do a few things:
* Change the title and author of the PDF documentation to be inline with the CITATION (#914). Most important  of these is that author is changed from "Gerald Evenden, Frank Warmerdam, and others" to "PROJ contributors".
* Rename a few things from PROJ.4 to PROJ. Most importantly is Sphinx's `project`. This is visible in a few places, such as the HTML documentation.
* Also renamed the PDF to `proj.pdf`, which will break a few external links to that URL. (Is it possible to create a symbolic link at `https://raw.githubusercontent.com/OSGeo/proj.4/gh-pages/proj4.pdf`?) Perhaps we may want to rename these PDF outputs with the version in the filename, e.g. `proj-5.1.0.pdf`, or is that going to be difficult to maintain?
* Tidy a few LaTeX things, so I can build it (I couldn't without these fixes). Also, the PDF version doesn't need to self-reference itself in the index.

Specific notes are tracked in each commit description.


kbevers:
> Rename a few things from PROJ.4 to PROJ. Most importantly is Sphinx's project. This is visible in a few places, such as the HTML documentation.

From the 5.0.0 release notes:

```
In recognition of the history of the software we are keeping PROJ.4 as
the *name of the organizing project*. The same project team also
produces the datum-grid package.

In summary:

o The PROJ.4 project provides the product PROJ, which is now at
  version 5.0.0.

o The foundational component of PROJ is the library libproj.

o Other PROJ components include the application proj, which provides
  a command line interface to libproj.

o The PROJ.4 project also distributes the datum-grid package,
  which at the time of writing is at version 1.6.0.
```

I have so far avoided changing the project name in Sphinx from PROJ.4 to PROJ for that reason. We do include information on the proj-datumgrid packages as well, so the documentation is not limited to PROJ only. I realise this is mostly semantics and hardly anyone cares. Thoughts on this?

Otherwise I think this looks fine. Having version number in the PDF file probably doesn't make sense for the auto-generated PDF but once the docs reach a more complete state it could be handy having documentation snapshots for each version on download.osgeo.org/proj/. For instance starting from version 6.0.0.

kbevers:
> Ah right, I knew the project would remain as PROJ.4, but I seem to have slipped that this was the same as Sphinx's project. I'll reverse this.

The lines are a bit blurred here, I'll admit that. The way I view it is that proj4.org is the project website that contain information about the project itself, documentation for PROJ and to some extend the proj-datumgrid packages. I am working on a plan for a better structure on the docs as a whole. I'll try to get that done soon and put it in an issue so we can discusss it in more details.

Let me know when you've reverted the PROJ.4 -> PROJ things and I'll merge this.

mwtoews:
I've restored project to PROJ.4. Two other modifications that I've also done:
* Make release the same as version. This make it simpler to maintain. (From what I've read, this distinction is important for some software that uses Sphinx, but I don't see this distinction necessary the PROJ.4 project.)
* Disable the default syntax highlighting, which was Python (when unspecified). Most code blocks now should look more plain now, unless indicated (i.e. `.. code-block:: console`). I've quickly checked html, latexpdf and epub outputs, and I think they less distracting. Perhaps a few blocks here-and-there would benefit from specified syntax highlighting.

Any luck looking into creating symbolic links from proj.pdf to proj4.pdf?

kbevers:
> Any luck looking into creating symbolic links from proj.pdf to proj4.pdf?

This should be as simple as adding the symlink to the `gh-pages` branch. I expect it to also work when you download files via raw.githubusercontent.com. I think it is going to be a hassle maintaining versionen pdf-files.

PR #947, opened 2018-04-22
PJ_isea.c: change local helper functions to void returns

### LINKED ISSUE #423: Invalid return 1; in PJ_isea.c:isea_hex

schwehr:
The return values are not used and do not mean anything.

### DIFF
--- a/src/PJ_isea.c
+++ b/src/PJ_isea.c
@@ -30,7 +30,7 @@ struct hex {
 /* y *must* be positive down as the xy /iso conversion assumes this */
 ISEA_STATIC
 void hex_xy(struct hex *h) {
-    if (!h->iso) return 1;
+    if (!h->iso) return;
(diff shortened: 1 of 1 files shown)


kbevers:
I agree with the changes. However I dont understand how this fix #423 ? It seems there's still a bunch of code that can't be reached after line 918...

schwehr:
Err... Probably doesn't fix it. My brain is shot

PR #951, opened 2018-04-24
cmake function proj_convert_grid_lla2gsb should use the target

malytomas:
use generated executable directly rather than trying to guess the full path


kbevers:
What problem does this solve?

kbevers:
My question has been unanswered for a month. Closing this on the basis that I don't understand the value this PR brings.

PR #953, opened 2018-04-25
rename proj4.pdf -> proj.pdf; create symbolic link for transition

mwtoews:
As discussed in #945


kbevers:
Let's see if this thing works :)

kbevers:
It didn't: https://raw.githubusercontent.com/OSGeo/proj.4/gh-pages/proj4.pdf

After a bit of digging I've learned that you can't use symlinks with gh-pages. That would have been nice. We will just have to live with the file name change. There's probably not many who rely on that link anyway.

mwtoews:
I guess it's an improvement over "404: Not Found" :)

PR #955, opened 2018-04-25
pj_transform: reset error state before each call to pj_inv/pj_fwd

rouault:
Fixes issues raised in https://lists.osgeo.org/pipermail/gdal-dev/2018-April/048446.html

The use case is that pj_transform() is called from geos projection to something else,
with multiple coordinates. If one of the coordinate tuple fails the inverse
transform of geos, it fails with PJD_ERR_TOLERANCE_CONDITION. Causing all following
coordinates to fail since the error state is not reset.

### DIFF
--- a/src/pj_transform.c
+++ b/src/pj_transform.c
@@ -42,6 +42,8 @@ enum PJ_DIRECTION {
 };
 typedef enum PJ_DIRECTION PJ_DIRECTION;
 
+/* Copied from proj.h FIXME */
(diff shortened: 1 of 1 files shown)


kbevers:
Looks good to me

borysiasty:
@jef-n please note this bug also affects QGIS' OTF reprojection of XYZ layers (although it's not a problem as long as QGIS builds use PROJ 4.9.3)

PR #965, opened 2018-04-30
Move include of math.h inside include guard.

schwehr:
No reason to include math.h more times.

And balance whitespace a bit


kbevers:
Nice catch!

PR #966, opened 2018-04-30
Doc improvements

kbevers:
This is mainly introduces a better way to document the individual parameters that can be set on a given operation. Instead of giving a short description of the parameters specific to the operation in the info table a new section called "parameters" is added. In the parameters section every parameter that can affect the given operation is added and described using the Sphinx `:option:` directive. This has the advantage of better formatting, linked referencing to the parameter and most importantly more space to describe what the parameter does. This also makes it possible to define default parameter in separate files that can be included on all the doc pages where it is needed making it immensely simpler to maintain across the ~150 operations so far added to PROJ.

Before:

screen shot 2018-04-30 at 22 59 44

After:

screen shot 2018-04-30 at 22 59 55
screen shot 2018-04-30 at 23 00 28

In addition to the refactoring of parameter descriptions I have also added some basic information to the table on each page such as the alias of the operation, the domain of the operation (2D, 3D and/or 4D) and what input and output types to use with the operation. The last part has been made more important with the recently added constraint on the pipeline operation that checks if the correct input is passed to each step in the pipeline.


kbevers:
I did think about adding an include file that would add the most basic projection parameters but soon realised that there are so many exceptions to the rule that it wouldn't be very feasible. Basically the only parameters you can be (almost) sure works for every projection is `+x_0`,  `x_y` and `+lon_0`. So in the end I decided to just be verbose and add what is needed. It comes with the added benefit of actually having to check that things work before writing the docs. I found a few problems that way checking up on the ~20 operations involved in this PR. I expect there to be more!

kbevers:
At some point in the future it would be cool to add custom PROJ directives for parameters. If done in a smart way it should be easy to generate a glossary of all the parameters that is used in PROJ. 
I haven't reached that level of Sphinx mastery yet so for now my focus is on just documenting the parameters.

Issue #967, opened 2018-05-01
Website isn't deployed when building PDF fails


kbevers:
After merging #966 it seems that the website is not deployed properly from Travis when commits are made to the master branch. I have noticed this before in the past without looking into why. I did that today and it would seem the `build_docs.sh` script fails during the making of the PDF file. An excerpt form the [build output](https://travis-ci.org/OSGeo/proj.4/jobs/373593240) is pasted below

```
LaTeX Warning: Hyper reference `apps/proj:cmdoption-proj-le' on page 29 undefin
ed on input line 2668.
! Extra alignment tab has been changed to \cr.
<recently read> \endtemplate 
                             
l.2729 \hline\end{tabulary}
                           
? 
! Emergency stop.
<recently read> \endtemplate 
                             
l.2729 \hline\end{tabulary}
                           
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on proj.log.
Makefile:67: recipe for target 'proj.pdf' failed
make[1]: *** [proj.pdf] Error 1
make[1]: Leaving directory '/data/docs/build/latex'
Makefile:136: recipe for target 'latexpdf' failed
make: *** [latexpdf] Error 2
Done. Your build exited with 0.
```

I can not reproduce locally. I am using a newer version of Sphinx on my computer which could be the cause.

@hobu Have you seen this before? Would it be possible for you to update the docker images to include a more recent version of Shpinx so we can test if that is the problem?

hobu:
It is best to run the docker build process rather than a local one as part of your local editing process:
```
$ cd dev/proj4
$ docker run -v `pwd`:/data -w /data/docs osgeo/proj.4 make latexpdf
```

When I did that with the existing `osgeo/proj.4` docker container, it failed in the same way for me.  I just pushed a new container based on Ubuntu 18.04 that is updated and works.

Failing to update the website without the PDF is the only reason you checked the PDF 😉.

kbevers:
> It is best to run the docker build process rather than a local one as part of your local editing process

Well... I don't really buy into all that fancy docker technology. I prefer to keep things simple around here :) (also, most of the time I'm on a computer where docker is not allowed).

> Failing to update the website without the PDF is the only reason you checked the PDF 😉.

Nah, I check it regularly locally. Especially after big changes such this recent one...

Thanks for updating the containers!

PR #968, opened 2018-05-01
Fix syntax in RST tables

mwtoews:
Fixes LaTeX issue in #967 and allows PDF to be built on older sphinx versions (like mine)


kbevers:
Nice catch. I wonder why Sphinx didn't complain about that here...?

mwtoews:
Yeah, I'd like to know a better way to parse .rst documents to raise possible errors or warnings. I only found these, and similar (410a515228166399f8c145568f20a7799a450757) by inspecting the latex log output to see which document it was parsing, then looking a the .rst document to manually inspect for oddities. There should be a better way to log these somehow...

kbevers:
Apparently there's [rstcheck](https://github.com/myint/rstcheck) which seems to do the trick. More or less at least. I'll play around with it a bit. I might make sense to add as a travis check.

Issue #903, opened 2018-03-26
Update JNI binding to new Proj.4 API


desruisseaux:
Java Native Interface (JNI) binding currently uses the old Proj.4 API, together with access to some structures not formally documented as public API. Those JNI binding should be upgraded to the new Proj.4 API.

This mapping is accessible to Java developers through the `org.proj4.PJ` Java class. With the new Proj version number, we may need to rename the package name, for example `org.osgeo.proj`.

Note: `PJ` is used by the `geoapi-proj4` module, but that module has never been formally released by the Open Geospatial Consortium (OGC). Up to now, it was only provided on GitHub as an example of GeoAPI implementation. That module my be part of GeoAPI 3.1 release (again mostly as an example), or alternatively it could move directly in the Proj project.

Related discussion in issue #900.

desruisseaux:
Some first observations:

* Package name is currently `org.proj4`. Do the project name has been renamed as "Proj", or does it stay "Proj.4"? In the former case and given that Proj is now an OSGeo project, the package name could be renamed as `org.osgeo.proj`. That would be closer to consistency with the Java version of that library, which is located in the `org.osgeo.proj4j` package. It would be an incompatible change however.
* File `src/org_proj4_Projections.h` is obsolete and can be deleted.
* https://github.com/opengeospatial/geoapi/issues/26 needs to be merged to `src/jniproj.c`.
* What is the license exactly for Proj?

desruisseaux:
Deletion of `src/org_proj4_Projections.h` submitted by pull request #956.

desruisseaux:
Some questions for the next steps:

* What should be the version number of the JAR file? Should it follow Proj version numbers or should it have different numbers on the basis that it will be updated less often?
* If the JAR file version number follows Proj version number, is there a way to know the Proj version at build time? (e.g. an environment variable). The intend is to inject correct information in the `META-INF/MANIFEST.MF` file.
* The `org_proj4_PJ.h` file is currently in the `src/` directory. Would it be okay to move it to the `jniwrap/org.osgeo.proj/` directory? (would be slightly more convenient for `javac` - but only as a minor point). If yes, do I only need to change the paths in `Makefile.am` and `lib_proj.cmake` (in addition of `jniproj.c` of course) before to run `configure` again, or is there something else I need to know?

kbevers:
> What should be the version number of the JAR file? Should it follow Proj version numbers or should it have different numbers on the basis that it will be updated less often?

[...]

PR #969, opened 2018-05-02
Update documentation and modularize the Java bindings

desruisseaux:
The 3 first commits are changes in the documentation only or in the use of a `#define` instruction without any API or functional change. The last commit moves all Java classes in an `org.osgeo.proj` sub-directory with no change in the content of those files. This move is for leveraging a new Java feature, which supports modularization since Java 9. By moving the classes in an `org.osgeo.proj` directory, we are creating a new Java module named "org.osgeo.proj". This module name is consistent with the package name proposed in #903 (module name and package name are two different things, but for a small project like those JNI wrappers they should be the same).

This merge request does not yet apply any API change. In particular the package name still have its old name (`org.proj4`). Package names will be addressed later.


kbevers:
Looks good to me. Since you are now touching the documentation it would be good if you could use the correct spelling of the library: PROJ (and not Proj). I am trying to keep it consistent in all of the documentation.

Am I correct in saying there is no breaking changes for users with this PR?

desruisseaux:
No problem (I didn't noticed that the PROJ was all upper-case). Do I update this merge request or do we wait for the next round?

There is no breaking change for users with this pull request, except that it requires Java 9 (current Java release is 10). If support for older Java versions is required, it would be possible at the cost of a more complex `build.xml` file.

kbevers:
> Do I update this merge request or do we wait for the next round?

Just change it now

> There is no breaking change for users with this pull request, except that it requires Java 9 (current Java release is 10). If support for older Java versions is required, it would be possible at the cost of a more complex build.xml file.

I think we're good as it is now then. I was mainly asking since we should not make breaking changes in minor version releases. PROJ 6 is scheduled for release early next year so that is a good opportunity to introduce breaking changes.

desruisseaux:
Done. Travis build passed, waiting for AppVeyor.

PR #971, opened 2018-05-03
Cleanup pj_phi2.c

schwehr:
- tabs -> spaces
- IWYU
- #define -> static const typed
- Add const
- At least one digit on either side of decimal points
- Combine definition and initialization
- Minimize scope of variables when possible
- Remove the odd continue / break usage inside the for
- Use the defined symbol rather than embedding bare literals
  -18 -> PJD_ERR_NON_CON_INV_PHI2

### DIFF
--- a/src/pj_phi2.c
+++ b/src/pj_phi2.c
@@ -1,26 +1,48 @@
-/* determine latitude angle phi-2 */
+/* Determine latitude angle phi-2. */
 
 #include <math.h>
 
 #include "projects.h"
 
-static const double TOL = 1.0e-10;
-static const int N_ITER = 15;
+/*****************************************************************************/
+double pj_phi2(projCtx ctx, double ts, double e) {
+/******************************************************************************
+Determine latitude angle phi-2.
 
-double
-pj_phi2(projCtx ctx, double ts, double e) {
-    const double eccnth = 0.5 * e;
+Inputs:
+  ts = exp(-psi) where psi is the isometric latitude (dimensionless)
+  e = eccentricity of the ellipsoid (dimensionless)
+Output:
+  phi = geographic latitude (radians)
+
+Here isometric latitude is defined by
+
+  psi = log( tan(pi/4 + phi/2) *
+             ( (1 - e*sin(phi)) / (1 + e*sin(phi)) )^(e/2) )
+      = asinh(tan(phi)) - e * atanh(e * sin(phi))
+
+This routine inverts this relation using the iterative scheme given
+by Snyder (1987), Eqs. (7-9) - (7-11)
+*******************************************************************************/
+    const double TOL = 1.0e-10;
+    const int N_ITER = 15;
     double phi = M_HALFPI - 2.0 * atan(ts);
-    int i = N_ITER;
-    for(;;) {
+    int i;
+
+    for (i = N_ITER;  i > 0;  i--) {
         const double con = e * sin(phi);
-        const double dphi = M_HALFPI - 2.0 * atan(ts * pow((1.0 - con) /
-           (1.0 + con), eccnth)) - phi;
+        const double dphi =
+            M_HALFPI -
+            2.0 * atan(ts * pow((1.0 - con) / (1.0 + con), 0.5 * e)) - phi;
+
         phi += dphi;
-        if( !(fabs(dphi) > TOL && --i) )
-            break;
+
+        if (fabs(dphi) < TOL)
+            return phi;
     }
-    if (i <= 0)
+
+    if (i == 0)
         pj_ctx_set_errno(ctx, PJD_ERR_NON_CON_INV_PHI2);
-    return phi;
+
+    return HUGE_VAL;
 }
--- a/src/pj_phi2.c
+++ b/src/pj_phi2.c
@@ -1,28 +1,26 @@
 /* determine latitude angle phi-2 */
+
+#include <math.h>
+
 #include "projects.h"
 
-#define TOL 1.0e-10
-#define N_ITER 15
+static const double TOL = 1.0e-10;
+static const int N_ITER = 15;
 
-	double
+double
 pj_phi2(projCtx ctx, double ts, double e) {
-	double eccnth, Phi, con;
-	int i;
-
-	eccnth = .5 * e;
-	Phi = M_HALFPI - 2. * atan (ts);
-	i = N_ITER;
-	for(;;) {
-		double dphi;
-		con = e * sin (Phi);
-		dphi = M_HALFPI - 2. * atan (ts * pow((1. - con) /
-		   (1. + con), eccnth)) - Phi;
-		Phi += dphi;
-		if( fabs(dphi) > TOL && --i )
-			continue;
-                break;
-	}
-	if (i <= 0)
-		pj_ctx_set_errno( ctx, -18 );
-	return Phi;
+    const double eccnth = 0.5 * e;
+    double phi = M_HALFPI - 2.0 * atan(ts);
+    int i = N_ITER;
+    for(;;) {
+        const double con = e * sin(phi);
+        const double dphi = M_HALFPI - 2.0 * atan(ts * pow((1.0 - con) /
+           (1.0 + con), eccnth)) - phi;
+        phi += dphi;
+        if( !(fabs(dphi) > TOL && --i) )
+            break;
+    }
+    if (i <= 0)
+        pj_ctx_set_errno(ctx, PJD_ERR_NON_CON_INV_PHI2);
+    return phi;
 }


busstoptaktik:
While combining definition and initialization is in general to be preferred, in this case I would suggest having the initialization of `i` go into the header of the `for`loop. I believe this is a more idiomatic way of spelling out that "here we do an iterative computation".

In C99 we could have used `for (int i = N_ITER;  i > 0;  i--)`. In C89/90, we need to split definition and initialization.

Also, since `eccnth` is used only once, and since its definition is evident, I think the code is more readable if just using its definition `0.5*e` at its single place of usage. I am aware that, formally, this introduces an unneeded multiplication inside the loop, but am sure that this (1) will be detected as loop invariant by even a 1980s compiler, and (2) is dwarfed by the cost of the atan call.

Hence, I suggest this (only slightly tested) modified form:

```C
/* determine latitude angle phi-2 */

#include <math.h>

#include "projects.h"

static const double TOL = 1.0e-10;
static const int N_ITER = 15;

double pj_phi2 (projCtx ctx, double ts, double e) {
    double phi = M_HALFPI - 2.0 * atan(ts);
    int i;

    for (i = N_ITER;  i > 0;  i--) {
        const double con = e * sin(phi);
        const double dphi = M_HALFPI - 
            2.0 * atan(ts * pow((1.0 - con) / (1.0 + con), 0.5*e)) - phi;

        phi += dphi;

        if (fabs (dphi) < TOL)
            return phi;
    }

    if (i == 0)
        pj_ctx_set_errno(ctx, PJD_ERR_NON_CON_INV_PHI2);
    
    return HUGE_VAL;
}
```

Note that I have added quite a lot of additional vertical space, to delimit the logical sub-blocks.

Also, the stop-condition and the test for convergence are not exactly identical wrt. the original version, but since `pj_phi2 ()` tends to converge rapidly (less than a handful of iterations), the exact stop condition should not matter much.

Finally, note that I have changed the return value in case of non-convergence to HUGE_VAL, since this is what `pj_merc.c` and `PJ_omerc.c` expects. But since the algorithm converges quickly for all reasonable input values, this will probably not make any difference. For the same reason, perhaps we could skip the context arg and the pj_ctx_set_errno call. But this will be for another day.

cffk:
This is a matter of taste...  But I would say that `con` and `dphi` should *not* be declared `const`.  I *know* that the variables are not subsequently changed.  However, these are *just* temporary variables in a loop and there values change on each iteration.  So the `const` declaration isn't helpful to the human reader.  (The compiler will do the same thing here anyway.)

Surely, the **biggest** problem with this code, is that there's *no documentation* on what this function does!  If there were, then maybe it would be easy to see how to replace the lousy `pow((1-x)/(1+x), e/2)` idiom with something involving `atanh`.

busstoptaktik:
> Surely, the biggest problem with this code, is that there's no documentation on what this function does! If there were, then maybe it would be easy to see how to replace the lousy pow((1-x)/(1+x), e/2) idiom with something involving atanh.

@cffk: I certainly agree wrt. documentation and (potential) algorithmic improvements.

Essentially, pj_phi2 implements the material around equation (7-9), p. 38 in Snyder: Map Projections - a Working Manual (included in proj.4/docs/old). Perhaps taking this into account, you can suggest improvements?

cffk:
For the record...!

Replace Snyder's (7-7) with

    psi = y/a = asinh(tan(phi)) - e * atanh(e * sin(phi))

replace (7-9) with

    phi = atan(sinh(psi + e * atanh(e * sin(phi))))

and iterate with an initial guess, `phi = atan(sinh(psi))`.

Advantages

* manifestly odd
* accurate for small `phi`/`psi`

This assumes decent implementations of `sinh`, `atanh`, `asinh`. These
are provided by `C99` and fall-back implementations are easy.

The tricky part of slotting this approach into PROJ is that the argument
of `pj_phi2` is

    t = exp(-psi)

which right away tosses out oddness and accuracy!  So it's necessary
to fix all the places that `pj_phi2` is invoked.

Alternatively, use, the conformal latitude `chi`

    psi = asinh(tan(chi))
    chi = atan(sinh(psi))

and transform between `phi` and `chi` with the series (given by Engsager
and Poder, 2007) used in `proj_etmerc.c`.  (E+P, 2007, use the symbol `Phi`
instead of  `chi`.)

schwehr:
A nice discussion.  Thanks!

I will push an updated version when I have a few.

+1 for the documentation.  I'm not a geodesist and my math is super rusty, so I'm refactoring this blind.  What are ts, e, eccnth, phi, con, dphi?  Units?  Acceptable input ranges?  Many readers of this code don't even have a spatial background of any kind (e.g. if a member of a security team is tracing an issue through this code).   A unittest in C or C++ would be a great complement to function comment and would allow people to try out refactorings.  And a microbenchmark would allow for seeing the impact of changes.  I should write those and stick them in autotest2.  Should there be checks to state what can come in?  I'm used to Google/Alphabet's ABSL DCKECK and QCHECK.  e.g.

```C
assert(!pj_is_nan(ts));
assert(!pj_is_nan(e));
assert(ts > -1.0e42 && ts < 1.0e42);  /* Insert the correct range checks here */
assert(e > -1.0e42 && e < 1.0e42);
```

Thanks for the change to move i to the for.  I had tried that and got it wrong somehow.  Sleep deprived me wasn't able to spot what I did wrong.

Extra spacing seems good.  Wishing Proj had an official clang-format spec so I wasn't worrying about the other formatting aspects.

For the const:

I would argue for keeping as many consts in as possible.  The code base has historically been overly clever/tricky/sneaky.  I have to assume that all code is doing things that I don't immediately see until proven that it is doing what I think it is doing.   And yes, I'm pretty sure those consts aren't going to change what the compiler is putting out in this case... to the point where I'm not going to check with godbolt.org.

An alternative middle ground is to be as aggressive with const as possible.  Then remove a subset of them down the road after cleaning the entire code base and letting it bake for 6 months or a year.

My C++ brain wants to make things constexpr if at all possible and if not const so that later refactors are as easy and obvious as possible.  And you sometimes (or eventually) get a bonus of using a constexpr function.  e.g. http://open-std.org/JTC1/SC22/WG21/docs/papers/2017/p0533r0.pdf

And an interesting discussion here: https://stackoverflow.com/questions/17347935/constexpr-math-functions

And now I'm having bad flashbacks to working on machines where you can't count on trig being implemented correctly... when you have to fix the OS providers atan2 and you just spent $50K US on the OS.

cffk:
How about this for the body of the loop

    double con = e * sin(phi), ophi = phi;
    phi = M_HALFPI - 2 * atan(ts * pow((1 - con) / (1 + con), e/2));
    
    if (!(fabs (phi - ophi) >= TOL))  /* Backwards test so that NaNs exit immediately */
        return phi;

Notes:

* I still find the arbitrary use of the const declaration jarring.
* I prefer to read ints over doubles (1 instead of 1.0) where I know the compiler will do the promotion; less "noise" in the code.
* It'll be a little more accurate here to use the formula for `phi` rather than artificially subtracting and adding it back.
* In this form, the equation for `phi` is an exact transcription of Synder's (7-9).

kbevers:
> BTW, I am so totally uncomfortable messing with this code without direct unittesting.

We all are. Just waiting for somebody to volunteer to change that :-)

cffk:
Sorry to be beating a dead horse here...  Regarding the `const`
declaration, the pattern is

    for (int i = 0; i < imax; ++i) {
      [const] double t = temp-var;
      ...
    }

an alternative style might be

    double t;
    for (int i = 0; i < imax; ++i) {
      t = temp-var;
      ...
    }

which I would regard as identical, but now the `const` is disallowed.
Maybe you discover that things can be speeded up by unrolling the loop
(e.g., when doing Clenshaw sum, this avoids some copies).  Then we have

    double x, y;
    for (int i = 0; i < imax; i += 2) {
      double t;
      t = temp-var;
      ...
      y = x;
      t = temp-var;
      ...
      x = y;    /* loop unrolled so that the role of x, y is unchanged */
    }

Possibly you need to modify the body of the loop, so it's now

    for (int i = 0; i < imax; ++i) {
       double t = temp-var;
       ...
       t += something;
       ...
     }

In these case, the (now disallowed) `const` declaration is supposed to
warn the code author "woah, you really don't want to do this".  That's
appropriate for M_PI, TOL, NITER but not in these cases.  The variable
`t` is merely a placeholder for an expression which *varies* from one
iteration to the next.

Finally, I urge caution with these code fixes and limit them to ones
that clearly improve the code (e.g., MACRO -> static const).  This is
especially the case with transformations which are made without any
attempt to understand the code!

hobu:
> Finally, I urge caution with these code fixes and limit them to ones  that clearly improve the code (e.g., MACRO -> static const). This is especially the case with transformations which are made without any
attempt to understand the code!

I very much agree with this sentiment. PROJ doesn't have enough test coverage for us to be doing gross refactoring of things for style and puritanical consistency reasons. Another consideration is a style change creates a discontinuity in the implementation, and it makes it difficult to go forward and backward in time for archeology reasons to determine when or why something might have changed.

schwehr:
I attempted and failed to understand this code.  I've been trying to stick to behavior invariant changes.

And @cffk, I honestly don't understand what you are getting at with your comment.  I have not changed the scope and behavior of the key `Phi` var.  `dphi` was always inside the for loop.

With the code:
```
double t;
for (int i = 0; i < imax; ++i) {
  t = temp-var;
  ...
}
```

I view that as saying to the human that the t in each iteration might *not* be independent.  Reader beware.  By moving the t inside the loop with or without a const, you are saying with certainty that t from each iteration can be treated independently if you'd like to.

So I read this function and figured out for each variable if the iterations were independent.  With the Proj coding history, that was not a trivial task.

I try to leave loop unrolling to the compiler as much as possible.  Or breakout godbolt plus test with real data as cache interactions can be surprising.

cffk:
Argument `ts < 0` should signal an error (because `ts = exp(-psi)`).  Argument `ts = 1/x` should return negative of result of `ts = x`.  Thus `ts = inf` should be moved out of the exceptional class.  Tests should include round trip comparisons `phi -> psi -> t -> phi`.  All proj functions should be tested with the `SRMax` flattening value of `f = 1/150` to test convergence etc.  The eccentricity is given by `sqrt(f * (2 - f)) = 0.11528`.

schwehr:
I think there is no point in me working on this function.  All I've done is prove that it is currently a mess.  Which is a tautology.  I was just trying to make the function as it exists easier to follow.

PR #973, opened 2018-05-04
Minor cleanup of pj_phi2.c

schwehr:
- tabs -> spaces
- IWYU
- #define -> static const typed
- Combine definition and initialization
- Define one var per statement
- Use the defined symbol rather than embedding bare literals
  -18 -> PJD_ERR_NON_CON_INV_PHI2

Subset of fixes discussed in #971


kbevers:
Looks good to me. I think it would be nice if you added the descriptive comment Charles suggested in the previous PR. Or something to that effect if it doesn't fit a 100% in this less radically changed version of the code.

schwehr:
kbevers@: Docs added

kbevers:
Thanks!

PR #995, opened 2018-05-13
ISEA_STATIC -> static

schwehr:
As this is a part of PROJ, having these functions possibly not static
is confusing.  Time to declare this is just a part of PROJ.

### DIFF
--- a/src/PJ_isea.c
+++ b/src/PJ_isea.c
@@ -12,24 +12,13 @@
 #  define M_PI 3.14159265358979323846
 #endif
 
-/*
- * Proj 4 provides its own entry points into
- * the code, so none of the library functions
- * need to be global
- */
-#define ISEA_STATIC static
-#ifndef ISEA_STATIC
-#define ISEA_STATIC
-#endif
-
 struct hex {
         int iso;
         int x, y, z;
 };
 
 /* y *must* be positive down as the xy /iso conversion assumes this */
-ISEA_STATIC
-void hex_xy(struct hex *h) {
+static void hex_xy(struct hex *h) {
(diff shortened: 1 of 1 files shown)


schwehr:
so much fun...
```
--2018-05-13 03:16:15--  (try:20)  http://download.osgeo.org/proj/proj-datumgrid-1.7.zip
Connecting to download.osgeo.org|140.211.15.132|:80... connected.
HTTP request sent, awaiting response... Read error (Connection reset by peer) in headers.
Giving up.
```

kbevers:
> so much fun...

Yeah. I guess the server has gone home for the weekend :-) SAC team is trying to sort things out:

https://lists.osgeo.org/pipermail/sac/2018-May/010024.html

src/PJ_isea.c
```diff
@@ -309,8 +291,7 @@ sph_azimuth(double f_lon, double f_lat, double t_lon, double t_lat)
 #endif
 
 /* coord needs to be in radians */
-ISEA_STATIC
-int
+static int
```
kbevers: A small style nit here. In some cases the `static int` (or whatever) is on a line by itself, in others it is on the same line as the function. I prefer the latter (which is also what is generally used in PROJ) but at least it should be consistent across the file.

schwehr: Done.  I was trying not to worry to much about formatting without the project having an official style (or better yet a clang-format definition I can just run before committing).

kbevers: I know. Thanks for doing the touch up. About the style guide... we'll get there eventually :-)

PR #998, opened 2018-05-15
IWYU: Partial PJ_aitoff.c..PJ_goode.c

schwehr:


busstoptaktik:
`#include`ing `math.h`  before `projects.h` bypasses the magic needed to get `M_PI` related constants defined properly under Windows. @micahcochran did a lot of work 2 years ago to get this working.

In most cases, MSVC failing to define `M_PI`  will be masked by conditional private definitions in `projects.h`, but I can imagine pathological cases where having miniscule differences in the M_PI definitions between the library and a user program will create havoc.

So probably, to enforce strict IWYU (which anyway is a violation of DRY), include this incantation:

    #ifdef _MSC_VER
    /* enable predefined math constants M_* for MS Visual Studio workaround */
    #  ifndef _USE_MATH_DEFINES
    #     define _USE_MATH_DEFINES
    #  endif
    #endif

before including math.h (it can probably safely be abbreviated to just defining `_USE_MATH_DEFINES` in most cases).

kbevers:
Wouldn't the section from `projects.h` below define the needed constants no matter the reason they are missing?

https://github.com/OSGeo/proj.4/blob/6284d19e1d2efdc60fd3752187c275e345ab423b/src/projects.h#L95-L117

I don't think the problem is very big if existing at all in practical use. Kurts work here is a good step towards making `projects.h` less important (and complicated) in the future.

busstoptaktik:
> Wouldn't the section from projects.h below define the needed constants no matter the reason they are missing?

Yes, it would define them - but not necessarily identically to the MSVC definitions. And as they are universally important in converting to/from angular coordinates, deviation is catastrophic (for a sufficiently restrictive definition of catastrophic). Hence my suggestion of adding (an unconditional) `#define _USE_MATH_DEFINES` prior to `#include <math.h>` (alternatively, we would have to conditionally define the M_PI constants in almost every PJ_LIB__ enabled file)

kbevers:
Yes, let's agree on that. And for the record my MSVC `M_PI` is the exact same as the one in `projects.h` so the implications are miniscule I would say. There is of course a change that Microsoft has been using different values for pi over the years but probably not.

Once we start working on 6.0 we can clean up `projects.h` and probably get rid of some of the `M_*` definitions.

schwehr:
`IWYU (which anyway is a violation of DRY)`

I disagree.  The whole point of DRY is to allow cleaner refactoring.  Having an include in the compilation unit is a statement that this compilation unit uses symbols from that header. The include guards prevent a DRY violation.

https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md

PR #1000, opened 2018-05-15
IWYU: Partial PJ_natearth.c..rtodms.c

schwehr:
Focused on the standard includes.
I ignored the include what you use suggestions for the internal headers.

### DIFF
--- a/src/PJ_natearth.c
+++ b/src/PJ_natearth.c
@@ -13,6 +13,9 @@ and designed in collaboration with Tom Patterson.
 Port to PROJ.4 by Bernhard Jenny, 6 June 2011
 */
 #define PJ_LIB__
+
+#include <math.h>
+
 #include "projects.h"
 
 PROJ_HEAD(natearth, "Natural Earth") "\n\tPCyl., Sph.";
@@ -95,4 +98,3 @@ PJ *PROJECTION(natearth) {
 
     return P;
 }
-
--- a/src/PJ_natearth2.c
+++ b/src/PJ_natearth2.c
@@ -6,6 +6,9 @@ and Atmospheric Sciences, Oregon State University.
 Port to PROJ.4 by Bojan Savric, 4 April 2016
 */
 #define PJ_LIB__
+
+#include <math.h>
+
 #include "projects.h"
 
 PROJ_HEAD(natearth2, "Natural Earth 2") "\n\tPCyl., Sph.";
@@ -92,4 +95,3 @@ PJ *PROJECTION(natearth2) {
 
     return P;
 }
-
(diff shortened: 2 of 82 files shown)


kbevers:
Is this the last batch?

> I ignored the include what you use suggestions for the internal headers.

Seems fine. I suspect the suggestions are mostly removing includes?

schwehr:
Yup, this is the last one.  It's mostly adding includes of math.h where there is no include of proj_math.h.  I feel like math.h should be included in those cases too, but wanted something simple to get through.  There are a few extra other includes like for `PJ_ob_tran.c`

The most complicated one is `proj_api.h` where the includes move out of `extern "C" {`.   If a C++ compiler doesn't have std C headers with the extern already in them, that compile is IMHO too broken to bother with.

kbevers:
> Yup, this is the last one. 

Awesome! I appreciate the effort. I'd like to ask you to add a section on IWYU [here](https://github.com/OSGeo/proj.4/blob/master/docs/source/community/code_contributions.rst#programming-language). Now that all the files follow the principle we can introduce it as coding convention in PROJ. Just a few words on why and how, preferably with a link to the IWYU tool you've used for this.

> It's mostly adding includes of math.h where there is no include of proj_math.h. I feel like math.h should be included in those cases too, but wanted something simple to get through. 

I guess it is easy to search out the files that include `proj_math.h` and add a `math.h` include as well. My original intention with `proj_math.h` was, as you know, as a replacement of `math.h` but now that it seems we'll have to change our C/C++ standard requirements anyway the lifespan of `proj_math.h` will be short. I expect Evens barn raising efforts to at least give us access to C99 math functions in the future. So we might as well make it easy to remove again when that time comes.

> The most complicated one is proj_api.h where the includes move out of extern "C" {. If a C++ compiler doesn't have std C headers with the extern already in them, that compile is IMHO too broken to bother 

Yeah, I think it is best not to touch that file unless absolutely necessary. There's  only 1.5 years left before it can be put away.

schwehr:
I will definitely add something about `IWYU` in those docs.

If you want, I can easily send a follow PR on with math.h for all the files with proj_math.h.  But it's probably better to just do a `#include "proj_math.h` -> `#include <math.h>` when proj_math.h is done.

kbevers:
> I will definitely add something about IWYU in those docs.

Thanks!

> If you want, I can easily send a follow PR on with math.h for all the files with proj_math.h. But it's probably better to just do a #include "proj_math.h -> #include <math.h> when proj_math.h is done.

Either way is fine. We will have to revisit those files nonetheless when `proj_math.h` can be removed.

PR #1004, opened 2018-05-19
Vertical grid shift: do not interpolate node values at nodata value (fixes #1002)

### LINKED ISSUE #1002: Vertical Shift Interpolation errors at edges
Hi All, 
This concerns the "geoidgrids" v shift functionality.
Maybe this is a known issue but what I have encountered from testing is that when a point lies within one or more gtx "no_data" points then the found value is reported valid but actually the no_data value of -88 was used in the interpolation. Only when all four gtx points are equal to "no_data" (-88) then it is reported to fail.

This code can be found in file: pj_apply_vgridshift.c

lines 114-122:

`  cvs = (float *) ct->cvs;
        value = cvs[grid_ix + grid_iy * ct->lim.lam]
            * (1.0-grid_x) * (1.0-grid_y)
            + cvs[grid_ix2 + grid_iy * ct->lim.lam]
            * (grid_x) * (1.0-grid_y)
            + cvs[grid_ix + grid_iy2 * ct->lim.lam]
            * (1.0-grid_x) * (grid_y)
            + cvs[grid_ix2 + grid_iy2 * ct->lim.lam]
            * (grid_x) * (grid_y);
`
Afterwards is looked if cvs equals "no_data".
It should first evaluate the four points, then if any of them equals "no_data" it should fail.

rouault:
I've made the choice of ignoring nodes that are the nodata value, but not completely fail the computation if at least one of the 4 nodes has a valid value (properly taking into account the weights of the valid nodes in the final value). I've no strong opinion if this is better than just failing if one of the value if nodata.


kbevers:
If I understand your implementation correct, it means that we are effectively extrapolating when determining the grid value where three of the nodes are NODATA, yes?

rouault:
> If I understand your implementation correct, it means that we are effectively extrapolating when determining the grid value where three of the nodes are NODATA, yes?

Yes, the final value in the case you mention will be the value of the only node != NODATA

kbevers:
I am not a big fan of extrapolation here. It can be handled nicely but the implementation is a bit more complicated. We need at least two "good" nodes before we can do a meaningful interpolation. This results in the following cases for interpolation:

1. 4/4 nodes: As previously. Never NODATA.
2. 3/4 nodes: If the interpolation point is within the triangle made up of the three "good" nodes we can interpolate a value. Otherwise NODATA.
3. 2/4 nodes: If the interpolation point is *exactly* on the line between the to "good" nodes we can interpolate a value. Otherwise NODATA.
4. 1/4 nodes: Interpolation impossible. Always NODATA.

Wether this more complicated approach is worthwhile I am not sure. It is the correct way to do it I think, but seeing as the issue that sparked this has just recently surfaced after 15-20 years it may not be that big a deal :-)

rouault:
Hum, your above suggestion sounds to me too complicated implementation-wise vs the value it adds. So I'd say either stick with what I implemented, or error out as soon as at least one node is NODATA.

By the way the interpolation I implemented here is what is done in GDAL in similar cases (which can be questioned indeed)

PR #1006, opened 2018-05-19
Add documentation for Include What You Use (IWYU)

schwehr:
Docs requested in discussion of #1000

### DIFF
--- a/docs/source/community/code_contributions.rst
+++ b/docs/source/community/code_contributions.rst
@@ -158,7 +158,7 @@ Include What You Use (IWYU)
 Managing C includes is a pain.  IWYU makes updating headers a bit
 easier.  IWYU scans the code for functions that are called and makes
 sure that the headers for all those functions are present and in
-sorted order.  However, you cannot blindly apply IWYU to Prow.  It
+sorted order.  However, you cannot blindly apply IWYU to PROJ.  It
 does not understand ifdefs, other platforms, or the order requirements
 of PROJ internal headers.  So the way to use it is to run it on a copy
 of the source and merge in only the changes that make sense.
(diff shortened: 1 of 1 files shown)


docs/source/community/code_contributions.rst
```diff
@@ -151,3 +151,19 @@ Typo detection and fixes
...
 
 Run ``scripts/fix_typos.sh``
+
+Include What You Use (IWYU)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Managing C includes is a pain.  IWYU makes updating headers a bit
+easier.  IWYU scans the code for functions that are called and makes
+sure that the headers for all those functions are present and in
+sorted order.  However, you cannot blindly apply IWYU to Prow.  It
```
kbevers: > Prow

PROJ?

rouault:
For people not familiar with the tool, can we have some scripts/iwyu.sh or something like that, that would run it on proj with appropriate switches (if that makes sense...) ? And if we are currently 'clean' regarding this, this would be great to have in a Travis target, otherwise this will never be enforced consistently

kbevers:
> @rouault PROJ is nowhere near clean for the internal headers. I think the way to do IWYU without taking it just as advice is to fix all the headers so that they work as expected with IWYU. That will be a lot of work. It will require some annotations and some forwards. I will work towards that over time if people are okay with the annotation comments being in the code base.

In general I am not super happy about annotation comments in the code. And since the workload in becoming "IWYU clean" is substantial perhaps it is better to simply have "include what you use" as a guiding principle in new files? So in practice moving the section on IWYU further up the document to the "coding conventions" section. This way over time we should converge towards a clean state and we would not require any automated checks. Of course this leaves room to deviate from the guidelines but such is the nature of guidelines.

Issue #991, opened 2018-05-11
Multiple definitions of the same thing


schwehr:
PJ_isea looks like most of the trouble.  But there are likely other repetitions that should be removed.

```
egrep '[^-0-9](3\.14|1\.57|6\.28)' *
geocent.c:#define PI         3.14159265358979323e0
PJ_isea.c:#  define M_PI 3.14159265358979323846
PJ_isea.c:#define DEG90 1.57079632679489661922
PJ_krovak.c:#define S90 1.570796326794896  /* 90 deg */
projects.h:#define M_PI            3.14159265358979323846
projects.h:#define M_PI_2          1.57079632679489661923
projects.h:#define M_TWOPI         6.28318530717958647693   /* 2*pi */
```

```
(egrep '#define.*180' * ; egrep -i 'torad|todeg' *.h ) | sort -u
PJ_isea.c:#define DEG180 M_PI
PJ_isea.c:#define DEG2RAD (M_PI/180.0)
PJ_isea.c:#define RAD2DEG (180.0/M_PI)
proj.h:double proj_todeg (double angle_in_radians);
proj.h:double proj_torad (double angle_in_degrees);
proj_internal.h:#define PJ_TODEG(rad)  ((rad)*180.0/M_PI)
proj_internal.h:#define PJ_TORAD(deg)  ((deg)*M_PI/180.0)
proj_internal.h:#ifndef PJ_TODEG
proj_internal.h:#ifndef PJ_TORAD
```

kbevers:
Thanks for reporting these. Surprised that PI is not defined in more places after all.

`PJ_isea.c` is just a big mess. I've been to the library to borrow a copy of the orignal paper by Snyder so I can clean that mess up properly. Unfortunately no test data...

schwehr:
I can try to dump interesting cases from my fuzzer corpus if that is helpful for generating test cases.  I so need to work on my gie fuzzer so I can make proper test cases at scale

kbevers:
Closed with https://github.com/OSGeo/proj.4/commit/446a4d67f35db27036ae16f0490dbc45228a20b5

PR #1007, opened 2018-05-19
Pj isea cleanup

schwehr:


src/PJ_isea.c
```diff
@@ -3,15 +3,49 @@
...
+#define TABLE_G 0.6615845383
+
+/* H = 0.25 R tan g = */
+#define TABLE_H 0.1909830056
+
+/* in radians */
+#define ISEA_STD_LAT 1.01722196792335072101
+#define ISEA_STD_LON .19634954084936207740
+
+#define RAD2DEG (180.0/M_PI)
```
kbevers: Since you've already started the clean up: `proj_internal.h` provides `PJ_TODEG` and `PJ_TORAD`. Use those and you can more or less close https://github.com/OSGeo/proj.4/issues/991.

schwehr: PJ_TO* done in 8d42763.  Still have proj_todeg and PJ_TODEG, so #991 isn't quite done yet.

kbevers: That's by design. `proj_todeg` is primarily there for API users and `PJ_TODEG` is for internal use only (saving a few function calls by using the preprocessor).

PR #1011, opened 2018-05-23
Limit the number of round trips to 1 million and check for underflows

### LINKED ISSUE #1010: gie.c:roundtrip() UndefinedBehaviorSanitizer: float-cast-overflow with ans
An issue for #1009 

What is `ans` supposed to represent?  What range should it have?  And why is it allowed to be negative?  It would be great to have some range checking on ans and give a helpful message if the value was nonsensical.
```
third_party/proj4/tests/gie_fuzzer.cc:505:20: runtime error: 3.84808e+09 is outside the range of representable values of type 'int'
    #0 0x55c8e0b6898f in roundtrip(char const*) third_party/proj4/tests/gie_fuzzer.cc:505:20
    #1 0x55c8e0b6603c in dispatch(char const*, char const*) third_party/proj4/tests/gie_fuzzer.cc:766:49
    #2 0x55c8e0b65a31 in process_file(char const*) third_party/proj4/tests/gie_fuzzer.cc:242:19

SUMMARY: UndefinedBehaviorSanitizer: float-cast-overflow third_party/proj4/tests/gie_fuzzer.cc:505:20
``` 

```c
    static int roundtrip (const char *args) {
    // ...
    double ans = proj_strtod (args, &endp);
    ntrips = (int) (endp==args? 100: fabs(ans));
```

schwehr:
Why was fabs applied to ntrips?  And why is it parsed as a float?
Also add documentation for roundtrip.
External rst docs do not match this function.

Fixes #1010

### DIFF
--- a/docs/source/apps/gie.rst
+++ b/docs/source/apps/gie.rst
@@ -207,10 +207,26 @@ gie command language
     to function. The accepted coordinate is passed to the operation first in
     it's forward mode, then the output from the forward operation is passed
     back to the inverse operation. This procedure is done ``n`` times. If the
-    resulting coordinate is within the set tolerance of the initial coordinate
+    resulting coordinate is within the set tolerance of the initial coordinate,
     the test is passed.
 
-    Example:
+    Example with the default 100 iterations and the default tolerance:
+
+    .. code-block:: console
+
+        operation       proj=merc
+        accept          12 55
+        roundtrip
+
+   Example with count and default tolerance:
+
+    .. code-block:: console
+
+        operation       proj=merc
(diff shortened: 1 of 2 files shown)


kbevers:
> Why was fabs applied to ntrips? And why is it parsed as a float?

The answer to this is likely to be something like "it worked in a hurry".

> External rst docs do not match this function.

Do you mind fixing it? You have more or less already done so in the function documentation anyway.

schwehr:
Add two more examples so that all of the 3 cases are covered.

kbevers:
Thanks!

PR #1013, opened 2018-05-23
Use proj_strtod.h for proj_strtod and proj_atof

schwehr:
Skipped dmstor.c that has a local proj_strtod.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -18,11 +18,11 @@ EXTRA_DIST = makefile.vc proj.def bin_cct.cmake bin_gie.cmake bin_cs2cs.cmake \
 
 proj_SOURCES = proj.c gen_cheb.c p_series.c
 cs2cs_SOURCES = cs2cs.c gen_cheb.c p_series.c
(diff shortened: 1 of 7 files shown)


kbevers:
We have a few other string functions specific to PROJ, e.g. `pj_chomp` and `pj_shrink`. Would it make sense to make a more generic `proj_string.h` header?

We also have `pj_strtod` and `pj_atof`. I think it is about time we make the switch.

schwehr:
A `pj_strings.[ch]` that was a part of the core lib, but with the header private seems like a reasonable idea.  I was aiming to just incrementally fix one small issue with this PR.  I am a fan of header file names that match the compilation unit (especially for internal code), so if you want proj_string.h, I would suggest moving all the related code into proj_string.c.  If you want to ditch PR and do something more major, that would be cool, but maybe just start small with this one?  As long as the headers in question are internal only, refactoring shouldn't cause issues, right?

But, any way you slice it, not having a header and using local definitions for symbols in other compilation units is likely cause a bad day from time to time

Issue #1001, opened 2018-05-18
Add support for deformation models


ccrook:
This is an enhancement proposal to provide a generic implementation of deformation models
into proj.

This briefly describes the reason for using deformation models and proposes 
a detailed format definition for encoding such a model.  The format definition 
is intended as a straw man for debate before implementation.  The intention
in detailing so precisely is to hopefully provide an unambiguous representation
of the proposal.  Also if it does provoke debate then it can be updated to converge on an implementation proposal.

## Background

Deformation models are used to relate coordinates or coordinate systems based
on physical objects to global references frames such ITRF/WGS84 used by GNSS
(global navigation satellite systems) such as GPS. 

For stable continental tectonic plates which are moving as a rigid body 
across the earth this movement can be represented as a simple rotation, 
and implemented using a helmert transformation with differential rotation 
parameters. https://proj4.org/operations/transformations/helmert.html

In some areas deformation can be described by a velocity model  For example 
GIA (glacial isostatic adjustment) in Scandinavia is well modelled by 
velocity grids.

However in many areas the secular velocity model is overlaid with complex
deformation due tectonic events such as earthquakes, which may include both
coseismic deformation (a step function) as well an ongoing post-seismic
deformation signal.

The New Zealand geodetic datum NZGD2000 is an example of a datum that 
incorporates a [deformation model](https://www.linz.govt.nz/data/geodetic-system/datums-projections-and-heights/geodetic-datums/new-zealand-geodetic-datum-2000-nzgd2000/nzgd2000-deformation-model).

While there are formats which describe aspects of deformation 
(NTv2, ctable, have been used) there is currently nothing which provides a
general complete contained model of deformation that can be implemented 
as a time dependent transformation step into a transformation chain.

## Functional model

A generic functional model which can represent most types of deformation consists of 
an arbitrary number of components, each of which consists of a 
spatial model multiplied by a piecewise linear time function.  The spatial
model is defined by an ordered sequence of one or more grids.  The displacement of the spatial model
can be defined in one of two ways, either by adding the values calculated for
each grid independently, or by calculating the value of the first grid in the sequence
which contains the calculation point.  This allows defining a set of nested grids (such
as used in the NTv2 model.  Each grid may define horizontal deformation, 
vertical deformation, or 3d deformation.

Multiple grids are provided to allow efficient representation of deformation such as 
that due to earthquakes with very variable complexity.  The spatial model 
can use a coarse grid representing the far field deformation, and
[...]

PR #1015, opened 2018-05-24
Temporal gridshifting

kbevers:
Extending the gridshift operations into the fourth dimension. The idea is to be able to use the grid shift operations as step functions when making transformations in areas subject to deformation after seismic events. This is useful when transforming coordinate from one epoch to another, e.g. in dynamic reference frames. This work is closely related to the proposed deformation model grid format in #1001.

I will expand the docs later to today, for now I am just checking that the code builds properly.


src/PJ_vgridshift.c
```diff
@@ -37,25 +44,60 @@ static LPZ reverse_3d(XYZ xyz, PJ *P) {
...
+        return pj_default_destructor (P, ENOMEM);
+    P->opaque = (void *) Q;
 
    if (!pj_param(P->ctx, P->params, "tgrids").i) {
         proj_log_error(P, "vgridshift: +grids parameter missing.");
         return pj_default_destructor(P, PJD_ERR_NO_ARGS);
     }
 
+   if (pj_param(P->ctx, P->params, "tt_final").i) {
+        Q->t_final = pj_param (P->ctx, P->params, "dt_final").f;
```
rouault: we should likely have a pj_eval_param_time(const char*) function to do the below processing, since it is repeated above.
localtime() is known on POSIX to be thread-unsafe. It is better to use localtime_r() instead, when available

src/PJ_hgridshift.c
```diff
@@ -34,21 +42,36 @@ static LPZ reverse_3d(XYZ xyz, PJ *P) {
...
 
-
-static PJ_COORD forward_4d (PJ_COORD obs, PJ *P) {
-    obs.xyz = forward_3d (obs.lpz, P);
-    return obs;
+static PJ_COORD forward_4d(PJ_COORD obs, PJ *P) {
+    struct pj_opaque_hgridshift *Q = (struct pj_opaque_hgridshift *) P->opaque;
+    PJ_COORD point = obs;
+    if (Q->t_final != 0.0  && Q->t_epoch != 0.0) {
+        if (obs.lpzt.t < Q->t_epoch && Q->t_final > Q->t_epoch)
```
rouault: shouldn't that be be <= and >= ?

Or so if t_final and t_epoch are specified, we apply the grid shift only if the observation time is below the epoch, and the epoch below the final epoch. That doesn't immediately seem obvious to me. Should probably be explained in the .rst files

docs/source/operations/transformations/vgridshift.rst
```diff
@@ -11,9 +11,11 @@ Change Vertical datum change by grid shift
 +-----------------+--------------------------------------------------------------------+
 | **Domain**      | 3D                                                                 |
 +-----------------+--------------------------------------------------------------------+
-| **Input type**  | Geodetic coordinates (horizontal), meters (vertical).              |
+| **Input type**  | Geodetic coordinates (horizontal), meters (vertical),              |
```
rouault: hgridshift.rst should also be similarly updated

src/PJ_hgridshift.c
kbevers: I have just updated the .rst files. Do they cover the topic better now?

From a physical point of view I would say that the case where `obs.lpzt.t == Q->t_epoch` is undefined, since the shift doesn't happen instantaneously but rather over a timespan of minutes to days. To me it makes the most sense to say that the offset is applied only if the observation epoch is before the event that caused the deformation. Should you happen to have acquired a coordinate *during* the deformation event it will be partly to fully shifted and we dont' really have the means to do anything about that.

A similar point can be made for the second part of the condition, but I think from a practical point of view it make sense to change it to `Q->t_final >= Q->t_epoch`.

src/PJ_vgridshift.c
kbevers: > we should likely have a pj_eval_param_time(const char*) function to do the below processing, since it is repeated above.

Good point. I intend to add similar functionality to `deformation` and possibly `helmert` as well. I will take care of it then.

> localtime() is known on POSIX to be thread-unsafe. It is better to use localtime_r() instead, when available

When can `localtime_r` be expected to be available?

src/PJ_hgridshift.c
rouault: Yes the docs make things really clear (sorry I started looking at the code before reading your PR general comment where you mentioned that would be later detailed :-) )
And actually no strong opinion if that should be strict or equal comparisons.

src/PJ_vgridshift.c
kbevers: I think it is a bit out of scope for this PR. How big a risk is it to include the code as it is now? localtime() is only called once during creation of the `PJ` object.

busstoptaktik (commented):
This review consists of just one comment about code flow and readability, but it applies 4 places :-)

src/PJ_vgridshift.c
rouault: The risk is probably low. A TODO in the code + issue to remember should be good enough

src/PJ_hgridshift.c
```diff
@@ -34,21 +42,36 @@ static LPZ reverse_3d(XYZ xyz, PJ *P) {
...
+    return point;
 }
 
-
-static PJ_COORD reverse_4d (PJ_COORD obs, PJ *P) {
-    obs.lpz = reverse_3d (obs.xyz, P);
-    return obs;
+static PJ_COORD reverse_4d(PJ_COORD obs, PJ *P) {
+    struct pj_opaque_hgridshift *Q = (struct pj_opaque_hgridshift *) P->opaque;
+    PJ_COORD point = obs;
```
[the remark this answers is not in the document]
kbevers: Good suggestion. Updated the code.

src/PJ_vgridshift.c
kbevers: Created #1016

busstoptaktik: You can probably significantly reduce (but not eliminate) the risk by saying:

    struct tm cucumber;
    ...
    cucumber = *localtime(...);

Essentially doing a copy of the the `localtime`internal `struct tm`, rather than storing a pointer to it.

kbevers:
@rouault and @busstoptaktik, thanks for the swift and helpful reviews. I am merging this now so that it can be included in 5.1.0.

PR #1018, opened 2018-05-24
gie T.fout output cleanup

schwehr:
- printf -> fprintf in more places
- puts -> fprintf
- Move list_err_codes to after where T.fout is set

This makes fuzzing easier by allowing more of the output to be sent to
/dev/null.  Found with autofuzz.

### DIFF
--- a/src/gie.c
+++ b/src/gie.c
@@ -260,12 +260,6 @@ int main (int argc, char **argv) {
         return 0;
     }
 
-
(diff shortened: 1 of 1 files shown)


kbevers:
This looks good. I'll merge when the final release of 5.1.0 is out next week.

schwehr:
Sounds good.

PR #1023, opened 2018-05-30
Fix buffer overflow in gie.c:append_args()

schwehr:
Observed a buffer overflow in append_args with autofuzz with the strcpy in append_args.
I think the +2 is required to account for both a nul char and the space.


kbevers:
> I think the +2 is required to account for both a nul char and the space.

That warrants a comment I think

kbevers:
Thank you :-)

PR #1024, opened 2018-05-31
Add googletest 1.8.0 framework in test/googletest, and use it with autoconf builds only (CMake integration to be done)

rouault:
catch.hpp removed consequently


kbevers:
I think you forgot to remove this guy: https://github.com/rouault/proj.4/blob/googletest/test/test_main.cpp

A few minor comments/questions:
- What strategy should we use for updating googletest? Once a year? When ever it is updated? Only in case of releases caused by security vulnerabilities?
- How will the packagers react to including googletest in the source-tree? I imagine some will discard it and use their own googletest packages. If I have googletest installed on my system, will that be used or the PROJ internal one?

Apart from that it looks okay to me. @schwehr @mloskot are you happy with this PR as it is now? I don't want to merge prematurely again...

rouault:
> What strategy should we use for updating googletest? Once a year? When ever it is updated? Only in case of releases caused by security vulnerabilities?

I'd say If we need it. Security vulnerabilities for a testing framework (that is to say used in controlled ways, and not delivered to actual users) aren't really relevant

> How will the packagers react to including googletest in the source-tree? I imagine some will discard it and use their own googletest packages.

Pinging @sebastic to see his thaughts about this. That would require work to do so

> If I have googletest installed on my system, will that be used or the PROJ internal one?

PROJ internal one only. I don't feel myself motivated enough to add ./configure complexity for *just* a testing framework. Looking at https://packages.debian.org/sid/amd64/googletest/filelist , it is a bit of a unusual package since it is delivered as sources only. Due to this non standard way, I can imagine that the location could vary according to distributions/systems.

Honestly the catch2 approach was much simpler to handle... I wouldn't care if we prefer to stay with it. Not sure why using catch2 would prevent anyone from using it to write tests.

kbevers:
> PROJ internal one only. I don't feel myself motivated enough to add ./configure complexity for just a testing framework. Looking at packages.debian.org/sid/amd64/googletest/filelist , it is a bit of a unusual package since it is delivered as sources only. Due to this non standard way, I can imagine that the location could vary according to distributions/systems.

I understand that. I also prefer to keep configuration as simple as possible.

> Honestly the catch2 approach was much simpler to handle... I wouldn't care if we prefer to stay with it. Not sure why using catch2 would prevent anyone from using it to write tests.

I don't think it prevents anyone from writing tests. I guess it all comes down to avoiding parallel work between PROJ and Google/@schwehr. Whether the added complexity is worth it is hard for me to judge. I guess it depends on the amount of tests that will be added to gdal-autotest2. At the moment the number of tests there is limited. Not sure if that justifies using a more complex framework.

sebastic:
For the proj Debian package I'm not likely to patch out the gtest files, and just use those included in the source tarball.

The googltest Debian package does have some patches for portability to non-Linux (GNU/Hurd & GNU/kFreeBSD) which may be required to make the proj test suite work on those architectures too.

The impact for packagers is mostly additional files for license & copyright review. Catch2 is less work in that respect.

mloskot:
@kbevers I'm away until Monday, no laptop and patchy mobile Internet. @rouault sorry for the trouble, I didn't try to force any changes. I'm sorry for the trouble caused by the extended discussion. I'll shut up. I'll contribute to any tests, regardless the framework used.

rouault:
@mloskot No problem. You weren't the source of my frustration. Enjoy your weekend.

schwehr:
Re: googletest

@sebastic Can you point me at the Debian patches?  Perhaps they can be integrated upstream?

Last I talked to the googletest folks (which was a few years back), their assumption was that each project should really have it's own local copy of googletest.  I was trying to package googletest in fink.

I can give a go at autoconf.  I will be just be slow for a while.  Looking forward to not being quite so sleep deprived sometime in the future :)

rouault:
@schwehr Not sure what you meant by "I can give a go at autoconf". In this PR I've integrated the build of the googletest lib in automake

sebastic:
> Can you point me at the Debian patches?

https://salsa.debian.org/debian/googletest/tree/master/debian/patches

schwehr:
@rouault Sorry. I am apparently too tired to follow much of anything right now.

kbevers:
I think we should get this resolved.

I am -0 on this PR. Google test seem a bit too complicated for the task for me taste but if it is favoured by the majority I think we should go with it.

rouault:
I'm mostly indifferent. From my perspective either framework should deliver what we need. The CMake integration of GTest is needed to make sure new tests are run on Windows.
By the way I've just committed a change so that cmake tests, our custom ones and the dummy one based on catch2, are actually work on AppVeyor

rouault:
> in what framework have you written the most test so far? Catch2 or Gtest?

Nothing in gtest now. A few in Catch2, but shouldn't be hard to convert to Gtest if we finally decide for it.

kbevers:
> Nothing in gtest now. A few in Catch2, but shouldn't be hard to convert to Gtest if we finally decide for it.

Okay. I was hoping that could be a deciding factor. It seems there are no strong opinions either way. It would be easier to make a decision if at least one person was super passionate about one of the frameworks :-)

mloskot:
As the catch vs gtest troublemaker, I may owe a statement. I've become passionate about @schwehr 's work, not gtest itself. I use both myself. For a moment I pushed for Kurt's work integration (I think I did similar shouting-out when Kurt arrived to GDAL with his tests). My goal is/would be to collaborate with Kurt, if gtest makes it in. Otherwise, I'm going to contribute into tests based on Catch. However, I'm far from being interested in disrupting efforts of the core development team, even if some of my posts may feel like trolling.

kbevers:
@mloskot,

> I've become passionate about @schwehr 's work, not gtest itself

For good reasons.

> Otherwise, I'm going to contribute into tests based on Catch. However, I'm far from being interested in disrupting efforts of the core development team

Kurt is en route to be a part of the "core development team" so it makes sense to keep things aligned.

> even if some of my posts may feel like trolling.

Also, we may have a few trolls in the community. You are not one of them!

All,

Should we just go for gtest and get this over with?

rouault:
okay, are we going for gtest then ? I'd like to be able to write tests in the final test framework. @mloskot still interested in writing the CMake stuff needed ?

hobu:
> On Jun 6, 2018, at 5:57 PM, Even Rouault <notifications@github.com> wrote:
> 
> @mloskot still interested in writing the CMake stuff needed ?

I can cover this too if Mat can’t.

kbevers:
> okay, are we going for gtest then ?

Yes, let's go with gtest.

mloskot:
On 7 June 2018 at 01:13, Howard Butler <notifications@github.com> wrote:

> > On Jun 6, 2018, at 5:57 PM, Even Rouault <notifications@github.com>
> wrote:
> >
> > @mloskot still interested in writing the CMake stuff needed ?
>
> I can cover this too if Mat can’t.
>
>
Yes, I can work on it later tonight (EU time), unless @hobu arrives with
the update earlier. Thanks.

mloskot:
@hobu First stap at the CMake setup for our googletest in #1034

Issue #970, opened 2018-05-02
Enable HTTPS


hobu:
GitHub pages now [supports HTTPS for custom domains](https://blog.github.com/2018-05-01-github-pages-custom-domains-https/). We will need to update some A records and retrigger the domain process within GitHub to achieve support for it.

hobu:
proj4.org DNS modifications now complete. We will need to wait for things to propagate before attempting to switch over.

hobu:
DNS looking good for me. I will attempt to retrigger the GitHub domain process in 12 hours or so.

```
[hobu@kasai proj4 (master)]$ dig +noall +answer proj4.org
proj4.org.		14399	IN	A	185.199.109.153
proj4.org.		14399	IN	A	185.199.110.153
proj4.org.		14399	IN	A	185.199.111.153
proj4.org.		14399	IN	A	185.199.108.153
```

PR #1027, opened 2018-06-02
Update URLs

mwtoews:
Most of this is essentially updating http:// to https:// related to #970

There are other updates to URLs, as noted in the commit log


kbevers:
Thanks, Mike

PR #1029, opened 2018-06-04
pj_transform.c: Use errno symbols, not numeric literals

schwehr:


src/pj_transform.c
```diff
@@ -206,12 +207,14 @@ static int geographic_to_projected (PJ *P, long n, int dist, double *x, double *
             projected_loc = pj_fwd3d( geodetic_loc, P);
             if( P->ctx->last_errno != 0 )
             {
-                if( (P->ctx->last_errno != 33 /*EDOM*/
-                        && P->ctx->last_errno != 34 /*ERANGE*/ )
+                if( (P->ctx->last_errno != EDOM /* 33 */
```
kbevers: Isn't it somewhat pointless to keep the errno values as comments?

kbevers (commented):
Thanks for taking care of this. It's a lot easier to read the code when the errno symbols are used!

src/pj_transform.c
schwehr: Yes, having them isn't useful.  I wasn't brave enough to remove them.  Will do it now.

PR #1030, opened 2018-06-05
Remove include(CTest) as unnecessary

mloskot:
The library does not require CDash server, so there is no point in generating all testing configurations (ie. Experimental, Continuous, Nightly)

[CTest](https://github.com/Kitware/CMake/blob/master/Modules/CTest.cmake) also adds implicit BUILD_TESTING option (default ON) which is not used by the library - there is custom `PROJ_TESTS`.

-----

IOW, calling [enable_testing](https://cmake.org/cmake/help/latest/command/enable_testing.html) is enough.


schwehr:
Do you need to then modify https://proj4.org/install.html#cmake where it has `Tests are run with: ctest`

mloskot:
@schwehr I don't think there is need to edit the docs. 

In order to configure tests and run them with `ctest.exe`, the only necessary configuration is like this:

```
enable_testing()
add_test(...)
add_test(...)
```

then

```
cmake ..
cmake --build .
ctest
```

The `include(CTest)` vs `ctest.exe` might be confusing, I guess - both are called "ctest", a typical CMake mess :) The latter does not require the former.  The `enable_testing` is the required bit:

https://cmake.org/cmake/help/latest/manual/ctest.1.html
> CMake-generated build trees created for projects that use the ENABLE_TESTING and ADD_TEST commands have testing support.

kbevers:
So, what is achieved by not including CTest? Faster builds? Not questioning the actual removal, just trying to understand.

rouault:
> Faster builds?

At least removal of a source of confusion. Looking at the AppVeyor logs, this doesn't indeed prevent ctest from working

kbevers:
Thanks, @mloskot. It is always nice get to unneeded code removed.

mloskot:
> So, what is achieved by not including CTest?

Cleans up large number of useless, unused and possibly confusing targets. Improves experience while developing with IDEs or `cmake --build . --target help`:



> Faster builds? 

I haven't measured, but I would expect it may save a few secs :)

schwehr:
I for one as a CMake and IntelliJ/CLion noob was totally confused.  I'm still totally confused, but about fewer things now!  And it's good to know that ctest is not yet supported in CLion:

https://twitter.com/clion_ide/status/1004101775333445633

PR #1031, opened 2018-06-05
Make gie read dms style coordinates

kbevers:
Before this PR gie would not read coordinates like

```
accept       -133d43'00.0 57d75'00.0 0.0
```

properly. Everything after `d` would be ignored and `-133.0 57.0` would be read. This is taken care of now. Not in the best way possible, but we will have to wait a bit to fix it correctly, as mentioned in the comment.

### DIFF
--- a/src/gie.c
+++ b/src/gie.c
@@ -681,7 +681,22 @@ Attempt to interpret args as a PJ_COORD.
 
     T.dimensions_given = 0;
     for (i = 0;   i < 4;   i++) {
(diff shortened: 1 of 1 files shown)


rouault:
A testcase could be useful

kbevers:
> A testcase could be useful

I know. A bunch of test cases are on the way. I discovered this problem while trying to convert the shell-script based tests in `nad/` to the gie format. Most of the input coordinates in those test are on DMS format so I needed to fix gie before moving on.

PR #1035, opened 2018-06-07
Add a scripts/gen_html_coverage.sh to generate local report of code c…  …overage with lcov/genhtml

rouault:
You can already check at https://rawgit.com/OSGeo/proj-coverage/master/coverage_html/src/index.html the result of a manual run of scripts/gen_html_coverage.sh (which can be used for local evaluation of the coverage), and ./travis/deploy_html_coverage.sh

### DIFF
--- a/scripts/gen_html_coverage.sh
+++ b/scripts/gen_html_coverage.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -eu
+rm -rf coverage_html
+lcov --directory src --capture --output-file proj.info
+genhtml -o ./coverage_html --ignore-errors source --num-spaces 2 proj.info 
--- a/travis/after_success.sh
+++ b/travis/after_success.sh
@@ -10,4 +10,10 @@ if test "$TRAVIS_SECURE_ENV_VARS" = "true" -a "$TRAVIS_BRANCH" = "master"; then
     echo "publish website";
     ./travis/add_deploy_key.sh;
     ./travis/deploy_website.sh $TRAVIS_BUILD_DIR/docs/build /tmp;
+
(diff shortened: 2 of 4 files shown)


rouault:
> Out of curosity, what does this do that coveralls isn't already doing? Local creation of hmtl coverage reports?

Yes, local creation + I find the UI of coveralls to be not so good if you want to browse the source. coveralls has still some advantages to compare evolutions of the coverage, but the genhtml output is more readable IMHO

kbevers:
>  I find the UI of coveralls to be not so good if you want to browse the source. coveralls has still some advantages to compare evolutions of the coverage, but the genhtml output is more readable IMHO

That's interesting. I find coveralls UI rather good, to be honest. That's obviously just a preference thing. Not too familiar with the genhtml output but I can see myself using it locally. For the online version I will be sticking to coveralls.

And yes, the coverage history is a cool feature of coveralls.

mloskot:
BTW, there is a pretty good alternative to coveralls, https://codecov.io - one cool feature is that it can aggregate/merge coverages from multiple builds for the same commit. For example, https://codecov.io/gh/osmcode/libosmium uses it

rouault:
@kbevers Do you see an issue in merging like this ? Or do you prefer that I disable the part in travis/after_success.sh ?

kbevers:
> Or do you prefer that I disable the part in travis/after_success.sh ?

I feel like it's just doing the same thing twice. I generally dislike that but if you feel the new coverage tool is valuable I am okay with merging.

At some point it could be worth looking into codecov.io to see if it works better overall.

rouault:
OK, I've disabled the Travis integration, and just kept the local script

PR #1038, opened 2018-06-08
Complete overhaul of HOWTO-RELEASE.

kbevers:
The release process is described in more details as well as adding
section covering release candidates and how to proceed after a
release.

`scripts/update_man.sh` updated to reflect changes in the release procedure.

### DIFF
--- a/HOWTO-RELEASE
+++ b/HOWTO-RELEASE
@@ -1,24 +1,37 @@
-	Preparing a PROJ.4 Release
-	==========================
+Preparing a PROJ Release
+===============================================================================
 
-1) Update the version number in configure.ac (in AC_INIT()).
+Preparing a PROJ release is a three-step process. Before the actual release is
+published we want to issue one or more release candidates to ensure that the
+code is in fact ready to be released. That is step one. Step two starts when a
+release candidate has been selected for promotion to final release. Step three
+updates version numbers in the master branch to avoid confusion between
+released code and code still in development.
 
-   1a) Update the version number in PROJ_BUILD_VERSION in CMakeLists.txt
-   1b) Update the version number in doc/source/conf.py
-   1c) Update version numbers in man pages in man/man1/ and man/man3/
+The three phases are described below.
 
-2) Update the version number in proj_api.h (#define PJ_VERSION).
+1 Release candidate
+===============================================================================
 
-   2a) Update PROJ_VERSION_MAJOR, PROJ_VERSION_MINOR and PROJ_VERSION_PATCH
-       in src/proj.h
+At least one release candidate is necessary before the final release is
+published. In case more than one release candidate is needed most steps below
+can be skipped depending on what prompted the release candidate 1+n. Steps 1.6
+and 1.7 are mandatory for all release candidates, otherwise it will in most
+cases only be necessary to update the NEWS section (1.3).
 
-3) Update the version number, and date in src/pj_release.c.
+Release candidates should be released about a week before the final release is
+scheduled. When preparing major version releases it may be wise to issue the
+release candidate earlier than that.
 
-4) Update the version number in the -version-info definition in
-   src/Makefile.am.  It consists of "current:revision:age".
+When a release candidate has not prompted any request for changes in 48 hours,
+a motion for promotion to final release should be issued on the PROJ mailing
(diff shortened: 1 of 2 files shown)


HOWTO-RELEASE
```diff
@@ -32,47 +45,191 @@
...
+
+2 Final release
+===============================================================================
+
+Barring a successfull promotion process we can proceed with the final release.
+
+2.1 Prepare and upload archives
+-------------------------------------------------------------------------------
+
+See step 1.6 above. Do not rename the archives with RC postfixes.
```
rouault, line 173: this assumes that no commit has made its way between the last RC and the time we promote it. For GDAL, I usually log in the download server and remove the RCx prefix from the filenames. That way the last RC and the final file are guaranteed to be binary identical

HOWTO-RELEASE
```diff
@@ -32,47 +45,191 @@
...
+-------------------------------------------------------------------------------
 
-   The script builds man pages from Sphinx docs, copies them to man/ and commits
-   them to git.
+Tag the release with:
 
-6) Add a note to the ChangeLog that a new release is being issued, and what
-   the release number is.
+```
+git tag x.y.z
```
rouault: An annotated tag is better. ```git tag -a -m "Create tag x.y.z" x.y.z```

HOWTO-RELEASE
```diff
@@ -32,47 +45,191 @@
...
+where `x.y.z` is the tag of the previous release. You may need to edit the list
+slightly since not all authors has properly configured their names in git.
+
+*Commit the changes to master.*
+
+
+1.4 Update `CITATION`
+-------------------------------------------------------------------------------
+
+If needed, update the year in `CITATION`.
```
mwtoews: the citation is also repeated in `docs/source/about.rst`, so perhaps add that here too.

HOWTO-RELEASE
```diff
@@ -32,47 +45,191 @@
...
+PROJ uses [semantic versioning](http://semver.org/). Depending on the changes
+in the release will be either a major, minor or patch release. Update the
+version numbers in the files.
 
-   https://github.com/OSGeo/proj.4/releases/new
+    - configure.ac:         Update the version number in AC_INIT().
+    - src/proj.h:           Update PROJ_VERSION_MAJOR, PROJ_VERSION_MINOR and
+                            PROJ_VERSION_PATCH.
+    - src/proj_api.h:       Update PJ_VERSION.
+    - src/pj_release.c:     Update date to the scheduled release date.
```
mwtoews: also use <4 spaces for the above four list items

mwtoews (approved):
Looks good, lots of details. Just a few minor inline comments that should be considered.

As this appears to be structured as a markdown document, I'm wondering if it should be renamed with a `.md` extension (otherwise I believe GitHub just shows an unformatted text file). This was done for `README.md`.

HOWTO-RELEASE
kbevers: That could work too. I've been doing it this way so far because it is quite easy to do. Also, I'd like to setup autotools so that the RCx postfix is added automatically when running make dist. The folder inside the archive would also have the postfix then and the renaming trick would not be so simple anymore. But yes, you have a point, there's a slight change that something was changed by mistake.

HOWTO-RELEASE
kbevers: Did not know about that feature. Updated.

HOWTO-RELEASE
```diff
@@ -1,24 +1,37 @@
...
-       in src/proj.h
+At least one release candidate is necessary before the final release is
+published. In case more than one release candidate is needed most steps below
+can be skipped depending on what prompted the release candidate 1+n. Steps 1.6
+and 1.7 are mandatory for all release candidates, otherwise it will in most
+cases only be necessary to update the NEWS section (1.3).
 
-3) Update the version number, and date in src/pj_release.c.
+Release candidates should be released about a week before the final release is
+scheduled. When preparing for big releases it may be wise to issue the release
```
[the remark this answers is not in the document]
kbevers: Probably, yes. I was really just trying to say that if there's a huge amount of changes it may be smart to plan for some more time. That could also happen for a minor version release but is less likely. I'll reword the text.

PR #1039, opened 2018-06-08
Add pj_phi2_test.

schwehr:
My first attempt at a GoogleTest based test in PROJ proper.

License intentionally changed from Apache 2 to match PROJ

Nigrated from:

https://github.com/schwehr/gdal-autotest2/blob/master/cpp/third_party/proj/pj_phi2_test.cc

### DIFF
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -57,3 +57,11 @@ target_link_libraries(proj_test_unit
   gtest
   ${PROJ_LIBRARIES})
 add_test(NAME proj_test_unit COMMAND proj_test_unit)
+
+add_executable(pj_phi2_test
+        main.cpp
+        pj_phi2_test.cpp)
+target_link_libraries(pj_phi2_test
+        gtest
+        ${PROJ_LIBRARIES})
+add_test(NAME pj_phi2_test COMMAND pj_phi2_test)
--- a/test/unit/Makefile.am
+++ b/test/unit/Makefile.am
@@ -23,3 +23,4 @@ pj_phi2_test-check: pj_phi2_test
 	./pj_phi2_test
 
 check-local: basic_test-check
+check-local: pj_phi2_test-check
--- a/test/unit/Makefile.am
+++ b/test/unit/Makefile.am
@@ -7,12 +7,19 @@ include_HEADERS = gtest_include.h
 AM_CPPFLAGS = -I$(top_srcdir)/src -I$(top_srcdir)/test -I$(top_srcdir)/test/googletest/include
 AM_CXXFLAGS = @CXX_WFLAGS@ @NO_ZERO_AS_NULL_POINTER_CONSTANT_FLAG@
 
-bin_PROGRAMS =	basic_test
+bin_PROGRAMS = basic_test
+bin_PROGRAMS += pj_phi2_test
 
 basic_test_SOURCES = basic_test.cpp main.cpp
 basic_test_LDADD = ../../src/libproj.la ../../test/googletest/libgtest.la
 
 basic_test-check: basic_test
 	./basic_test
+
+pj_phi2_test_SOURCES = pj_phi2_test.cpp main.cpp
+pj_phi2_test_LDADD = ../../src/libproj.la ../../test/googletest/libgtest.la
+
+pj_phi2_test-check: pj_phi2_test
+	./pj_phi2_test
 
 check-local: basic_test-check
--- a/test/unit/pj_phi2_test.cpp
+++ b/test/unit/pj_phi2_test.cpp
@@ -1,11 +1,11 @@
 /******************************************************************************
  *
  * Project:  PROJ
- * Purpose:  Test pj_phi2.c
+ * Purpose:  Test pj_phi2 function.
  * Author:   Kurt Schwehr <schwehr@google.com>
  *
  ******************************************************************************
- * Copyright (c) 2018, Google Inc
+ * Copyright (c) 2018, Google Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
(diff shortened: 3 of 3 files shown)


schwehr:
I expect this to take a few iterations to get correct

rouault:
Please include it in test/unit/Makefile.am and test/unit/CMakeLists.txt

schwehr:
Arg... using CLion for git I missed that it didn't add those

test/unit/Makefile.am
```diff
@@ -7,12 +7,19 @@ include_HEADERS = gtest_include.h
...
 basic_test-check: basic_test
 	./basic_test
 
+pj_phi2_test_SOURCES = pj_phi2_test.cpp main.cpp
+pj_phi2_test_LDADD = ../../src/libproj.la ../../test/googletest/libgtest.la
+
+pj_phi2_test-check: pj_phi2_test
+	./pj_phi2_test
+
 check-local: basic_test-check
```
schwehr, line 25: What is check-local?

test/unit/pj_phi2_test.cpp
```diff
@@ -0,0 +1,85 @@
+/******************************************************************************
+ *
+ * Project:  PROJ
+ * Purpose:  Test pj_phi2.c
```
mloskot: I think this should read "Test pj_phi2 function".

test/unit/CMakeLists.txt
```diff
@@ -57,3 +57,11 @@ target_link_libraries(proj_test_unit
   gtest
   ${PROJ_LIBRARIES})
 add_test(NAME proj_test_unit COMMAND proj_test_unit)
+
+add_executable(pj_phi2_test
```
rouault, line 61: I'm wondering if we need to create a dedicated test program for that ? Seems a bit overkill. We could put that as an extra source of the existing test. Same for automake

schwehr: I think we should start with separate targets.  Keep test files as isolated from each other as possible.  A test could accidentally mess up errno, the default ctx, default logging or other state.  In fact tests should be in there that do that kind of thing.

test/unit/pj_phi2_test.cpp
```diff
@@ -0,0 +1,85 @@
...
+    EXPECT_DOUBLE_EQ(M_PI_2, pj_phi2(ctx, 0.0, -1.0));
+
+    EXPECT_DOUBLE_EQ(0.0, pj_phi2(ctx, 1.0, 1.0));
+    EXPECT_DOUBLE_EQ(M_PI, pj_phi2(ctx, -1.0, -1.0));
+
+    // TODO(schwehr): M_PI_4, M_PI_2, M_PI, M_E
+    // https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html
+
+    EXPECT_DOUBLE_EQ(-0.95445818456292697, pj_phi2(ctx, M_PI, 0.0));
+    EXPECT_TRUE(isnan(pj_phi2(ctx, 0.0, M_PI)));
```
rouault: should be std::isnan() to please x86_64-w64-mingw32

schwehr:
Fixing... https://travis-ci.com/OSGeo/proj.4/jobs/128419573
```
../../../test/unit/pj_phi2_test.cpp:82:47: error: ‘isnan’ was not declared in this scope
     EXPECT_TRUE(isnan(pj_phi2(ctx, -inf, -inf)));
```

rouault:
The MSVC cmake builds with ```pj_phi2_test.obj : error LNK2019: unresolved external symbol pj_phi2 referenced in function```
The issue is that pj_phi2() isn't exported in proj.def. So perhaps this test should be linked against the static version of the library (hoping that the symbols would then be visible) ? Or if that's too annoying to do, skip that test in the MSVC cmake case ?

schwehr:
@rouault There are going to be a lot of tests like this.  I don't want to be changing what symbols are exported for externally facing builds.  How do I change the MSVC build to be static?

Anyone else have other ideas on how to handle this cleanly?

rouault:
In appveyor.yml, we do -DBUILD_LIBPROJ_SHARED=ON and looking at src/lib_proj.cmake, I've the feeling that it is either a shared or a static lib build, but not both at the same time. We do want appveyor to test DLL builds. What we could potentially do is have appveyor do 2 builds: one shared one, without that test, and on static one with that test.

rouault:
I've implemented my above proposal in https://github.com/rouault/proj.4/commit/6b0c94e47962164f20d858bda344f6fa6aeaf625 . Results in successful builds : https://ci.appveyor.com/project/rouault/proj-4/build/1.0.110 and https://travis-ci.org/rouault/proj.4/builds/389972491

test/unit/pj_phi2_test.cpp
```diff
@@ -0,0 +1,85 @@
...
+    EXPECT_TRUE(std::isnan(pj_phi2(ctx, 0.0, inf)));
+
+    EXPECT_DOUBLE_EQ(4.7123889803846897, pj_phi2(ctx, -inf, 0.0));
+    EXPECT_TRUE(std::isnan(pj_phi2(ctx, 0.0, -inf)));
+
+    EXPECT_TRUE(std::isnan(pj_phi2(ctx, inf, inf)));
+    EXPECT_TRUE(std::isnan(pj_phi2(ctx, -inf, -inf)));
+}
+
+}  // namespace
```
kbevers, line 85: weird line ending here...

mloskot:
I may be missing something, but AppVeyor already sets common `CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE` for EXEs and DLLs, so running the shared build should work fine, I'd expect.

Typically, within single build directory, it is either shared or static. 

BTW, CMake offers [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html). There should be no need for custom ` BUILD_LIBPROJ_SHARED` option.

kbevers:
Added in #1045. Closing this one.

PR #1040, opened 2018-06-08
Initial version of WKT2 support and ISO-19111:2018 / OGC Topic 2

### LINKED ISSUE #394: Please add support for EPSG:4979, 3D version of EPSG:4326
On Mac OS X, PROJ.4 version 4.9.2, EPSG:4979 is not a valid CRS identifier:

```
$ cs2cs +init=EPSG:4979
Using from definition: init=EPSG:4979 
Rel. 4.9.2, 08 September 2015
<cs2cs>: 
cause: no options found in 'init' file
program abnormally terminated
```

EPSG:4979 is cited in various RFC as the proper identifier of the 3D Geodetic WGS84, when using ellipsoid height on top of EPSG:4326, so I'd like to use the proper identifier in code handling ellipsoid heights.

For example:
- https://tools.ietf.org/html/rfc5870
- https://tools.ietf.org/html/rfc6225

This is full the definition on the EPSG registry:
http://www.epsg-registry.org/report.htm?type=selection&entity=urn:ogc:def:crs:EPSG::4979&reportDetail=long&title=WGS%2084&style=urn:uuid:report-style:default-with-code&style_name=OGP%20Default%20With%20Code

As a workaround I'm using the 2D definition "EPSG:4326" that appears to produce ellipsoid heights, but I'm not sure how stable this is.

### LINKED ISSUE #1167: Missing datum shift for DB_REF coordinate systems
Follow-up to https://gis.stackexchange.com/questions/299837/db-ref-in-qgis-position-offset :+1: 
All DB_REF based coordinate systems come without datum shift parameters.

DB_REF is similar to the German DHDN, but with a different parameter set. The offset is between 0.5 and 3 meters.

According to the EPSG registry tfm code 5826, parameters should be:

+towgs84=584.9636,107.7175,413.8067,1.1155214628,0.2824339890,-3.1384490633,7.992235

assuming that ETRS89 and WGS84 are the same.

Without any datum shift, PROJ and GDAL will not reproject correctly to any other CRS.

rouault:
It is known to be incomplete, but this is already enormous to review...
Lots of TODO and FIXME. See TODO.TXT for a non exhaustive list.

Short summary of the main changes file-wise
- include/proj directory added with the new c++ headers. Public ones are installed in $prefix/include/proj, so as to avoid pollutiong $prefix/include
- src has the new .cpp files
- test/unit has new tests
Everything indented with scripts/reformat_cpp.sh (llvm based with 4 space for indentation)

As discussed with @kbevers , in a preliminary step, that work and following will land in the OSGeo/proj.4/iso19111_dev staging branch (just forked from master), (this PR is against that branch) and when the whole work is ready, it will be merged back to master.
I'm not sure how to deal with git history wise. Probably that I will stack new commits on top of that one for the iso19111_dev branch, and when we will decide to merge it back to master, we'll  decide if we squash all the intermediate commits or not depending on how messy this history will be.

Resolves #394 
Resolves #1167


mloskot:
I see you aim to distinguish C from C++ headers. Good.
Why you prefer `.hh` instead of `.hpp` which nicely matches the three-letter `.cpp`?

include/proj/common.hh
```diff
@@ -0,0 +1,256 @@
...
+        NONE,
+        /** Angular unit of measure */
+        ANGULAR,
+        /** Linear unit of measure */
+        LINEAR,
+        /** Scale unit of measure */
+        SCALE
+    };
+
+    PROJ_DLL UnitOfMeasure(const std::string &nameIn = std::string(),
```
mloskot: Regarding `PROJ_DLL`. Do you aim to fine-tune the exportable interface by selective exporting versus exporting the whole classes?

src/internal.cpp
```diff
@@ -0,0 +1,124 @@
...
+    std::string ret(osStr);
+    for (size_t i = 0; i < ret.size(); i++)
+        ret[i] = static_cast<char>(::toupper(ret[i]));
+    return ret;
+}
+
+// ---------------------------------------------------------------------------
+
+/** Strip leading and trailing double quote characters */
+std::string stripQuotes(const std::string &osStr) {
```
mloskot: `osStr` - is this hungarian notation necessary?

I think it's redundant to say `o` for object and `s` for string. Dropping those would also help to stick to the line length limit,

mloskot: I see the naming follows GDAL hungarian notation.

Feel free to ignore my subjective earlier comment.

include/proj/common.hh
rouault: I tried initially to export whole classes but VC emits a lot of warnings due to the hidden "class Private" used for the pimpl paradigm. Exporting only methods avoid that warning.

src/metadata.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+using namespace NS_PROJ::metadata;
+using namespace NS_PROJ::internal;
+using namespace NS_PROJ::util;
+
+// ---------------------------------------------------------------------------
+
+//! @cond Doxygen_Suppress
+struct Citation::Private {
+    optional<std::string> title{};
```
mloskot, line 70: `title{};` is redundant I think, just `title;`

rouault: This is needed to make -Weffc++ happy.

include/proj/coordinatesystem.hh
```diff
@@ -0,0 +1,302 @@
...
+    PROJ_OPAQUE_PRIVATE_DATA
+
+    CoordinateSystemAxis();
+    INLINED_MAKE_SHARED
+};
+
+// ---------------------------------------------------------------------------
+
+class CoordinateSystem : public common::IdentifiedObject,
+                         public io::IWKTExportable {
```
mloskot: Dumb question: can `CoordinateSystem` be something-else-exportable?

rouault: At some point I guess there will be a IPROJStringExportable interface for things that can be converted to PROJ strings, like CRS or Conversion, but a CoordinateSystem itself probably not

src/metadata.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+void Identifier::setProperties(
+    const PropertyMap &properties) // throw(InvalidValueTypeException)
+{
+    {
+        auto oIter = properties.find(AUTHORITY_KEY);
+        if (oIter != properties.end()) {
+            auto genVal =
+                util::nn_dynamic_pointer_cast<BoxedValue>(oIter->second);
+            if (genVal) {
```
mloskot: FYI, This shorter version of if-statement with single variable declaration and initialization is C++ idiomatic, no compiler should warn:

```
if (auto genVal = util::nn_dynamic_pointer_cast<BoxedValue>(oIter->second))
{
  assert(genVal);
}
else
{
  assert(!genVal);
}
```

include/proj/util.hh
```diff
@@ -0,0 +1,396 @@
...
+    PROJ_DLL static LocalNameNNPtr createLocalName(const NameSpacePtr &scope,
+                                                   const std::string &name);
+    PROJ_DLL static GenericNameNNPtr
+    createGenericName(const NameSpacePtr &scope,
+                      const std::vector<std::string> &parsedNames);
+};
+
+// ---------------------------------------------------------------------------
+
+class CodeList {
```
mloskot: The `CodeList` feels like it could be `using CodeList = std::string` :)

rouault: Not completely. This is the base class to create enumeration of strings. See AxisDirection in coordinatesystem.hpp

include/proj/coordinateoperation.hpp
```diff
@@ -0,0 +1,366 @@
...
+    PROJ_DLL const util::optional<std::string> &operationVersion() const;
+    PROJ_DLL const std::vector<metadata::PositionalAccuracy> &
+    coordinateOperationAccuracies() const;
+
+    // In the case of a derivingConversion of a DerivedCRS, sourceCRS() and
+    // targetCRS()
+    // will return null pointers if the owning DerivedCRS has been destroyed.
+    PROJ_DLL const crs::CRSPtr sourceCRS() const;
+    PROJ_DLL const crs::CRSPtr targetCRS() const;
+    PROJ_DLL const crs::CRSPtr &interplationCRS() const;
```
kbevers: Is this a spelling error? `interpOlation`?

kbevers (commented):
This is a lot to grasp so I have only looked it over briefly and it looks good to me. It seems like the overall structure is good and that is probably the most important thing at this point. Very impressed with how much you've manage to do already!

include/proj/coordinateoperation.hpp
rouault: yes, fixed

mloskot:
Even, this is a huge and impressive overhaul indeed. It's clear the stage is very early and please don't consider my question as high priority, ignore them if answering would be too much effort.
What is relation of this implementation to the ISO 19111; what is different, what is improved, what does your implementation do not cover, etc. Basically some bit of documenting background and rationale would be helpful during reviews. Especially for those who do not analyse the ISO document thoroughly (I'd assume it's majority of reviewers).

rouault:
@mloskot  Yes I can imagine that it is hard to make sense of this without looking at the ISO-19111/OGC Topic 2 doc. At some point I'll add some documentation notes to each class to relate the C++ implementation with the standard (or document the departure). But I won't repeat the standard document, so I think readers will have to look at the UML schemas of the standard (or the hiearchy diagram generated by Doxygen) to get the whole picture.

But at that point of the implementation, there isn't much departure; this is really close. There will probably be some to model WKT2 BoundCRS concept, generalization of TOWGS84, that has no direct equivalent in the 19111 modelization. There are likely simplifications in the metadata/common packages regarding to other OGC/ISO standards since I didn't study them closely, but just put the strict minimum needed for the CRS / CoordinateSystem / CoordinateOperation packages.

Anyway I don' think most users will have to directly deal with this level of modelling. They will be likely deal with utility classes that offer higher level services: like import/export from/to WKT1,WKT2,PROJ strings, EPSG database. And the actual C++ objects will mostly be opaque to users.

mloskot:
@rouault 
> (...) so I think readers will have to look at the UML schemas of the standard 

This is fair play.

> But at that point of the implementation, there isn't much departure; this is really close

This is an important note and it should be motivating for reviewers to actually gain deeper familiarity with the standard.

Those who don't, because they can't or don't want to, may still act as extra pair of debugging eyes and  review the low level C++ aspects of the implementation.

include/proj/crs.hpp
```diff
@@ -346,6 +358,9 @@ class BoundCRS : public CRS {
              const operation::TransformationNNPtr &transformationIn);
     INLINED_MAKE_SHARED
 
+    bool isTOWGS84Compatible() const;
+    std::string getVDatumPROJ4GRIDS() const;
```
kbevers: Is there a particular reason for using including the "PROJ4" name here? Why not  ``getVDatumPROJGRIDS`` or simply ``getVDatumGrids``?

rouault: Yes, that's the name used by the GDAL WKT1 EXTENSION. The purpose here is to be compatible with the past.

```
VERT_CS["EGM2008 geoid height",
    VERT_DATUM["EGM2008 geoid",2005,
        EXTENSION["PROJ4_GRIDS","egm08_25.gtx"],
        AUTHORITY["EPSG","1027"]],
    UNIT["metre",1,
        AUTHORITY["EPSG","9001"]],
    AXIS["Up",UP],
    AUTHORITY["EPSG","3855"]]
```
The equivalent WKT2 will be
```
BOUNDCRS[
    SOURCECRS[
        VERTCRS["EGM2008 geoid height",
            VDATUM["EGM2008 geoid"],
            CS[vertical,1],
                AXIS["up",up,
                    LENGTHUNIT["metre",1]],
            ID["EPSG",3855]]],
    TARGETCRS[
        GEODCRS["WGS 84",
            DATUM["WGS_1984",
                    LENGTHUNIT["metre",1]]],
            PRIMEM["Greenwich",0,
            CS[ellipsoidal,3],
                AXIS["latitude",north,
                    ORDER[1],
                AXIS["longitude",east,
                    ORDER[2],
                AXIS["ellipsoidal height",up,
                    ORDER[3],
                    LENGTHUNIT["metre",1]],
            ID["EPSG",4979]]],
    ABRIDGEDTRANSFORMATION["EGM2008 geoid height height to WGS84 ellipsoidal height",
        METHOD["GravityRelatedHeight to Geographic3D"],
        PARAMETERFILE["Geoid (height correction) model file","egm08_25.gtx",
            ID["EPSG",8666]]]]
```

kbevers: Ah, okay. Is the function going to be used to create both WKT1 and WKT2 strings? Just wondering if we'll be stuck with a hardcoded version number in a (API?) function that will stay around for a long time.

rouault: The function is mostly to re-create GDAL WKT1 string from the internal model (using BoundCRS approach internally). And I've also used it for now to regenerate PROJ strings using the +geoidgrids params, but that might change later to use pipeline and +proj=vgridshift

rouault (commented):
@kbevers Might be of interest for you. With that changeset

```
$ PROJ_LIB=data  src/projinfo  -s EPSG:8267 -t EPSG:4909
PROJ string: 
+proj=vgridshift +grids=gvr2016.gtx

WKT2_2015 string: 
COORDINATEOPERATION["Inverse of GR96 to GVR2016 height (1)",
    SOURCECRS[
        VERTCRS["GVR2016 height",
            VDATUM["Greenland Vertical Reference 2016"],
            CS[vertical,1],
                AXIS["gravity-related height (H)",up,
                    LENGTHUNIT["metre",1]]]],
    TARGETCRS[
        GEODCRS["GR96",
            DATUM["Greenland 1996",
                    LENGTHUNIT["metre",1]]],
            PRIMEM["Greenwich",0,
            CS[ellipsoidal,3],
                    ORDER[1],
                    ORDER[2],
                AXIS["ellipsoidal height (h)",up,
                    ORDER[3],
                    LENGTHUNIT["metre",1]]]],
    METHOD["Geographic3D to GravityRelatedHeight (Gravsoft)",
        ID["EPSG",1047]],
    PARAMETERFILE["Geoid (height correction) model file","gvr2016.gtx"],
    OPERATIONACCURACY[0.1],
    AREA["Greenland - 58°N to 85°N"],
    BBOX[58,-75,85.01,-6.99],
    ID["INVERSE(DERIVED_FROM(EPSG))",8269]]
```

```
$ PROJ_LIB=data  src/projinfo  -s EPSG:5733 -t EPSG:4937
PROJ string: 
+proj=vgridshift +grids=dnn.gtx

WKT2_2015 string: 
COORDINATEOPERATION["DNN height to ETRS89",
    SOURCECRS[
        VERTCRS["DNN height",
            VDATUM["Dansk Normal Nul"],
            CS[vertical,1],
                AXIS["gravity-related height (H)",up,
                    LENGTHUNIT["metre",1]]]],
    TARGETCRS[
        GEODCRS["ETRS89",
            DATUM["European Terrestrial Reference System 1989",
                    LENGTHUNIT["metre",1]]],
            PRIMEM["Greenwich",0,
            CS[ellipsoidal,3],
                    ORDER[1],
                    ORDER[2],
                AXIS["ellipsoidal height (h)",up,
                    ORDER[3],
                    LENGTHUNIT["metre",1]]]],
    METHOD["GravityRelatedHeight to Geographic3D",
        ID["PROJ","HEIGHT_TO_GEOGRAPHIC3D"]],
    PARAMETERFILE["Geoid (height correction) model file","dnn.gtx"],
    AREA["Denmark - onshore"],
    BBOX[54.51,8,57.8,15.24],
    ID["PROJ","EPSG_5733_TO_EPSG_4937"]]
```

rouault (commented):
@busstoptaktik I've hopefully fixed the whitespaces issues in proj.h

kbevers:
In light of the ongoing RFC2 motion I figured I would take a closer look at the work that has been done so far. Again, I must stress how impressive this whole thing is. I am amazed at how much you have done in so little time!

So far I have come across two things:

1. I am doing an out of source CMake build. `proj.db` seem to be create in the source tree and not in the build tree. Is this on purpose? It doesn't seem like the correct thing to do to me. `all.sql.in` is also created in `data/` in the source tree, which I guess should either go in the build tree or be cleaned up after `proj.db` has been generated.

2. When I try to build the docs I get an error complaining about a `xml/index.xml` missing. I assume this is related to Doxygen/breathe, maybe because I need to run a script beforehand?

```
kevers@falcon:~/dev/proj/doc$ make html
sphinx-build -b html -d build/doctrees  -W source build/html
Running Sphinx v1.8.1
loading pickled environment... failed: build environment version not current
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 205 source files that are out of date
updating environment: 205 added, 0 changed, 0 removed
reading sources... [ 10%] development/reference/cpp/cpp_general
Warning, treated as error:
/Users/kevers/dev/proj/docs/source/development/reference/cpp/common.rst:6:doxygennamespace: Cannot find file: /Users/kevers/dev/proj/xml/index.xml
make: *** [html] Error 2
```

I'll expand the list if I stumble on something else later on.

rouault:
>     1. `proj.db` seem to be create in the source tree and not in the build tree. Is this on purpose?

Not really. I'm lame at CMake... I think I tried to make it create it in the build tree, but either I failed or there was an issue with "make test". In the current situation, proj.db does get installed by "make install" and "make test" works (except the Builtins2, but that's unrelated to my work), so that's not so bad. Help appreciated to improve



>     2\. maybe because I need to run a script beforehand?

Yes, you need to run "scripts/doxygen.sh" before

rouault:
>     1. `all.sql.in` is also created in `data/` in the source tree, which I guess should either go in the build tree or be cleaned up after `proj.db` has been generated.

If I remove all.sql.in, then the next "make" will recreate it again, and then recreate proj.db, which takes ~ 10 seconds each time, so that would be painful. Presumably a CMake master could improve that. And as far as generating it in the build tree, I also failed. I had to create an external script "data/generate_all_sql_in.cmake" . The automake solution is easier since I can use commands like "cat" etc, that completely avoids the temporary file, but for portability on Windows, everything becomes so complicated... Sigh

kbevers:
> Not really. I'm lame at CMake...

Yeah, me too. Can you make a list of candidates for improvements? Maybe @mloskot or someone else skilled in the ways of CMake can take a look at it and improve things.

> Yes, you need to run "scripts/doxygen.sh" before

Okay, do you mind putting that in `docs/Makefile`?

> If I remove all.sql.in, then the next "make" will recreate it again, and then recreate proj.db, which takes ~ 10 seconds each time, so that would be painful. Presumably a CMake master could improve that. And as far as generating it in the build tree, I also failed. I had to create an external script "data/generate_all_sql_in.cmake" . The automake solution is easier since I can use commands like "cat" etc, that completely avoids the temporary file, but for portability on Windows, everything becomes so complicated... Sigh

Okay, I see the problem. This is definitely a candidate for CMake improvements. Unfortunately I haven't got the skills to fix it either.

rouault:
> Can you make a list of candidates for improvements?

https://github.com/OSGeo/proj.4/issues/1165

> do you mind putting that in docs/Makefile?

I've just pushed a commit for that

kbevers:
> I've just pushed a commit for that

Thanks. Regarding the `doxygen.sh` script, `sed -i ...` behaves differently on Linux and OSX. I fixed this by changing it to `sed -i "" ...` based on the answer [here](https://stackoverflow.com/questions/29081799/sed-1-invalid-command-code-f). I don't know if this also works in gnu sed or if it is necessary to add a if checking which platform the script is run on.

Unless there's a good reason for outputting the doxygen files to the root of the repo, I would prefer that it get's created in `docs/build/`.

kbevers:
With your recent changes to the html doc build I can confirm that it works on OS X.

Locally I have set up conda environment for all PROJ doc related tools and libraries. I can dump it to a file if that is useful to anyone. It is an extension to what is currently in `docs/plot/environment.yml`.

rouault:
> I can confirm that it works on OS X.

cool. I've just added "make html" in the Travis-CI OSX config as well

kbevers:
> I've just added "make html" in the Travis-CI OSX config as well

Yeah, I saw that. I figured you were just using it as a test? Or do you intend to keep it there as a test that it works?

rouault:
> Or do you intend to keep it there as a test that it works?

Yes, to make sure we don't regress on this since apparently there are people using it ;-)

kbevers:
I decided to go through the docs as my starting point to this massive body of work.

It's cool that we can use Doxygen to generate Sphinx pages. I am not really familiar with doxygen but I'll understand it soon enough I am sure. One comment so far on the division of labour between sphinx and doxygen, is that it seems a bit backwards to me to write `src/general_doc.dox` and transform it into a `.h` file that Doxygen can parse into sphinx rst files. As far as I can tell doxygen is not needed here. A better option would be to just write this as a Sphinx rst file from the beginning. This would also make it easier to use better formatting that doesn't force everything into a `namespace` construct instead of using proper headings.

I realize that `general_doc.dox` was likely made before the addition of breathe to the doc setup. I just wanted to point out that this is an area with room for improvement and simplification (we can drop `generate_breathe_friendly_general_doc.py`).

rouault:
> It's cool that we can use Doxygen to generate Sphinx pages

Honestly I prefer the Doyxgen HTML output to what is generated by Breathe, which lacks diagrams and easier navigation. So ultimately we may want to offer the Doxygen HTML build as well on the website.

Yes the general_doc.dox modification for Breathe is a hack. The thing is that the Doxygen comments of methods refer to links in general_doc.dox (for example the references to the standards), and Breathe can't figure out those \ref unfortunately due to limitations of it. The namespace hack was the only solution I found.
Presumably for a rst only solution, we could include references to rst sections from the Doxygen section and Breathe could do the linking, but then the Doxygen output would no longer work

kbevers:
> Honestly I prefer the Doyxgen HTML output to what is generated by Breathe, which lacks diagrams and easier navigation. So ultimately we may want to offer the Doxygen HTML build as well on the website.

[Exhale](https://exhale.readthedocs.io/en/latest/index.html) seems like an interesting option to get the best of both worlds. I think it is worth looking into that at some point.

For now let's just keep what you already made. While not perfect at least it works without a bunch of problems.

rouault (commented):
@kbevers FYI https://github.com/OSGeo/proj.4/pull/1040/commits/2f37ea9ff7f884a9c71d76cdd6dc749e51499a35 also impact the C side of PROJ. Now on MSVC and Unix builds, all symbols are hidden by default, unless exported with PROJ_DLL. proj.def is no longer needed for MSVC

rouault (commented):
@kbevers FYI

rouault:
This PR will be discontinued. A "all-in-one-commit" version of it is in https://github.com/OSGeo/proj.4/pull/1175

PR #1042, opened 2018-06-13
Fileapi fgets b110119937

schwehr:
Do not scan past the end of the read data in pj_ctx_fgets
    
use-of-uninitialized-value third_party/proj4/proj/src/pj_fileapi.c:pj_ctx_fgets
    
Found with autofuzz msan

### DIFF
--- a/src/pj_fileapi.c
+++ b/src/pj_fileapi.c
@@ -188,6 +188,7 @@ char *pj_ctx_fgets(projCtx ctx, char *line, int size, PAFile file)
     long start = pj_ctx_ftell(ctx, file);
     size_t bytes_read;
     int i;
+    int max_size;
 
     line[size-1] = '\0';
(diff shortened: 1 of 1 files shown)


kbevers:
Looks good to me. Do you mind squashing the last two commits to one before I merge? We should try to have as few commits in the history that breaks the build.

mloskot:
👍 squash squash :)

schwehr:
Will do on the squashing... now to find a block of time longer than 2 minutes

schwehr:
3 became 2

mloskot:
(off topic)
@schwehr Before kids, 30 min free slot wasn't worth booting my laptop. With kids, 10-15 min breaks can make wonders - https://github.com/mloskot/qt-creator-plugin-boostbuild/blob/master/README.md#why-boostbuild-plugin-for-qt-creator

kid napping, laptop warming laps while sat next to cot bed :)

PR #1043, opened 2018-06-14
Start a `customstyle` for bibliographic references

mwtoews:
This PR follows on from #1041

Modifies reference label to use BibTeX key name, as it was last week. There is a check to ensure the BibTeX key name is just letters and numbers, for the RST link. Also, it's my preference to format names in References section with [`lastfirst`](https://bitbucket.org/pybtex-devs/pybtex/src/master/pybtex/style/names/lastfirst.py), e.g. switch from "F. W. Bessel" to "Bessel, F. W." However, if anyone has a strong preference to keep it `plain` (as it is right now), I can do this before merge.

TODO (if we get there) is to modify the reference style to be more Harvard-like, with year after name(s). This looks to be much more work, but it could be put into `bibstyle.py` too, or potentially committed upstream to Pybtex.

Other misc. minor additions in this commit is to fix an uppercase title for an entry, and add a few more file types to `.gitignore`. The `*.bak` is from JabRef.

### DIFF
--- a/docs/source/bibstyle.py
+++ b/docs/source/bibstyle.py
@@ -1,0 +1,64 @@
+###############################################################################
+#
+#  Purpose: Configure custom bibliography style for sphinxcontrib-bibtex
+#  Author:  Mike Toews <mwtoews at gmail.com>
+#
+###############################################################################
+#  Copyright (c) 2018, Mike Toews <mwtoews at gmail.com>
+#
+#  Permission is hereby granted, free of charge, to any person obtaining a
+#  copy of this software and associated documentation files (the "Software"),
+#  to deal in the Software without restriction, including without limitation
+#  the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#  and/or sell copies of the Software, and to permit persons to whom the
+#  Software is furnished to do so, subject to the following conditions:
+#
+#  The above copyright notice and this permission notice shall be included
+#  in all copies or substantial portions of the Software.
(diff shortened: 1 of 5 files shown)


docs/source/bibstyle.py
```diff
@@ -0,0 +1,64 @@
...
+
+        Raises KeyError if BibTeX key has other characters other than letters
+        and numbers.
+        """
+        label = entry.key
+        if not self.re_char_nums.match(label):
+            raise KeyError(
+                'BibTeX key must contain only letters and numbers '
+                '(found {0!r})'.format(label))
+	return label
```
kbevers: There's a tab before `return label` which trips up my Python:

```
(base) C:\dev\proj\docs>make html
sphinx-build -b html -d build/doctrees   source build/html
Running Sphinx v1.6.6

Configuration error:
There is a syntax error in your configuration file: inconsistent use of tabs and spaces in indentation (bibstyle.py, line 51)
Did you change the syntax from 2.x to 3.x?
Makefile:56: recipe for target 'html' failed
make: *** [html] Error 1
```

kbevers:
> Also, it's my preference to format names in References section with lastfirst, e.g. switch from "F. W. Bessel" to "Bessel, F. W." However, if anyone has a strong preference to keep it plain (as it is right now), I can do this before merge.

+1 to this!

kbevers:
Great. Thank you.

PR #1044, opened 2018-06-18
Set error codes in proj_create and proj_create_argv

aaronpuchert:
When there are no arguments, we set `PJD_ERR_NO_ARGS`, just like the old API does in that case.

On allocation failure we set `ENOMEM` as usual.

### DIFF
--- a/src/proj_4D_api.c
+++ b/src/proj_4D_api.c
@@ -560,13 +560,16 @@ PJ *proj_create (PJ_CONTEXT *ctx, const char *definition) {
     /* Make a copy that we can manipulate */
     n = strlen (definition);
     args = (char *) malloc (n + 1);
-    if (0==args)
+    if (0==args) {
+        proj_context_errno_set(ctx, ENOMEM);
         return 0;
+    }
     strcpy (args, definition);
 
     argc = pj_trim_argc (args);
     if (argc==0) {
         pj_dealloc (args);
+        proj_context_errno_set(ctx, PJD_ERR_NO_ARGS);
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


kbevers:
Looks good to me. Good idea to set the error numbers.

It would be good if we mentioned in the docs that the `proj_create_*` family of functions will set the error number when something goes wrong. Would you be willing to add that?

aaronpuchert:
> It would be good if we mentioned in the docs that the `proj_create_*` family of functions will set the error number when something goes wrong. Would you be willing to add that?

Do you mean the documentation in `docs/source/development/reference/functions.rst`? There is already some indication in the section for `proj_errno`: "An non-zero error codes indicates an error either with the *transformation setup* or during a transformation." (Emphasis mine.)

Just to be clear, `proj_create` does already set the error code in other cases, for example when parameters are out of bounds, or when we run out of memory. (That is, the projection constructor sets the error code then.)

kbevers:
> Do you mean the documentation in docs/source/development/reference/functions.rst?

Yes. There's some indication yes. I think it is better to state it explicitly under the `proj_create*` functions as well. If you are not up for it I'll do myself at some point - just wanted to give you the chance to do so first.

aaronpuchert:
It's probably better if you do it then. I'm not so familiar with RST and the conventions that go with it. I agree that it should be documented.

The backstory is that we kind of relied on the fact that there is an error code set if `proj_create` returns the null pointer, but that was not the case for an empty string. It seems we can not in general rely on that, since it is basically impossible to be sure that all error cases set an error code, but I wanted to fix at least these two locations.

So maybe you should write that there should be an error code set if the function returns `NULL`, but there is no guarantee for that. So the following is dangerous:

```c
targetProjection = proj_create(context, string);
if (!targetProjection)
{
    int error = proj_context_errno(context_);  // Can return 0.
    printf("%s", proj_errno_string(error));    // Returns NULL then -> segfault.
}
```

Instead, one should do the following:

```c
targetProjection = proj_create(context, string);
if (!targetProjection)
{
    const char* error = proj_errno_string(proj_context_errno(context_));
    if (error)
        printf("%s", error);
    else
        printf("unknown error");
}
```

kbevers:
> It seems we can not in general rely on that, since it is basically impossible to be sure that all error cases set an error code, but I wanted to fix at least these two locations.

What cases are not covered after your fix is applied?

aaronpuchert:
I'm not aware of any, but it seems hard to verify. Would be great to have a mechanism to ensure that, but I haven't really thought about that yet.

kbevers:
Probably, yes. I think it is fine for us to claim that an error code is set when null is returned by a `proj_create*` function. If that isn't the case in some particular situation it is a bug that should be fixed.

aaronpuchert:
Agreed, it's definitely a reasonable assumption from a user's perspective. I'll have a closer look at the error handling when I find some time. Maybe it's not as hard as I think.

Issue #993, opened 2018-05-12
Error message enum


schwehr:
The error messages currently are just #defines.  It would be good to have the option to use an enum instead.  This would also help track down where bare literals are being used for error conditions.  Also, I think there should be a symbol for no error and possibly also a we don't know code of some sort.

kbevers:
I don't understand this. Can you add some context?

schwehr:
e.g. https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogr_core.h#L273

```C++
#ifdef STRICT_OGRERR_TYPE
/** Type for a OGR error */
typedef enum
{
    OGRERR_NONE,                       /**< Success */
    OGRERR_NOT_ENOUGH_DATA,            /**< Not enough data to deserialize */
    OGRERR_NOT_ENOUGH_MEMORY,          /**< Not enough memory */
    OGRERR_UNSUPPORTED_GEOMETRY_TYPE,  /**< Unsupported geometry type */
    OGRERR_UNSUPPORTED_OPERATION,      /**< Unsupported operation */
    OGRERR_CORRUPT_DATA,               /**< Corrupt data */
    OGRERR_FAILURE,                    /**< Failure */
    OGRERR_UNSUPPORTED_SRS,            /**< Unsupported SRS */
    OGRERR_INVALID_HANDLE,             /**< Invalid handle */
    OGRERR_NON_EXISTING_FEATURE        /**< Non existing feature. Added in GDAL 2.0 */
} OGRErr;
#else
/** Type for a OGR error */
typedef int OGRErr;

#define OGRERR_NONE                0       /**< Success */
#define OGRERR_NOT_ENOUGH_DATA     1       /**< Not enough data to deserialize */
#define OGRERR_NOT_ENOUGH_MEMORY   2       /**< Not enough memory */
#define OGRERR_UNSUPPORTED_GEOMETRY_TYPE 3 /**< Unsupported geometry type */
#define OGRERR_UNSUPPORTED_OPERATION 4     /**< Unsupported operation */
#define OGRERR_CORRUPT_DATA        5       /**< Corrupt data */
#define OGRERR_FAILURE             6       /**< Failure */
#define OGRERR_UNSUPPORTED_SRS     7       /**< Unsupported SRS */
#define OGRERR_INVALID_HANDLE      8       /**< Invalid handle */
#define OGRERR_NON_EXISTING_FEATURE 9      /**< Non existing feature. Added in GDAL 2.0 */

[...]

PR #1048, opened 2018-06-19
pj_transform.c:projected_to_geographic: Convert literal -17 to PJD_ERR_NON_CONV_INV_MERI_DIST (#993)

schwehr:


kbevers:
You're really good at finding these! :)

PR #1050, opened 2018-06-20
Add proj_errno_string() to proj.def

### LINKED ISSUE #1049: proj_errno_string() link error with MSVC
MSVC linker complains definition not found for `proj_errno_string()`.

`proj_errno_string()` is missing in `proj.def` file and maybe related?

kbevers:
Resolves #1049

### DIFF
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -58,6 +58,15 @@ target_link_libraries(proj_test_unit
   ${PROJ_LIBRARIES})
 add_test(NAME proj_test_unit COMMAND proj_test_unit)
 
+add_executable(proj_errno_string_test
+  main.cpp
+  proj_errno_string_test.cpp)
+target_link_libraries(proj_errno_string_test
+  gtest
+  ${PROJ_LIBRARIES})
+add_test(NAME proj_errno_string_test COMMAND proj_errno_string_test)
+
+
 if (MSVC AND BUILD_LIBPROJ_SHARED)
 # ph_phi2_test not compatible of a .dll build
 else()
--- a/test/unit/proj_errno_string_test.cpp
+++ b/test/unit/proj_errno_string_test.cpp
@@ -1,0 +1,68 @@
+/******************************************************************************
+ *
+ * Project:  PROJ
+ * Purpose:  Unit test for proj_errno_string()
+ * Author:   Kristian Evers <kristianevers@gmail.com>
+ *
+ ******************************************************************************
+ * Copyright (c) 2018, Kristian Evers.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ ****************************************************************************/
+
+#include <cstring>
+
+#include "proj.h"
+
+#include "gtest_include.h"
+
+namespace {
+
+TEST(ProjErrnoStringTest, NoError) {
+    EXPECT_EQ(0, proj_errno_string(0));
+}
+
+TEST(ProjErrnoStringTest, ProjErrnos) {
+    EXPECT_STREQ("no arguments in initialization list", proj_errno_string(-1));
+    EXPECT_STREQ("invalid projection system error (-1000)", proj_errno_string(-1000));
+    EXPECT_STREQ("invalid projection system error (-9999)", proj_errno_string(-9999));
+    // for errnos < -9999, -9999 is always returned
+    constexpr int min = std::numeric_limits<int>::min();
+    EXPECT_STREQ("invalid projection system error (-9999)",proj_errno_string(min));
+    EXPECT_STREQ("invalid projection system error (-9999)", proj_errno_string(-10000));
+}
+
+TEST(ProjErrnoStringTest, SystemErrnos) {
+    constexpr int max = std::numeric_limits<int>::max();
+
+#ifdef HAVE_STRERROR
+    EXPECT_STREQ(strerror(5), proj_errno_string(5));
+    EXPECT_STREQ(strerror(9999), proj_errno_string(9999));
+    EXPECT_STREQ(strerror(10000), proj_errno_string(10000));
+    EXPECT_STREQ(strerror(max), proj_errno_string(max));
+#else
+    EXPECT_STREQ("no system list, errno: 5\n", proj_errno_string(5));
+    EXPECT_STREQ("no system list, errno: 9999\n", proj_errno_string(9999));
+    // for errnos > 9999, 9999 is always returned
+    EXPECT_STREQ("no system list, errno: 9999\n", proj_errno_string(10000));
+    EXPECT_STREQ("no system list, errno: 9999\n", proj_errno_string(max));
+#endif
+}
+
+}  // namespace
--- a/src/proj.def
+++ b/src/proj.def
@@ -157,3 +157,5 @@ EXPORTS
     proj_context_errno             @141
 
     pj_isnan                       @142
+
+    proj_errno_string              @143
--- a/test/unit/Makefile.am
+++ b/test/unit/Makefile.am
@@ -9,6 +9,7 @@ AM_CXXFLAGS = @CXX_WFLAGS@ @NO_ZERO_AS_NULL_POINTER_CONSTANT_FLAG@
 
 bin_PROGRAMS = basic_test
 bin_PROGRAMS += pj_phi2_test
+bin_PROGRAMS += proj_errno_string_test
 
 basic_test_SOURCES = basic_test.cpp main.cpp
 basic_test_LDADD = ../../src/libproj.la ../../test/googletest/libgtest.la
@@ -22,5 +23,11 @@ pj_phi2_test_LDADD = ../../src/libproj.la ../../test/googletest/libgtest.la
 pj_phi2_test-check: pj_phi2_test
 	./pj_phi2_test
 
+proj_errno_string_test_SOURCES = proj_errno_string_test.cpp main.cpp
+proj_errno_string_test_LDADD= ../../src/libproj.la ../../test/googletest/libgtest.la
+
+proj_errno_string_test-check: proj_errno_string_test
+	./proj_errno_string_test
+
 check-local: basic_test-check
-check-local: pj_phi2_test-check
+check-local: pj_phi2_test-check proj_errno_string_test-check


rouault:
May I suggest to add a call to the function in https://github.com/OSGeo/proj.4/blob/master/test/unit/basic_test.cpp ?

kbevers:
Yes you may. I'll see if I can find the time to add those later today.

test/unit/CMakeLists.txt
```diff
@@ -52,7 +52,8 @@ unset(_save_cxx_flags)
 #
 add_executable(proj_test_unit
   main.cpp
-  basic_test.cpp)
+  basic_test.cpp
+  proj_errno_string_test.cpp)
```
schwehr: I strongly encourage you to not combine compilation units into the same test target.  You now must make absolutely sure that proj_errno_string_test.cpp leaves any global in a state that exactly matches as if the tests were never run.  If you fail to cleanup or have a test barf, you risk making the problem much harder to follow.  Also, if you separate them out, you can run them in parallel in separate processes... I need to figure out how to parallelize that.  I'm guess that this doesn't parallelize the unit tests :(

```
rm -rf build-ninja; (mkdir -p build-ninja && cd build-ninja && cmake -GNinja .. && cmake --build . && ctest .)
```

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,57 @@
+/******************************************************************************
+ *
+ * Project:  PROJ
+ * Purpose:  Unit test for proj_errno_string()
+ * Author:   Kurt Schwehr <schwehr@google.com>
```
schwehr: Please change the author.

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace {
+
+TEST(ProjErrnoStringTest, Basic) {
+
+    // Test the no error result
+    ASSERT_EQ(0, proj_errno_string(0));
+
+    // Test both valid and invalid PROJ error numbers
+    ASSERT_STREQ("no arguments in initialization list", proj_errno_string(-1));
```
schwehr: ASSERT_STREQ -> EXPECT_EQ.  Try to reserve asserts for things were it is totally unsafe to proceed.  By using expect, you get to keep generating test output and fail later.  That often makes debugging easier when I do something dumb to the code under test.

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+    ASSERT_STREQ("invalid projection system error (-9999)", proj_errno_string(-9999));
+    ASSERT_STREQ("invalid projection system error (-9999)", proj_errno_string(-99999));
+    ASSERT_STREQ("invalid projection system error (-1000)", proj_errno_string(-1000));
+
+    // Test POSIX error numbers
+    #ifdef HAVE_STRERROR
```
schwehr: I prefer to always have # start at column 0... it's not C++ code.  No indent is a big warning to my brain to beware the C preprocessor.

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+namespace {
+
+TEST(ProjErrnoStringTest, Basic) {
+
+    // Test the no error result
+    ASSERT_EQ(0, proj_errno_string(0));
+
+    // Test both valid and invalid PROJ error numbers
+    ASSERT_STREQ("no arguments in initialization list", proj_errno_string(-1));
+    ASSERT_STREQ("invalid projection system error (-9999)", proj_errno_string(-9999));
```
schwehr: This is the kind of place where a gmock matcher is useful.  Match part of the message and in a separate expect, match the number.  You can then even make a range based for loop with a std::initializer_list and say something like this that should be less brittle.

```
#include <limits>

// Warning: untested code...
for (const int val : {-1, -1000, -9999, std::numeric_limits<int>::min()}) {
  // Messages are of the form:
  //     invalid projection system error (-1000)
  const char *msg = proj_errno_string(val);
  EXPECT_THAT(msg, testing::ContainsRegex("invalid projection"));
  // Do this part better.  I can't just use my normal absl.
  char buf[100];
  sprintf(buf, "(%d)", val);
  EXPECT_THAT(msg, testing::ContainsRegex("invalid projection"));
}
```

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+    ASSERT_STREQ("invalid projection system error (-9999)", proj_errno_string(-99999));
```
schwehr: Whoops... `-9999` != `-99999`

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "gtest_include.h"
+
+namespace {
+
+TEST(ProjErrnoStringTest, Basic) {
+
+    // Test the no error result
+    ASSERT_EQ(0, proj_errno_string(0));
+
+    // Test both valid and invalid PROJ error numbers
```
schwehr: More small tests with good names means that we don't need comments...
```
TEST(ProjErrnoStringTest, NegativeErrorNumbers) {
```

schwehr (changes requested):
My initial pj_phi2_test.cpp wasn't particularly good. Sorry for a crap example. It needs major help. Most of these comments apply to that test file too.

test/unit/proj_errno_string_test.cpp
kbevers: This is on purpose. The function returns -9999 on errnos < -9999. I can change it to another number that is less confusing.

test/unit/proj_errno_string_test.cpp
kbevers: Okay, that makes sense. Thanks.

test/unit/proj_errno_string_test.cpp
kbevers: I just did a search for how to test if two strings are equal and this came up. I'll change the ASSERTs to EXPECTs.

test/unit/proj_errno_string_test.cpp
kbevers: Don't want credit for my crappy test? :)

test/unit/CMakeLists.txt
kbevers: Okay, I didn't realise what I was doing here. I'll change this to a separate test target.

test/unit/proj_errno_string_test.cpp
schwehr: Then a comment so the reader knows that without having to hunt for it.

test/unit/CMakeLists.txt
rouault: There will be a significant cost in term of linking time if we add a binary for each test we write. For my C++ SRS stuff, I intend to have just one test binary. Normally PROJ shouldn't cause global side effects, except perhaps the global default context (but in that case we should have a way to reset it between runs)

test/unit/proj_errno_string_test.cpp
schwehr: Yes, _STREQ is the way for strings, but exact full matches on strings are brittle.  What if later you rephrase the message or change parens to square brackets, yada yada.  I recommend trying to find the most stable parts and probe those.

test/unit/CMakeLists.txt
schwehr: I've not done testing at scale with cmake or autoconf much, so I differ to @rouault .  Go with one test bin and if the world gets too painful, we can pull out a key test file or two to separate binaries.  I just want to take advantage of the large number of cores that workstations are starting to have (I've got 28 physical cores for 56 hyperthreads), but I know that the web based CI systems will go really slow if link time is large.

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ ****************************************************************************/
+
+#ifdef HAVE_STRERROR
```
schwehr: I'm wondering if we should leave off the `#ifdef` for system headers like this.  It's like pulled in elsewhere anyway.

kbevers: It was mostly meant as a communicate device to say that cstring is only needed in certain cases. I am sure it is included elsewhere. I can remove it if it is just a distraction.

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+TEST(ProjErrnoStringTest, NoError) {
+    EXPECT_EQ(0, proj_errno_string(0));
+}
+
+TEST(ProjErrnoStringTest, ProjErrnos) {
+    EXPECT_STREQ("no arguments in initialization list", proj_errno_string(-1));
+    EXPECT_STREQ("invalid projection system error (-9999)", proj_errno_string(-9999));
+    // for errnos < -9999, -9999 is always returned
+    EXPECT_STREQ("invalid projection system error (-9999)", proj_errno_string(-50000));
```
schwehr: I would suggest using `-10000` and `std::numeric_limits<int>::min()` to cover the range of possible values.  Using one past makes sure the transition happens where you think it does. 
 And somethings bad things happen at the extremes.

test/unit/CMakeLists.txt
kbevers: I am indifferent at the moment - I'll just go with whatever you guys decide is best. I might come back with a stronger opinion after having played around with the test framework for a while.

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+    EXPECT_STREQ("invalid projection system error (-1000)", proj_errno_string(-1000));
+}
+
+TEST(ProjErrnoStringTest, SystemErrnos) {
```
schwehr: Is there a maximum positive value that can be used?  What if I call it with `std::numeric_limits<int>::max()` ?

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+#ifdef HAVE_STRERROR
+    EXPECT_STREQ(strerror(5), proj_errno_string(5));
+    EXPECT_STREQ(strerror(9999), proj_errno_string(9999));
```
schwehr: So the behavior over `9999` is different between having `HAVE_STRERROR` and not?

test/unit/proj_errno_string_test.cpp
schwehr: I think it might be a distraction, but maybe @rouault or @mloskot could offer an opinion on if it's helpful or just noise.  Or just go with what you think is best.  It's not really that important :)

test/unit/proj_errno_string_test.cpp
kbevers: I assume so, yes. Haven't checked the docs for `strerror()` but `proj_errno_string()`[0] returns the error message verbatim from `strerror`.

[0] really it is `pj_strerrno()` that does it. `proj_errno_string` is just a wrapper.

test/unit/proj_errno_string_test.cpp
kbevers: I'll just remove it

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+TEST(ProjErrnoStringTest, NoError) {
+    EXPECT_EQ(0, proj_errno_string(0));
+}
+
+TEST(ProjErrnoStringTest, ProjErrnos) {
+    EXPECT_STREQ("no arguments in initialization list", proj_errno_string(-1));
+    EXPECT_STREQ("invalid projection system error (-1000)", proj_errno_string(-1000));
+    EXPECT_STREQ("invalid projection system error (-9999)", proj_errno_string(-9999));
+    // for errnos < -9999, -9999 is always returned
+    int min = std::numeric_limits<int>::min();
```
schwehr: constexpr isn't super important here, but it's a good habit.

`constexpr int min = std::numeric_limits<int>::min();`

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+    EXPECT_STREQ("invalid projection system error (-9999)",proj_errno_string(min));
+    EXPECT_STREQ("invalid projection system error (-9999)", proj_errno_string(-10000));
+}
+
+TEST(ProjErrnoStringTest, SystemErrnos) {
+    int max = std::numeric_limits<int>::max();
```
schwehr: `constexpr int`

test/unit/proj_errno_string_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+#ifdef HAVE_STRERROR
+    EXPECT_STREQ(strerror(5), proj_errno_string(5));
+    EXPECT_STREQ(strerror(9999), proj_errno_string(9999));
+    EXPECT_STREQ(strerror(10000), proj_errno_string(10000));
+    EXPECT_STREQ(strerror(max), proj_errno_string(max));
+#else
+    EXPECT_STREQ("no system list, errno: 5\n", proj_errno_string(5));
+    EXPECT_STREQ("no system list, errno: 9999\n", proj_errno_string(9999));
+    // for errnos > 9999, 9999 is always returned
+    EXPECT_STREQ("no system list, errno: 9999\n", proj_errno_string(max));
```
schwehr: Switch order for maximum clarity... 9999, 10000, max.

mloskot:
Why `basic_test.cpp` is needed at all?

schwehr:
I would argue that `basic_test.cpp` isn't necessary once we have a range of tests in place, but no need to rush into removing it.

rouault:
Looks good to me

kbevers:
Thanks for your reviews everyone. They was most helpful!

PR #1051, opened 2018-06-20
Introduce proj_strtod_test.cpp.

schwehr:
Also cleanup proj_strtod.h to work with C++

### DIFF
--- a/src/proj_strtod.h
+++ b/src/proj_strtod.h
@@ -1,4 +1,17 @@
 /* Internal header for proj_strtod.c */
 
+#ifndef PROJ_STRTOD_H
+#define PROJ_STRTOD_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 double proj_strtod(const char *str, char **endptr);
 double proj_atof(const char *str);
+
+#ifdef __cplusplus
+}
+#endif
+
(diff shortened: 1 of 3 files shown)


schwehr:
WARNING: This does not yet work on my local machine.

kbevers (commented):
Looks good to me. Haven't looked into why the test is broken but the overall structure is clear to me.

Beware that currently we have two(!) versions of `proj_strtod()`. It might be useful to add a comment on which one of the two is used.

I see you've decided to include this into the same compilation unit as the basic test.  Is this how we should do be doing it contrary to your initial suggestion of keeping them in separate executables?

Also, I would suggest to remove the test code in `src/proj_strtod.c` once you are sure to cover the same cases as it touches. It seems that you have already most, if not all, covered already although slightly different input is used.

kbevers:
I think perhaps Travis is failing because `test/unit/proj_strtod_test.cpp` isn't included in the autoconf makefiles. On Travis we test that the source tarball can be build with both autotools and CMake. Currently the tarball does not include this test file.

kbevers:
This is horribly out of date. @schwehr I am closing this as it seems you won't get round to cleaning this up - feel free to reopen if/when you find renewed interest in this in the future.

Issue #1053, opened 2018-06-21
Enhancing unitconvert to support arbitrary conversion factor


rouault:
Currently, +proj=unitconvert only supports a fixed set of linear units. The EPSG database currently supports 59 length units, whereas pj_unit 21 (and potentially someone could create a WKT with a completely custom unit)
In the old pj_init() interface there was the +to_meter= and +vto_meter= options to be able to specify an arbitrary conversion factor.
Why not reinteroducing this capability in +proj=unitconvert with +xy_in_factor, +xy_out_factor, +z_in_factor and +z_out_factor whose values would be te conversion factor to metre ?

rouault:
Or just keep existing parameter names, but accept numeric values instead of the unit names

kbevers:
`+to_meter` and `+vto_meter` works with the new API:

```
$ echo 12 56 10 | cct +proj=merc
 1335833.8895   7522963.2413       10.0000           inf
$ echo 12 56 10 | cct +proj=merc +to_meter=2.0 +vto_meter=2.0
  667916.9448   3761481.6206        5.0000           inf
```

`unitconvert` was really only introduced to handle time units. It was easy to add distance units as well so I did. I have been thinking instead introducing a `+tunits` parameter for handling time units instead of `unitconvert`.

rouault:
OK, I actually tried with +proj=cart and they are ignored
```
$ echo "90 0 0" | src/cct +proj=pipeline +step +proj=cart +a=1000 +b=1000  +to_meter=1000 +vto_meter=1000
       0.0000      1000.0000        0.0000           inf
```
Same for +proj=geocent

rouault:
ok, inv_prepare() and fwd_finalize() do ignore fr_meter/to_meter/etc. in the PJ_IO_UNITS_CARTESIAN case. That should be fixed right ?

rouault:
Actually for cartesian, we should only take into account +to_meter and apply it to the 3 axis (doesn't make sense to use one unit for XY, and another one for Z)

kbevers:
> ok, inv_prepare() and fwd_finalize() do ignore fr_meter/to_meter/etc. in the PJ_IO_UNITS_CARTESIAN case. That should be fixed right ?

Not entirely sure it should. Maybe. In what case do you need to change the units of cartesian coordinates? 

> Actually for cartesian, we should only take into account +to_meter and apply it to the 3 axis (doesn't make sense to use one unit for XY, and another one for Z)

[...]

PR #1054, opened 2018-06-21
Add support for deg, rad and grad in unitconvert (fixes #1052), and document that it supports numeric factors (refs #1053)

### LINKED ISSUE #1052: Enhancing unitconvert to support angular units ?
I'm in the process of adding a "export to PROJ string" method to my ISO-19111 based objects, and I'm considering in particular the export of GeographicCRS
For a few esoteric / old GeographicCRS, the unit of the ellipsoidal coordinate system is not degree. For example EPSG:4807 / "NTF (Paris)" uses grad. So users should normally provide coordinates in grad. A possibility to handle that would to extend the xy_in and xy_out parameters to support angular units (probably restricted to radian, grad and degree for now)

The export of EPSG:4807 would then be
```
+proj=pipeline +step +proj=unitconvert +xy_in=grad +xy_out=radian +step=axisswap +order=2,1 +step +proj=longlat +ellps=clrk80ign +pm=paris
```

Thoughts ?

rouault:


docs/source/operations/conversions/unitconvert.rst
```diff
@@ -124,12 +128,30 @@ The same list can also be produced on the command line with :program:`proj` or
 | ind-ch   | Indian Chain                    |
 +----------+---------------------------------+
 
+.. _angular_units:
+
+Angular units
+###############################################################################
+
+In the table below all angular units supported by PROJ `unitconvert` are listed.
+(since PROJ 5.2.0)
```
kbevers: Please use `.. versionadded:: 5.2.0` instead. Place it below the section header. That's consistent with the rest of the docs.

docs/source/operations/conversions/unitconvert.rst
```diff
@@ -41,25 +41,29 @@ expected to be in units of decimalyears. This can be fixed with `unitconvert`::
...
 ################################################################################
 
-.. option:: +xy_in=<unit>
+.. option:: +xy_in=<unit> or <conversion_factor>
 
-    Horizontal input units. See :ref:`distance_units` for a list of available
-    units.
+    Horizontal input units. See :ref:`distance_units` and :ref:`angular_units`
+    for a list of available units. `<conversion_factor>` is the conversion factor
+    from the input unit to the International System unit.
```
kbevers: Perhaps just say that it is the conversion factor to meters? Otherwise I would say that "SI unit" is closer to standard nomenclature. Also valid below.

src/PJ_unitconvert.c
```diff
@@ -382,6 +382,49 @@ static PJ_COORD reverse_4d(PJ_COORD obs, PJ *P) {
...
+pj_angular_units[] = {
+    {"rad",         "1.0",                   "Radian", 1.0},
+    {"deg",         "0.017453292519943296",  "Degree", DEG_TO_RAD},
+    {"grad",        "0.015707963267948967",  "Grad",   GRAD_TO_RAD},
+    {NULL,          NULL,                     NULL,    0.0}
+};
+
+
+/***********************************************************************/
+static double getUnitConversionFactor(const char* name,
```
kbevers: Good idea to add this function. Makes the code a bit easier to read. Small style nitpick: The rest of the functions use snake case so it looks a bit odd that this is named in camel case.

kbevers (changes requested):
What happens if I do something like `+proj=unitconvert +xy_in=us-ft +xy_out=grad` or `+proj=unitconvert +z_in=m +z_out=deg`?

Apart from that consideration I only have a few minor details to correct.

docs/source/operations/conversions/unitconvert.rst
rouault: "conversion factor to meters"... if it is linear unit, otherwise it is to radian. I can add that precision

rouault:
> What happens if I do something like +proj=unitconvert +xy_in=us-ft +xy_out=grad or +proj=unitconvert +z_in=m +z_out=deg

Non-sense, but should we try to protect against this ?

docs/source/operations/conversions/unitconvert.rst
rouault: Done

src/PJ_unitconvert.c
rouault: Done

docs/source/operations/conversions/unitconvert.rst
rouault: Done

kbevers:
> Non-sense, but should we try to protect against this ?

I didn't think it would be necessary to protect against creation of non-sense pipelines but of course there were complaints that you could do just that. This is somewhat similar so maybe it is better to just take care of that in advance. It is not a top priority for me - it can also be handled by explaining the behaviour in the docs.

docs/source/operations/conversions/unitconvert.rst
```diff
@@ -124,12 +132,31 @@ The same list can also be produced on the command line with :program:`proj` or
...
 +----------+---------------------------------+
 
+.. _angular_units:
+
+Angular units
+###############################################################################
+
+In the table below all angular units supported by PROJ `unitconvert` are listed.
+
+.. versionadded:: 5.2.0
```
kbevers: If you put it right below the heading it is more clear that it is the whole section that was added in version 5.2.0

rouault: fixed

rouault:
ok, I've added a consistency check for the unit types in xy_in vs xy_out and z_in vs z_out

src/projects.h
```diff
@@ -583,6 +583,7 @@ enum deprecated_constants_for_now_dropped_analytical_factors {
 #define PJD_ERR_ELLIPSOIDAL_UNSUPPORTED -56
 #define PJD_ERR_TOO_MANY_INITS          -57
 #define PJD_ERR_INVALID_ARG             -58
+#define PJD_ERR_INCONSISTENT_UNIT       -59
```
kbevers, line 586: You should add a string representation in `src/pj_strerrno.c` as well

kbevers:
Add the string representation of the new error code and I am happy with this. It's a nice new feature to have!

kbevers (approved):

src/projects.h
rouault: Good point. I've added a comment at the end of the enumeration to update pj_strerrno.c AND pj_transform.c as well (and I've sanitized the access to the transient_error array in it)

PR #1058, opened 2018-06-25
Add inverse lagrange projection

kbevers:
Courtesy of Michael Stumpf <mi12st34@gmail.com>


kbevers:
The failing scenarios should also be tested. I will add those later today or tomorrow when I have time.

src/PJ_lagrng.c
```diff
@@ -39,27 +41,53 @@ static XY s_forward (LP lp, PJ *P) {           /* Spheroidal, forward */
...
+    double c, x2, y2p, y2m;
+
+    if (fabs(fabs(xy.y) - 2.) < TOL) {
+        lp.phi = xy.y < 0 ? -M_HALFPI : M_HALFPI;
+        lp.lam = 0;
+    } else {
+        x2 = xy.x * xy.x;
+        y2p = 2. + xy.y;
+        y2m = 2. - xy.y;
+        if (fabs(c = y2p * y2m - x2) < TOL) {
```
schwehr: Please don't do assignments in ifs.

kbevers: Oh yeah, I'll fix that right away

PR #1063, opened 2018-06-27
Declutter frontpage

kbevers:
A first go at resolving https://github.com/OSGeo/proj.4/issues/1056

At the moment the News site will also be included in the PDF. It wasn't before. I don't think it fits particularly well in the PDF but I haven't found a way to exclude certain chapters from the PDF. It's not the end of the world so for now I am fine with it being included in the PDF...


kbevers:
Most of the text on the frontpage is grabbed from the about section.

hobu (approved):
looks great!

kbevers:
Great. I'll merge it right away and leave my computer until Monday morning!

PR #1065, opened 2018-06-28
Helmert: consider that xyzt.t == HUGE_VAL means t_epoch

rouault:
Currently when doing
```
echo "2 49 0" | src/cct  +proj=pipeline +ellps=GRS80 +step +proj=cart +step +proj=helmert +x=10 +y=3 +z=1
```
we get as a result:
```
         -nan           -nan          -nan           inf
```
This is due to cct initializing the xyzt.t to HUGE_VAL

### DIFF
--- a/src/PJ_helmert.c
+++ b/src/PJ_helmert.c
@@ -431,8 +431,9 @@ static PJ_COORD helmert_forward_4d (PJ_COORD point, PJ *P) {
 
     /* We only need to rebuild the rotation matrix if the
      * observation time is different from the last call */
-    if (point.xyzt.t != Q->t_obs) {
-        Q->t_obs = point.xyzt.t;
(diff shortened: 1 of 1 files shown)


rouault:
Actually digging more, the particular case is needed since in update_parameters() we multiply the differencial values by dt. But when those differencial values are 0 and dt == HUGE_VAL, the multiplication of both result in a NaN.

kbevers:
I think this looks alright. There's a grey area when doing something like `echo 1 2 3 | cct +proj=helmert +x=10 +dx=`. Here a t-coordinate is necessary to do the transformation correctly but not supplying it will default to the time-invariant version of the Helmert. 
Perhaps it would be better to fail in those cases. Which was probably my intention with the way the code was written initially (although not considering all cases of possible "bad" input data).

PR #1069, opened 2018-07-08
improve lagrange inverse projection

StumpfMichael:
Error.


StumpfMichael:
Error.

rouault:
you don't have to close a pull request that fails the tests. You can add new commits to it and push them, and this will update the pull request. Or you can git commit --amend and force push and this will also update

kbevers:
The reason for the failed test, I believe, is that the error condition that was removed in this PR is being tested against. Should simply be a matter of adjusting the test to reflect the new behaviour of the code. The tests are located here:

https://github.com/OSGeo/proj.4/blob/81bfa7551ffb3f0406760f566c50b84370c3d020/test/gie/builtins.gie#L2412-L2467

Issue #1071, opened 2018-07-10
vertcon .gtx grids are in millimeters instead of meters


rouault:
Investigating vertical datum transformations, I tried the gtx grids of https://download.osgeo.org/proj/vdatum/vertcon/ and came to the same conclusion as http://lists.maptools.org/pipermail/proj/2011-December/006041.html , that is that the vertical offsets in the grids is in millimeters and not in meters. This comes directly from the raw vertconc.94, vertcone.94 and vertconw.94 files of https://www.ngs.noaa.gov/PC_PROD/VERTCON/VERTCON.zip that are themselves in millimeters (the Fortran code with them does a x 0.001 conversion). There is nothing in the .gtx header for the vertical unit, so we are a bit stuck.
There are 2 options I can see:
- regenerate the grids with values in meters instead of millimeters (this shouldn't cause too much issue since the values are in Float32 and presumably only 3 decimal figures are signficant, so Float32 precision is enough)
- leave them as such, and as ad-hoc code in the PROJ code to detect this situation (we can check the filenames + one value in them to ensure it is the expected value in millimeters) and apply on the fly the mm to m conversion.

Thoughts ?

kbevers:
I don't like the second option. I think re-creating the grids using meters instead of milimeters is the way to go. At the same time they could be added to [OSGeo/proj-datumgrid]( https://github.com/OSGeo/proj-datumgrid) in a new NorthAmerica sub-package. It's probably a good idea to change the names slightly. That also gives you a chance to better clarify that the units have changed.

rouault:
OK, another issue is the sign convention. The original grid and thus the .gtx is the offset to add to NGVD 29 to get a NAVD 88 height, and this matches the https://www.epsg-registry.org/export.htm?gml=urn:ogc:def:coordinateOperation:EPSG::7969 transformation method. But in the forward_3D() of vgridshift, we actually substract values. I guess I can workaround this in my WKT to PROJString code to use the inverse step...

> a new NorthAmerica sub-package

You mean there https://github.com/OSGeo/proj-datumgrid/tree/master/north-america ? Or a subdirectory of it ?

kbevers:
[...]

PR #1072, opened 2018-07-10
vgridshift: add a +multiplier={value}

rouault:
As mentionned in #1071, it is often unclear how the offset of a vertical grid
is applied.

### DIFF
--- a/docs/source/operations/transformations/vgridshift.rst
+++ b/docs/source/operations/transformations/vgridshift.rst
@@ -113,3 +113,19 @@ Optional
 .. include:: ../options/t_final.rst
 .. versionadded:: 5.1.0
 
+.. option:: +multiplier=<value>
+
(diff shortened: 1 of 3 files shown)


rouault:
Another possibility would be to add a +multiplier parameter, and then people would be able to control both direction and multiplier value (in case of non-meter units)

kbevers:
I think the `operation` parameter is too easily confused with a "coordinate operation". Especially when used in gie files: `operation proj = vgridshift grids = ... operation = add`. I'm leaning towards the `+multiplier` parameter. It would also fix your original mm problem.

rouault:
> I'm leaning towards the +multiplier parameter.

Done

kbevers:
Looks good to me. We should reconsider if it makes sense to add the vertcon grids with adjusted units to proj-datumgrids.

PR #1080, opened 2018-07-20
Implementation of webmerc projection

dtutic:
TBD: webmerc is projection tied to WGS84 ellipsoid. Ideal would be that it is specified only with +webmerc without any additional parameter. Also if +R is given together with +ellps, output of -V option prints that +ellps is ignored, but correct final figure is printed. In projection implementation WGS 84 is forced no matter of values of +ellps or +R but this is not communicated with user who tries to use some other ellipsoid or sphere.

1. Proposal is to rename full name of webmerc projection to "WGS 84 Web Mercator / WGS 84 Pseudo Mercator" instead of "Web Mercator / Pseudo Mercator"
REASON: Official name of projection is "WGS 84 / Pseudo Mercator", see https://epsg.io/3857. Projection is tied to WGS 84 ellipsoid.

2. webmerc projection for now is implemented as projection of a sphere. Consequences are that projection factors are calculated for Mercator projection of a sphere (i.e. no angular distortion) e.g.:

```
./proj +proj=webmerc +ellps=WGS84 -V
#Web Mercator / Pseudo Mercator
#	Cyl, Sph
#	
# +proj=webmerc +ellps=WGS84
#Final Earth figure: sphere
#  Radius: 6378137.000
0 0
Longitude: 0dE [ 0 ]
Latitude:  0dN [ 0 ]
Easting (x):   0.00
Northing (y):  0.00
Meridian scale (h) : 1.00000000  ( 2.319e-09 % error )
Parallel scale (k) : 1.00000000  ( 0 % error )
Areal scale (s):     1.00000000  ( 2.319e-09 % error )
Angular distortion (w): 0.000
Meridian/Parallel angle: 90.00000
Convergence : 0d [ -0.00000000 ]
Max-min (Tissot axis a-b) scale error: 1.00000 1.00000
```

These are not properties of Web Mercator projection which is non-conformal projection. This pull request gives implementation for ellipsoid, e.g.

```
PJ *PROJECTION(webmerc) {

    /* Overriding k_0, lat_0, lon_0, x_0 and y_0 with fixed parameters */
    P->k0 = 1.0;
    P->phi0 = 0.0;
    P->lam0 = 0.0;
    P->x0 = 0;
    P->y0 = 0;

    /* Overriding ellipsoid and using WGS 84 */
    P->f = 1.0/298.257223563;
    P->a  = 6378137.0;
    P->es = P->f*(2-P->f);
    pj_calc_ellipsoid_params (P, P->a, P->es);

    P->inv = s_inverse;
    P->fwd = s_forward;
    return P;
}
```

which gives proper projection factors and final Earth figure, e.g.

```
./proj +proj=webmerc +ellps=WGS84 -V
#WGS 84 Web Mercator / WGS 84 Pseudo Mercator
#	Cyl, Ell
#	
# +proj=webmerc +ellps=WGS84
#Final Earth figure: ellipsoid
#  Major axis (a): 6378137.000
#  1/flattening: 298.257224
#  squared eccentricity: 0.006694379990
0 0
Longitude: 0dE [ 0 ]
Latitude:  0dN [ 0 ]
Easting (x):   0.00
Northing (y):  0.00
Meridian scale (h) : 1.00673950  ( 0.6739 % error )
Parallel scale (k) : 1.00000000  ( 0 % error )
Areal scale (s):     1.00673950  ( 0.6739 % error )
Angular distortion (w): 0.385
Meridian/Parallel angle: 90.00000
Convergence : 0d [ -0.00000000 ]
Max-min (Tissot axis a-b) scale error: 1.00674 1.00000
```

Also line

`PROJ_HEAD(webmerc, "Web Mercator / Pseudo Mercator") "\n\tCyl, Sph\n\t";`

is changed to

`PROJ_HEAD(webmerc, "Web Mercator / Pseudo Mercator") "\n\tCyl, Ell\n\t";`

because there is no (at least I am not aware of) definition of Web Mercator projection for sphere, so sphere should be avoided when we speak of Web Mercator projection. It is projection of rotational ellipsoid defined with specific equations which are chosen to be same as equations for Mercator of sphere, but it is different projection with different properties (cylindrical non-conformal, non-equareal, non-equidistant).

Avoiding introduction of sphere when talking and defining Web Mercator projection will remove confusion that user uses ellipsoid as reference surface but projection is for sphere (which, in fact, is not)!

3. Updated description of Web Mercator projection to reflect its properties.


kbevers:
> Proposal is to rename full name of webmerc projection to "WGS 84 Web Mercator / WGS 84 Pseudo Mercator" instead of "Web Mercator / Pseudo Mercator"
REASON: Official name of projection is "WGS 84 / Pseudo Mercator", see epsg.io/3857. Projection is tied to WGS 84 ellipsoid.

I am not a fan of this proposal. The name you are refering to is a mix of reference frame and projection. Similar to "ETRS89/UTM Zone 32". That makes sense when you are describing a specific CRS in e.g. a catalogue like epsg.io but not when only talking about the projection (I agree that the waters are muddy here). In the EPSG guidance note 7 the projection is referred to as "Popular Visualisation Pseudo-Mercator ("Web Mercator")". Several other names are in use as well. So let's just stick with "Web Mercator". The ties to the WGS84 ellipsoid can be described in the documentation.

docs/source/operations/projections/webmerc.rst
```diff
@@ -21,21 +20,21 @@ From `Wikipedia <https://en.wikipedia.org/wiki/Web_Mercator>`_:
...
-| **Input type**      | Geodetic coordinates                                     |
-+---------------------+----------------------------------------------------------+
-| **Output type**     | Projected coordinates                                    |
-+---------------------+----------------------------------------------------------+
++---------------------+-----------------------------------------------------------------+
+| **Classification**  | Non-conformal cylindrical projection of an WGS 84 ellipsoid     |
++---------------------+-----------------------------------------------------------------+
+| **Available forms** | Forward and inverse, ellipsoidal projection                     |
++---------------------+-----------------------------------------------------------------+
+| **Defined area**    | Global, low linear distortions near the equator                 |
```
kbevers, line 28: I think it is better to just say that it is globally defined and not go into where distortions are at a minimum. The original text is just a copy of what's written for Mercator which is kind of not valid here anyway.

dtutic:
No problem. It was just a suggestion. Definitelly map projection is not the same thing as coordinate reference system.

But as you said, this is fuzzy distinction in this case. It can be seen in several ways, and that adds up to the confusion.

Retorical questions. Is there a Web Mercator projection of a Bessel ellipsoid? What use it would serve and why? Do we need Web Mercator projection outside of CRS it is tied to?

---- Kristian Evers wrote ----


Proposal is to rename full name of webmerc projection to "WGS 84 Web Mercator / WGS 84 Pseudo Mercator" instead of "Web Mercator / Pseudo Mercator"
REASON: Official name of projection is "WGS 84 / Pseudo Mercator", see epsg.io/3857. Projection is tied to WGS 84 ellipsoid.

I am not a fan of this proposal. The name you are refering to is a mix of reference frame and projection. Similar to "ETRS89/UTM Zone 32". That makes sense when you are describing a specific CRS in e.g. a catalogue like epsg.io but not when only talking about the projection (I agree that the waters are muddy here). In the EPSG guidance note 7 the projection is referred to as "Popular Visualisation Pseudo-Mercator ("Web Mercator")". Several other names are in use as well. So let's just stick with "Web Mercator". The ties to the WGS84 ellipsoid can be described in the documentation.

docs/source/operations/projections/webmerc.rst
```diff
@@ -49,13 +48,8 @@ Example::
 Parameters
 ################################################################################
 
-.. note:: All parameters for the projection are optional.
+.. note:: Projection does not have any additional parameters. See `EPSG.io  <http://epsg.io/3857>`_
```
kbevers, line 51: I think it is better to refer to [Wikipedia](https://en.wikipedia.org/wiki/Web_Mercator) here? The epsg.io site doesn't go into much detail. Also, it is easily mistaken as official outlet of the EPSG (which it is not). It is already mentioned under "Further reading".

kbevers:
Apart from my inlined comments and the thing about the name, I think this looks good. Solid first PR! Thanks for dealing with this.

dtutic:
Again, only suggestion. Mercator projection is used with no problems from south to north seas, i.e. big range of latitudes, same holds for Web Mercator.

That is why it is limiting to say thay it is recommended only for areas near equator. Its special properties makes it usable at most latitudes as long the area on the map is kept relatively small.

Definitelly unsuitable for world or most of continent maps.

---- Kristian Evers wrote ----


@kbevers commented on this pull request.

________________________________

In docs/source/operations/projections/webmerc.rst<https://github.com/OSGeo/proj.4/pull/1080#discussion_r204143427>:

> -| **Defined area**    | Global, but best used near the equator                   |
-+---------------------+----------------------------------------------------------+
-| **Alias**           | webmerc                                                  |
-+---------------------+----------------------------------------------------------+
-| **Domain**          | 2D                                                       |
-+---------------------+----------------------------------------------------------+
-| **Input type**      | Geodetic coordinates                                     |
-+---------------------+----------------------------------------------------------+
-| **Output type**     | Projected coordinates                                    |
-+---------------------+----------------------------------------------------------+
++---------------------+-----------------------------------------------------------------+
+| **Classification**  | Non-conformal cylindrical projection of an WGS 84 ellipsoid     |
++---------------------+-----------------------------------------------------------------+
+| **Available forms** | Forward and inverse, ellipsoidal projection                     |
++---------------------+-----------------------------------------------------------------+
+| **Defined area**    | Global, low linear distortions near the equator                 |


I think it is better to just say that it is globally defined and not go into where distortions are at a minimum. The original text is just a copy of what's written for Mercator which is kind of not valid here anyway.

rouault:
Was superseded by https://github.com/OSGeo/proj.4/pull/1095

PR #1087, opened 2018-08-10
Fix wrong behaviour of torad_coord() with gcc 8.1 -O2 (fixes #1084)

### LINKED ISSUE #1084: 4D-API_cs2cs-style.gie fails on Debian unstable (with GCC 8)
`4D-API_cs2cs-style.gie` started to fail on Debian unstable recently. Quite possibly related to GCC 8 becoming the default compiler.

```
Reading file 'GDA.gie'
total:  3 tests succeeded,  0 tests skipped,  0 tests failed.
PROJ_LIB=../../nad ../../src/gie builtins.gie
Reading file 'axisswap.gie'
total: 27 tests succeeded,  0 tests skipped,  0 tests failed.
PROJ_LIB=../../nad ../../src/gie deformation.gie
Reading file '4D-API_cs2cs-style.gie'
     -----
     FAILURE in 4D-API_cs2cs-style.gie(114):
     expected: 12.5 55.5 0
     got:      inf   inf   inf
     deviation:  999999999.999000 mm,  expected:  150.000000 mm
total: 28 tests succeeded,  0 tests skipped,  1 tests FAILED!
Reading file 'deformation.gie'
total: 37 tests succeeded,  0 tests skipped,  0 tests failed.
make[5]: *** [Makefile:455: 4D-API-cs2cs-style] Error 1
make[5]: *** Waiting for unfinished jobs....
Reading file 'builtins.gie'
total: 1568 tests succeeded,  0 tests skipped,  0 tests failed.
```

rouault:
torad_coord() of gie.c has this sequence:
```
if( cond )
    axis = l->param + strlen ("axis=");
n = strlen (axis);
```

When the if branch is evaluated, n is always zero
even if on inspection axis is non empty

The reason is that the struct ARG_list which is the
type of l use a variable-length array for the param member

struct ARG_list {
    paralist *next;
    char used;
    char param[1];
};

But this is not a proper way of declaring it, and
gcc 8 has apparently optimizations to detect that l->param + 5
points out of the array, hence it optimizes strlen() to 0.

According to https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html,
the proper way of declaring such arrays is to use [0]

### DIFF
--- a/src/pj_param.c
+++ b/src/pj_param.c
@@ -42,7 +42,7 @@ paralist *pj_mkparam_ws (char *str) {
     }
 
     /* Use calloc to automagically 0-terminate the copy */
-    newitem = (paralist *) pj_calloc (1, sizeof(paralist) + len);
+    newitem = (paralist *) pj_calloc (1, sizeof(paralist) + len + 1);
     if (0==newitem)
         return 0;
     memmove(newitem->param, str, len);
--- a/src/projects.h
+++ b/src/projects.h
@@ -458,7 +458,12 @@ struct PJconsts {
 struct ARG_list {
     paralist *next;
     char used;
-    char param[1];
+#ifdef __GNUC__
+    char param[0]; /* variable-length member */
+    /* Safer to use [0] for gcc. See https://github.com/OSGeo/proj.4/pull/1087 */
+#else
+    char param[1]; /* variable-length member */
+#endif
 };
 
 


rouault:
looking closer, looks like a too aggressive optimization in that particular case. Filed as https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86914

kbevers:
Do you still want to merge this pull request or just wait until the GCC guys sort their things out?

rouault:
I think that it is reasonable to merge it, even if it is a GCC bug. This version is out and so even if it is fixed, there's always the risk someone sticking with this version and hitting the issue. After all the use of one-element array was *discouraged* by the gcc docs, so the use of zero-length is a better practice.
(an even better practice would be to just avoid the use of such tricky features)

PR #1088, opened 2018-08-11
Assorted fixes: Makefile, tests, gie enhancement, Travis-CI simplification

rouault:


nad/Makefile.am
```diff
@@ -24,7 +24,7 @@ EXTRA_DIST = GL27 nad.lst proj_def.dat nad27 nad83 pj_out27.dist pj_out83.dist t
 		esri.extra other.extra \
 		CH IGNF testIGNF proj_outIGNF.dist \
 		ITRF2000 ITRF2008 ITRF2014 \
-		makefile.vc CMakeLists.txt tests/test_nodata.gtx
+		makefile.vc CMakeLists.txt tests/test_nodata.gtx null.lla
```
kbevers, line 27: What's the reasoning behind including it here? The `null` grid is included in the grid distributions.

src/gie.c
```diff
@@ -516,6 +517,20 @@ static int ignore (const char *args) {
     return 0;
 }
 
+static int require_grid (const char *args) {
```
kbevers, line 520: This is a great idea! Previously we've handled missing grids by adding `ignore pjd_failed_to_load_grid`. I think those would be handled better by this new `require_grid` keyword.

kbevers (changes requested):
The new gie keyword should be documented in `docs/source/apps/gie.rst`. 

It doesn't seem to be possible to require more than one grid, is that true? I think it would be useful if you were able to require several grids, since that is needed in some transformations.

rouault:
> It doesn't seem to be possible to require more than one grid, is that true?

It is possible. You just to have repeat the require_grid keyword several times

nad/Makefile.am
rouault: The reasoning is that null.lla is in git, and if we don't had it in EXTRA_DIST it is not part of the generated tarball by make dist. And thus if someone use the tarball and does "make check" without downloading the grids, it won't get the null grid (since 'make' in nad create null from null.lla) and that causes a number of test failures. I think it is reasonable to have the tiny null grid generated by proj itself

kbevers (approved):

nad/Makefile.am
kbevers: All right. I agree, it is reasonable to include it.

Issue #807, opened 2018-02-19
Add user-specifiable output format in cct


kbevers:
In `proj` and `cs2cs` we have the option to add the `-f` flag and specify how output coordinates are printed on screen. This should also be possible in `cct`. Preferably in a way that avoids recreating https://github.com/OSGeo/proj.4/issues/124.

The most important thing here is to let users control the number of decimals `cct` returns.

busstoptaktik:
I intend to implement this by letting the user specify required precision **in meter**, e.g. specifying millimetric resolution by saying -p 0.001, -p mm or -p 1e-3.

For linear output, this translates straightforwardly to number of decimals.

For angular output I would probably start by saying 1 degree == 100000 m, so -p mm would translate into 8 decimals.

kbevers:
Just to make things clear, this is best handled by @busstoptaktik's approach. Ideally this would also be extended to `proj` and `cs2cs` at some point.

sgh:
Ok. So `-p 1e-4` would give 9 decimals on angles if I follow you correctly?

rouault:
@busstoptaktik The approximation 1 degree == 100000 m only holds for Earth, and PROJ is definitely used with other celestial bodies. That would mean we would need to examine the pipeline to fetch the first mention of a semi-major axis ?

kbevers:
> That would mean we would need to examine the pipeline to fetch the first mention of a semi-major axis ?

Yes. Should be quite easy since the main `PJ` object always has ellipsoid parameters attached to it. In case of pipelines there is the possibility that the "main ellipsoid" isn't the same as the ellipsoid representing the output coordinates but I am betting it will always at least resemble the celestial body that it is meant to represent. So, `deg_length = 2 * M_PI * P->a * 1/360` should be good enough for this application.

busstoptaktik:
> So, deg_length = 2 * M_PI * P->a * 1/360 should be good enough for this application.

Yes, that was what I intended, although for semimajor axes in the Earth like interval 6200 km-6600 km, I would default directly to L<sub>deg</sub>=100 km

kbevers:
> although for semimajor axes in the Earth like interval 6200 km-6600 km, I would default directly to Ldeg=100 km

So you would have a special case for Earth and do it generically for longer/shorter semimajor axes?

busstoptaktik:
> So you would have a special case for Earth and do it generically for longer/shorter semimajor axes?

[...]

PR #1089, opened 2018-08-14
cct: Added -d option to allow extra decimals in output.

sgh:


rouault:
I'm wondering if specifying the number of *extra* decimals is the preferable interface. Wouldn't the absolute number be more obvious ? We would then probably need distinct parameters for the x,y and z components

rouault:
I'll let @kbevers @busstoptaktik give their opinion

sgh:
hmm. So In case I wanted 14 decimals I would say "-f %.14f" ?

kbevers:
No, for linear output you would say `-p 1e-14` or `-p 10 fm`. In case of angular output you would get more decimals by specifying a 10 femtometer precision.

There are two purposes to this approach. 1) Make it simpler to specify a certain level of precision when outputting angular coordinates and 2) make users think a little about what they are doing instead of just outputting a bunch of nonsense decimals. It is tempting to output as many decimals as possible but realistically only the first 3 or 4 are sensible for linear output (and this is still rather extreme) and maybe 8 for angular output.

sgh:
I think `-p 1e-14` makes sense. I do not quite follow you as to how angular precision would be increased in your example.

busstoptaktik:
@kbevers said:

> We've discussed the issue before: #807. While a bit more complicated to implement, this is the solution I prefer.

Being the originator of both the original *cct* design and the implementation suggestion over at #807 it is probably not surprising that I agree on this, although after having giving it a second thought, I would prefer to use the flag `-q` (indicating quantification, which it is) rather than `-p` (indicating precision, which it's not). Although this will be in conflict woth the `-q, --quiet` flag of `gie`... there's no free lunch :-)

@sgh said:
> I do not quite follow you as to how angular precision would be increased in your example.

With an arbitrary semimajor axis, *a*, and a flattening, *f*, close to 1, the circumference of the planet considered (i.e. typically the Earth) is on the order of *2πa*. For the Earth, with *a≈6400 km* the circumference, *O≈40000 km*, making one degree of latitude correspond to *40000 km/360° ≈ 111 km ≈ 100000 m*.

Hence to a good approximation, if your output is in degrees and you specify millimetric quantification (-p mm) *cct* should interpret that as you asking for 8 digits after the decimal separator (the integer to the left of the separator counts the number of degrees, i.e. 100 km units The first digit after the separator, i.e. the decidegrees, counts the number of 10 km units, a.s.o.). For the longitudinal coordinate at non-equatoreal latitudes, this will give you progressively more digits than actually needed. You're supposed to live with that.

With modern day GNSS observations originating as 3D cartesian coordinates, linear units really are the most reasonable to use for quantification anyway. If your system provides millimetric accuracy, you are interested in __millimetric__ quantification (makes sense for a human consumer), not __nanodegree__ quantification, which is rather hard to relate to in anthropometric terms.

@sgh, I appreciate your effort, and you're certainly addressing a known shortcomming of *cct*, but I suggest to close this PR and continue the discussion over at #807, in order to plan a more generally useful implementation.

sgh:
I'll close the request. My actual use for this extra ouput resolution is to test how well proj agree with other implementations of the transforms. It has no use in any actual usecase.

kbevers:
>  My actual use for this extra ouput resolution is to test how well proj agree with other implementations of the transforms.

I suggest you use [gie](https://proj4.org/apps/gie.html) for that instead. It gives you all the tools you that exact purpose.

sgh:
Thank you for the hint. Unfortunately the stuff I'm trying to do does to stay within proj. I need the data out for processing of external tools.

kbevers:
re https://github.com/OSGeo/proj.4/issues/807 I'm reopening this again. @sgh you mentioned that you were willing to also implement the `-d` flag in `proj` and `cs2cs`. If possible it would be nice to have those included in this PR as well.

sgh:
Yes - I'll do it.

kbevers:
> Yes - I'll do it.

Awesome! Thanks.

PR #1090, opened 2018-08-17
Implementation of Equal Earth projection (closes #1085)

### LINKED ISSUE #1085: Add Equal Earth projection
Equal Earth is a projection inspired by the Robinson projection, but unlike the Robinson projection retains the relative size of areas. The projection is described in this publication:

Bojan Savric, Tom Patterson & Bernhard Jenny (2018). 
The Equal Earth map projection.
International Journal of Geographical Information Science.
[http://doi.org/10.1080/13658816.2018.1504949](http://doi.org/10.1080/13658816.2018.1504949)

Java code is available here: [https://github.com/OSUCartography/JMapProjLib/blob/master/src/com/jhlabs/map/proj/EqualEarthProjection.java](https://github.com/OSUCartography/JMapProjLib/blob/master/src/com/jhlabs/map/proj/EqualEarthProjection.java)

jdugge:
Implement the Equal Earth projection


docs/plot/plotdefs.json
```diff
@@ -307,6 +307,17 @@
         "res": "low",
         "type": "poly"
     },
+    {
+        "filename": "eqearth.png",
+        "latmax": 90,
+        "latmin": -90,
+        "lonmax": 180,
+        "lonmin": -180,
+        "name": "aea",
```
rouault: wrong name

test/gie/more_builtins.gie
```diff
@@ -408,6 +408,59 @@ accept       -81d00'00.000 34d32'30.000 0.0
 expect       2_000_000.000 561_019.077 0.0
 -------------------------------------------------------------------------------
 
+-------------------------------------------------------------------------------
+Some tests from PJ_eqearth.c
+-------------------------------------------------------------------------------
+operation +proj=eqearth +ellips=WGS84
```
rouault: typo: should be +ellps=

kbevers:
@jdugge This is a very impressive first contribution. Good job! 

I have a few corrections in the nit-picking department:

1) Could you add an inverse test where the y-coordinate is smaller than `-MAX_Y`? This is a special case and I want to make sure we hit that in the tests as well.

2) In case the inverse mapping doesn't converge an error is reported. Could you make a test for that case as well?

src/PJ_eqearth.c
```diff
@@ -0,0 +1,91 @@
...
+        f = yc * (A1 + A2 * y2 + y6 * (A3 + A4 * y2)) - xy.y;
+        fder = A1 + 3 * A2 * y2 + y6 * (7 * A3 + 9 * A4 * y2);
+        tol = f / fder;
+        yc -= tol;
+        if (fabs(tol) < EPS) {
+            break;
+        }
+    }
+    if( i == 0 )
+        pj_ctx_set_errno( P->ctx, PJD_ERR_NON_CONVERGENT );
```
kbevers: Return `lp` here. We already know that the output is erroneous and we might as well save a few clock cycles and not hit the next couple of lines.

jdugge:
Thanks!
I'll add in a test that covers an inverse projection exceeding the negative bound, but I haven't been able to find a test case where the Newton-Raphson didn't converge. In the worst case, it took three iterations. Do you have an idea for finding a case that breaks the algorithm?

jdugge:
@kbevers: Alright, should be all set now

kbevers:
> @kbevers: Alright, should be all set now

Awesome. Thanks for adding this!

PR #1093, opened 2018-08-19
Assorted fixes related to +geoc flag handing

rouault:
While digging geocentric latitude support in PROJ, I discovered there was an existing (and somewhat unknown) +geoc flag whose semantic is: the input latitude is a geocentric one.

* The first commit fixes a regression of PROJ 5 in the inverse code path where the geodetic -> geocentric conversion was missing
* The second commit is somewhat independant of the first one, but was found when investigating the issue: it appears there is dead code in the fwd_finalize() and inv_prepare(), so remove it since looking at it, I believe it wouldn't behave correctly if hit
* The third commit is independent of the two above ones, and for consistency, make +geoc honoured in the legacy pj_transform() / cs2cs case (was easier for me to implement that, so as to check the correctness of the fix in the first commit)


kbevers:
As far as I can tell this looks all right but I am not too familiar with the geocentric code so let's hear what @busstoptaktik thinks?

PR #1100, opened 2018-08-21
Update helmert.rst

phidrho:


busstoptaktik (commented):
IIRC, the EPSG uses the 15 parameter description: The 15th is the epoch

kbevers:
> IIRC, the EPSG uses the 15 parameter description: The 15th is the epoch

This is correct. Here's what the EPSG Guidance Note 7-2 has to say on this subject:

> To account for tectonic plate motion (but not deformation) the rate of change of each of the seven parameters
may be included in a time-dependent variation of the Helmert 7-parameter transformation. This leads to a 15-
parameter transformation: the three translations, three rotations and the scale change, the rates for these
seven parameters and the parameter reference epoch for these parameters

I'll close this PR on the basis on that.

Issue #1097, opened 2018-08-21
Equal Earth projection is missing ellipsoidal implementation


beuan:
Thank you @jdugge @kbevers @rouault on your awesome work getting the Equal Earth projection so fast available in PROJ library. I am supper amazed how this is moving forward. 

The current PROJ implementation is sphere-only, but I noticed [people are starting to use it with WGS84 ellipsoid](https://github.com/geoserver/geoserver/blob/master/data/release/user_projections/epsg.properties#L32), which might cause issues later if this projection is released as this. Used with WGS84 and current implementation, it also loses its equal-area property.  

In the paper we suggest: 
> For projecting geographic coordinate on the ellipsoids, such as WGS 1984, we suggest transforming geographic latitudes to authalic latitudes (Snyder 1987, p. 16) before projecting with the spherical Equal Earth projection. (Šavrič et al. 2018)

I believe functions for this conversions already exists in PROJ code base:
```
double pj_qsfn(double sinphi, double e, double one_es)
double pj_authlat(double beta, double *APA)
```
I am not very well familiar with the whole code base, but with provided functions we might be able to add ellipsoidal implementation before its release on Sep 1st. I am willing to assist as much as possible.

I assume we need new `e_forward()` and `e_inverse()` functions, updated `PJ *PROJECTION(eqearth)` function and new tests. Would there be anything else?

Org Issue: #1085
Pull Request: #1090

cc: @berniejenny

beuan:
I forked PROJ and implemented ellipsoidal equations for the Equal Earth projection. Unfortunately, I do not know how to run tests not using `CMAKE`. I was able to build new version and test points using `NMAKE`. Is there any other way I can verify that tests are passing? Thanks.
Forked branch with proposed fix: https://github.com/OSGeo/proj.4/compare/master...beuan:EqEarth_ellips

beuan:
Thank you @jdugge for running the tests! Looks like your tests are not passing due to the missing tolerance (I think). 
Would you mind running them again? I just added the tolerance parameter.  I hope this fixes the issue.

kbevers:
Closed with #1101

PR #1101, opened 2018-08-22
Adding ellipsoidal equations for the Equal Earth

beuan:
This changes is for #1097.


src/PJ_eqearth.c
```diff
@@ -9,14 +9,16 @@ projection, International Journal of Geographical Information Science,
...
 */
 #define PJ_LIB__
 
+#include <errno.h>
 #include <math.h>
 
 #include "projects.h"
 
-PROJ_HEAD(eqearth, "Equal Earth") "\n\tPCyl., Sph.";
+PROJ_HEAD(eqearth, "Equal Earth") "\n\tPCyl., Sph&Ell\n\th=";
```
kbevers: What's up with the `\n\th=` at the end here? As far as I can tell there's no `+h` parameter present in this projection. Bad cut & paste job?

kbevers (changes requested):
The code seems fine to be, apart from my inlined comments.

The docs needs updating to reflect that the projection also exist in an ellipsoidal form.

src/PJ_eqearth.c
```diff
@@ -28,28 +30,59 @@ PROJ_HEAD(eqearth, "Equal Earth") "\n\tPCyl., Sph.";
...
+    struct pj_opaque *Q = P->opaque;
+    double sbeta;
+    double psi, psi2, psi6;
+
+    if (P->es != 0.0) {
+       /* Ellipsoidal case, converting to authalic latitude */
+       sbeta = pj_qsfn(sin(lp.phi), P->e, 1.0 - P->es) / Q->qp;
+       if (fabs(sbeta) > 1) {
+          /* Rounding error. */
+          sbeta = sbeta > 0 ? 1 : -1;
```
kbevers, line 50: It would be nice to have a test case that exercises this condition.

src/PJ_eqearth.c
beuan: Yes, bad cut & paste.

src/PJ_eqearth.c
beuan: I am not sure I can hit this test case. The best testing case is probably phi = +/-90. In this case pj_qsfn() function returns +/-Q->qp and for some rounding error it might hit this case. Those tests are already in and looked they are passing.

beuan: Using phi > 90 would not help, because sin function would already "fix" this.

kbevers: All right, we'll live without a test for that

kbevers:
We also need `+ellps` listed under parameters.

kbevers:
> I added +ellps parameter. I hope this is what you had in mind...

Indeed it was. Thanks!

kbevers (approved):

beuan:
@kbevers @rouault, thank you for your comments, review and assistance!
Once all checks pass, should I merge or you prefer to be the one who merges?

kbevers:
> Once all checks pass, should I merge or you prefer to be the one who merges?

I just did. You shouldn't be able to do it. Thanks for adding this.

beuan:
Thanks!

src/PJ_eqearth.c
```diff
@@ -28,28 +30,59 @@ PROJ_HEAD(eqearth, "Equal Earth") "\n\tPCyl., Sph.";
...
+
+static XY e_forward (LP lp, PJ *P) {           /* Ellipsoidal/Spheroidal, forward */
     XY xy = {0.0,0.0};
-    double phi, phi2, phi6;
-    (void) P;
+    struct pj_opaque *Q = P->opaque;
+    double sbeta;
+    double psi, psi2, psi6;
+
+    if (P->es != 0.0) {
```
busstoptaktik, line 45: A little bit of nitpicking here:  For clarity, I would prefer to have block-scoped comments *precede* the block they refer to:

```
/* Ellipsoidal case, converting to authalic latitude */
if (P->es != 0.0) {
    sbeta = pj_qsfn(sin(lp.phi), P->e, 1.0 - P->es) / Q->qp;

    /* Rounding error. */
    if (fabs(sbeta) > 1) 
        sbeta = sbeta > 0 ? 1 : -1;
}

/* Spheroidal case, using latitude */
else 
    sbeta = sin(lp.phi);
```

Or with even less branching:

```
sbeta = sin(lp.phi);

/* In the ellipsoidal case, we convert sbeta to authalic latitude */
if (P->es != 0.0) {
    sbeta = pj_qsfn(sbeta, P->e, 1.0 - P->es) / Q->qp;

    /* Rounding error. */
    if (fabs(sbeta) > 1) 
        sbeta = sbeta > 0 ? 1 : -1;
}
```

kbevers: Sorry Thomas, I merged this while you were writing...

busstoptaktik: Don't worry - although I think my suggestion has some merit in terms of clarity, it also is mostly cosmetical and may be simply a matter of personal preference

beuan: I happy to do the change and another PR if you would like to...

busstoptaktik: @beuan: Well - I wrote the comment because *I* would prefer to minimize branching and have block-scope comments precede the block.

But it's your code, and it's already merged, so please only make the change if you personally find my (cosmetical) arguments highly convincing, and would *like* to see the change - in which case I will obviously feel honoured, that you found the comment convincing and/or useful :-)

beuan: #1102

PR #1104, opened 2018-08-24
Ensure correct axis ratio for images in documentation

jdugge:
The example images in the documentation have incorrect axis ratios:

This happens for two reasons:
1. When generating the images using `plot.py` an image size of 2400 x 1800 is used and the plot is set to fill the available space.
Example image of the orthographic projection

2. In the `.rst` files for the projections the option `:scale: 50%` is used, which in the generated HTML documentation sets the `style`  of the images to `width=1200` and `height=900`. The images are also affected by a cascaded attribute `max-width=100%`, which limits the image width to the width of the main page column (800px). Because the `height` attribute is explicitly set to `900`, this will squeeze the image horizontally, resulting in an overall size of 800 x 900.
Documentation for the orthographic projection

This pull request fixes these issues by enforcing a 1:1 axis ratio when generating the images, discarding unnecessary margins around the image, and by setting only the image `width` in the documentation to a reasonable value, which ensures that the axis ratio will remain unchanged.


kbevers:
This looks great! I'm not too about regenerating all the figures but I guess it is necessary to sort this out properly.

A few things before this can be merged:

1) The images are not centered on the pages. Can be fixed by adding `:align: center`.

2) The file-sizes on the images are way to big (it increases the repo size by a significant amount, unfortunately). You should be able to reduce the file sizes by running `/docs/plot/greyscale.sh`.

jdugge:
`:align: center` is added
I made the images a bit smaller and converted them to grayscale. I also adapted the `greyscale.sh` script a bit, using `mogrify` from the Imagemagick toolset instead of a loop and removing the `-fx` option, which slows down the conversion progress a lot without influencing the result.

I amended and force-pushed the new version, to avoid having the large files in the git history.

kbevers:
Thanks, this looks much better than the current docs!

I was thinking that since you are updating all the figures anyway it might be a good time to change something that has been bothering for a long time. When I originally made all the figures I added the projection setup string as a header for the figures just as a simple way to see what was going on. Since then a lot has changed and I don't think that is particularly smart any longer. It would be better to write the setup string as text in figure descriptions beneath the images instead of hardcoding it in the images. Would you be up for making that change as well?

jdugge:
Done!
I've added a script (`figure_captions.py`) that updates the figure captions in the documentation. This script should be called when a plot definition in `plotdefs.json` is changed.

kbevers:
That is so awesome! Thank you very much. Good idea adding the update script.

It looks like the images are updated twice, one time for each commit. Is that correct?

jdugge:
Ah yes, sorry! I just fixed it, now there's only a single commit for all changes.

kbevers:
Thanks!

PR #1106, opened 2018-08-29
4D API: honour vto_meter / vunits for proj=longlat

rouault:
This worked for cs2cs / pj_transform(), but not the new API


kbevers:
Nice catch!

PR #1108, opened 2018-09-02
be more expansive in ignorance

greg-minshall:
hi.  just some additions to .gitignore to ignore more things after running some tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,15 +89,36 @@ test-driver
 /src/nad2bin
 /src/proj
 /src/geod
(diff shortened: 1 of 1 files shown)


kbevers:
I guess this is needed when you do in-source builds, yes?

greg-minshall:
yes, when you pull from the repository and do a build and then say "git status" to see what you might need to cleanup.  i find it nice to have as little polluting the status output, so that whatever i may have modified/added/deleted stands out.

PR #1109, opened 2018-09-03
Use -d to specify number of decimals in output.

sgh:
#807 I have done the necessary changes. I do however not know how to sufficiently test them. In addition to that I was hoping that the tools we more similar. The way things are done in `cct` seem better that the way thins are done in `proj` and `cs2cs` and I was not really up to migrating all of them to the cct-way.

### DIFF
--- a/src/cs2cs.c
+++ b/src/cs2cs.c
@@ -53,7 +53,7 @@ tag = '#';	/* beginning of line tag character */
  oform_buffer[16],	/* buffer for oform when using -d */
 *oterr = "*\t*",	/* output line for unprojectable input */
 *usage =
-"%s\nusage: %s [ -eEfIlrstvwW [args] ] [ +opts[=arg] ]\n"
+"%s\nusage: %s [ -dDeEfIlrstvwW [args] ] [ +opts[=arg] ]\n"
 "                   [+to [+opts[=arg] [ files ]\n";
 
 static double (*informat)(const char *, 
--- a/src/cs2cs.c
+++ b/src/cs2cs.c
@@ -31,7 +31,6 @@
 #include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <stdio.h>
 #include <string.h>
 
 #include "proj.h"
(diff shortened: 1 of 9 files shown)


kbevers (changes requested):
I think this looks all right. There's a few things missing documentation wise:

1) Add `-d` to the list of available flags in `proj.c` (line 53)
2) Add `-d` and `-D` to the list of available flags in `cs2cs` (line 56)
3) Add `-d` option to `cct`, `cs2cs` and `proj` doc-files
4) Add `-D` option `cs2cs` doc file.

Regarding testing, that is a bit complicated at the moment. The `proj` and `cs2cs` apps can be tested by using the test scripts in `nad/`. See https://github.com/OSGeo/proj.4/pull/1106/files for an example. A similar thing could be done for `cct` but I think that should be avoided and instead we need to make a proper test setup of the CLI apps. Hopefully we will get round to doing that soon.

sgh:
Yeah - I did see the autogenerate comment ....

src/cs2cs.c
```diff
@@ -293,10 +294,15 @@ int main(int argc, char **argv)
...
                 reverseout = 1;
                 continue;
-              case 'd': /* set debug level */
+              case 'D': /* set debug level */
                 if (--argc <= 0) goto noargument;
                 pj_ctx_set_debug( pj_get_default_ctx(), atoi(*++argv));
                 continue;
+              case 'd':
+                if (--argc <= 0) goto noargument;
+                snprintf(oform_buffer, sizeof(oform_buffer), "%%.%df", atoi(*++argv));
```
kbevers: The use of `snprintf` is what is causing Travis to trip up. Unfortunately `snprintf` is not C89  which is a requirement at the moment. While a bit more cumbersome it should be possible to do this with `sprintf` instead.

sgh: Oh ... that was the reason :) thanks.

kbevers:
Thanks for adding this!

sgh:
Thank you for the feedback ping-pong.

PR #1114, opened 2018-09-11
Fix SONAME decrement from libproj.so.13 to libproj.so.12.

sebastic:
Increment age instead of revision for added interfaces, see:

 https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -41,7 +41,7 @@ geodtest_LDADD = libproj.la
 
 lib_LTLIBRARIES = libproj.la
 
-libproj_la_LDFLAGS = -no-undefined -version-info 14:0:2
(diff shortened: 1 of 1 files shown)


kbevers:
What is this, the third time in a row I've gotten this wrong? I may never learn, so thanks for having my back on this!

sebastic:
Libtool versioning is non-trivial and the `version-info` format is a bit counter-intuitive, so don't worry too much about it.

Some kind of sanity check may be in order, though. Perhaps add a step to the release howto that the SOVERSION must never decrement by checking that current - age (first minus last version-info number) is greater than or equal to the current SOVERSION.

PR #1117, opened 2018-09-11
Modifications for Intel compilers

mwtoews:
Some changes to allow compilation with Intel compilers

- Specify C++11 for Intel, and also workaround a CMake Error with empty variables
- Intel 17.0 does not have good support for `constexpr`, but it seems `const` is fine for these contexts


kbevers:
Is this a fix related to PROJ 5.2 RC1?

mwtoews:
yes for 5.2 RC1. After these modifications, all ll tests pass with Intel's `CC=icc CXX=icpc`.

CMakeLists.txt
```diff
@@ -41,8 +41,10 @@ endif()
 if (CMAKE_C_COMPILER_ID STREQUAL "Intel")
   if (MSVC)
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /fp:precise")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Qstd=c++11")
   else ()
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fp-model precise")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
hobu, line 47: I wonder if we shouldn't set a CMake 3.1 minimum and do this sometime soon (not for this patch, which looks fine IMO)
```
set (CMAKE_CXX_STANDARD 11)
```

mwtoews: I looked into this briefly, and it seems that this was brought in after CMake 2.6.0 (the current minimum version at the top of the file)

kbevers:
@mwtoews 

> yes for 5.2 RC1. After these modifications, all ll tests pass with Intel's CC=icc CXX=icpc.

Great. Thanks for fixing this!

@hobu 

> I wonder if we shouldn't set a CMake 3.1 minimum and do this sometime soon (not for this > patch, which looks fine IMO)
> 
> ```set (CMAKE_CXX_STANDARD 11)``` 

This especially makes sense once the WKT code is merged. Please add a ticket on this so we remember :-)

PR #1124, opened 2018-09-18
Version 6 preparation

kbevers:
Knocking out the first few issues related to version 6: Removing `projects.h` and requiring the `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` macro to be defined before `proj_api.h` can be included.

I am open for better names than `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H`.


rouault:
src/lib_proj.cmake should also no longer install projects.h

kbevers:
> src/lib_proj.cmake should also no longer install projects.h

That's were it was! I couldn't for the life of me figure out where it was located in the CMake setup. Thanks.

PR #1125, opened 2018-09-18
Rename nad/ as data/ and move nad/test* to test/old/*

rouault:
This cleans up the ill-named nad/ directory
We rename it as data/ and keep only data files in it. The test scripts and their expected results  are moved to test/old/


kbevers (approved):
I am very happy with these changes! Thanks for doing the grunt work.

As far as I can tell you haven't altered the documentation to match this. That's good since we shouldn't publish that info on the website just yet but better write a ticket so it isn't forgotten. There's at least the "installation from source" in the Sphinx docs and possibly also the README.

rouault:
Doc ticket created: https://github.com/OSGeo/proj.4/issues/1127

PR #1129, opened 2018-09-20
Update proj.def reflecting depublication of projects.h

kbevers:
Since `projects.h` is no longer available to the public we shouldn't expose it's functions in the DLL.

There may be smarter ways to handle which functions get exposed in the DLL once we remove the nmake build system but for now this should do.


hobu:
I'd be nice to get rid of the def file altogether someday. It is so easy to forget to update.

kbevers:
> I'd be nice to get rid of the def file altogether someday. It is so easy to forget to update.

Yes, it is ridiculously annoying. I think we can get away with it if we instead prefix DLL-functions with a prefix that something along the lines of `__cdecl`. I don't really know how that works...

hobu:
> prefix DLL-functions with a prefix that something along the lines of `__cdecl`.

Yes, that is the modern way to do it. GDAL and GEOS have good examples that could be followed.

rouault:
I've introduced this for my C++ PROJ stuff
```
#ifndef PROJ_DLL
#ifdef PROJ_MSVC_DLL_EXPORT
#define PROJ_DLL __declspec(dllexport)
#elif defined(PROJ_MSVC_DLL_IMPORT)
#define PROJ_DLL __declspec(dllimport)
#else
#define PROJ_DLL
#endif
#endif
```
and src/CMakeLists.txt has
```
ifdef(MSVC)
    add_definitions(-DPROJ_MSVC_DLL_EXPORT=1)
endif()
```

kbevers:
> I've introduced this for my C++ PROJ stuff

Let's see if we can expand that to the existing API once the WKT2 work is merged

hobu:
Maybe rename to simply `PROJ_DLL` in case anyone wants to do other kinds of visibility stuff like 
`_attribute__ ((visibility("default")))`

cffk:
CMake simplifies this using

   set (CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)

This obviates the need to modify the source code or to manually
maintain a .def file.

Downsides:

   * requires building with CMake
   * requires CMake version 3.4 or later

hobu:
> I'm not super keen on exported all symbols by default. Having control is a good thing to avoid users using stuff we don't intend to be for public consumption

I agree. Especially for data structures.

kbevers:
Right, I don't think there's any point in merging this since very few functions can be removed from the DLL as is. I'll keep the branch around as a reference of which functions from `projects.h` that are use in `proj` and `cs2cs` and should eventually be refactored away.

So, moving forward I think we need to gradually migrate the CLI apps to using `proj.h` functions AND replace `proj.def` with use of the `PROJ_DLL` prefix.

PR #1133, opened 2018-09-21
the Bertin 1953 projection

### LINKED ISSUE #1120: Add Bertin1953 projection
I documented and added the Bertin1953 projection to D3 a while ago. Would there be interest in porting it to proj.4?

https://visionscarto.net/bertin-projection-1953
https://github.com/d3/d3-geo-projection#geoBertin1953

(It is very often used by French cartographers.)

Fil:
closes #1120


src/PJ_bertin1953.c
```diff
@@ -0,0 +1,109 @@
...
+
+static XY s_forward (LP lp, PJ *P) {
+    XY xy = {0.0,0.0};
+    struct pj_opaque *Q = P->opaque;
+
+    /* Projection constants */
+    double fu = 1.4, k = 12., w = 1.68;
+
+    /* Aliases */
+    double lambda = lp.lam, phi = lp.phi;
```
kbevers: This is not particularly idiomatic PROJ code. If you just use `lp.lam` and `lp.phi` I immediately know what you are doing, with this I have to think about it every time I come across `lambda` and `phi`. I suggest you just stick to using the existing variable names.

src/PJ_bertin1953.c
```diff
@@ -0,0 +1,109 @@
...
+    
+    /* Variable */
```
kbevers: It's not really necessary to explain that this is a variable - C-programmers will be familiar the syntax :-)

docs/source/operations/projections/bertin1953.rst
```diff
@@ -0,0 +1,64 @@
...
+.. include:: ../options/R.rst
+
+.. include:: ../options/x_0.rst
+
+.. include:: ../options/y_0.rst
+
+Further reading
+################################################################################
+
+#. Philippe Rivière (2017). `Bertin Projection (1953) <https://visionscarto.net/bertin-projection-1953>`, Visionscarto.net.
```
kbevers, line 64: Is this the best reference we have? If possible, I'd like to have a reference to the mathematical description as well, preferably an academic paper.

docs/source/operations/projections/bertin1953.rst
```diff
@@ -0,0 +1,64 @@
+.. _bertin1953:
+
+********************************************************************************
+Bertin 1953
+********************************************************************************
+
+.. versionadded:: 5.2.0
```
kbevers: This will be included in version 6.0.0, not 5.2.0 (which was just released).

src/PJ_bertin1953.c
```diff
@@ -0,0 +1,109 @@
...
+    lambda = adjlon(lambda);
+
+    /* Adjust pre-projection */
+    if (lambda + phi < -fu) {
+      d = (lambda - phi + 1.6) * (lambda + phi + fu) / 8.;
+      lambda += d;
+      phi -= 0.8 * d * sin(phi + M_PI / 2.);
+    }
+
+    /* Project with Hammer (1.68,2) */
```
kbevers, line 57: So this is the same Hammer projection as defined in `PJ_hammer.c`, yes? Why not use that directly by instantiating a Hammer specific `PJ` object with `Q->P_hammer = proj_create(P->ctx, '+proj=hammer ...')` in the projection setup and call `pj_fwd(Q->P_hammer, lp)` here?

This avoids duplicate code and potential bugs in Hammer will be fixed both places if necessary.

docs/source/operations/projections/bertin1953.rst
```diff
@@ -0,0 +1,64 @@
...
+
++---------------------+-------------------------------+
+| **Classification**  |                               |
```
kbevers: I'd say this is categorized as miscellaneous

kbevers (changes requested):
Thanks for opening this PR.

I have added some inline comments which I think will improve the quality of this PR. In addition I a few general comments/changes I'd like to see before this gets merged:

1. We need tests.
2. I'd prefer if the code didn't use camelCase variable names. Generally speaking (there are exceptions), PROJ uses snake_case.

I would like to see a more formal mathematical description of this projection. As I understand it, the projection is basically the Hammer projected subjected to a rotation and some inital warping. I'd be interested to see if we couldn't to the same thing by constructing a pipeline of some more basic operations. This is just me thinking aloud, not something that should be pursued in this PR :-)

docs/source/operations/projections/bertin1953.rst
Fil: Well if you read that modest blog post, you'll know :)
(I wish I had time and resources to publish this as an academic paper.)

src/PJ_bertin1953.c
Fil: Yes I wanted to do that, but couldn't find an example in the src/

kbevers: Something similar is done in `PJ_deformationc.`:

https://github.com/OSGeo/proj.4/blob/0b32b840d3161c64d8f1f487b153178528c8c124/src/PJ_deformation.c#L270-L272

https://github.com/OSGeo/proj.4/blob/0b32b840d3161c64d8f1f487b153178528c8c124/src/PJ_deformation.c#L89

You'll have to adjust it a bit but the general idea is there. Also notice that you need to create a custom descructor for the `PJ` object. There's an example for that in the same file as well.

docs/source/operations/projections/bertin1953.rst
kbevers: I did, but it wasn't clear to me if something else was published.

You are welcome to use the PROJ docs as a platform for describing the mathematics behind the projection.

src/PJ_bertin1953.c
Fil: Feel free to add this! It's my first contribution to Proj.4 and it had been a looong time since I coded anything in C :-)

docs/source/operations/projections/bertin1953.rst
Fil: PS: I agree that my construction is totally ad-hoc, and that another formula might be much nicer mathematically (esp. with second and third derivatives). It's the best I could come up with as a first approximation, and I'm very open to making an upgraded formula at some point (provided it still respects the original drawing).

kbevers (approved):

Fil:
thank you!

PR #1136, opened 2018-09-27
cs2cs_emulation_setup: fix issue with non C-locale

rouault:
in +towgs84 case, we use sprintf() with floating-point formatter to output
the ellipsoid parameters. For a locale with decimal separtor != dot, the
resulting string will not be parsed correctly by proj_atof(), leading to
wrong numeric result.

The fix is similar to the one done in pj_latlong_from_proj()

Note for later: if using C++, we could use a locale-independent formatting
solution to avoid such issue.

### DIFF
--- a/src/proj_4D_api.c
+++ b/src/proj_4D_api.c
@@ -519,6 +519,17 @@ Returns 1 on success, 0 on failure
     if (P->is_geocent || P->helmert || do_cart) {
         char def[150];
         sprintf (def, "break_cs2cs_recursion     proj=cart   a=%40.20g  es=%40.20g", P->a_orig, P->es_orig);
+        {
+            /* In case the current locale does not use dot but comma as decimal */
+            /* separator, replace it with dot, so that proj_atof() behaves */
+            /* correctly. */
+            /* TODO later: use C++ ostringstream with imbue(std::locale::classic()) */
+            /* to be locale unaware */
+            char* next_pos;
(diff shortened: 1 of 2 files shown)


src/proj_4D_api.c
```diff
@@ -519,6 +519,19 @@ Returns 1 on success, 0 on failure
     if (P->is_geocent || P->helmert || do_cart) {
         char def[150];
         sprintf (def, "break_cs2cs_recursion     proj=cart   a=%40.20g  es=%40.20g", P->a_orig, P->es_orig);
+        {
```
busstoptaktik, line 522: This block is stylistically quite different from its surroundings. I would prefer something like this (untested):
```
if (P->is_geocent || P->helmert || do_cart) {
    char def[150];
    char *next_pos = def;

    sprintf (def, "break_cs2cs_recursion     proj=cart   a=%40.20g  es=%40.20g", P->a_orig, P->es_orig);

    /* In case the current locale does not use dot but comma as decimal */
    /* separator, replace it with dot, so that proj_atof() behaves */
    /* correctly. */
    /* TODO later: use C++ ostringstream with imbue(std::locale::classic()) */
    /* to be locale unaware */
    while ( (next_pos =  strchr (next_pos, ',')))
        *next_pos++ = '.';

    ...
```
or even a for rather than while:

    for (next_pos = def; next_pos = strchr (next_pos, ','); next_pos++)
        *next_pos = '.'

Alternatively, we could build comma-dot agnosticism into the alternative proj_strtod in proj_strtod.c

PR #1137, opened 2018-09-28
Remove nmake build system

rouault:
This was announced as scheduled for PROJ 6.

My ongoing work on the iso19111 branch hasn't updated the makefile.vc stuff and only supports cmake builds for Windows


kbevers:
I tried to do something similar a week ago but never pushed the changes because I realized that our CMake setup can't produce a drop-in replacement for NMake. As far as I can tell from this PR it will also be the case here. I am okay with merging this but we need to make sure that we have a setup that is as close to being a drop-in replacement for users relying on NMake builds.

rouault:
What is the issue exactly with the cmake build ?

kbevers:
> What is the issue exactly with the cmake build ?

I failed to create a NMake generator that would output a DLL. I think we need the `__cdecl` qualifiers before functions to include in a DLL before that is possible. Or it may be that I am just a complete idiot with CMake and I have missed something obvious at the time...

mloskot:
How did you configure to build DLL? Have you set https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

kbevers:
> How did you configure to build DLL?

I didn't do any changes to the CMake files, I just ran a `cmake -G "NMake Makefiles"` and then build with `nmake`. As I recall, I ran into some problems, the DLL was one of them. I am not near a windows computer right now so I can't be more specific than this. I'll look into this again on monday.

mloskot:
Typically, to build shared library it takes `-DBUILD_SHARED_LIBS=ON` which is a native cmake option, but some projects especially ancient cmake configs define their own custom equivalents. I don't recall what CMakeLists.txt of PROJ do, writing from mobile now but will try to check later

rouault:
There's a dedicated -DBUILD_LIBPROJ_SHARED=ON switch which is used by appveyor.yml to test DLL builds

kbevers:
Maybe things aren't as bad as I thought. I'm probably not the only one who can't find their way around this stuff so maybe we should add something to the install/build instructions on how to do this.

mloskot:
> There's a dedicated -DBUILD_LIBPROJ_SHARED=ON 

Any reason to not to switch over to the official variable?
https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

I think less custom variables/options the better, self-descriptive, knowledge-transitive setup.

rouault:
> Any reason to not to switch over to the official variable?

I don't think so unless @hobu has another opinion (this comes from c90714e3b141e2ee02d110a73818f71f726a35fb)

kbevers:
Okay, so to follow up on this, here's what I did in my previous test:

```
> C:\dev\build\nmakeproj>cmake -G "NMake Makefiles" ..\..\proj
> nmake
```

which after compilation doesn't createa DLL. Running cmake with `-DBUILD_LIBPROJ_SHARED=ON` produces a DLL so I guess that we can use the cmake generated nmake makefiles as a drop in replacement. At some point it would be nice to get it confirmed by someone that actually uses the DLL that it works for them if build this way.

PR #1139, opened 2018-09-29
Documentation building fixes

rouault:
*  Doc building: fix warning regarding Altamimi2002 citation not found
   There was an issue since transformation.rst which uses this citation is
    alphabetically after references.rst.
    The documented workaround in mcmtroffaes/sphinxcontrib-bibtex#8
     is to rename references.rst as zrefereces.rst, so it is alphabetically last.
*  Doc building: enforce warning as errors mode


kbevers:
Seems good to me. Nice have those intermittent warnings removed!

PR #1142, opened 2018-10-04
Add Lambert Conic Conformal (2SP Michigan) projection

### LINKED ISSUE #940: Add Lambert Conic Conformal (2SP Michigan)
As described on page 23 of the [EPSG guidance note 7-2](http://www.epsg.org/Portals/0/373-07-2.pdf). The only difference from this to the normal 2SP LCC is a scaling factor on the ellipsoid. Should be straight forward to implement, preferably as a new operation, e.g. `+proj=lccm`.

It is used in a few systems in the EPSG database which justifies adding this otherwise non-standard projection.

sphynx:
Fixes #940


docs/source/operations/options/scale.rst
```diff
@@ -0,0 +1,5 @@
+.. option:: +scale=<value>
+
+    Ellipsoid scale factor. Determines ellipsoid scale factor used in the projection.
```
rouault: As this is a very specific options to this projection method, I would keep it inline in lccm.rst

docs/source/operations/projections/lccm.rst
```diff
@@ -0,0 +1,35 @@
...
+********************************************************************************
+Lambert Conformal Conic (2SP Michigan)
+********************************************************************************
+
+.. figure:: ./images/lcc.png
+   :width: 500 px
+   :align: center
+   :alt:   Lambert Conformal Conic (2SP Michigan)
+
+   proj-string: ``+proj=lccm +lon_0=-90 +k=1.0000382``
```
rouault: should be scale and not k ?
(I was wondering if 'k' shouldn't be used instead of 'scale', but its meaning in LCC_2SP Michigan is very specific, so perhaps scale is fine)

src/PJ_lccm.c
```diff
@@ -0,0 +1,149 @@
...
+ *****************************************************************************/
+
+#define PJ_LIB__
+#include <errno.h>
+#include "proj.h"
+#include "projects.h"
+#include "proj_math.h"
+
+PROJ_HEAD(lccm, "Lambert Conformal Conic (2SP Michigan)")
+    "\n\tConic, Sph&Ell\n\tlat_1= and lat_2= or lat_0";
```
rouault: Perhaps scale should be mentioned too

docs/source/operations/options/scale.rst
sphynx: Yes, it makes sense, will change that.

I was also not sure about the name 'scale'. I think 'k_0' (and 'k' in the past) is generally used for scaling, but since this is a different kind of scaling I decided that it would be better to have a separate option than to reuse 'k_0' somehow ('k_0' can still be used in conjunction with 'scale' though).

docs/source/operations/projections/lccm.rst
sphynx: Good catch! As I said in the previous comment I tried 'k' first, but then switched to 'scale', but forgot to change the doc.

src/PJ_lccm.c
sphynx: Ok, should I just add it to the end? i.e. "lat_1= and lat_2= or lat_0, scale="?

src/PJ_lccm.c
```diff
@@ -0,0 +1,149 @@
...
+PJ *PROJECTION(lccm) {
+    double cosphi, sinphi;
+    int secant;
+    struct pj_opaque *Q = pj_calloc (1, sizeof (struct pj_opaque));
+
+    if (0==Q)
+        return pj_default_destructor (P, ENOMEM);
+    P->opaque = Q;
+
+    if (pj_param(P->ctx, P->params, "tscale").i)
```
kbevers: A bit simpler if written like:

```
    Q->scale = 1.;
    if (pj_param(P->ctx, P->params, "tscale").i)
      Q->scale = pj_param(P->ctx, P->params, "dscale").f;
```

test/gie/builtins.gie
```diff
@@ -2536,6 +2535,61 @@ accept  -200 -100
...
 
+===============================================================================
+Lambert Conformal Conic (2SP Michigan)
+	Conic, Sph&Ell
+	lat_1= and lat_2= or lat_0
+===============================================================================
+
+Should behave like normal LCC when scale is set to 1 (or not used).
+-------------------------------------------------------------------------------
+operation +proj=lccm +ellps=GRS80 +lat_1=0.5 +lat_2=2
```
kbevers: Instead of these generic test coordinates, maybe try to add some that are picked more deliberately. See https://coveralls.io/builds/19347707/source?filename=src%2FPJ_lccm.c#L43 for an overview of what lines you are not hitting at the moment.

The story behind the existing tests similar to these ones is that we auto-generated them all before a major restructure of the code in order to have some rudimentary regression tests. To be honest we didn't pick the most suitable coordinates, the only criteria here was to hit each quadrant of the coordinate system. When writing new test by hand more care should be taken in my opinion.

test/gie/more_builtins.gie
```diff
@@ -243,6 +243,23 @@ expect    failure errno no_args
 
 
 
+-------------------------------------------------------------------------------
+Tests for PJ_lccm.c
+-------------------------------------------------------------------------------
+This test is taken from EPSG guidance note 7-2 (version 54, August 2018,
```
kbevers, line 249: Is this the only test coordinate in guidance note 7-2? If so, can you dig out some coordinates from another source?

docs/source/operations/projections/lccm.rst
```diff
@@ -0,0 +1,39 @@
...
+Lambert Conformal Conic (2SP Michigan)
+********************************************************************************
+
+.. figure:: ./images/lcc.png
+   :width: 500 px
+   :align: center
+   :alt:   Lambert Conformal Conic (2SP Michigan)
+
+   proj-string: ``+proj=lccm +lon_0=-90 +scale=1.0000382``
+
```
kbevers: Please write a short description of the projection here. If possible it would be nice to have a reference to an academic paper that describes the projection. Otherwise a reference the EPSG guidance note 7-2 or Wikipedia will do as well.

docs/source/operations/projections/lccm.rst
```diff
@@ -0,0 +1,39 @@
+.. _lccm:
+
+********************************************************************************
+Lambert Conformal Conic (2SP Michigan)
+********************************************************************************
+
```
kbevers: Please add a table similar to this one from `merc.rst`:

https://github.com/OSGeo/proj.4/blob/3378c5ec96fb3f067beffc6c3acd15765d5ce5d1/docs/source/operations/projections/merc.rst#L12-L26

kbevers (changes requested):
Thanks for adding this. I've inlined some suggestions for improvements to the code, tests and docs.

src/PJ_lccm.c
```diff
@@ -0,0 +1,149 @@
...
+    if (fabs(Q->phi1 + Q->phi2) < EPS10)
+        return pj_default_destructor(P, PJD_ERR_CONIC_LAT_EQUAL);
+
+    Q->n = sinphi = sin(Q->phi1);
+    cosphi = cos(Q->phi1);
+    secant = fabs(Q->phi1 - Q->phi2) >= EPS10;
+    if( (Q->ellips = (P->es != 0.)) ) {
+        double ml1, m1;
+
+        P->e = sqrt(P->es);
```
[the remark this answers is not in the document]
sphynx: Hi Kristian! Thanks for your feedback. Actually, this code is mostly just a copy of PJ_lcc.c, so I tried not to touch it all, since, as you mention in the issue #940, the only difference between LCCM and LCC is that scaling factor which appears in three places in the code. I haven't even touched any stylistic issues which I found, trying to keep the diff between PJ_lcc and PJ_lccm to a bare minimum: just those three lines in question. 

Therefore to apply your proposals, it would be nice:

- either to reuse all that projection logic from normal LCC in LCCM (I'm not sure what is the best way to achieve this with existing code structure). We can probably create a common file and reuse it in both LCC and LCCM. Or we can just express LCC via LCCM with +scale=1.
- or to keep them in sync, i.e. to apply your proposed changes in both files. Keeping both of them would obviously increase the maintenance cost (people have to apply changes in two places and eventually someone will forget to do that).

Please let me know what is your preferred way. Perhaps I can do that refactoring work in a separate branch.

src/PJ_lccm.c
sphynx: Agreed.

test/gie/builtins.gie
sphynx: Yes, makes sense. I took a look on coveralls report after seeing that the coverage has decreased and took note of the lines which were not tested (mostly corner cases). I'll add some more interesting tests.

test/gie/more_builtins.gie
sphynx: Yes, it's the only one test in that document. I'll try to google that projection and see if I can find any other test coordinates. But I think just one serious test for the common case is enough (other "normal" coordinates probably won't add much in terms of testing if they are not corner cases)

docs/source/operations/projections/lccm.rst
sphynx: I'll do that.

docs/source/operations/projections/lccm.rst
sphynx: OK.

src/PJ_lccm.c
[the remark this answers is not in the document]
rouault: Oh, I didn't realize that LCC_2SP Michigan was so close to LCC ! I'd definitely think that we should just merge them together. ie add the few lines needed to PJ_lcc.c and take into account the ellipsoidal scale in the e_forward and e_inverse. And update LCC documentation to properly document the LCC_2SP Michigan trick. What do you think @kbevers  ?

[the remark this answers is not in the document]
kbevers: +1 on merging the two. If I understand correctly the scale factor is main difference and that is definitely smarter to hand in one file. A clean up of the code is welcome :-)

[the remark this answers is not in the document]
rouault: We probably don't need a dedicated lccm. EPSG distinguishes LCC_1SP and LCC_2SP, but we have a single PROJ method for both, so that would fit that model. The +scale != 1 will be the signature of LCC_2SP Michigan. But I don't have a strong opinion on this. I'll defer to @kbevers judgment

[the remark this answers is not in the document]
kbevers: Let's keep it in the same projection. Sounds like we would just introduce a bunch of duplicate code otherwise.

src/PJ_lcc.c
```diff
@@ -5,22 +5,22 @@
 #include "proj_math.h"
 
 PROJ_HEAD(lcc, "Lambert Conformal Conic")
-    "\n\tConic, Sph&Ell\n\tlat_1= and lat_2= or lat_0";
+    "\n\tConic, Sph&Ell\n\tlat_1= and lat_2= or lat_0, scale=";
```
rouault: could you update  docs/source/operations/projections/lcc.rst to document scale ?

sphynx: Sure, see the next commit.

test/gie/more_builtins.gie
sphynx: I've tried to find more "official" test coordinates, but haven't found any. While doing my research I've found this very relevant and interesting message:

http://lists.maptools.org/pipermail/proj/2015-April/007108.html

It provides three ways of emulating 2SP Michigan in existing proj.4. Author of the message provides several coordinates which he used for his testing, but he warns that they are not official and he may have made a mistake somewhere. 

However, what is more interesting - is that one can just use normal `k_0` option instead of my new `scale` and it magically works! I've just tried it in my tests and indeed, if we just replace `+scale=1.0000382` with `+k_0=1.0000382` it still gives the same results for the official test coordinates! Even though it obviously hits the different code path.

docs/source/operations/projections/lcc.rst
```diff
@@ -31,3 +64,19 @@ Parameters
 .. include:: ../options/x_0.rst
 
 .. include:: ../options/y_0.rst
+
```
rouault, line 67: would be good to also point to k_0.rst which is used in the LCC_1SP formulation

test/gie/more_builtins.gie
rouault: Actually it is not magic at all. When looking closely at the formulas, I can convince myself that k_0 incorporates the effect of the new scale parameter. In theory we could merge both together, as LCC_2SP Michigan doesn't use the k_0 = scale factor at natural origin which is used only by LCC_1SP. In practice, it might be good to have the different parameters to distinguish the different semantics (and k_0 is typically < 1, whereas scale is typically > 1). I don't have a strong opinion on this however

test/gie/builtins.gie
sphynx: Done

src/PJ_lccm.c
sphynx: Done

src/PJ_lccm.c
```diff
@@ -0,0 +1,149 @@
...
+        if (!pj_param(P->ctx, P->params, "tlat_0").i)
+            P->phi0 = Q->phi1;
+    }
+    if (fabs(Q->phi1 + Q->phi2) < EPS10)
+        return pj_default_destructor(P, PJD_ERR_CONIC_LAT_EQUAL);
+
+    Q->n = sinphi = sin(Q->phi1);
+    cosphi = cos(Q->phi1);
+    secant = fabs(Q->phi1 - Q->phi2) >= EPS10;
+    if( (Q->ellips = (P->es != 0.)) ) {
```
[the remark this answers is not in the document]
sphynx: Done

src/PJ_lccm.c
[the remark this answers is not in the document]
sphynx: Done

docs/source/operations/projections/lccm.rst
sphynx: Done in PJ_lcc.rst

test/gie/more_builtins.gie
kbevers: Are there scenarios where you would use both the `+k_0` and `+scale` factors?

If not I prefer to reuse `+k_0`. It's the canonical way to define a scale factor in PROJ and has been since forever. Care should be taken to explain in the examples how it can be used in different circumstances.

rouault: > I am not sure about scenarios of simultaneous use of both `+k_0` and `+scale` factors

Neither LCC_1SP, LCC_2SP or LCC_2SP_Michigan require both, so we can probably forget about tose

> Frankly, I feel somewhat disheartened by this issue

Sorry for that. We don't really know about the real difficulty (or triviality!) unless really digging into things. That wasn't mean to trap you. Your work on this has been very helpful, at least to demistify LCC_2SP Michigan... If you could just remove the scale parameter in the PJ_lcc.c file and replace scale by k_0 in the documentation and tests, I think your PR should be good to merge !

kbevers: > Frankly, I feel somewhat disheartened by this issue.

I am sorry to hear that. To be honest, I assumed that this would need to be a new projection in it's own right when I wrote that issue. You have helped clear that up, so your efforts haven't been in vain. Thank you for that.

I still feel like this is a "good first issue". Turns out that it is rather simple to fix :-) But in all seriousness, I somewhat expect a bit of dialogue before new contributors take on issues. Not that you aren't allowed to just crack on but sometimes we can help the process along from the start so a bunch of unnecessary work is avoided. In this case we would probably have realised that there's wasn't much else to this issue than documentation.

I hope you aren't completely discouraged and will consider contributing on other issues. There are plenty of them, even for programmers that don't know much about geography and cartography :-)

sphynx: @rouault @kbevers Thank you for your support and prompt responses, much appreciated.

> If you could just remove the scale parameter in the PJ_lcc.c file and replace scale by k_0 in the documentation and tests, I think your PR should be good to merge !

Ok, I've done that and updated everything related to `+scale`. I have also explicitly included `+k_0` in the LCC documentation providing slightly different, more specific description for this projection, mentioning SP2 Michigan. 

> I somewhat expect a bit of dialogue before new contributors take on issues. 

Makes sense. Before I started I hadn't expected that the maintainers of the package would respond that quickly, so I thought it's better to do something first and then start discussing to make the discussion more concrete :)

> I hope you aren't completely discouraged and will consider contributing on other issues. 

Haha, yes, I am tempted by the famous Dymaxion, but it looks somewhat scary :)

docs/source/operations/projections/lcc.rst
```diff
@@ -65,9 +65,7 @@ Parameters
 
 .. include:: ../options/y_0.rst
 
-.. include:: ../options/k_0.rst
-
-.. option:: +scale=<value>
+.. option:: +k_0=<value>
```
rouault, line 68: The other use of k_0 should be described. In LCC_1SP, this represents the "Scale factor at natural origin" parameter .

sphynx: OK. Is this definition acceptable?

    This parameter can represent two different values depending on the
    form of the projection. In LCC 1SP it determines the scale factor
    at natural origin. In LCC 2SP Michigan it determines the ellipsoid
    scale factor.

rouault:
Looks good to me! Would you mind squashing all the commits into a single one ? (I can also do it with the github UI but then the commit would appear as having you and me as authors)

docs/source/operations/projections/lcc.rst
sphynx: Done.

sphynx:
> Looks good to me! Would you mind squashing all the commits into a single one ? (I can also do it with the github UI but then the commit would appear as having you and me as authors)

Squashed and pushed with `--force`.

kbevers (approved):

kbevers:
It also looks good to me. Thanks for putting up with our annoying requests. I think this has turned out to be a great PR giving us both better documentation, tests and understanding of the projection. Good job!

PR #1143, opened 2018-10-06
Make PROJ_HEAD descriptions more uniform

sphynx:
While working with proj.4 I've noticed that the output of descriptions from `proj -lP` is not very uniform: some of the projections have periods in their descriptions, i.e. 'Cyl., Sph.' while others don't (and the split is approximately half/half). It does not look very good, so I've decided to fix it and make it more uniform with this commit. I've dropped the dots from 'Cyl.', 'Sph.' and 'inv.' and applied a couple of other minor changes, like 'Ellps. -> Ell' in one or two cases.

### DIFF
--- a/src/PJ_airy.c
+++ b/src/PJ_airy.c
@@ -31,7 +31,7 @@
 #include <errno.h>
 #include "projects.h"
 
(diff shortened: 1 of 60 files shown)


kbevers:
Nice! Thanks for doing this.

PR #1145, opened 2018-10-08
update docker pull command

mbernasocchi:
changed docker pull osgeo/proj.4 to changed docker pull osgeo/proj


hobu:
This needs to be against the reStructured text which generates the docs, not generated output, which exists in the `gh-pages` branch.

PR #1148, opened 2018-10-10
Move struct definitions for proj_list_* functions to proj.h

### LINKED ISSUE #1147: incomplete type 'PJ_ELLPS' when using public API from "proj.h"
My code just includes "proj.h". The types like `PJ_COORD` is found and complete, i.e. i can work with it's members (documented in `datatypes.html`); but the following code

```cpp
    auto p =  proj_list_ellps();
    while (p) {
        std::count << p->id << "\n";
    ...
    }
```

leads to error 

```
error: invalid use of incomplete type 'const struct PJ_ELLPS'
         std::count << p->id << "\n";
                        ^~
````

### PR DESCRIPTION
**Pull request by kbevers**
With projects.h not being available to outside users anymore we need to
define PJ_UNITS, PJ_ELLPS, PJ_PRIME_MERIDIANS and PJ_OPERATIONS
elsewhere.

Fixes #1147

### DIFF
--- a/src/proj.h
+++ b/src/proj.h
@@ -180,16 +180,34 @@ struct PJ_INIT_INFO;
 typedef struct PJ_INIT_INFO PJ_INIT_INFO;
 
 /* Data types for list of operations, ellipsoids, datums and units used in PROJ.4 */
-struct PJ_LIST;
+struct PJ_LIST {
+    const char  *id;                /* projection keyword */
+    PJ          *(*proj)(PJ *);     /* projection entry point */
+    const char  * const *descr;     /* description text */
+};
+
 typedef struct PJ_LIST PJ_OPERATIONS;
 
-struct PJ_ELLPS;
+struct PJ_ELLPS {
+    const char  *id;    /* ellipse keyword name */
+    const char  *major; /* a= value */
+    const char  *ell;   /* elliptical parameter */
+    const char  *name;  /* comments */
+};
 typedef struct PJ_ELLPS PJ_ELLPS;
 
-struct PJ_UNITS;
+struct PJ_UNITS {
+    const char  *id;        /* units keyword */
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: src/proj.h [hunk 1]
```diff
@@ -180,16 +180,34 @@ struct PJ_INIT_INFO;
 typedef struct PJ_INIT_INFO PJ_INIT_INFO;
 
 /* Data types for list of operations, ellipsoids, datums and units used in PROJ.4 */
-struct PJ_LIST;
+struct PJ_LIST {
+    char    *id;                 /* projection keyword */
```
rouault: would be nice if we could use const char*

kbevers: I don't see why we couldn't. Do you want to change the other `char *`s in `PJ_ELLPS` and friends as well?

kbevers:
There's also `PJ_DATUMS` and `pj_datums_ref()` which I so far have not touched. My feeling is that in the long term the `+datum` parameter will be used less and less (unless we repurpose it in conjunction with the EPSG DB stuff) and it will end up as dead weight in the `proj.h` API.

In the short to medium term not having it in `proj.h` may make things a bit more difficult in the code for `cs2cs`. I am sure if we have any outside consumers of that particular function and datastructure.

Thoughts?

src/proj.h
kbevers: @rouault ping :-)

kbevers: I gave it a go. It wasn't as complicated as I feared. Do you mind taking another look at the updated commits @rouault ?

rouault:
LGTM.

That said, on a higher level, there might be a risk of confusion for users between those functions that return hard-coded prime-meridian / ellipsoid / datum, and the new ones that will query the database (but will return standard names). But I guess some people (my own new code included) will still need to query PROJ for its hard-coded strings.

PR #1154, opened 2018-10-14
Make plot.py error message more helpful

sphynx:
I tried to generate images for the new projection and was somewhat confused by the original error message. I think it's worth improving with providing some suggestion to the user. I've seen this suggestion in the source code, but I think makes sense to put it into the error message as well.


kbevers:
Yeah, that probably make it a bit more clear what's going on. Thanks.

PR #1162, opened 2018-10-26
Disallow use of non-projection operations in proj (was: Don't convert non-angular inputs to proj to radians)

kbevers:
[As reported on the mailing list](http://lists.maptools.org/pipermail/proj/2018-October/008436.html), calling `proj` with a operation that doesn't take angular input will not return the expected result. This is changed with this commit.

### DIFF
--- a/src/proj.c
+++ b/src/proj.c
@@ -160,12 +160,12 @@ static void process(FILE *fid) {
             }
         } else {    /* x-y or decimal degree ascii output, scale if warranted by output units */
             if (inverse) {
-                if (Proj->left == PJ_IO_UNITS_ANGULAR) {
+                if (proj_angular_input(Proj, PJ_FWD)) {
                     data.uv.v *= RAD_TO_DEG;
                     data.uv.u *= RAD_TO_DEG;
                 }
             } else {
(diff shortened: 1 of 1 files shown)


src/proj.c
```diff
@@ -132,6 +131,11 @@ static void process(FILE *fid) {
                 facs_bad = proj_errno(Proj);
             }
 
+            /* undo radian concersion in case we don't get angular input coordinates */
```
rouault: typo: conversion.Also below

kbevers: Thanks

kbevers:
Thinking more about this, I am having doubts that this is the best solution. I am starting to lean towards `proj` only accepting projection operations (that is, operations with `P->left = PJ_UNITS_ANGULAR` and `P->right = PJ_UNITS_PROJECTED`). When a non-projection operation is initialized, `proj` should fail and tell the user that `proj` is only for projections and reference `cct`. Thoughts?

kbevers:
I decided to change the logic so that `proj` will only accept operations requiring angular inputs and produces non-angular output. I think this is a simpler approach and it is also more in line with how `proj` originally was intended to be used.

PR #1163, opened 2018-10-26
cmake: remove unused code

mathstuf:
These functions aren't called from anywhere else in the codebase.


mathstuf:
This does bump the minimum CMake version to 2.8.3, but since it is from 2010 and proj requires C++11, this seems OK to me.

kbevers:
Thanks for cleaning this up!

PR #1164, opened 2018-10-26
cmake: detect PTHREAD_MUTEX_RECURSIVE by compiling

### LINKED ISSUE #1158: Use check_c_source_compiles for PTHREAD_MUTEX_RECURSIVE
As explained in [this VTK issue](https://gitlab.kitware.com/vtk/vtk/issues/17178), PTHREAD_MUTEX_RECURSIVE might be an rvalue in some implementations and therefore incompatible with `check_symbol_exists`. Instead, `check_c_source_compiles` should be used instead.

mathstuf:
The `check_symbol_exists` cannot detect `#define symbol 1` because it
takes the address of the symbol to detect it, but the address of a
literal is not allowed. Some platforms define `PTHREAD_MUTEX_RECURSIVE`
by such a literal.

Fixes #1158 

---
Cc: @kbevers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,11 +82,6 @@ set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 find_package (Threads)
 
 include(CheckIncludeFiles)
-include(CheckSymbolExists)
-CHECK_SYMBOL_EXISTS(PTHREAD_MUTEX_RECURSIVE pthread.h HAVE_PTHREAD_MUTEX_RECURSIVE_DEFN)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -112,6 +107,23 @@ else ()
   add_definitions (-DHAVE_C99_MATH=0)
 endif ()
 
+find_package(Threads)
```
rouault: find_package(Threads) is also at line 82. Probably the one at line 82 should be removed then ?

kbevers:
Thanks for taking care of this!

Issue #1118, opened 2018-09-11
CMake C++11 configuration


hobu:
We are manually setting CFLAGS for `-std=c++11` all over the place. This is problematic for a wide constelation of compilers. CMake takes care of it for us if we use 

```
set (CMAKE_CXX_STANDARD 11)
```

but we need to increase our CMake minimum version to 3.1+ to take advantage of it.

QuLogic:
It seems [relatively good](https://repology.org/metapackage/cmake/versions), except for those few old systems that aren't even >=3.

mloskot:
BTW, it's trivially easy to (re)install the latest CMake on (old) Linux:

```bash
#!/bin/sh
set -x

sudo rm -rf /usr/local/bin/cmake*
sudo rm -rf /usr/local/bin/ctest*
sudo rm -rf /usr/local/doc/cmake
sudo rm -rf /usr/local/man/man1/cmake*
sudo rm -rf /usr/local/man/man1/ctest*
sudo rm -rf /usr/local/man/man7/cmake*
sudo rm -rf /usr/local/share/cmake*

wget -O cmake-linux.sh https://cmake.org/files/v3.12/cmake-3.12.2-Linux-x86_64.sh
sudo sh cmake-linux.sh -- --skip-license --prefix=/usr/local
cmake --version
```

mloskot:
Ubuntu LTS could be a reference point perhaps, ie. 14.04 LTS offers CMake 3.5

IMHO, there switching 2.8 to 3.1 is not worth the CMake modernisation overhaul. Better do it when jumping to 3.5 at least.

kbevers:
> Ubuntu LTS could be a reference point perhaps, ie. 14.04 LTS offers CMake 3.5

Debian stable is usually quite conservative too, it seem to be at CMake 3.6 so yeah, 3.5 is probably not a bad choice. Going over the list @QuLogic supplied it generally seems that CMake 3.5 or newer is available on all major distributions. And most lesser known as well.

No objection from me on going for a minimum version 3.5. I am interested to hear if there are arguments against that I haven't considered.

kbevers:
Let's hear some of the usual suspects opinion on this. @hobu, @cffk, @schwehr Any arguments against requiring a minimum version of CMake 3.5?

schwehr:
On (almost) debian testing, I get

```
cmake --version
cmake version 3.10.2
```

[...]

PR #1168, opened 2018-10-31
[ADOPTED] RFC3: Dependency management

kbevers:
Here is a first attempt at formalising how we deal with updates to dependencies of PROJ.

This is an offspring of the conversation in #1118 where it was suggested to keep some sort of rolling update scheme. I have tried to turn that into an RFC. Most of it is based on gut feeling, while I think my choices are quite sane there may be concerns I haven't thought about.

### DIFF
--- a/docs/source/community/rfc/rfc-3.rst
+++ b/docs/source/community/rfc/rfc-3.rst
@@ -6,8 +6,8 @@ PROJ RFC 3: Dependency management
 
 :Author: Kristian Evers
 :Contact: kreve@sdfe.dk
-:Status: Proposed to the PSC
-:Last Updated: 2019-01-14
+:Status: Adopted
+:Last Updated: 2019-01-16
 
 Summary
 -------------------------------------------------------------------------------
--- a/docs/source/community/rfc/rfc-3.rst
+++ b/docs/source/community/rfc/rfc-3.rst
@@ -7,18 +7,18 @@ PROJ RFC 3: Dependency management
 :Author: Kristian Evers
 :Contact: kreve@sdfe.dk
 :Status: Draft
-:Last Updated: 2018-10-31
+:Last Updated: 2018-11-10
 
 Summary
 -------------------------------------------------------------------------------
 
-This document defines a rule set for dependency management in PROJ. With PROJ
-being a core component in many downstream software packages clearly stating
-which dependencies the library has is of great value. This document concern
-both programming language standards as well as minimum required versions of
-library dependencies and build tools.
+This document defines a set of guidelines for dependency management in PROJ.
+With PROJ being a core component in many downstream software packages clearly
+stating which dependencies the library has is of great value. This document
+concern both programming language standards as well as minimum required
+versions of library dependencies and build tools.
 
-It is propsed to adopt a rolling update scheme that ensures that PROJ is
+It is proposed to adopt a rolling update scheme that ensures that PROJ is
 sufficiently accessible, even on older systems, as well as keeping up with the
 technological evolution. The scheme is divided in two parts, one concerning
 versions of used programming languages within PROJ and the other concerning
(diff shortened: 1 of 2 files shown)


docs/source/community/rfc/rfc-3.rst
```diff
@@ -0,0 +1,112 @@
...
+behind the most recent standard. C++ is two iterations behind. After adoption of
+this RFC the required standard for compilation of PROJ will always be two
+iterations behind the most recent standard. This will result in a change of C
+standard immediately after adoption of the RFC. When a new standard for either C
+or C++ is released PROJ should change its requirement to the next standard in
+the line. For C++ that means a change in standard roughly every three years, for
+C the periods between standard updates is expected to be longer. Adaptation of
+new programming language standards should be coordinated with a major version
+release of PROJ. The initial switch to C99 will be coordinated with the release
+of PROJ 6.0.0. Adaptation of C++14 is possible with PROJ 7.0.0, provided that
```
rouault: Beyond the age of the standards, I think we should also look at what the compilers actually implement (and Microsoft is the one traditionally lagging behind). It might be also prudent to have a sort of X-version-behind rule for the compilers themselves (e.g with C++11, that is a 1-version-behind for Visual Studio, since you need VS >= 2015) . And perhaps take into consideration what is available in LTS versions of Linux distributions and in the CI environments we use

kbevers: > Beyond the age of the standards, I think we should also look at what the compilers actually implement (and Microsoft is the one traditionally lagging behind). It might be also prudent to have a sort of X-version-behind rule for the compilers themselves (e.g with C++11, that is a 1-version-behind for Visual Studio, since you need VS >= 2015) . 

I would prefer to keep this in as generic terms as possible but you have a point. My thinking was that by lagging behind two iterations of the standard compilers would be more or less caught up. It seems to me that Microsoft has started moving a bit faster in that regard, but maybe that is just my imagination? Both GCC and LLVM seem to be quite good at following along the standards.
Which compilers should be considered? MSVC, GCC, LLVM?

mloskot: FYI, VS2017 conforms to C++11, 14 and 17 now with -std=c++14 being the default mode.

hobu: I think the discussion about compiler versions should be oriented toward the C++ standard it supports and nothing else. Otherwise, the support matrix becomes convoluted and every developer ends up being responsible for cross-checking their development against the entire thing.

The matrix as defined right now covers "PROJ requires C++11" as a statement. I think there should be an RFC vote to change that to "PROJ requires C++14" or any other future revision.

kbevers: Thanks for your input guys. So summing up, it sounds like there's some agreement that we shouldn't just automatically switch to a newer revision of a standard because a new one is released. At least not without confirming that there's broad support of said standard revision first. That was sort also what I was going for but I think I'll have to reword the text to make that clearer. The example about switching to C++14 with version 7 was just that, an example demonstrating the idea. I would not suggest a switch at that point, but it might be beneficial for PROJ 8. I Agree that the PSC should confirm an update to this RFC before moving to a new C/C++ standard.

kbevers:
I have updated the draft with less strict requirements for updates of C/C++ standards, hopefully in line with what was expressed in the comments for the first draft. I have tried to maintain a fairly generic tone that doesn't go into details with specific compilers etc, basically leaving those considerations to the PSC when a change in C/C++ standard is required.

Additionally I have added a bootstrapping section that clearly states which dependencies PROJ has after the RFC has passed. The list is missing autoconf. It should probably be added too, I just have no idea what versions we support at the moment (presumably the minimum version supported is rather ancient).

kbevers:
Everyone, this has been standing still for quite some time now. I would like to put this up for a vote on the mailing list in the beginning of the new year. Would you please go over the RFC draft once more and see if it makes sense in the current form?

PR #1174, opened 2018-11-13
DOC: remove old docs that are published by USGS

mwtoews:
The external PDF https://pubs.usgs.gov/of/1990/of90-284/ofr90-284.pdf is much better than the local docs/old/of90-284.pdf file, which I've removed. The USGS file is 1/10 of the file size and allows text search.

The former local file was previously converted from a PS source (d8d578965b4cbcbf60ef7c2b36f87951e989cfb6). We're freeing up 21.4 MB with this PR.


mwtoews:
I've updated this PR to also remove a 21.4 MB docs/old/USGS-Snyder-Map-Projections-A-Working-Manual-1987.pdf file which appears to be identical to a USGS published manual from https://doi.org/10.3133/pp1395

So in all, we should have 42.8 MB cleared up, as these documents are published by USGS.

QuLogic:
Not quite, due to compression though. It's about 3MB more (out of 239!)

mwtoews:
correct, it would effectively add a bit more to the total repo size. Is rewriting history even an option?

QuLogic:
It would invalidate all forks and clones, and probably cause issues for infrequent contributors who might have a dormant fork. So I'd like to think it's unlikely.

kbevers:
> That would not even be a new clone, just git remote set-url origin <new place>, so I'm not sure that's even enough justification.

"some place else" doesn't have to be git-based ;-) But yeah, it is hard to change things without it being extremely annoying for a lot of people.

rouault:
Note: my remark about git repo size was not to oppose to this PR

kbevers:
I am rather indifferent I think. On one hand it is always nice to rely on the proper source of the documents, especially since they are in better shape. On the other hand it doesn't do any difference in terms of disk space used and you are at least guaranteed to have some basic background material readily at hand when you have a clone of the repo (comes in handy on airplanes etc.). So, it's a +0 from me.

mwtoews:
@kbevers I've added a README.md to the latest (squashed) commit to differentiate between the internal resources and external resources that were previously located here. Does this better help locate these documents?

My main comment on reduced directory size is that if one were to "Download ZIP" the resulting download is now over half the size as it is with the largish files. And as I previously stated, these PDFs are much more usable since they now have text access for (e.g.) searching.

PR #1181, opened 2018-11-20
Database: use official IGNF.xml registry to create content from the IGNF authority

rouault:
* Add grid alternatives for French vertical grids added per https://github.com/OSGeo/proj-datumgrid/pull/40

* Database: use official IGNF.xml registry to create content from the IGNF authority
    
    Up to now, we re-processed the data/IGNF PROJ.4 definition to ingest it into proj.db,
    but this file originally come from a processing of IGNF.xml
    ( http://librairies.ign.fr/geoportail/resources/IGNF.xml )
    
    The end result is not strictly equivalent, as data/IGNF has some 'magic' to create
    towgs84 / nadgrids, since IGNF.xml doesn't necessary contain all transformations
    from its geodetic systems to WGS84. I've tried to re-add some of those missing transforms
    (null Helmert transforms), so it can be used for pivoting, but that might be incomplete.


rouault:
@dgricci FYI

PR #1182, opened 2018-11-26
Remove data/epsg, IGNF and esri.* files / support legacy +init=epsg:XXXX syntax

rouault:
This is an attempt at resolving the issues raised in https://github.com/OSGeo/proj.4/issues/1178

- compatibility emulation of +init=epsg:XXXX / +init=IGNF:XXXX syntax using the database to do the resolution, and with the assumptions we traditionnaly did on axis order and units. This emulation is enabled in the following cases:
      - using cs2cs,
      - using proj_api.h API: pj_init(), pj_init_ctx(), pj_init_plus(), pj_init_plux_ctx()
      - or setting PROJ_USE_PROJ4_INIT_RULES to YES / proj_context_use_proj4_init_rules(ctx, TRUE)
- cs2cs modified to use proj_create_crs_to_crs()
- cs2cs now accepts syntax like "cs2cs EPSG:4326 +to EPSG:32631" in which case, it will assume EPSG axis order and unit. If using "cs2cs +init=epsg:4326 ...", it will assume traditional order.
- proj_create_crs_to_crs() modified to use the C++ code to compute the best transformation pipeline.
- proj_create_crs_to_crs() can accept as source and target values, and string accepted by the C++ createFromUserInput() API, that is a AUTH:CODE, a PROJ string or a WKT string.
- note: so, as cs2cs is using proj_create_crs_to_crs(), one can use cs2cs with WKT strings as well: "cs2cs 'PROJ_CS[....]' +to 'PROJ_CS[...]'"
- cs2cs extended to support `cs2cs [-options] <SOURCE> <TARGET>` (without +to), but in that mode no input files may be specified (only from standard input)
- if using proj_create_crs_to_crs() with "EPSG:XXXX" values, it will assume EPSG axis order and unit (so typically lat-long degree for geodetic CRS): this is a breaking change. If using it with "+init=epsg:XXXX" values, it will assume traditional order and unit (long-lat radians for geodetic CRS)
- and all needed changes to be able to remove the data/epsg, IGNF and esri.* legacy files


kbevers (changes requested):
This PR seems important enough that I (and others) should give a proper review. This is just to say that I haven't done so and please don't merge before I (and others) have had a change to look it over. I am travelling the next couple of days so there's a good chance I'll have a few hours of uninterrupted quality review time in an airport some where :-)

src/proj_4D_api.c
```diff
@@ -794,7 +794,12 @@ PJ  *proj_create_crs_to_crs (PJ_CONTEXT *ctx, const char *srid_from, const char
         return NULL;
     }
 
-    P = proj_create(ctx, proj_string);
+    if( proj_string[0] == '\0' ) {
+        /* Null transform ? */
```
kbevers, line 813: Maybe we should add a proper null operation? We can call it `noop`.

kbevers (changes requested):
Overall, I think you have done a good job finding a nice solution to a very difficult problem and I have only minor details to discuss.

I don't have that many specific comments to the code, only a few in `cs2cs.cpp`.

Now that `proj_create_crs_to_crs()` accepts proj-strings, EPSG-like codes and WKT(2) strings the documentation should be adjusted accordingly. I also would suggest renaming the arguments `srid_from`/`srid_to` to `from`/`to` to make it clear that more than just an EPSG-code can be supplied.

Concerning the input to `cs2cs`, I think that using `+to` looks weird when used with EPSG-codes and WKT strings. How about dropping the `+to`, so that calls to `cs2cs` are on the form `cs2cs <FROM> <TO>`? The `+to` will still have to be used when using proj-strings as input, unless they are in quotes: `cs2cs "+proj=latlong +ellps=GRS80" "+proj=utm +zone=32"`. This also opens the door for a removal of the `+to` syntax in the future.

In case there are several transformations available between two crs'es, is it possible with `cs2cs` to select which one to use? Or would that require a combination of `projinfo` and `cct`?

Before this is merged the `cs2cs` docs should be updated.

rouault:
>  I also would suggest renaming the arguments `srid_from`/`srid_to` to `from`/`to`

ok

>  I think that using `+to` looks weird when used with EPSG-codes and WKT strings. How about dropping the `+to`, so that calls to `cs2cs` are on the form `cs2cs <FROM> <TO>`?

Hum, I initially had the same idea, but I discovered while refactoring it that cs2cs could accept input files : ```cs2cs +proj= .. +to +proj= ... first_file second_file ...``` . And it can also support omission of the source CRS before +to or the omission of the target CRS after the +to. So the +to is central to determine which if and which of the CRS have been specified and distinguish them from input files. That said if we have only 2 arguments (ie not prefixed by '-'), and none of them is +to, then we can assume they are `<FROM>` and `<TO>`

> In case there are several transformations available between two crs'es, is it possible with cs2cs to select which one to use? 

No, only the one that is sorted first by create_operations() will be used.

> Or would that require a combination of projinfo and cct?

Yes

> Before this is merged the cs2cs docs should be updated.

If you don't mind, I'll rather create a ticket about that. I'd like to advance on the coding side of GDAL, and then libgeotiff, to check that the PROJ side of things has no holes.

src/proj_4D_api.c
rouault: Traced as https://github.com/OSGeo/proj.4/issues/1185 . Low priority

src/pj_fwd.c
```diff
@@ -138,6 +137,18 @@ static PJ_COORD fwd_finalize (PJ *P, PJ_COORD coo) {
 
     case PJ_IO_UNITS_ANGULAR:
         coo.lpz.z = P->vfr_meter * (coo.lpz.z + P->z0);
+
+        if( P->is_long_wrap_set ) {
+            if( coo.lpz.lam != HUGE_VAL ) {
+                double val = fmod(coo.lpz.lam , M_TWOPI);
```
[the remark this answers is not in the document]
rouault: Similar indeed. adjlon() reframe in [-PI,PI], whereas the long_wrap mechanism reframe in [long_wrap - PI, long_wrap + PI]. I just copied this snippet from pj_transform.c. I've modified this as P->long_wrap_center +  adjlon(coo.lpz.lam - P->long_wrap_center);

rouault:
> That said if we have only 2 arguments (ie not prefixed by '-'), and none of them is +to, then we can assume they are `<FROM>` and `<TO>`

Done

kbevers:
> If you don't mind, I'll rather create a ticket about that. I'd like to advance on the coding side of GDAL, and then libgeotiff, to check that the PROJ side of things has no holes.

Sure, it can wait, as long as you are confident you can remember the detail in a couple of months when you get around to doing it. Also, I am a bit unsure, how does the workings of `cs2cs` affect GDAL and libgeotiff?

rouault:
> as long as you are confident you can remember the detail in a couple of months

The intro text of this PR onto which I've pointed https://github.com/OSGeo/proj.4/issues/1186 captures well the changes

> Also, I am a bit unsure, how does the workings of `cs2cs` affect GDAL and libgeotiff?

Wasn't specific to cs2cs, but more a general remark on being sure PROJ is ready for GDAL and libgeotiff needs. I feel I'm behind the schedule I anticipated and want to be sure that the PROJ 6 release will be OK for their needs. PROJ received all love until now, so time to give some to the other projects :-)

I've just pushed the changes regarding function renaming and moving them at bottom ofin proj.h

rouault:
@kbevers I believe I've addressed your comments (or created tickets for defered actions)

src/proj.h
```diff
@@ -559,27 +568,490 @@ PJ_OBJ_LIST PROJ_DLL *proj_obj_create_from_name(PJ_CONTEXT *ctx,
...
+        * all grids were available. */
+    PROJ_GRID_AVAILABILITY_IGNORED,
+} PROJ_GRID_AVAILABILITY_USE;
+
+void PROJ_DLL proj_operation_factory_context_set_grid_availability_use(
+                                            PJ_OPERATION_FACTORY_CONTEXT *ctxt,
+                                            PROJ_GRID_AVAILABILITY_USE use);
+
+void PROJ_DLL proj_operation_factory_context_set_use_proj_alternative_grid_names(
+    PJ_OPERATION_FACTORY_CONTEXT *ctxt,
```
kbevers: Usually we just use `ctx` for the argument name of contexts. Is the extra "t" in `ctxt` signaling that this context is different from a `PJ_CONTEXT`?

src/proj.h
```diff
@@ -559,27 +568,490 @@ PJ_OBJ_LIST PROJ_DLL *proj_obj_create_from_name(PJ_CONTEXT *ctx,
...
+    PJ_CS_TYPE_TEMPORALCOUNT,
+    PJ_CS_TYPE_TEMPORALMEASURE
+} PJ_COORDINATE_SYSTEM_TYPE;
+
+PJ_COORDINATE_SYSTEM_TYPE PROJ_DLL proj_obj_cs_get_type(const PJ_OBJ* cs);
+
+int PROJ_DLL proj_obj_cs_get_axis_count(const PJ_OBJ *cs);
+
+int PROJ_DLL proj_obj_cs_get_axis_info(const PJ_OBJ *cs, int index,
+                                       const char **pName,
```
kbevers: You work much faster than I can think, so I only just now notice that the argument names also are en camelCase, using some sort of Hungarian notation. A leftover from GDAL I am sure. Similar to the function names this is not consistent with the rest of the API, although not as important as the function names. I feel bad asking this, but is it possible to change this also?

kbevers (commented):
> @kbevers I believe I've addressed your comments (or created tickets for defered actions)

Indeed you have, thanks! In the meantime I have come up with a few more things. Sorry about that, I know you are eager to move on to other things, but it is hard to grasp everything at once in these huge PRs.

src/proj.h
rouault: With the 'p', I wanted to underline the fact that it is a pointer to a const char* name, that is an output argument, to distinguish from input arguments. I feel that if we just use "const char** name" people might be confused and not realize it is an output argument.  I'm open to other suggestions.

rouault: Another option I can see would be to prefix the argument name with out_  : out_name ?

kbevers: I see, that makes sense. At least to me it is more clear what is meant by `out_name` compared to `pName`. I hadn't made that connection until your explanation just now. Let's go with that.

kbevers:
Yet another thing that I've stumbled across. Two actually. 

First, I am transforming coordinates from ETRS89 3D to ETRS89/UTM32+DVR90 (see below). The horizontal part is fine, the vertical is ignored, even though the height *should* change. At the moment PROJ has no way of knowing what to do with the height, so not much else can be expected. I am quite sure though that we should be able to tell that this particular transformation can't in fact be made. I see two options: 1) Refuse to instantiate the transformation or 2) show the user a warning saying that the height is not transformed.

```
$ PROJ_LIB=/usr/local/share/proj ./bin/cs2cs EPSG:4937 EPSG:7416
56 12 0
687071.44	6210141.33 0.00
```

In this next example I am transforming from ETRS89 2D to ETRS89/UTM32+DVR90 which is a 2D to 3D conversion. I am not sure what EPSG says about transformations like that, but this *can* be viewed as two incompatible systems since the source CRS can't include a height, we can't meaningfully assign a height to the destination CRS. So, what do we do about that? Refuse to do the transformation? Implicitly upgrade the source CRS to it's 3D equivalent?

```
$ PROJ_LIB=/usr/local/share/proj ./bin/cs2cs EPSG:4258 EPSG:7416
56 12 0
687071.44	6210141.33 0.00
```

I think this goes deeper than this PR but in any case it is something to consider. So far we have decided to strictly honour axis orders of systems, so being strict about the number of axes is keeping in line with that.

src/proj.h
rouault: Done

rouault:
Regarding EPSG:4937 to EPSG:7416, if you add the dvr90.gtx grid in your data/, that should work since
```
$ src/projinfo -s EPSG:4937 -t EPSG:7416
 +proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +inv +proj=vgridshift +grids=dvr90.gtx +step +proj=utm +zone=32 +ellps=GRS80
```
Note that the above will be returned even if the grid is not installed. By default projinfo return transformations that have known grids, ie ones that are registered in the `alternative_grids`, as having priority since it might be possible for the users to have them. But when used in the cs2cs / proj_create_crs_to_crs() context, we must be more realistic and thus only consider grids that are effectively available when it is executed, so PROJ_GRID_AVAILABILITY_DISCARD_OPERATION_IF_MISSING_GRID is used, which is equivalent to
```
$ src/projinfo -s EPSG:4937 -t EPSG:7416 --grid-check discard_missing
PROJ string: 
+proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +z_in=m +xy_out=rad +z_out=m +step +proj=utm +zone=32 +ellps=GRS80
```

There's no infrastructure to emit warnings, and if we do that on the end result, that would imply to know that a given pipeline doesn't contain a vertical transform step. Not trivial (the EPSG dataset could for example have an explicit 0 vertical offset between 2 vdatums that would be optimized out by the PROJ string optimizer in io.cpp). A user could also add a custom entry in the transformation tables of the database for that tuple of CRS that could be a PROJ pipeline and we would have no clue if it contains a dummy or real vertical transform. But if we synthesize a null transform, we could know that the vertical transform is a dummy one. Actually if you add --summary, you'll see "Inverse of Null geographic offset from ETRS89 to ETRS89 + UTM zone 32N"

Regarding EPSG:4258 to EPSG:7416, yes, I've chosen to be lax on this and upgrade Geographic2D CRS to Geographic3D. It wouldn't be too difficult to add checks to refuse doing this.

kbevers:
Ah yes, much better! I was certain that I had the grids on my system, but apparently not. Very cool that this (more or less) works out of the box!

```
$ PROJ_LIB=/usr/local/share/proj ./bin/cs2cs EPSG:4937 EPSG:7416
56 12 0
687071.44	6210141.33 36.59
```

In case EPSG define a 3D system but PROJ isn't aware of the grid that might be needed, will the vertical component then just be treated as a zero offset?

> if you add --summary, you'll see "Inverse of Null geographic offset from ETRS89 to ETRS89 + UTM zone 32N"

What does the "Null geographic offset" mean in this context? 


> Regarding EPSG:4258 to EPSG:7416, yes, I've chosen to be lax on this and upgrade Geographic2D CRS to Geographic3D. It wouldn't be too difficult to add checks to refuse doing this.

Geodetically speaking, being strict is the correct thing to do but from a practical point of view I am not so sure. Maybe @busstoptaktik has some thoughts on this topic?

kbevers:
Hmm... this doesn't seem right. So I guess it wants to apply the grid offset to the first coordinate since EPSG:5733 is a 1D system. Probably technically correct, but I reckon that most would be surprised by this.

```
$ PROJ_LIB=/usr/local/share/proj ./bin/cs2cs EPSG:4937 EPSG:5733
56 12 0
-0.55	12.00 inf

$ PROJ_LIB=/usr/local/share/proj ./bin/projinfo --identify -s EPSG:4937 -t EPSG:5733
PROJ string:
+proj=pipeline +step +inv +proj=vgridshift +grids=dnn.gtx

WKT2_2015 string:
COORDINATEOPERATION["Inverse of DNN height to ETRS89",
    SOURCECRS[
        GEODCRS["ETRS89",
            DATUM["European Terrestrial Reference System 1989",
                    LENGTHUNIT["metre",1]]],
            PRIMEM["Greenwich",0,
            CS[ellipsoidal,3],
                    ORDER[1],
                    ORDER[2],
                AXIS["ellipsoidal height (h)",up,
                    ORDER[3],
                    LENGTHUNIT["metre",1]]]],
    TARGETCRS[
        VERTCRS["DNN height",
            VDATUM["Dansk Normal Nul"],
            CS[vertical,1],
                AXIS["gravity-related height (H)",up,
                    LENGTHUNIT["metre",1]]]],
    METHOD["Inverse of GravityRelatedHeight to Geographic3D",
        ID["INVERSE(PROJ)","HEIGHT_TO_GEOGRAPHIC3D"]],
    PARAMETERFILE["Geoid (height correction) model file","dnn.gtx"],
    AREA["Denmark - onshore"],
    BBOX[54.51,8,57.8,15.24],
    ID["INVERSE(PROJ)","EPSG_5733_TO_EPSG_4937"]]
```

src/proj.h
kbevers: Beautiful! Thanks.

rouault:
> In case EPSG define a 3D system but PROJ isn't aware of the grid that might be needed, will the vertical component then just be treated as a zero offset?

Yes. For example if you transform from EPSG:5951( "ETRS89 / NTM zone 11 + NN2000 height" which is EPSG:5111 + EPSG:5941) to EPSG:7416, it will only consider the horizontal part, given there is no transform between the vertical parts.

> What does the "Null geographic offset" mean in this context?

This is the term I've coinced to describe a synthetized no-op transformation between 2 geographicCRS

> -s EPSG:4937 -t EPSG:5733

Yes this is non intuitive. If you want a better understanding on how things work under the hood, I'd suggest you have a look at CoordinateOperationFactory::Private::createOperations(), around line 10360 of coordinateoperation.cpp. This is where all the complex stuff happens.
Internally, as createOperations() call itself recursively,  when doing Geog3D to Projected+Vertical, I need to allow Geog3D to VerticalCRS (since EPSG has such transformations) to compose it with the Geog3D to Projected step . But I could potentially disallow this in the most external user-visible call, or maybe prepend initial steps to do lat/long swapping and deg->rad, and append with the rad->deg and axis swapping, so that lat, long, Z is returned

kbevers:
> I need to allow Geog3D to VerticalCRS (since EPSG has such transformations) 

That makes sense. VerticalCRS is a quasi-3D system since the height isn't useful unless it has a position, so this type of transformation is necessary to be allowed. 

> or maybe prepend initial steps to do lat/long swapping and deg->rad, and append with the rad->deg and axis swapping, so that lat, long, Z is returned

I think this is the correct way to go.

busstoptaktik:
Sorry for being late to the party - due to external circumstances, I have not been able to spend much time with PROJ recently.

There is a lot of nice stuff in this PR which nevertheless leaves me **deeply** worried. I have no way of getting through the entire thing, but have been looking into proj.h, which is supposed to define the (C-style) API of the library, which has grown by a huge factor, introducing a large number of functions that, if accepted into the API, will tie our hands and feet for decades to come.

An API should expose a high level abstraction of the library functionality. And from a sufficiently high level of abstraction, the barn raising code provides *no new functionality at all*. It provides a new family of *syntaxes* (WKTn) for accessing existing functionality, but basically any transformation that can be expressed in WKT2 can be expressed as a `proj=pipeline ...`.

And with no new functionality, neither should the API change. At least not by much. I see no need for the plethora of new API functions, when basically feeding a WKTn-string into proj_create could do the work.

I am not against extending the API, but it should be done cautiously, and not based on code that is not yet sufficiently mature and well understood. Adding hundreds of lines of new API entries gives me an impression of exposing library guts and veins, along with the tubing, plumbing, and wiring, needed to make it survive on the outside of the library body.

In short it smells like material that needs to get into experimental production, to gain enough practical experience to actually wrap it into a generically abstracted small (or better: tiny) set of API calls.

Dumping too specialized/low level stuff into the API is the road to turning `proj.h` into the new `projects.h`, which we have now spent years to get rid of.

If really needed, put the PJ_OBJ stuff into some other non-API header file, then gain experience, invent better abstractions, and deploy that.

Apart from PJ_OBJ stuff, a lot of the new stuff in proj.h is about object introspection - a field that has already been sprawling for the last year or two. I think this indicates that we need a new mechanism/architecture for querying about internals. Perhaps a good occasion to consider this as well.

rouault:
> And from a sufficiently high level of abstraction, the barn raising code provides _no new functionality at all_

Yes... but not quite. One of the objective was to "move" code for WKT edition (now actually abstracted to act on ISO19111 objects), from GDAL down to PROJ so that PROJ becomes a true "libcrs" and GDAL no longer needs (ideally) to mess with direct WKT manipulation, so from this point of view, the functionnal scope of PROJ is greatly extended, hence this shows in the API.  Well, this is was what I had in mind. Not sure if it was clearly communicated, but we discussed it with @hobu, at least the transition of PROJ from a mostly-projection-related library to a "libcrs"

> In short it smells like material that needs to get into experimental production, to gain enough practical experience to actually wrap it into a generically abstracted small (or better: tiny) set of API calls.

OK, would moving all the stuff that in the `Binding in C of advanced methods from the C++ API` section starting at line 905 of proj.h into a proj_advanced.h header would satisfy you ? Those functions are the one for advanced creation of PJ_OBJ, and are probably needed only by GDAL. The basic ones doing conversion from/into WKT and PROJ strings, computation of candidate transformations between 2 CRS, and inspection of the results, would remain in proj.h. Those are likely to be useful for projects like GRASS or QGIS.

> If really needed

All the recent additions to proj.h are driven because GDAL needs them. (that's the "and whatever else changes are needed to address GDAL and libgeotiff needs" at the end of RFC2)

~
Folks, I knew this would be a huge effort, but didn't really measure how much. I'm doing >~ 12h work days since a few weeks on this. Have only started on the GDAL integration, and each change in GDAL regarding SRS tends to subtlety break and require updating tens or more of unit tests. I'd really hate to see all this trashed.

busstoptaktik:
@rouault, I’m not trying to block your work - I’m impressed at what you achieve, and I obviously agree in the overarching goal of turning PROJ into a generic libcrs. That’s what I’ve been working on myself for a few years now.

Being generally in favour of evolutionary design, rather than revolutionary, I have tried to clean up stuff that was blocking evolution towards a fully functional libcrs.

But the evolutionary road leads through many roundabouts, and I think we have a lot of code that should be cleaned out now. Not the least because the introduction of C++11 as “least acceptable” common code platform, has made it possible to rationalize a lot of text handling (RAII and C++strings are way better tools than hand crafted memory handling and C style strxxx() operations).

So on the PROJ plumbing side a lot of stuff can be re-rationalized. I’m working on this, but developing the ideas in a separate project, since they are quite extensive, and hard to develop in parallel to your massive effort. But due to the remaining cache of accumulated PROJ cruft, I’m a bit nervous about extending the API, risking blocking necessary (or at least highly desirable) code cleanups in the intermediate term.

So not being able to understand the need for PJ_OBJ for handling 19111-objects (I believe PJ is perfectly ready for that), I am not happy to see PJ_OBJ creeping into API space, and I really have a hard time understanding why the PJ-like PJ_OBJs cannot be created by a proj_create-like API entry point (probably even proj_create herself).

You may convince me later, but for now it looks like an unnecessary complication. The disclaimer here being that while I have spent quite a lot of time with 19111, you have spent quite a lot more time implementing it, and probably have your reasons for selecting the design you selected.

But if the design mostly reflects the existing state of things in GDAL I am tempted to suggest that perhaps the architecture and abstractions could probably be both simplified and improved over the coming years: Since GDAL and PROJ have developed alongside, there is probably also common historical decisions having (in hindsight) led to less than optimal designs and implementations.

But you cannot argue with actual, productive code, and I acknowledge the enormous effort you are putting into this, and I am eager to see it completed.

But I have also spent quite a few years fighting code bloat, followed by code rot, in legacy geodetic systems, so I am reluctant to just accept that “new functionality requires more code and new abstractions”. This may be OK for the first iteration, but once hacked out in code, we may be able to see *better* abstractions leading to *less* code (or at least much better functionality for approximately the same amount of code).

My vision for a libcrs is different from yours, and probably would not plug as seamlessly as yours into the GDAL crs-surface. But my vision doesn’t really matter, since **you’re doing the work and providing the results**. I would just be very unhappy to see a too large API surface block us in making rationalizations over the coming years: Generalizing/streamlining  the abstractions and simplifying and reducing the code base (at least attempting at it...) are still very important activities to keep the library maintainable and evolvable.

So to conclude: I **do not** want to block your work - I think it is both herculean, fantastic, and highly useful. But please limit API growth: If, as you state, much of the material is probably useful for GDAL only, it could be an indication that it is probably too specific in its design. So I’m very happy with having that part placed in a “for GDAL’s eyes only” pseudo-API.

rouault:
> I believe PJ is perfectly ready for that

I'm less optimistic. A high level vision of PJ is it is just a mathematical formula to express a coordinate transformation. It doesn't capture the nature of objects (hard to tell if +proj=utm is a projected CRS using a UTM conversion or the UTM conversion itself), neither all metadata like names, codes, axis abbreviations, area of uses, etc that are needed to generate WKT or do queries in the database.

I can imagine there could be a way to have PJ being an alias of a PJ_OBJ of CoordinateTransformation type, but reconciling the existing C code and the new C++ ISO19111-based would be another significant task on itself (and hacking into the pure C code is really intimidating, as you probably better know than me when you've added the pipeline capabilities). We cannot rewrite all PROJ in one day (one day meaning six month:-) ) So yes the current state of PROJ master is a hybrid beast between the "pure" mathematical historical C code and all the new C++ metadata stuff. One approach could be that the Conversion class could do the math itself (or actually be sub-classed in as many classes as they are conversion methods), instead of generating a PROJ string that is instanciated as a PJ object later.

> I have spent quite a lot of time with 19111, you have spent quite a lot more time implementing it, and probably have your reasons for selecting the design you selected.

To be able to convert into/from WKT, and support several WKT variants, you need an object modelling that is close to 19111. Well, that seemed to be the best solution when I envisioned things initially, and code shows that it can work in practice. Not sure about alternative designs.
Actually if we look at the GDAL side that handled WKT itself, the OGRSpatialReference object used to be a just a tree of WKT nodes, where you would directly add / modify / delete nodes. There are two downsides to this approach: it is very easy to generate invalid WKT if you forget to add a node (and my current re-plug of GDAL to PROJ shows that such situations happened, at least in unit tests), and it is strongly tied to a particular WKT version since it doesn't abstract the representation at all

>  I would just be very unhappy to see a too large API surface block us in making rationalizations over the coming years

yes, I know. I can just reassure you by telling I'm not payed by the number of new functions / LOC I add (on the contrary it costs me time to write, document and write tests for them). The C API just binds on the corresponding C++ objects which are strongly based on the ISO19111 models, so it is not a "random" API.

So if we take
```
PJ_OBJ PROJ_DLL *proj_obj_create_projected_crs(PJ_CONTEXT *ctx,
                                               const char* crs_name,
                                               const PJ_OBJ* geodetic_crs,
                                               const PJ_OBJ* conversion,
                                               const PJ_OBJ* coordinate_system);
```
it truly reflects that a ProjectedCRS is a composition of a name, a base GeodeticCRS, a Conversion and a CoordinateSystem.

Similarly, @kbevers wasn't really enthusiastic about all the proj_obj_create_conversion_xxxx() functions, but I believe it is better to have such explicit instanciations where the prototype clearly indicates the needed arguments (projection parameters), rather than just relying on the user to use the proper PROJ keyword when compositing a PROJ string (I regularly wonder if I must use lat_ts, lat_0 or lat_1 ... , for a particular method, and the setup functions aren't particularly strict at runtime on checking extra and missing parameters) (or similarly compositing a correct WKT string at hand is error prone)

The potential risk I see is if someday we say: "ISO19111 modelling is not appropriate, let's use something else". Then yes, you'll have mostly to ditch all this. I've preferred using a standard based abstraction rather than a crappy one of my own invention.

> So I’m very happy with having that part placed in a “for GDAL’s eyes only” pseudo-API.

OK, I'll move that to a dedicated header, with a warning that it is considered as experimental material, driven by GDAL needs and may evolve, etc...

busstoptaktik:
> A high level vision of PJ is it is just a mathematical formula to express a coordinate transformation. It doesn't capture the nature of objects 

Actually, to me a PJ is just a bag of bits with a few embedded function pointers defining how to interpret the bits (actually a kind of limited RTTI implemented in C). At this level of (lack of) abstraction, it can represent anything.

The "one API entry per operation type" helps enforcing correct initialization, even though a sufficiently clever user can ram garbage into any parameter anyway if not understanding it properly.

But at least the individual API entries provide a reference frame for providing the relevant documentation, so as I said: You may very well convince me that this is the better choice.

On the other hand it was also you, a few years ago, who convinced me that the first versions of proj.h/proj_4D_api.c included too much material into the "proj_..." API namespace, so I learned this API surface skepticism from a highly experienced and competent schoolmaster :-)

> So yes the current state of PROJ master is a hybrid beast between the "pure" mathematical historical C code and all the new C++ metadata stuff

Which is of course A-OK in an evolutionary world. And having taken the step and accepting C++11 code into the PROJ code base really opens up for some major code reductions and improvements in the historical department of the code base.

> One approach could be that the Conversion class could do the math itself (or actually be sub-classed in as many classes as they are conversion methods), instead of generating a PROJ string that is instanciated as a PJ object later.

Yes, that would probably be a superior design, but would require a re-implementation of most of the library.

Entirely dropping the PROJ syntax on the other hand, would probably not be good either: Despite all its quirkiness, to me it really relates to WKT as JSON or YAML relates to XML: The less strict, but typing friendly, and human-readable version of the formal uncle. Instantiating in any syntax, and letting the object serialize itself in whatever syntax preferred is a nice feature.

rouault:
> I learned this API surface skepticism from a highly experienced and competent schoolmaster :-)

"Do what I say, not what I do" :-) That said, the API, even if large, is still an abstraction. For example, it does not expose that the database implementation is SQLite (and even in the C++ code, the interaction with SQLite has been carefully isolated in the DatabaseContext class). The C++ API itself tries to limit what is visible as public exported symbols.

> Entirely dropping the PROJ syntax on the other hand, would probably not be good either

Wasn't suggesting that either. There are a number of data formats using PROJ string to express the CRS of the data. It has the benefit of being compact (with the associated drawback of lacking some usefu metadata). My view is that PROJ strings are good to express coordinate transformation, especially since the pipeline introduction, but inappropriate to express a CRS object.

rouault:
@kbevers @busstoptaktik How do we go forward with this PR ? (my recent commits aren't related. fixes there and there)

kbevers (approved):

PR #1204, opened 2018-12-20
Add markdown .md support for Sphinx to osgeo/proj-docs docker container

hobu:


hobu:
I have been using `osgeo/proj-docs` for https://entwine.io and https://pdal.io in an attempt to not have an out-of-date docker container corresponding to every website. We added a number of markdown documents to the Entwine website for this release, and the ability to render it into (especially wiki page stuff), could be useful.

kbevers:
Okay, that's the reason. Perhaps the docker container should be renamed saying then? There's a risk that someone removes the markdown table support because it seems like it is not used in the PROJ docs.

hobu:
> There's a risk that someone removes the markdown table support because it seems like it is not used in the PROJ docs.

Sure, but only you, me, and @rouault have the ability to push that container to dockerhub at the moment.

kbevers:
Well, I may be allowed to push to dockerhub but I certainly do not have the ability to do so :-) Maybe just add a comment somewhere saying that this container is used elsewhere too so that it isn't messed up by mistake in a spring cleaning frenzy two years from now.

hobu:
> this container is used elsewhere

It used to be the other way, btw. PROJ was using **PDAL's** doc container and people were suspicious of that 😄. Just trying to make things fail in only one location...

kbevers:
> Just trying to make things fail in only one location...

Or many, depending on how you look at it :-)

Issue #1200, opened 2018-12-17
proj_context_guess_dialect


snowman2:
Is there something analogous to `proj_context_guess_wkt_dialect` for determining the `PJ_PROJ_STRING_TYPE`? 

If not, this would be nice to have added to `proj.h`. One potential name could be `proj_context_guess_dialect`.

rouault:
No, there isn't such capability.

rouault:
I don't have such need, so unlikely to come from me. Why would that be needed ?
Anyway, from my experience, I would discourage use of PROJ strings at all to express CRS objects for example because they aren't rich enough for that, and suffer from ambiguities (axis order, units, etc,) I tried to remedy a bit with the "PROJ 5" variations, but that's far from being perfect. PROJ strings however are much better used to express coordinate transformation / pipelines.

snowman2:
The idea would be to modify a parameter in the string based on the version. For example, in `pyproj` it has a `preserve_units` option where it makes sure the `+units` parameter is `+units=m` (I don't know the history or reasoning behind it, it would be for backwards compatibility). However, that syntax is non-existent in the version 5 of the proj string and a different set of parameters would be needed to be modified.

>> I would discourage use of PROJ strings at all to express CRS objects

That is good to know. However, for compatibility with libraries that do so such as (rasterio, fiona, geopandas, etc...) it is a useful feature to have. This would be useful to have documented somewhere.

kbevers:
> However, that syntax is non-existent in the version 5 of the proj string

What syntax is that exactly?

> However, for compatibility with libraries that do so such as (rasterio, fiona, geopandas, etc...) it is a useful feature to have

That's why it's still there ;) It is likely though that a future version of PROJ removes support for parameters such as `+datum` and `+towgs84`.

snowman2:
The original pyproj version adjusts it [here](https://github.com/jswhit/pyproj/blob/78540f5ff40da92160f80860416c91ee74b7643c/lib/pyproj/__init__.py#L352).

The workaround I have is [here](https://github.com/snowman2/pyproj/blob/3663f6dcc83746ccd948a215799240aa7b16648e/pyproj/__init__.py#L401).

kbevers:
You didn't answer my question? What syntax is it that you think is unavailable in PROJ 5 compared to PROJ.4? `+units`? That's supported in PROJ 5.  `preserve_units` is not a PROJ construct.

snowman2:
[...]

PR #1205, opened 2018-12-20
Better handling of pipelines including step with PJ_IO_UNITS_WHATEVER operations

kbevers:
Solves issues mentioned in #1200. See commit texts for details.

### DIFF
--- a/src/conversions/unitconvert.cpp
+++ b/src/conversions/unitconvert.cpp
@@ -457,6 +457,8 @@ PJ *CONVERSION(unitconvert,0) {
 
     P->left  = PJ_IO_UNITS_WHATEVER;
     P->right = PJ_IO_UNITS_WHATEVER;
+    P->skip_fwd_prepare = 1;
+    P->skip_inv_prepare = 1;
 
     /* if no time input/output unit is specified we can skip them */
     Q->t_in_id = -1;
@@ -476,6 +478,8 @@ PJ *CONVERSION(unitconvert,0) {
         }
         if (f != 0.0)
             Q->xy_factor *= f;
+        if (normalized_name != nullptr && strcmp(normalized_name, "Radian") == 0)
+            P->left = PJ_IO_UNITS_RADIANS;
     }
 
     if ((name = pj_param (P->ctx, P->params, "sxy_out").s) != nullptr) {
@@ -489,6 +493,8 @@ PJ *CONVERSION(unitconvert,0) {
         }
         if (f != 0.0)
             Q->xy_factor /= f;
+        if (normalized_name != nullptr && strcmp(normalized_name, "Radian") == 0)
+            P->right= PJ_IO_UNITS_RADIANS;
     }
 
     if( xy_in_is_linear >= 0 && xy_out_is_linear >= 0 &&
--- a/src/pipeline.cpp
+++ b/src/pipeline.cpp
@@ -329,6 +329,33 @@ static void set_ellipsoid(PJ *P) {
 }
 
 
+static enum pj_io_units get_next_non_whatever_unit(void *pipeline_data, int step, PJ_DIRECTION dir) {
+    PJ **pipeline = static_cast<struct pj_opaque*>(pipeline_data)->pipeline;
+    int nsteps = static_cast<struct pj_opaque*>(pipeline_data)->steps;
+    int i;
+
+    if (dir == PJ_FWD) {
+        for (i = step+1; i<=nsteps; i++) {
+            if (pj_left(pipeline[i]) != pj_right(pipeline[i]))
+                return pj_left(pipeline[i]);
+            if (pj_left(pipeline[i]) != PJ_IO_UNITS_WHATEVER)
+                return pj_left(pipeline[i]);
+            if (pj_right(pipeline[i]) != PJ_IO_UNITS_WHATEVER)
+                return pj_right(pipeline[i]);
+        }
+    } else {
+        for (i=step; i>1; i--) {
+            if (pj_right(pipeline[i]) != pj_left(pipeline[i]))
+                return pj_right(pipeline[i]);
+            if (pj_right(pipeline[i]) != PJ_IO_UNITS_WHATEVER)
+                return pj_right(pipeline[i]);
+            if (pj_left(pipeline[i]) != PJ_IO_UNITS_WHATEVER)
+                return pj_left(pipeline[i]);
+        }
+    }
+    return PJ_IO_UNITS_WHATEVER;
+}
+
 
 
 PJ *OPERATION(pipeline,0) {
@@ -480,14 +507,36 @@ PJ *OPERATION(pipeline,0) {
         }
     }
 
-    /* Check that output units from step i are compatible with expected units in step i+1 */
-    for (i = 1; i < nsteps; i++) {
-        enum pj_io_units unit_returned = pj_right (static_cast<struct pj_opaque*>(P->opaque)->pipeline[i]);
-        enum pj_io_units unit_expected = pj_left  (static_cast<struct pj_opaque*>(P->opaque)->pipeline[i+1]);
 
-        if ( unit_returned == PJ_IO_UNITS_WHATEVER || unit_expected == PJ_IO_UNITS_WHATEVER )
+    /* Replace PJ_IO_UNITS_WHATEVER with input/output units of neighbouring steps where */
+    /* it make sense. It does in most cases but not always, for instance                */
+    /*      proj=pipeline step proj=unitconvert xy_in=deg xy_out=rad step ...           */
+    /* where the left-hand side units of the first step shouldn't be changed to RADIANS */
+    /* as it will result in deg->rad conversions in cs2cs and other applications.       */
+    PJ **pipeline = static_cast<struct pj_opaque*>(P->opaque)->pipeline;
+    for (i=1; i<=nsteps; i++) {
+        if (pj_left(pipeline[i]) == PJ_IO_UNITS_WHATEVER && pj_right(pipeline[i]) == PJ_IO_UNITS_WHATEVER) {
+            pipeline[i]->left = get_next_non_whatever_unit(P->opaque, i, PJ_FWD);
+            pipeline[i]->right = get_next_non_whatever_unit(P->opaque, i, PJ_FWD);
+        }
+    }
+
+    for (i=nsteps; i>0; i--) {
+        if (pj_left(pipeline[i]) == PJ_IO_UNITS_WHATEVER && pj_right(pipeline[i]) == PJ_IO_UNITS_WHATEVER) {
+            pipeline[i]->right = get_next_non_whatever_unit(P->opaque, i, PJ_INV);
+            pipeline[i]->left = get_next_non_whatever_unit(P->opaque, i, PJ_INV);
+        }
+    }
+
+    /* Check that units between each steps match each other, fail if they don't */
+    for (i = 1; i < nsteps; i++) {
+        enum pj_io_units curr_step_output = pj_right (pipeline[i]);
+        enum pj_io_units next_step_input  = pj_left  (pipeline[i+1]);
+
+        if ( curr_step_output == PJ_IO_UNITS_WHATEVER || next_step_input == PJ_IO_UNITS_WHATEVER )
             continue;
-        if ( unit_returned != unit_expected ) {
+
+        if ( curr_step_output != next_step_input ) {
             proj_log_error (P, "Pipeline: Mismatched units between step %d and %d", i, i+1);
             return destructor (P, PJD_ERR_MALFORMED_PIPELINE);
         }
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -74,6 +74,13 @@ target_link_libraries(proj_errno_string_test
   ${PROJ_LIBRARIES})
 add_test(NAME proj_errno_string_test COMMAND proj_errno_string_test)
 
+add_executable(proj_angular_io_test
+    main.cpp
+    proj_angular_io_test.cpp)
+target_link_libraries(proj_angular_io_test
+    gtest
+    ${PROJ_LIBRARIES})
+add_test(NAME proj_angular_io_test COMMAND proj_angular_io_test)
 
 if (MSVC AND BUILD_LIBPROJ_SHARED)
 # ph_phi2_test not compatible of a .dll build
--- a/test/unit/proj_angular_io_test.cpp
+++ b/test/unit/proj_angular_io_test.cpp
@@ -1,0 +1,101 @@
+/******************************************************************************
+ *
+ * Project:  PROJ
+ * Purpose:  Test that in- and output units behave correctly, especially
+ *           angular units that need special treatment.
+ * Author:   Kristian Evers <kristianevers@gmail.com>
+ *
+ ******************************************************************************
+ * Copyright (c) 2019, Kristian Evers
+ *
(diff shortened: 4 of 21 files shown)


kbevers:
So this fail tests with `cs2cs` that output degrees. I believe the problem is caused by `cs2cs` wanting to convert radians to degrees when receiving output units of `PJ_IO_UNITS_ANGULAR` but since that can both mean radians and degrees it is a bit difficult to determine what to do. `cs2cs` seem to be working around that somehow by using `destToRadians`.

For instance:

```
$ echo 79 45 0.0 | ./bin/cs2cs -v +proj=latlong +ellps=bessel +towgs84=5,0,0 +to +proj=latlong +ellps=bessel +towgs84=1,0,0
# ---- From Coordinate System ----
+proj=latlong +ellps=bessel +towgs84=5,0,0
# ---- To Coordinate System ----
+proj=latlong +ellps=bessel +towgs84=1,0,0
*	* inf

$ ./bin/projinfo -s "+proj=latlong +ellps=bessel +towgs84=5,0,0" -t "+proj=latlong +ellps=bessel +towgs84=1,0,0"
unknown id, Transformation from unknown to WGS84 + Inverse of Transformation from unknown to WGS84, unknown accuracy, unknown domain of validity

PROJ string:
+proj=pipeline +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=cart +ellps=bessel +step +proj=helmert +x=4 +y=0 +z=0 +step +inv +proj=cart +ellps=bessel +step +proj=unitconvert +xy_in=rad +xy_out=deg

$ echo 70 45 0 0 |./bin/cct +proj=pipeline +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=cart +ellps=bessel +step +proj=helmert +x=4 +y=0 +z=0 +step +inv +proj=cart +ellps=bessel +step +proj=unitconvert +xy_in=rad +xy_out=deg
 69.9999560942   44.9999715931        3.9987        0.0000
```

I think this is a fundamental problem that needs to be solved. Depending on how an operation/pipeline is initialized it may need conversion from radians to degrees or not. Traditionally it has always been the job of the user of the API to convert radians to degrees but that is not always the case anymore. It should be the same in all cases. At this point it is probably smartest to not have the users do the radian->degree and degree->radian conversions. At least that is what the new C++ code does as far as I can tell (by more or less always including `unitconvert` steps in either end of pipelines). Thoughts?

src/PJ_pipeline.c
```diff
@@ -477,14 +477,35 @@ PJ *OPERATION(pipeline,0) {
...
     for (i = 1; i < nsteps; i++) {
-        enum pj_io_units unit_returned = pj_right (P->opaque->pipeline[i]);
-        enum pj_io_units unit_expected = pj_left  (P->opaque->pipeline[i+1]);
+        enum pj_io_units previous_step_input  = i>1 ? pj_left  (P->opaque->pipeline[i-1]) : PJ_IO_UNITS_WHATEVER;
+        enum pj_io_units previous_step_output = i>1 ? pj_right (P->opaque->pipeline[i-1]) : PJ_IO_UNITS_WHATEVER;
+
+        enum pj_io_units current_step_input  = pj_left  (P->opaque->pipeline[i]);
+        enum pj_io_units current_step_output = pj_right (P->opaque->pipeline[i]);
+
+        enum pj_io_units next_step_input  = pj_left  (P->opaque->pipeline[i+1]);
```
rouault: should i+1 < nsteps being tested ?

src/PJ_pipeline.c
```diff
@@ -477,14 +477,35 @@ PJ *OPERATION(pipeline,0) {
...
+        /* we override the units with the values of the neighbouring steps      */
+        if (current_step_input == PJ_IO_UNITS_WHATEVER && current_step_output == PJ_IO_UNITS_WHATEVER) {
+            if (i==1) {
+                P->opaque->pipeline[i]->left  = next_step_input;
+                P->opaque->pipeline[i]->right = next_step_input;
+            } else if (i==nsteps-1) {
+                P->opaque->pipeline[i]->left  = previous_step_output;
+                P->opaque->pipeline[i]->right = previous_step_output;
+            } else {
+                P->opaque->pipeline[i]->left  = previous_step_input;
```
rouault: shouldn't that be previous_step_output ?

src/PJ_pipeline.c
kbevers: Are you referring to the if/else statements below?

rouault: no, it seems to me that P->opaque->pipeline[i+1] could be an out-of-bound access when i == steps - 1

src/PJ_pipeline.c
kbevers: Yes. I'll fix it right away.

kbevers:
I've updated the code now. It doesn't fix the original problem mentioned in https://github.com/OSGeo/proj.4/issues/1200#issuecomment-448853865 but it ensures that `proj_angular_input` and `proj_angular_output` can be trusted since input/output units across the pipeline now are consistent.

It is still failing `testvarious`, specifically tests that apply `+lon_wrap=180`. I am not sure how to best deal with that. It shows up in tests like:

```
$ echo 181d00\'00.000\"E 45d00\'00.000\"N |./bin/cs2cs +proj=latlong +datum=WGS84 +to +proj=latlong +datum=WGS84 +lon_wrap=180 -E
181d00'00.000"E 45d00'00.000"N	179dW	45dN 0.000
```

which internally is created by this pipeline:

```
$ ./bin/projinfo -s "+proj=latlong +datum=WGS84" -t "+proj=latlong +datum=WGS84 +lon_wrap=180"
unknown id, Transformation from unknown to unknown, unknown accuracy, unknown domain of validity

PROJ string:
+proj=pipeline +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +inv +proj=longlat +datum=WGS84 +step +proj=latlong +datum=WGS84 +lon_wrap=180 +step +proj=unitconvert +xy_in=rad +xy_out=deg
```

It can be seen here that while the `latlong` step applies `lon_wrap` it isn't respected by `unitconvert`. Maybe the best fix is to add an `+lon_wrap` option for `unitconvert` as well?

Regarding the original issue of `proj` not accepting a complicated pipeline, well, I think the correct thing to do is to disallow pipelines specified on the command line as input to `proj`. We can still allow stuff like `proj +init=epsg:32667` as long as the units work out.

rouault:
> Maybe the best fix is to add an `+lon_wrap` option for `unitconvert` as well?

Huh, that doesn't look right ! Why should be unitconvert be aware of lon_wrap, and what would it do with it ? There must be an issue somewhere else... In fact lon_wrap is originally a pj_transform() hack, not directly handled by +proj=lonlat itself. I believe I added some support for lon_wrap in the preparation or postprocessing steps of the pj_fwd / pj_inv code for a use case (presumably when converting cs2cs to use proj_create_crs_to_crs() rather than pj_transform(), but it might be missing in other places. Yep, support for long_wrap in fwd_finalize() was added per https://github.com/OSGeo/proj.4/commit/a66c12277666489cac74535bad8d2cf565ad542d#diff-71791a0be2e11da6fe12ceb08c34e7e1R140 . Or perhaps the issue is that OUTPUT_UNITS != PJ_IO_UNITS_ANGULAR with your changes

kbevers:
I don't know, but this at least suggest that `unitconvert` is doing *something*:
```
$ echo 181 45 0 0 |./bin/cct +proj=pipeline +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +inv +proj=longlat +datum=WGS84 +step +proj=latlong +datum=WGS84 +lon_wrap=180
181.0000000000   45.0000000000        0.0000        0.0000

$ echo 181 45 0 0 |./bin/cct +proj=pipeline +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +inv +proj=longlat +datum=WGS84 +step +proj=latlong +datum=WGS84 +lon_wrap=180 +step +proj=unitconvert +xy_in=rad +xy_out=deg
    -179.0000        45.0000        0.0000        0.0000
```

But really I guess it is somewhere else but related to that last step. It could very well be in  `fwd_finalize()`.

kbevers:
> Or perhaps the issue is that OUTPUT_UNITS != PJ_IO_UNITS_ANGULAR with your changes

I don't think so, because before the output units would have been `PJ_IO_UNITS_WHATEVER`, just as they are now. The input units of `unitconvert` have changed though (because `xy_in=rad` here), so maybe that is the issue.

kbevers:
I have added

```
    P->skip_fwd_prepare = 1;
    P->skip_inv_prepare = 1;
```

to `unitconvert.cpp` which seems to solve the problem. It was indeed the prepare function that was manipulating the input to `unitconvert`. All tests pass locally, hopefully also on the CI targets!

test/unit/CMakeLists.txt
```diff
@@ -74,6 +74,13 @@ target_link_libraries(proj_errno_string_test
   ${PROJ_LIBRARIES})
 add_test(NAME proj_errno_string_test COMMAND proj_errno_string_test)
 
+add_executable(proj_angular_io_test
```
rouault, line 77: should be added to Makefile.am as well

test/unit/proj_angular_io_test.cpp
```diff
@@ -0,0 +1,101 @@
+/******************************************************************************
+ *
+ * Project:  PROJ
+ * Purpose:  Test that in- and output units behave correctly, especially
+ *           angular units that need special treatment.
+ * Author:   Kristian Evers <kristianevers@gmail.com>
+ *
+ ******************************************************************************
+ * Copyright (c) 2019, Kristian Eversi
```
rouault: typo or name change :-) ?

src/pipeline.cpp
```diff
@@ -480,14 +507,36 @@ PJ *OPERATION(pipeline,0) {
         }
     }
 
-    /* Check that output units from step i are compatible with expected units in step i+1 */
+
+    /* Replace PJ_IO_UNITS_WHATEVER with input/output units of neighbouring steps where */
+    /* it make sense. It does in most cases but not always, for instance                */
+    /*      proj=pipeline step proj=unitconvert xy_in=deg xy_out=rad step ...           */
+    /* where the left-hand side units of the first step shouldn't be chanced to RADIANS */
```
rouault: s/chanced/changed/

src/conversions/unitconvert.cpp
```diff
@@ -476,6 +478,8 @@ PJ *CONVERSION(unitconvert,0) {
         }
         if (f != 0.0)
             Q->xy_factor *= f;
+        if (strcmp(normalized_name, "Radian") == 0)
```
rouault: in case a numeric value is passed, normalized_name will be nullptr, so this should be tested
Same below.
Strange that it doesn't crash on tests

rouault: > Strange that it doesn't crash on tests

Actually, it does on Linux & Mac :-) Windows must have some nullptr protection in its strcmp() implementation

kbevers: Yes, of course... Is there a better way to do this using C++ functions?

rouault: An alternative might be to initialize normalized_name = "" and make get_unit_conversion_factor() return "" instead of nullptr

As far as C++ is concerned, you can compare std::string with the natural == operator. But there are caveats: constructing a std::string from a const char* requires for the later to be non-null. And in theory std::string construction from const char* can fail with a out-of-memory C++ exception that should be caught somewhere. c_api.cpp normally makes sure to wrap in try / catch all calls to C++ code that can throw exceptions, but using C++ from "traditional" PROJ code that can be directly called from C exported symbols has the risk of having uncaugh exceptions (I'd note that this is the theory. In practice GDAL for example makes great use of std::string and doesn't bother catching out-of-memory exceptions when the strings are known to be of small size. so there's a theoretical risk of crash in case of low-mem situation).
Furthermore starting to use C++ & exceptions in code that makes manual memory allocations (to be opposed to C++ objects like std::vector that manage their memory) expose to memory leaks if you fail to free objects when exceptions are thrown.

kbevers: I've just added a check for `nullptr`, that is probably the simplest way to go. 

Okay, many good reasons for not mixing in C++ here. Thanks for the explanation!

kbevers:
So this is troubling me now:

https://github.com/OSGeo/proj.4/blob/d89a6ab2f3f26d1451971369b0ccfc4eb322729d/test/gie/unitconvert.gie#L28-L31

I didn't run into that before as somehow the `unitconvert.gie` check is not added to CMake. I'll do that shortly.

The problem with the test is that it now converts the returned `0.017..` radians to degrees before making the check. Because the output unit now tells `gie` (and the other cli apps) to do so. So the question is then, are there any scenarios where we would want radians as the final output? Are there any known coordinate systems that work in radians instead of degrees? If not, I think it is okay to change or remove this test.

rouault:
> Are there any known coordinate systems that work in radians instead of degrees?

Not to my knowledge. At least not in EPSG.  That said, in my recent work, the "+init=epsg:XXXX" syntax in "PROJ4 emulation mode" expects geographic coordinates to be in radians, as an attempt of having proj_create_crs_to_crs() create PJ objects that work mostly as pj_transform() did, but that might be something to reconsider.

```
$ PROJ_USE_PROJ4_INIT_RULES=YES src/projinfo "+init=epsg:4326"
PROJ string:
+proj=pipeline +step +proj=longlat +ellps=WGS84 +step +proj=unitconvert +xy_in=rad +xy_out=rad

WKT2_2015 string:
GEODCRS["WGS 84",
    DATUM["World Geodetic System 1984",
            LENGTHUNIT["metre",1]],
        ID["EPSG",6326]],
    PRIMEM["Greenwich",0,
        ID["EPSG",8901]],
    CS[ellipsoidal,2],
        AXIS["longitude",east,
            ORDER[1],
            ANGLEUNIT["radian",1,
                ID["EPSG",9101]]],
        AXIS["latitude",north,
            ORDER[2],
            ANGLEUNIT["radian",1,
                ID["EPSG",9101]]],
    AREA["World"],
    BBOX[-90,-180,90,180]]
```

> If not, I think it is okay to change or remove this test.

It would still be good to have a unit test testing that aspect of unitconvert, since the deg->rad step is a common step in generated pipelines. Possibly as a non-gie test if that makes it easier. Or have a no_unit_convert option to gie ?

kbevers:
> It would still be good to have a unit test testing that aspect of unitconvert, since the deg->rad step is a common step in generated pipelines. Possibly as a non-gie test if that makes it easier. Or have a no_unit_convert option to gie ?

It is easily tested by simply expecting the same output as input:

```
operation   proj=unitconvert xy_in=deg xy_out=rad
tolerance   0.000000000001
accept      1       1       1   1
expect      1       1       1   1	# gie does a rad->deg conversion behind the scenes
```

A more generic solution could be to add the option to override i/o units to pipeline steps, for instance:

```
+proj=pipeline +step proj=unitconvert xy_in=deg xy_out=rad +override_unit=whatever
```

or something to that effect.

rouault:
Looks reasonable to me

snowman2:
>  in my recent work, the "+init=epsg:XXXX" syntax in "PROJ4 emulation mode" expects geographic coordinates to be in radians, as an attempt of having proj_create_crs_to_crs() create PJ objects that work mostly as pj_transform() did, but that might be something to reconsider.

I think it would be a good idea to reconsider this behavior. I think with the major version change, breaking changes will be expected and could be better in the long run. But, I have limited experience in the matter and can only offer the perspective of a new consumer.

PR #1208, opened 2018-12-26
Unify PJ_OBJ and PJ structures

rouault:
Implements what was discussed in https://lists.osgeo.org/pipermail/proj/2018-December/thread.html#8109
And also:
* cleanup legacy structures of projects.h
* merge projects.h into proj_internal.h


kbevers:
I think this generally looks very good. Is there still a purpose for the `proj_obj_*` namespace or are those leftovers?

kbevers:
> Is there still a purpose for the proj_obj_* namespace or are those leftovers?

Same question for datatypes: `PJ_OBJ_*`.

rouault:
> > Is there still a purpose for the proj_obj_* namespace or are those leftovers?
> 
> Same question for datatypes: `PJ_OBJ_*`.

Done, except for PJ_OBJ_LIST since there is a conflict with an existing struct PJ_LIST that is used for listing operations

kbevers (approved):
I am happy with this PR as it is now. There may be some minor details that I'll come across later but that's is not reason enough for holding this back. I am very happy that everything can be contained within the `PJ` struct.

rouault:
@kbevers Should I merge this ?

kbevers:
> @kbevers Should I merge this ?

Yes, if you are happy with it. I just didn't want to pull the trigger if you weren't ready yet :)

snowman2:
One thing that was nice about the separate namespace for the proj_obj versus just proj was that it differentiated the functions what went together.

For example. the PJ object created with proj_create does not play well with the other functions that used to have the proj_obj namespace. However, if the user uses proj_create_from_user_input (or a similar function), then the PJ object plays nicely with the other functions.

Thoughts on this?

rouault:
@snowman2 Yes, I agree with that. We could perhaps improve that in the future, but it may be a bit risky for now to plug proj_create() and proj_create_argv() on proj_create_from_user_input () since the ISO191111 ingestion of PROJ string might be a bit lossy in its attempt to build a ISO19111 object if using some not so frequent PROJ keywords.

PR #1210, opened 2018-12-27
Remove proj_def.dat (#201)

rouault:
This also fixes a regression recently added in PROJ master now that cs2cs is using underneath the new ISO19111 code to generate pipeline, which doesn't emit +no_defs, and thus cause issues with CRS using LCC_1SP projections that erroneously use the lat_2=45 default of proj_def.dat


test/cli/proj_outIGNF.dist
```diff
@@ -1,22 +1,22 @@
 +init=IGNF:NTFG +to +init=IGNF:RGF93G
  3.300866856 43.4477976569 0.0000	3d18'0.915"E	43d26'52.077"N 0.000
 +init=IGNF:LAMBE +to +init=IGNF:LAMB93
- 600000.0000 2600545.4523  0.0000	652759.036	7033588.609 0.000
- 135638.3592 2418760.4094  0.0000	187444.148	6856142.911 0.000
- 998137.3947 2413822.2844  0.0000	1048843.997	6843923.913 0.000
+ 600000.0000 2600545.4523  0.0000	652760.737	7033791.243 0.000
```
kbevers, line 4: Coordinates are changing quite in this section. I would expect some differences at the cm to dm level but not meters or even tens of meters. Is something wrong here?

rouault: No, as mentionned in the intro text of this PR, this actually fixes the regression introduced by a66c12277666489cac74535bad8d2cf565ad542d. The updated values by this commit are the original ones of the 5.2 branch (https://github.com/OSGeo/proj.4/blob/f3134c202a9765c899fbd5d656de6ac2d3e545f0/nad/proj_outIGNF.dist#L4)

kbevers: Oh sorry, I totally missed that.

test/gie/ellipsoid.gie
```diff
@@ -57,13 +57,8 @@ expect    failure errno unknown_ellp_param
...
 expect    failure errno major_axis_not_given
 
-operation proj=merc no_defs
-expect    failure errno major_axis_not_given
-
-# This one should succeed due to ellps=WGS84 in proj_def.dat
 operation proj=merc
-accept    0 0
-expect    0 0
+expect    failure errno major_axis_not_given
```
kbevers: Is this because we now don't have a default ellipsoid?

rouault: yes

kbevers: I guess that would leave many projections default-less with regards to ellipsoids. It seems to me that `+proj=merc` not working without adding `+ellps=...` is a step in the wrong direction. I think setting ellipsoid parameters for `GRS80` when initialising a `PJ` would be a sane default. If `+ellps=bessel` (or whatever) is set it will obviously be overridden.

kbevers:
Looks good to me!

PR #1212, opened 2019-01-01
Split ISO19111 C API docs into functions and types

kbevers:
The rest of the C API is structured such that type definitions goes into
one page and functions to another. This commit makes sure the same is
done for the C bindings to the ISO19111 C++ API by defining two doxygen
groups in proj.h, which is then organized into the proper pages with
Sphinx and breathe.

Unfortunately breathe insists on putting the group name as the namespace for the enum definitions. I haven't found a way to avoid this other than not adding `:content-only:` directive which then displays all types as part of the group. This feels like a bug in breathe which should probably be reported.

On my local build I also experience problems with links not being properly set up between the datatypes used in function prototypes and their format description elsewhere in the docs. This works fine for the manually written Sphinx docs. The problem will likely go away if all API documentation is moved to Doxygen/breathe. I am not sure if that is desirable though.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -50,6 +50,7 @@ help:
 .PHONY: clean
 clean:
 	rm -rf $(BUILDDIR)/*
+	rm -rf .doxygen_up_to_date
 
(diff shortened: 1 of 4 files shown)


rouault:
LGTM

PR #1216, opened 2019-01-05
Make sure that coveralls reports coverage from both c and cpp files

kbevers:
Coveralls hasn't been working for a while. I suspect because we've renamed most of the files to `.cpp`.

### DIFF
--- a/travis/after_success.sh
+++ b/travis/after_success.sh
@@ -2,10 +2,10 @@
 
 set -e
 
-
 if [ "$BUILD_NAME" != "linux_clang" ]; then
     # coveralls falsely reports .c-files in the build directories as having 100% coverage so we exclude them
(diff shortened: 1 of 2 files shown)


kbevers:
So that didn't work. `coveralls` fails with:

```
  File "/home/travis/.local/bin/coveralls", line 11, in <module>
    sys.exit(run())
  File "/home/travis/.local/lib/python2.7/site-packages/cpp_coveralls/__init__.py", line 97, in run
    cov_report = coverage.collect(args)
  File "/home/travis/.local/lib/python2.7/site-packages/cpp_coveralls/coverage.py", line 448, in collect
    with io.open(source_file_path, mode='rb') as src_file:
IOError: [Errno 2] No such file or directory: '/home/travis/build/OSGeo/proj.4/test/googletest/src/src/gtest-internal-inl.h'
```

trying to find a way to avoid it

Issue #1214, opened 2019-01-01
Resolve differences with PJ objects created with proj_create versus PJ objects created with proj_create_from_user_input


snowman2:
PJ object created with `proj_create()` do not play well with the other functions that used to have the proj_obj namespace (i.e. objects created with `proj_create_from_user_input()`).

Also, PJ objects created with `proj_create_from_user_input()` do not play well with functions that were meant to be used with PJ objects created with `proj_create()`.

It would be good to merge this behavior when/if it makes sense.

Also, a for reference: https://github.com/OSGeo/proj.4/pull/1208#issuecomment-450754124

kbevers:
What are the discrepancies between the `proj_create`-style `PJ`'s and the `proj_create_from_user_input`-style `PJ`'s?  What will it bridge the gap?

Was it a mistake to merge `PJ` and `PJ_OBJ` into one type?

rouault:
- PJ objects create with proj_create() have their iso_obj member set to NULL so all functions of c_api.cpp will return an error when called on such objects
- Most PJ objects created by proj_create_from_user_input() or other function of c_api.cpp have all the member of PJconst set to 0 / nullptr, except the iso_obj member. Consequently PROJ 5 proj.h functions will return garbage / error out. The only exception to this is when the iso_obj object is of type CoordinateOperation and has a PROJ string representation, in which case a fully valid PJ object is returned having both the 'traditional' member and iso_obj members set

> What will it bridge the gap?

I've considered this a bit this morning. A first idea would be to redirect proj_create() onto proj_create_from_user_input() when this is possible, but there are some subtelties. For example in GDAL, importFromProj4("+proj=longlat +datum=WGS84") would create a CRS with longitude, latitude degree coordinate system, and proj_create_from_user_input() does the same. So the PJ* object that corresponds to this would expect degrees. Whereas proj_create("+proj=longlat +datum=WGS84") expects radians.

> Was it a mistake to merge PJ and PJ_OBJ into one type?

I was midly enthousiastic to be honest. This enables further enhancements though in case we have inspiration on how to do it.

kbevers:
> So the PJ* object that corresponds to this would expect degrees. Whereas proj_create("+proj=longlat +datum=WGS84") expects radians.

[...]

PR #1217, opened 2019-01-08
ISO19111: remove PROJ.5 specific format for CRS (refs #1214)

rouault:
As discussed in https://github.com/OSGeo/proj.4/issues/1214#issuecomment-452084720,
the introduction of a new PROJ.5 format to export CRS using pipeline/unitconvert/axisswap
as an attempt of improving the PROJ.4 format used by GDAL and other products is
likely a dead-end since it is still lossy in many aspects and can cause confusion
with coodinate operations.

Consequently the PROJ_5 convention will be identical to PROJ_4 for CRS export.

Note: on the import side, I've kept the code that could parse unitconvert and
axisswap when building a CRS definition from a pipeline. It is there as a hidden
feature as it was kind of a tear to remove that code in case it might still be
useful...

### DIFF
--- a/include/proj/crs.hpp
+++ b/include/proj/crs.hpp
@@ -502,9 +502,6 @@ class PROJ_GCC_DLL DerivedCRS : virtual public SingleCRS {
 
     PROJ_INTERNAL void setDerivingConversionCRS();
 
-    PROJ_INTERNAL void baseExportToPROJString(
-        io::PROJStringFormatter *formatter) const; // throw(FormattingException)
-
     PROJ_INTERNAL void baseExportToWKT(
         io::WKTFormatter *&formatter, const std::string &keyword,
         const std::string &baseKeyword) const; // throw(FormattingException)
@@ -977,9 +974,6 @@ class PROJ_GCC_DLL DerivedGeodeticCRS final : public GeodeticCRS,
                        const cs::SphericalCSNNPtr &csIn);
     PROJ_INTERNAL DerivedGeodeticCRS(const DerivedGeodeticCRS &other);
 
-    PROJ_INTERNAL void _exportToPROJString(io::PROJStringFormatter *formatter)
-        const override; // throw(FormattingException)
-
     PROJ_INTERNAL CRSNNPtr _shallowClone() const override;
 
     PROJ_INTERNAL bool
@@ -989,6 +983,10 @@ class PROJ_GCC_DLL DerivedGeodeticCRS final : public GeodeticCRS,
 
     PROJ_INTERNAL std::list<std::pair<CRSNNPtr, int>>
     _identify(const io::AuthorityFactoryPtr &authorityFactory) const override;
+
+    // cppcheck-suppress functionStatic
+    PROJ_INTERNAL void _exportToPROJString(io::PROJStringFormatter *formatter)
+        const override; // throw(FormattingException)
 
     INLINED_MAKE_SHARED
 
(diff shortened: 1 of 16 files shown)


kbevers:
> Note: on the import side, I've kept the code that could parse unitconvert and
axisswap when building a CRS definition from a pipeline. It is there as a hidden
feature as it was kind of a tear to remove that code in case it might still be
useful...

Sure, makes sense. I didn't see a comment mentioning this, would that be possible or is the functionality spread across several places (and thus difficult to comment on)?

> Consequently the PROJ_5 convention will be identical to PROJ_4 for CRS export.

What is the reason for keeping both then?

A general comment on the subject of PROJ.4 strings, we should do as much as possible to discourage the use of them. You also mention this in a comment in this PR. We should also stress this in the documentation.

rouault:
> What is the reason for keeping both then?

Export of coordinate operation / pipelines shouldn't be allowed for the PROJ.4 case

> We should also stress this in the documentation.

Yes, done too in the Doxygen of PROJStringFormatter::Convention
```
     * <ul>
     * <li>For PROJStringFormatter::Convention::PROJ_5 (the default),
     * <ul>
     * <li>For a crs::CRS, returns the same as
     * PROJStringFormatter::Convention::PROJ_4. It should be noted that the
     * export of a CRS as a PROJ string may cause loss of many important aspects
     * of a CRS definition. Consequently it is discouraged to use it for
     * interoperability in newer projects. The choice of a WKT representation
     * will be a better option.</li>
     * <li>For operation::CoordinateOperation, returns a PROJ
     * pipeline.</li>
     * </ul>

```

rouault:
> would that be possible or is the functionality spread across several places (and thus difficult to comment on)?

This functionnality is PROJStringParser::Private of io.cpp, more precisely in the PROJStringParser::Private::processAxisSwap(), some unitconvert related code in various methods of this class, and the logic that identifies the various pipeline steps in PROJStringParser::createFromPROJString(). It could be removed. The most painful would actually be to remove the corresponding test cases.

kbevers:
> Export of coordinate operation / pipelines shouldn't be allowed for the PROJ.4 case

Okay, I still don't understand. `projinfo` tells me there's only a `PROJ` format not. I thought that was where the `PROJ_4` and `PROJ_5` convention showed up. What did I miss?

Related to `projinfo`, wouldn't be better in the output below to replace `PROJ` with `PROJ.4` and `PROJ.5` with `PROJ`? Again, this is an attempt at stressing the legacy nature of PROJ.4 strings.

```
$ ./bin/projinfo -o PROJ EPSG:25832
PROJ string:
Error when exporting to PROJ string: BoundCRS cannot be exported as a PROJ.5 string, but its baseCRS might

$ ./bin/projinfo -o PROJ EPSG:32632
PROJ string:
+proj=utm +zone=32 +datum=WGS84 +units=m +no_defs
```

> Yes, done too in the Doxygen of PROJStringFormatter::Convention

Great, I missed that. I was also thinking about somewhere in the more general documentation but can be done at another time. I think we need to do a better job of describing how proj-strings work, both legacy PROJ.4 strings and newer ones. Thinking about this topic has been a background process for me for a long time so hopefully I will soon get round to actually doing it.

> The most painful would actually be to remove the corresponding test cases.

At least it is easy to tell which tests are affected :-) And probably not easy to take them apart...

kbevers:
> But your example shows that there's still an issue in the case of BoundCRS. I need to investigate. Do you prefer that I revert the change in projinfo and keep both formats available as before ?

Ah, I didn't even realize I found a bug. No, I think it is fine to just keep the one format. I think I prefer to call it `PROJ.4` though.

rouault:
> Ah, I didn't even realize I found a bug

Just fixed.

> No, I think it is fine to just keep the one format. I think I prefer to call it PROJ.4 though.

This wouldn't be always appropriate, since when exporting coordinate operation, this is not PROJ.4 compatible. What I could do is change the label dynamically: if it is a CRS, then use "PROJ.4". If It is a coordinate operation, then "PROJ" ?

kbevers:
> This wouldn't be always appropriate, since when exporting coordinate operation, this is not PROJ.4 compatible. What I could do is change the label dynamically: if it is a CRS, then use "PROJ.4". If It is a coordinate operation, then "PROJ" ?

Of course. Yes, while it's a subtle difference it conveys the message that this is a legacy style string.

PR #1218, opened 2019-01-09
Add API in proj.h to set a file finder callback and search paths; support multiple directories in PROJ_LIB

rouault:


kbevers:
Looks good to me. Very nice to be able to have more directories in `PROJ_LIB` now!

PR #1230, opened 2019-01-19
 Add support for spaces in grid name parameters (fixes #1152) 

### LINKED ISSUE #1152: proj_create: +geogrids breaks Windows pathnames with embedded spaces
proj_create() calls pj_trim_argc(), which breaks the create string into string tokens. In particular, Windows pathnames with embedded spaces are thus broken.

Example input: +geoidgrids=C:\Program Files\foo\Geoid Files\foobar.gtx

rouault:


kbevers:
Seems good to me. Just one question: Can folders with spaces in them also be added to `PROJ_LIB`?

kbevers:
Cool, just wanted to make sure that also worked.

PR #1233, opened 2019-01-22
Coordinate operation computation with boundcrs / wktext: drop useless early bindins terms in generated pipeline (fixes #1232)

### LINKED ISSUE #1232: projinfo forwards unused parameters
Consider this call to `projinfo` that uses legacy PROJ.4 strings:

```
projinfo -s "+proj=utm +zone=55 +south +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +wktext  +nadgrids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb +type=crs" -t  "+proj=utm +zone=55 +south +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +wktext +type=crs"
```

which results in the following output:
```
unknown id, Transformation from unknown to unknown + unknown to WGS84 + Inverse of Transformation from unknown to WGS84 + Transformation from unknown to unknown, unknown accuracy, unknown domain of validity

PROJ string:
+proj=pipeline +step +inv +proj=utm +zone=55 +south +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +wktext +nadgrids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb +step +proj=hgridshift +grids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb +step +proj=utm +zone=55 +south +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +wktext
```

All input parameters are forwarded even though they effectively are replaced by newer alternatives. For instance `+nadgrids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb` and `+proj=hgridshift +grids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb`. Parameters like `+towgs84` and `+wktext` are also present in the output but unused. 

It shouldn't be difficult to weed out the unused parameters, as that is kept track of in `PJ->params`.

rouault:


kbevers:
Nice. I notice though that only a few specifically named keywords are dropped. See for instance here where I have added a few parameters of my own invention:

```
>projinfo.exe -o PROJ -s "+proj=utm +zone=55 +south +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +wktext  +nadgrids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb +type=crs +param=random" -t  "+proj=utm +zone=55 +south +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +wktext +type=crs +nonstandardparameter"
unknown id, Transformation from unknown to unknown + unknown to WGS84 + Inverse of Transformation from unknown to WGS84 + Transformation from unknown to unknown, unknown accuracy, unknown domain of validity

PROJ string:
+proj=pipeline +step +inv +proj=utm +zone=55 +south +ellps=GRS80 +units=m +no_defs +param=random +step +proj=hgridshift +grids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb +step +proj=utm +zone=55 +south +ellps=GRS80 +units=m +no_defs +nonstandardparameter
```

I think it would make the most sense to only output parameters that PROJ actually uses. If possible. I thought that it would be a matter of checking against `PJ->params` but I see that is not the solution you have chosen. I guess there's a reason for that?

Another (unrelated) thing: I think it would improve readability of the output if pipelines where split across several lines, i.e. one per step:

```
>projinfo.exe -o PROJ -s "+proj=utm +zone=55 +south +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +wktext  +nadgrids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb +type=crs +param=random" -t  "+proj=utm +zone=55 +south +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +wktext +type=crs +nonstandardparameter"
unknown id, Transformation from unknown to unknown + unknown to WGS84 + Inverse of Transformation from unknown to WGS84 + Transformation from unknown to unknown, unknown accuracy, unknown domain of validity

PROJ string:
+proj=pipeline
+step +inv +proj=utm +zone=55 +south +ellps=GRS80 +units=m +no_defs +param=random
+step +proj=hgridshift +grids=C:\dev\proj-datumgrid\oceania\GDA94_GDA2020_conformal.gsb 
+step +proj=utm +zone=55 +south +ellps=GRS80 +units=m +no_defs +nonstandardparameter
```

rouault:
>  I thought that it would be a matter of checking against `PJ->params` but I see that is not the solution you have chosen. I guess there's a reason for that?

Yes, the ISO19111 code uses from the beginning its own PROJ string parsing code, because I was unconfortable with using the historical C code for that. Probably that the two should be reconciled at some point, or the C++ code could be assisted wth the C code to know which parameters are actually used. I'm going to look a bit if I can do something "quickly", or if this is more involved and has to be defered to "later"...

> I think it would improve readability of the output if pipelines where split across several lines, i.e. one per step

The PROJStringFormatter class could receive a setMultiLine() method to set up this (single line formating should probably be the default in the API). Exercice left to the suggester ? ;-)

kbevers:
> Yes, the ISO19111 code uses from the beginning its own PROJ string parsing code, because I was unconfortable with using the historical C code for that. Probably that the two should be reconciled at some point, or the C++ code could be assisted wth the C code to know which parameters are actually used. I'm going to look a bit if I can do something "quickly", or if this is more involved and has to be defered to "later"...

Sure, if it is complicated it isn't a priority. What you've done already probably handles >95% of all encounters of non-used parameters projinfo will ever see.

> The PROJStringFormatter class could receive a setMultiLine() method to set up this (single line formating should probably be the default in the API). Exercice left to the suggester ? ;-)

Definitely an exercise for the suggester. Not sure if I agree that single-line should be the default though. If `\`'s or `^`'s are placed at the ende of the line the string should still be copy'n'pasteable. But I guess that is easier to decide when there's an implementation. I'll get back to this when I have bit of surplus time :-)

rouault:
> Not sure if I agree that single-line should be the default though

Making multi-line the default in the API will for sure break several tens of expected results in test/unit/. Not necessarily a good argument by itself :-)

kbevers:
> Making multi-line the default in the API will for sure break several tens of expected results in test/unit/. Not necessarily a good argument by itself :-)

Sorry, I misread that. I was talking about the command line output. Yes, I agree that changing the API default is a bad idea - at least I am not keen on updating a million test cases :-)

src/iso19111/io.cpp
```diff
@@ -5352,13 +5354,17 @@ PROJStringSyntaxParser(const std::string &projString, std::vector<Step> &steps,
...
                 const auto pos = word.find('=');
                 auto key = word.substr(0, pos);
-                auto pair = (pos != std::string::npos)
-                                ? Step::KeyValue(key, word.substr(pos + 1))
-                                : Step::KeyValue(key);
-                if (steps.empty()) {
-                    globalParamValues.push_back(pair);
-                } else {
-                    steps.back().paramValues.push_back(pair);
+                if (!(dropEarlyBindingsTerms &&
```
kbevers, line 5359: I wonder, With the changes you have pushed in c048292f3b8e408e6a81700a74d9b44d532227ac, is it really necessary to give `+towgs84`, `+nadgrids`, `+geoidgrids`, `+wktext` special treatment? If it is, I guess `+datum` should be added to the list as well.

rouault: `+towgs84`, `+nadgrids`, `+geoidgrids` should at least receive special treatment since they might be stored in the preserved PROJ string in case there is a parameter that prevents building a full ISO19111 compliant CRS (we must preserve them when exporting the CRS as a PROJ string,but not when integrating it in a pipeline)
Regarding `datum`: if we drop it, then we must substitute it by its `ellps`. I'll try to do that.
+wktext dropping should indeed no longer be neeed

kbevers: Sure, makes sense. Thanks for the clarification.

PR #1234, opened 2019-01-22
Make tmerc an alias for etmerc.

### LINKED ISSUE #404: Changing Transverse Mercator Algorithm for +proj=tmerc
The issue to make tmerc an alias for etmerc @cffk 's implementation, and rename Snyder's algorithm from tmerc to stmerc.  This discussion came up in #374.

Excerpts from the aforementioned issue:

---

**busstoptaktik** commented on Apr 26
Regarding Transverse Mercator: are you using the high precision etmerc implementation, or the slightly faster, but less precise, original tmerc version?

**cffk** commented on Apr 26
etmerc will be used in the next release (and this is the checked in version).

**hobu** commented on Apr 27

> Strictly speaking, I believe the proper solution will be to rename what is now known as tmerc, to stmerc (indicating its John Snyder heritage), and renaming etmerc to tmerc

:+1:

**rouault** commented on Apr 27
For backward compatibility, I'd suggest not to remove etmerc, but make etmerc and tmerc aliases.

**busstoptaktik** commented on Apr 27
@rouault: yes - that is really a much better solution (and requires only a few additional lines of code).

kbevers:
etmerc is superior to tmerc and is therefore now the default transverse
mercator implementation in PROJ. The transverse mercator of
Snyder/Evenden origin is kept under the name stmerc for backwards
compatibility.

Closes #404


rouault:
`$ grep etmerc ../src/iso19111/*` suggest a few places that will also likely need to be reviewed/modified
```
../src/iso19111/c_api.cpp: * +proj=etmerc instead of +proj=tmerc (or USE_ETMERC=NO to disable implicit
../src/iso19111/c_api.cpp: * use of etmerc by utm conversions)
../src/iso19111/coordinateoperation.cpp:        if (l_method->getPrivate()->projMethodOverride_ == "etmerc" &&
../src/iso19111/coordinateoperation.cpp:                    l_method->getPrivate()->projMethodOverride_ == "etmerc";
../src/iso19111/coordinateoperation.cpp:            formatter->addStep(useETMerc ? "etmerc" : mapping->proj_name_main);
../src/iso19111/io.cpp:/** \brief Set whether Extended Transverse Mercator (etmerc) should be used
../src/iso19111/io.cpp:    if (name == "etmerc" || name == "utm" ||
../src/iso19111/io.cpp:    } else if (step.name == "etmerc") {
../src/iso19111/io.cpp:        if (step.name == "etmerc") {
../src/iso19111/io.cpp:            methodMap.set("proj_method", "etmerc");
```

src/iso19111/io.cpp
```diff
@@ -6070,7 +6070,7 @@ static bool isGeocentricStep(const std::string &name) {
 // ---------------------------------------------------------------------------
 
 static bool isProjectedStep(const std::string &name) {
-    if (name == "etmerc" || name == "utm" ||
+    if (name == "stmerc" || name == "utm" ||
```
rouault: name == "etmerc" should probably be kept as well. projectionMethodMappings[] in include/proj/internal/coordinateoperation_constants.hpp has only mapping for "tmerc"

rouault: Around line 6819
```
} else if (step.name == "etmerc") {
        mapping = getMapping(EPSG_CODE_METHOD_TRANSVERSE_MERCATOR);
```
should also test || step.name == "stmerc"

rouault: And around line 7071
```
        if (step.name == "etmerc") {
            methodMap.set("proj_method", "etmerc");
        }
```
there should be a similar case for stmerc

test/unit/test_operation.cpp
```diff
@@ -1237,7 +1237,7 @@ TEST(operation, tmerc_export) {
 
     {
         auto formatter = PROJStringFormatter::create();
-        formatter->setUseETMercForTMerc(true);
+        //formatter->setUseSTMercForTMerc(true);
```
rouault: I guess this commented part is temporary ?

src/iso19111/c_api.cpp
```diff
@@ -1204,9 +1204,9 @@ const char *proj_as_wkt(PJ_CONTEXT *ctx, const PJ *obj, PJ_WKT_TYPE type,
  * @param type PROJ String version.
  * @param options NULL-terminated list of strings with "KEY=VALUE" format. or
  * NULL.
- * The currently recognized option is USE_ETMERC=YES to use
- * +proj=etmerc instead of +proj=tmerc (or USE_ETMERC=NO to disable implicit
- * use of etmerc by utm conversions)
+ * The currently recognized option is USE_STMERC=YES to use
+ * +proj=stmerc instead of +proj=tmerc (or USE_STMERC=NO to disable implicit
```
rouault: The part "(or USE_STMERC=NO to disable implicit use of tmerc by utm conversions)" doesn't make sense. We probably no longer need a specific behaviour for USE_STMERC=NO

src/iso19111/coordinateoperation.cpp
```diff
@@ -5471,15 +5471,15 @@ void Conversion::_exportToPROJString(
...
         int zone = 0;
         bool north = true;
-        bool etMercSettingSet = false;
-        useETMerc = formatter->getUseETMercForTMerc(etMercSettingSet) ||
-                    l_method->getPrivate()->projMethodOverride_ == "etmerc";
-        if (isUTM(zone, north) && !(etMercSettingSet && !useETMerc)) {
+        bool stMercSettingSet = false;
+        useSTMerc = formatter->getUseSTMercForTMerc(stMercSettingSet) ||
+                    l_method->getPrivate()->projMethodOverride_ == "stmerc";
+        if (isUTM(zone, north) && !(stMercSettingSet && !useSTMerc)) {
```
rouault: the negation before useSTMerc should likely be removed. If the user forces STMerc, we don't want UTM to be expressed as +proj=utm, but rather +proj=stmerc ...

kbevers:
I've been fiddling around with this on and off the last couple of weeks and I must admit that I am too unfamiliar with the ISO19111 code to make these changes in a reasonable amount of time. That's unfortunate but time is a limited resource. I see two ways forward:

1) Close the PR and keep status quo
2) I revert my changes in the ISO19111 code and let someone else take care of that.

I don't expect anyone to clean up after me, so 1) is the likely outcome. Any takers for 2)?

rouault:
Didn't you want to use a `+proj=tmerc +approx` syntax actually ? In which case, you could open a PR for that in the 'PROJ classic' code and I'll try to fill in the gaps in the ISO19111 code when I've a chance to (but I'm terribly stretched at the moment)

kbevers:
> Didn't you want to use a `+proj=tmerc +approx` syntax actually ? In which case, you could open a PR for that in the 'PROJ classic' code and I'll try to fill in the gaps in the ISO19111 code when I've a chance to (but I'm terribly stretched at the moment)

Yeah, I did, but like you I am tied up in too much at the moment. Initially I thought this would be a minor simple thing that I could do on a lunchbreak. I wouldn't have started if I realised the complexity.

I can try to do the `+proj=tmerc +approx` thing within the next couple of days. Does that leave you enough time to fill in the gaps before the 15th?

rouault:
> Does that leave you enough time to fill in the gaps before the 15th?

I'll try. This also impacts GDAL if we change the USE_ETMERC option (not sure if we should or not)

kbevers:
> I'll try. This also impacts GDAL if we change the USE_ETMERC option (not sure if we should or not)

If it's too much trouble it's better to not do this. The functional gains are minimal and should not take priority over more important things. I for one would rather spend my time on getting the pipeline stack ready for 6.0.

kbevers:
> I can try to do the `+proj=tmerc +approx` thing within the next couple of days.

@rouault I have now made the changes. Assuming the tests pass, this should be ready for you to take over. Note that I have added the `+approx` flag to the UTM projection as well. This possibly simplifies things in the ISO19111 code.

kbevers:
Looks like I have to adjust the testvarious test. I'll do that later tonight when I am near a non-windows computer...

rouault:
@kbevers I've done the adjustments per https://github.com/rouault/proj.4/commit/8160417fdb459794b2dbc1229baaa79b949e9f6b

rouault:
I've also added a new commit https://github.com/rouault/proj.4/commit/0c3be95 for the ISO19111 part. You might merge my continued rouault/tmerc-alias branch into yours.

kbevers:
> I've also added a new commit [rouault@0c3be95](https://github.com/rouault/proj.4/commit/0c3be95) for the ISO19111 part. You might merge my continued rouault/tmerc-alias branch into yours.

Thanks, I owe you one!

Just a general question, is the documentation adequate as it is now? I wonder if I should leave out that `+proj=etmerc` can be used and just let it silently disappear from the documentation.

rouault:
> Just a general question, is the documentation adequate as it is now? I wonder if I should leave out that `+proj=etmerc` can be used and just let it silently disappear from the documentation.

I'm indifferent if we should let it mentionned as an alias like you did or remove it

rouault:
@kbevers  commit 5995098 doesn't fix the leak. Please pull from my tree for https://github.com/rouault/proj.4/commit/300d25b

kbevers:
> @kbevers commit [5995098](https://github.com/OSGeo/proj.4/commit/5995098f59d366fdafdce55d494e69b9be56b454) doesn't fix the leak. Please pull from my tree for [rouault@300d25b](https://github.com/rouault/proj.4/commit/300d25b)

You are quite right. It did fix some other problems though :-)

Hopefully the tests pass in this run. Do you mind if I rebase or squash the commits? They are all over the place right now.

rouault:
Please squash

PR #1238, opened 2019-01-26
simplify appveyor script

snowman2:
I spent a little too much time tinkering with this for the pyproj appveyor build and thought I might share updates if there is interest.


kbevers:
This does indeed simplify things a bit. Thanks for the improvements!

PR #1248, opened 2019-02-04
Remove all traces of nad2bin and nad2nad

### LINKED ISSUE #1235: nad2bin being run while building prevents cross compiling
I am trying to compile Proj 5.2 for Android and I am facing this error:

image

It tries to run nad2bin on the environnement, but nad2bin has been compiled for Android.

What is the proper approach?
I am using autotools, but I failed to disable building nad2bin (is it possible?). But is it the right approach?

### LINKED ISSUE #1236: Remove nad2bin?
Instead of hijacking the conversion in https://github.com/OSGeo/proj.4/issues/1235, let's continue the talk here.

In response to my comment @sebastic wrote:

> > We can move nad2bin to proj-datumgrid. I feel like it has served it's purpose as part of the main PROJ package now that all the *.lla files have moved elsewhere.

> In my opinion it belongs in the PROJ source tree. No need to complicate the proj-datumgrid project by having to build C programs.

In my opinon there's no need to distribute a tool in a package in which it isn't used. Additionally, it's a binary we install but there's no trace of documentation for the application. It is unlikely that documentation is going to be written and I suspect that it is also unlikely that anyone will create new grids using `nad2bin`.

Loosing `nad2bin` from PROJ would also be a welcome simplification of two build setups in contrast to only maintaining CMake in proj-datumgrids.

I am of course volunteering to make this change myself.

kbevers:
The source material for the default grids used by
PROJ has been moved to the proj-datumgrid repository. For that
reason it is no longer necessary to include the nad2bin program
in the PROJ repository and source distribution. From now on the
nad2bin application will be kept in the proj-datumgrid repo.

Previously the null grid was generated by running nad2bin on the
null.lla file. Since nad2bin is no longer available null.lla has
been replaced by its binary counterpart null. This file will be
distributed and installed alongside PROJ.

Build scripts and documenation has been adjusted so that nad2bin
is not mentioned anywhere. Additionally all references to nad2nad
has been removed as well. nad2nad has not been part of the PROJ
distribution for quite some time so this has been long overdue.

Resolves #1235 
Resolves #1236


kbevers:
This seem to be working now. It would be nice if someone could go through my changes and confirm that I haven't removed anything important.

PR #1251, opened 2019-02-06
Add proj_get_crs_info_list_from_database()

rouault:
This method is intended to be used typically by GUI that lists all possible CRS.

What is does could be done by previously existing functions, but it is much faster.
It typically runs in less than 0.1s (hot run) versus ~0.5s with the method that
consists in enumerating all codes and instanciating a PJ object for each of them.

### DIFF
--- a/src/proj.h
+++ b/src/proj.h
@@ -645,6 +645,74 @@ typedef enum
     PJ_CS_TYPE_TEMPORALMEASURE
 } PJ_COORDINATE_SYSTEM_TYPE;
 
+/** \brief Structure given overall description of a CRS.
+ *
+ * This structure may grow over time, and should not be directly allocated by
+ * client code.
+ */
+typedef struct
+{
+    /** Authority name. */
+    char* auth_name;
+    /** Object code. */
+    char* code;
+    /** Object name. */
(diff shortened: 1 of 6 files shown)


src/proj.h
```diff
@@ -645,6 +645,74 @@ typedef enum
...
+    /** Southern-most latitude of the area of use, in degrees. */
+    double south_lat_degree;
+    /** Eastern-most longitude of the area of use, in degrees. */
+    double east_lon_degree;
+    /** Northern-most latitude of the area of use, in degrees. */
+    double north_lat_degree;
+
+    /** Whether deprecated objects are allowed. Default to FALSE. */
+    int allow_deprecated;
+} PROJ_GET_CRS_LIST_PARAMETERS;
```
kbevers: The "GET" in this name is a little confusing. Why not simply call it `PROJ_CRS_LIST_PARAMETERS`?

kbevers (commented):
Seems good to me. Definitely useful for GUI applications.

src/proj.h
rouault: Yeah, will do that

PR #1253, opened 2019-02-08
Remove data/install.in.

QuLogic:
The files referenced in it (and more) are already installed by listing them in `pkgdata_DATA` (for autotools) and in the `install` command (for CMake).


QuLogic:
Err, why are there 4 Travis CI builds?

kbevers:
> Err, why are there 4 Travis CI builds?

No idea. They are all referencing the same build. Probably just a fluke.

Thanks for clearing up what was going on with install.in. And removing it :-)

PR #1254, opened 2019-02-10
Completely remove Chebychev remains from codebase

rouault:
c6ab83f5742bc5ac6f9cb9a8b2a4f1ea241b6f63 already removed their availability
in user facing application, but the library code remained, and appeared to be
unused by the rest of the library, and not available to library users,
the API being only in proj_internal.h. So remove all remains.


src/proj_internal.h
```diff
@@ -677,8 +677,8 @@ struct FACTORS {
 #define PJD_ERR_TOO_MANY_INITS          -57
 #define PJD_ERR_INVALID_ARG             -58
 #define PJD_ERR_INCONSISTENT_UNIT       -59
-/* NOTE: Remember to update pj_strerrno.c and transient_error in */
-/* pj_transform.c when adding new value */
+/* NOTE: Remember to update strerrno.cpp and transient_error in */
+/* transform.cpp when adding new value */
```
kbevers: gie.cpp also need to be added to this list (incidentally I did the same correction as you in https://github.com/OSGeo/proj.4/pull/1250/)

kbevers:
Thanks for doing this. I didn't realize that all this code was left unused. This is a great simplification!

src/proj_internal.h
rouault: Done

PR #1262, opened 2019-02-13
Set CMake variable CXX_STANDARD with C++11 as default

### LINKED ISSUE #1118: CMake C++11 configuration

mloskot:
Set required C++ standard version with CMake variable instead of hand-rolled compiler-specific flags.

Allow user to override the default with `-DCMAKE_CXX_STANDARD={98,11,14,17,20}`

Closes #1118

---

This PR is a basic and minimal approach, and more thorough refinements are possible in future: see  #1263

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@
 #################################################################################
 # General settings
 #################################################################################
-cmake_minimum_required(VERSION 2.8.3 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 # For historic reasons, the CMake PROJECT-NAME is PROJ4
 project(PROJ4 LANGUAGES C CXX)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -9,7 +9,7 @@
 #################################################################################
 # General settings
 #################################################################################
-cmake_minimum_required(VERSION 2.8.3 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
```
kbevers: As per RFC3 this can go all the way to 3.5. No need to raise it that high unless there's a reason for it but if it makes things easier for you here please go ahead.

kbevers:
Looks good to me. Thanks for taking the time to do this!

Your perspectives for the future are also really good. It would probably be a good idea to keep track of that stuff in a "CMake modernization" issue.

mloskot:
@kbevers 
> Your perspectives for the future are also really good. It would probably be a good idea to keep track of that stuff in a "CMake modernization" issue.

Do you mean a new issue or any existing one?

CMakeLists.txt
mloskot: Done

kbevers:
> Do you mean a new issue or any existing one?

I looked around for an existing ticket but I couldn't find one so I guess a new one.

mloskot:
@kbevers Done in #1263

Issue #1221, opened 2019-01-15
Update build instructions


kbevers:
With the switch to C++ and the inclusion of sqlite as a dependency we need to update the build instructions in `README.md` and `docs/source/install.rst`. Neither mention that sqlite is required, nor how to configure it (probably only relevant for windows builds).

PR #1268, opened 2019-02-14
Improve CMake sqlite3 dependency docs

mwtoews:
Hopefully address #1221 

Furthermore, CMake's configure message for "sqlite3 binary/dependency not found" is elevated to FATAL_ERROR, so it does not continue to (e.g.) download googletest, etc., which takes a while.


README.md
```diff
@@ -30,6 +30,11 @@ The following command line utilities are included in the PROJ package:
 
 
 ## Installation
+### Build dependencies
+
+PROJ requires a C and C++ compiler that supports the C++11 standard.
```
mloskot: or shorter "PROJ requires a C and C++11 compiler", also makes it clear C++11 standard requirement does not apply to the C compiler :)

README.md
```diff
@@ -30,6 +30,11 @@ The following command line utilities are included in the PROJ package:
...
+It also requires SQLite3 (includes, library and executable).
```
mloskot: s/includes/headers/

CMakeLists.txt
```diff
@@ -88,12 +88,12 @@ include(policies)
...
 if (NOT EXE_SQLITE3)
-MESSAGE (SEND_ERROR "sqlite3 binary not found!")
+MESSAGE (FATAL_ERROR "sqlite3 binary not found!")
 ENDIF (NOT EXE_SQLITE3)
 
 FIND_PACKAGE(Sqlite3)
 IF (NOT SQLITE3_FOUND)
-MESSAGE (SEND_ERROR "sqlite3 dependency not found!")
+MESSAGE (FATAL_ERROR "sqlite3 dependency not found!")
 ENDIF (NOT SQLITE3_FOUND)
```
dg0yt, line 97: I wonder if `SEND_ERROR` was used intentionally to collect more errors, before failing.

Unlike `find_program`, `find_package` supports a `REQUIRED` argument which stops processing with an error message if the package cannot be found. (And this is implemented in `cmake/FindSqlite3.cmake`.)

mwtoews: I tested this, and it does not stop as is documented/expected. CMake shows the message and continues. [Some other folks appear to see this behaviour too](https://stackoverflow.com/a/2711683), with the implication that `FindSqlite3.cmake` is potentially broken. This modification to message seems to work fine.

dg0yt: I don't understand.

> I tested this, and it does not stop as is documented/expected. CMake shows the message and continues.

This is expected. CMake has to continue in order to "collect more errors, before failing.". But eventually it will end with an error. This allows to report both the missing executable and package, instead of stopping already at the missing executable.

> Some other folks appear to see this behaviour too, with the implication that FindSqlite3.cmake is potentially broken.

I can't see how this is related. The linked Q&A is from 2010 and neither adressing this project nor SQLite. For concerns about `REQUIRED` not being support cf.:
https://github.com/OSGeo/proj.4/blob/a813ad65cfb819bfa295dfbacdf577b258377c93/cmake/FindSqlite3.cmake#L60-L66

mwtoews: I'll revert the message level back to `SEND_ERROR`, and add `REQUIRED` as recommended. I initially found it confusing that CMake would fail to find a required dependency, then continue to download googletest and perform other checks as though everything were normal.

mloskot: IMHO, I would suggest avoid last-minute pre-release fiddling with `FindSQLite3.cmake`, unless you are deadly seriously certain you know what you are doing.

mwtoews: > IMHO, I would suggest avoid last-minute pre-release fiddling with `FindSQLite3.cmake`, unless you are deadly seriously certain you know what you are doing.

Agreed. The only modification is to add `REQUIRED` to `FIND_PACKAGE`, which seems to have been since old versions.

README.md
```diff
@@ -42,6 +47,11 @@ On Windows, one may need to specify generator:
 
     cmake -G "Visual Studio 15 2017" ..
 
+If the SQLite3 dependency is installed in a custom location, specify the
+paths to the include directory and the library::
+
+    cmake -DSQLITE3_INCLUDE_DIR=/opt/SQLite/include -DSQLITE3_LIBRARY=/opt/SQLite/lib/libsqlite3.so ..
```
dg0yt, line 53: Or use the `CMAKE_PREFIX_PATH` cache variable (which is a list BTW):
~~~
cmake -DCMAKE_PREFIX_PATH=/opt/SQLite ..
~~~

This will also help find the executable (`find_program`).

mwtoews: I've never had problems finding `sqlite3` (it's on the `PATH`). Just the other components.

This suggestion works too, so I've added it to docs/source/install.rst

Issue #1177, opened 2018-11-15
cmake: default include and lib subdirs for Windows


rouault:
cmake/ProjInstallPath.cmake has set(DEFAULT_LIB_SUBDIR local/lib) and set(DEFAULT_INCLUDE_SUBDIR local/include) for the WIN32 case. @hobu I see you were the one to commit those lines ;-) : any reason for that local/ prefix, and not using the Unix layout ?

hobu:
This pattern originated in libLAS, and I blame @mloskot for it 😄 

It can be cleaned up.

PR #1274, opened 2019-02-16
Windows CMAKE Paths to match  osx/linux

snowman2:
Addressing issue #1177.


kbevers:
Thanks for taking care of this!

PR #1276, opened 2019-02-17
Modify the default strategy of researching intermediate CRS to do it only if there is no direct transformation

rouault:


pramsey:
Wonderful!

PR #1279, opened 2019-02-20
Vertcrs transform improvements

rouault:
Assorted set of improvements/fixes for VertCRS / CompoundCRS transformations following @pramsey 's use case of https://lists.osgeo.org/pipermail/proj/2019-February/008196.html


src/apply_vgridshift.cpp
```diff
@@ -35,16 +35,16 @@
...
     /* nodata?  */
     /* GTX official nodata value if  -88.88880f, but some grids also */
     /* use other  big values for nodata (e.g naptrans2008.gtx has */
     /* nodata values like -2147479936), so test them too */
-    return value > 1000 || value < -1000 || value == -88.88880f;
+    return value * vmultiplier > 1000 || value * vmultiplier < -1000 || value == -88.88880f;
 }
 
-static double read_vgrid_value( PJ *defn, PJ_LP input, int *gridlist_count_p, PJ_GRIDINFO **tables, struct CTABLE *ct) {
+static double read_vgrid_value( PJ *defn, PJ_LP input, double vmultiplier, int *gridlist_count_p, PJ_GRIDINFO **tables, struct CTABLE *ct) {
```
kbevers, line 47: Is the `vmultiplier` only used here for checking if a grid value is nodata?

rouault: yes

src/transformations/vgridshift.cpp
```diff
@@ -26,7 +26,7 @@ static PJ_XYZ forward_3d(PJ_LPZ lpz, PJ *P) {
     if (P->vgridlist_geoid != nullptr) {
         /* Only try the gridshift if at least one grid is loaded,
          * otherwise just pass the coordinate through unchanged. */
-        point.xyz.z += Q->forward_multiplier * proj_vgrid_value(P, point.lp);
+        point.xyz.z += Q->forward_multiplier * proj_vgrid_value(P, point.lp, Q->forward_multiplier);
```
kbevers: Related to my other comment above, it looks like a mistake here that `Q->forward_multiplier` is used twice. If I understand correctly the multiplier isn't actually applied in `proj_grid_value` but just casually reading this line of code I would expect it to be applied first in `proj_vgrid_value` and then once more in this statement. Shouldn't the multiplier be applied in `proj_vgrid_value`?

rouault: you're right. that will make things less confusing

kbevers:
@rouault You want this included in PROJ 6RC2, yes?

rouault:
> You want this included in PROJ 6RC2, yes?

yes

kbevers:
Do you mind requesting a review once you are finished? Just so that I don't merge this prematurely (there's a new "draft pull request" now that seems useful in this regard, btw).

rouault:
I guess I'm done on this

include/proj/coordinateoperation.hpp
```diff
@@ -148,6 +148,8 @@ class PROJ_GCC_DLL CoordinateOperation : public common::ObjectUsage,
     PROJ_DLL bool
     isPROJInstanciable(const io::DatabaseContextPtr &databaseContext) const;
 
+    PROJ_DLL bool hasRoughTransformation() const;
```
kbevers: I'm probably not a good enough English speaker to be a judge of this but "rough" sounds a bit off to me. I understand the intention it just doesn't seem like the right word to me for some reason. Maybe "approximate" or "imprecise"?

kbevers: Also, it sounds a bit like "Hough transformation" and that's something entirely different :)

mwtoews: how about `hasApproximateTransformation`?

src/iso19111/coordinateoperation.cpp
```diff
@@ -106,6 +106,12 @@ constexpr double UTM_SOUTH_FALSE_NORTHING = 10000000.0;
 static const std::string INVERSE_OF = "Inverse of ";
 static const char *NULL_GEOCENTRIC_TRANSLATION = "Null geocentric translation";
 static const char *NULL_GEOGRAPHIC_OFFSET = "Null geographic offset";
+static const char *APPROXIMATE_TRANSFORMATION_PREFIX =
+    " (approximate transformation";
+static const char *APPROXIMATE_TRANSFORMATION = " (approximate transformation)";
```
mwtoews: just to be sure, this "approximate transformation" is the same as the "rough transformation"? If so, then I guess make all of other instances of "rough" to "approximate".

include/proj/coordinateoperation.hpp
rouault: yeah, I wasn't sure about the appropriate term. All transformations are "approximate" / "imprecise" by nature. I also considered "crude". @mwtoews any suggestion of which term might convey the information that this is *very* approximate ?

mwtoews: yeah, English doesn't have relative words for _more_ or _less_ approximate. All I can think of is a list of other words similar to rough, such as ballpark, fuzzy, etc.

rouault: Hum, maybe hasUnpreciseTransformation() ?

mwtoews: Ok, it's silly, but I'm appreciating the adjective ["ballpark"](https://www.collinsdictionary.com/dictionary/english/ballpark) at the moment. It's illustrative of spatial inaccuracies by it's very definition, which in some cases (e.g. ~40 m) is in the range of the size of a ballpark.

rouault: @kbevers Are you fine with ballpark ?

kbevers: Yes, I think it is appropriate to include a bit of silliness from time to time :-)

busstoptaktik: In network adjustments you would start out with "preliminary coordinates" before refining through the adjustment, so perhaps a "preliminary transformation" would work?

And while doing linguistic nitpicking,  I think "instanciation" and multiple other derived forms all over the code tree should be spelled "instantiation", i.e. with a "t". Probably most important to correct this in function\method names.

rouault:
Latest commit amended to use the ballpark terminology

kbevers:
Looks good to me. Let's get this merged once the test passes!

rouault:
Merged and backported to 6.0

PR #1280, opened 2019-02-20
proj.db: add missing custom entries for vertical transform and fix a few wrong ones

rouault:


test/cli/testprojinfo
```diff
@@ -29,6 +29,8 @@ OUT=testprojinfo_out
 
 rm -f ${OUT}
 
+export PROJINFO_NO_GRID_CHECK=YES
```
kbevers, line 32: Is this a new env var? If so: Docs, please.

rouault: This is just for testing purposes, and shouldn't be advertized. The goal is to avoid the output of test/cli/testprojinfo to depend on whether grids are installed or not.

kbevers: Okay. Ideally we should avoid this sort of hack and fix the issue properly. What would it take to that? Not saying that this necessarily should be done now, I am just curious if there's a better solution.

rouault: I can't think of a better solution for now. Now that projinfo outputs whether grids are missing, this all depends on what you've in your data/ directory. For example travis/install.sh runs the tests with and without the proj-datumgrid package.

kbevers: > I can't think of a better solution for now

Alright, then this is how it will be done :)

> For example travis/install.sh runs the tests with and without the proj-datumgrid package.

It probably is about time that we stop doing that. We basically have a hard dependency on the datumgrid package so why jump through a bunch of hoops to make thing work without it? A change for after the 6.0 release I would say.

rouault: Having proj-datumgrid a hard dependency (at least for our CI) would not necessarily solve all issues, since testprojinfo could test transformations involving other grids, but for your local testing, you may have them available, whereas the CI wouldn't have them.

kbevers: Alright, I see. Thanks for the explanation.

kbevers (approved):
Looks good to me

rouault:
Merged, and backported to 6.0

Issue #523, opened 2017-06-07
Parameter +no_rot broken for +proj=omerc


poidl:
Hi, I just started using proj

`Rel. 4.9.3, 15 August 2016`

for numerical ocean modeling, and want to create oblique Mercator projections that align with "tilted" coastlines, i.e. coastlines that are neither oriented N-S nor W-E. The following command

`proj +proj=omerc +alpha=-45.00000 +lat_0=50.00000 +lonc=-141.00000`

works as expected to align the central line with e.g. the North American west coast. But if I want to look at the rotated frame using this (identical to above but with `+no_rot`)

`proj +proj=omerc +alpha=-45.00000 +no_rot +lat_0=50.00000 +lonc=-141.00000`

the picture is flipped along a line orthogonal to the central line. For example, the North American west coast is "horizontal" and the North American continent is in the upper part of the picture and the North-East Pacific in the lower part. This is all as expected, but the problem is that true North points to the right of the picture, while it should point to the left.

Do you have any suggestions how to fix this?

There is a follow-up question, in case this is not easy to fix: Do I understand correctly that it's easy to do the `+no_rot` part myself by just tossing the x,y coordinates of the output **without** `+no_rot` through a simple rotation matrix? If that's the case, then this issue might be low(er)-priority.

Thanks a lot for your help!
![figure_grid_rotation_tutorial_omerc_proj](https://user-images.githubusercontent.com/1216556/26868105-254c5aa4-4bab-11e7-9901-810c46d75079.png)

kbevers:
[This old post from the mailing list](http://lists.maptools.org/pipermail/proj/2013-February/006575.html) suggest that you use `+gamma=0.0` instead of `+no_rot`. Please try that and confirm if it gives you the expected results

kbevers:
Also, this bit of code from the `omerc` forward calculation suggest that the axes have been switched in the `+no_rot` case:

```
    if (Q->no_rot) {
        xy.x = u;
        xy.y = v;
    } else {
        u -= Q->u_0;
        xy.x = v * Q->cosrot + u * Q->sinrot;
        xy.y = u * Q->cosrot - v * Q->sinrot;
    }
``` 

I looked very briefly for a mathematical description of the projection but couldn't easily find an answer. The above mail list post suggests that there are some confusion about the intermediate coordinate system being left- or right-handed.

poidl:
Thanks for your help! I tried

[...]

PR #1281, opened 2019-02-22
Some improvements to the docs for tmerc and omerc

busstoptaktik:
Over at #523, @ralphtee gave some very nice explanations for omerc. I have added some of this material + a some descriptional text to the docs for omerc.

Additionally clarified a few things about algorithms in the docs for tmerc.


kbevers (commented):
Overall this is very good. I would suggest only keeping the first paragraph ("The Oblique Mercator projection is a cylindrical map projection that closes the
gap between the Mercator and the Transverse Mercator projections") above the classification table and moving the rest of the intro-text below the table. This conforms to the style that is used in most other properly documented projection as well as giving a quick overview of the table without scrolling down the page.

There's some good figures in #523. I don't know the source of the figures but if we are allowed to include them (or reproductions of them) I think they would help illustrate some of the points made about he different types of rotations that can be applied. It's probably a job for another day but at least it is worth considering this as a future improvement.

docs/source/operations/projections/omerc.rst
```diff
@@ -4,13 +4,99 @@
...
+The tilt angle (azimuth) of the central line can be given in two different ways.
+In the first case, the azimuth is given directly, using the option :option:`+alpha`
+and defining the centre of projection using the options :option:`+lonc` and
+:option:`+lat_0`.
+In the second case, the azimuth is given indirectly by specifying two points on
+the central line, using the options
+:option:`+lat_1`, :option:`+lon_1`, :option:`+lat_2`, and :option:`+lon_2`.
+
+Example: Verify that the Mercator projection is a limiting form of the Oblique
+Mercator
```
kbevers, line 75: Add `::` to the end of this line to get proper formatting of the example below

docs/source/operations/projections/omerc.rst
```diff
@@ -4,13 +4,99 @@
...
+
+    $ echo 12 55 | proj +proj=merc +ellps=GRS80
+    1335833.89   7326837.71
+
+    $ echo 12 55 | proj +proj=omerc +lonc=0 +alpha=90 +ellps=GRS80
+    1335833.89   7326837.71
+
+Example: Second case - indirectly given azimuth
```
kbevers, line 85: Add `::` to the end of this line to get proper formatting of the example below

docs/source/operations/projections/omerc.rst
```diff
@@ -4,13 +4,99 @@
...
+
+    $ echo 12 55 | proj +proj=omerc +lon_1=-1 +lat_1=1 +lon_2=0 +lat_2=0 +ellps=GRS80
+      349567.57   6839490.50
+
+
+Example: An approximation of the Danish "System 34" from :cite:`Rittri2012`
```
kbevers, line 93: Add `::` to the end of this line to get proper formatting of the example below

docs/source/operations/projections/omerc.rst
```diff
@@ -62,7 +148,10 @@ Optional
 
 .. option:: +no_rot
 
-    Do not rotate axis.
+    No rectification (not "no rotation" as one may well assume).
```
kbevers, line 157: Consider using a note or warning (`.. note:: this is a note`) for the disclaimer to the `+no_rot` option. This makes it more visible.

kbevers:
The updates to `omerc` and `tmerc` are not related, right? In that case they would be better placed in separate commits instead of the same.

busstoptaktik:
> The updates to `omerc` and `tmerc` are not related, right?

Not as related as I expected when starting the work: While omerc clearly being mostly in need of a touch up, I thought of it as an occasion for a general overhaul of omerc, tmerc, and merc, giving an overall better feeling that they are in close family, although looking much different.

I think I got part of that message through in the intro to omerc, but I ran short of time before getting through everything.

So, yes, it was the idea to make changes logically tied across all 3 projections, but that is probably not terribly evident from this PR.

docs/source/references.bib
```diff
@@ -197,6 +197,27 @@ @TechReport{ONeilLaubscher1976
...
+  volume       = {6},
+  isbn         = {87-7866-085-8},
+  pages        = {63}
+}
+
+@Misc{Rittri2012,
+    author = {Mikael Rittri},
+    title  = {New omerc approximations of Denmark System 34},
+    year   = {2012},
+    url    = {http://lists.maptools.org/pipermail/proj/2012-June/006323.html}
```
rouault: can you change this to https://lists.osgeo.org/pipermail/proj/2012-June/005926.html  , since maptools.org is probably going to be retired at some point

busstoptaktik: Thanks for noticing this - I had planned to change it, but forgot about it

Issue #1269, opened 2019-02-15
Compile warnings from Intel compilers


mwtoews:
With Linux Intel compilers, a few warnings are raised:
```
$ cmake -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc ..
$ make
...
In file included from /tmp/src/proj.4/src/jniproj.cpp(45):
/tmp/src/proj.4/src/proj_config.h(32): warning #47: incompatible redefinition of macro "HAVE_PTHREAD_MUTEX_RECURSIVE"
  #define HAVE_PTHREAD_MUTEX_RECURSIVE /**/
          ^
...
In file included from /tmp/src/proj.4/src/mutex.cpp(43):
/tmp/src/proj.4/src/proj_config.h(32): warning #47: incompatible redefinition of macro "HAVE_PTHREAD_MUTEX_RECURSIVE"
  #define HAVE_PTHREAD_MUTEX_RECURSIVE /**/

...
In file included from /tmp/src/proj.4/src/strtod.cpp(41):
/tmp/src/proj.4/src/proj_config.h(32): warning #47: incompatible redefinition of macro "HAVE_PTHREAD_MUTEX_RECURSIVE"
  #define HAVE_PTHREAD_MUTEX_RECURSIVE /**/
...
/tmp/src/proj.4/src/iso19111/io.cpp(289): warning #3924: attribute namespace "clang" is unrecognized
          PROJ_FALLTHROUGH
          ^
...
/tmp/src/proj.4/src/iso19111/io.cpp(297): warning #3924: attribute namespace "clang" is unrecognized
          PROJ_FALLTHROUGH
          ^
...
```
No errors are raised, and all tests pass!

---

With GCC 7.1 on the same system, some similar warnings are also raised:
```
...
In file included from /tmp/src/proj.4/src/jniproj.cpp:45:0:
/tmp/src/proj.4/src/proj_config.h:32:0: warning: "HAVE_PTHREAD_MUTEX_RECURSIVE" redefined
 #define HAVE_PTHREAD_MUTEX_RECURSIVE /**/
 
<command-line>:0:0: note: this is the location of the previous definition
[ 20%] Building CXX object src/CMakeFiles/proj.dir/mutex.cpp.o
In file included from /tmp/src/proj.4/src/mutex.cpp:43:0:
[...]

PR #1282, opened 2019-02-22
CMake warning about removing second proj_config.h

mwtoews:
As discovered / discussed in #1269

The behaviour here is to first warn about removing a second `proj_config.h` file, suggesting also run `make distclean`, then remove the file. This file may not be removed successfully in strange scenarios where the file permissions to the source directory are different.

### DIFF
--- a/cmake/ProjConfig.cmake
+++ b/cmake/ProjConfig.cmake
@@ -37,6 +37,14 @@ set(PACKAGE_TARNAME "proj")
 set(PACKAGE_URL "https://proj4.org")
 set(PACKAGE_VERSION "${${PROJECT_INTERN_NAME}_VERSION}")
 
-configure_file(cmake/proj_config.cmake.in src/proj_config.h)
+# check if a second proj_config.h exists (created by ./configure)
+# as this is within CMake's C_INCLUDES / CXX_INCLUDES
+set(AUTOCONF_PROJ_CONFIG_H "${CMAKE_SOURCE_DIR}/src/proj_config.h")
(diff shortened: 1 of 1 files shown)


mwtoews:
I've backed out the "remove" part, and just provide a warning. This is a safer option to pick, leaving the user to clean up the situation.

kbevers:
Will the warning be immediately visible for the user or is this one of those things that will get buried by thousands out lines of text on STDOUT?

mwtoews:
It's in the `cmake ..` step, which isn't too long
```
$ cmake ..
...
-- Looking for ceil in m
-- Looking for ceil in m - found
CMake Warning at cmake/ProjConfig.cmake:44 (message):
  Autoconf's /tmp/src/proj.4/src/proj_config.h may
  interfere with this CMake build.  Run 'make distclean' in the source
  directory before CMake's build.
  CMakeLists.txt:82 (include)


-- Found Sqlite3: /usr/lib/x86_64-linux-gnu/libsqlite3.so
-- Looking for pthread.h
-- Looking for pthread.h - found
...
```
also it's sent to stderr, so it's pretty much the only thing you see with `cmake .. > /dev/null`

kbevers:
> so it's pretty much the only thing you see with `cmake .. > /dev/null`

Yeah, but who does that? :)

I think this is an acceptable solution to a problem that only occur under rare circumstances. Thanks for figuring out the problem at providing a solution!

kbevers:
Cherry-picked into `6.0`

PR #1285, opened 2019-02-23
Doc: add minimal documentation for patterson, compact miller, misrsom, natearth2, sch and times projections

rouault:


docs/source/operations/projections/patterson.rst
```diff
@@ -0,0 +1,46 @@
+.. _patterson:
+
+********************************************************************************
+Patterson
+********************************************************************************
+
+The Patterson projection is a new cylindrical map projection designed for
```
kbevers: Drop the "new" - According to the reference it is at this time already 4 years old. I don't know when the shift from new to old/established happens but I am certain that we won't remember to adjust the documentation when it happens :-) Let's just stick to the facts here.

rouault: Done

rouault:
I believe that now all projection methods implemented have some documentation !

Based on comparing the output of
```
for i in src/projections/*.cpp; do echo `basename $i .cpp`; done | sort > /tmp/proj.txt
for i in docs/source/operations/projections/*.rst; do echo `basename $i .rst`; done | sort > /tmp/proj_doc.txt
```
with manual checking since some .cpp files are actually a collection of projection methods

busstoptaktik:
> I believe that now all projection methods implemented have some documentation !

Awesome. @kbevers: Wouldn't this be reason enough to cut an RC3 including this PR? It's a beautiful, end user significant landmark achievement which would be worth celebrating in a major release, rather than delayed to 6.1!

kbevers:
> Awesome. @kbevers: Wouldn't this be reason enough to cut an RC3 including this PR? It's a beautiful, end user significant landmark achievement which would be worth celebrating in a major release, rather than delayed to 6.1!

The docs actually doesn't go into the source distribution, so I don't think there's any reason to make a third release candidate based on this. But it should be merged and cherry-picked to the 6.0 branch so that it is published on the website when 6.0 is released. The same goes for your tmerc/omerc doc improvements.

rouault:
I don't think this is worth a RC3 either, otherwise we'll never release 6.0 :-)

busstoptaktik:
> I don't think this is worth a RC3 either, otherwise we'll never release 6.0 :-)

Good point - and certainly you ,@rouault  have plenty of other enormous contributions for this release to celebrate, and my suggestion for another RC was certainly not to belittle those.

PR #1288, opened 2019-02-24
Fix data path used by cli tests

QuLogic:
All other tests use `PROJ_LIB`, and allow it to be overridden from the command-line, so do the same here.


kbevers:
Thanks for streamlining this!

PR #1295, opened 2019-02-27
configure.ac: replace '==' by '=' to avoid Bashism (patch by Greg Troxel)

rouault:


kbevers:
So... Another RC?

rouault:
No, I don't think that's worth it. That's just a warning in Greg's linting tools, and he patched it on his side. But we can backport this to 6.0 branch for the next bugfix release

kbevers:
Okay, great. I think we've had enough RC's for now :) I'll call for a vote for of RC4 promotion to final today then.

rouault:
Merged and cherry-picked to 6.0

PR #1298, opened 2019-02-27
Fix a couple issues with default context fallback

QuLogic:
The first is this null pointer dereference warning:
```
open_lib.cpp: In function 'int* pj_open_lib_ex(projCtx, const char*, const char*, char*, size_t)':
open_lib.cpp:263:18: warning: null pointer dereference [-Wnull-dereference]
  263 |         if( ctx->last_errno == 0 && errno != 0 )
      |             ~~~~~^~~~~~~~~~
```
g++ is confused because of all the other checks for `ctx != nullptr`, and thinks that this one line should also have one, but [either `ctx != nullptr` or `ctx = pj_get_default_ctx()`](https://github.com/OSGeo/proj.4/blob/38f7a68edbb0a5d4a64d98b8297f4281a5d400d0/src/open_lib.cpp#L172-L174) so there really should be no checks at all.

Second, the fallback in `proj_log_func` didn't do anything since it wasn't re-assigning the variable.

### DIFF
--- a/src/internal.cpp
+++ b/src/internal.cpp
@@ -491,9 +491,7 @@ void proj_log_func (PJ_CONTEXT *ctx, void *app_data, PJ_LOG_FUNCTION logf) {
     passed as first arg at each call to the logger
 ******************************************************************************/
     if (nullptr==ctx)
-        pj_get_default_ctx ();
-    if (nullptr==ctx)
-        return;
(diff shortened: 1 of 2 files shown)


rouault:
Merged and cherry-picked in 6.0. Thanks

PR #1300, opened 2019-02-27
Miscellaneous documentation fixes

mwtoews:
Notably a few corrections to the BibTeX references, which is fussy format regarding title case, name formats, and other subtle details. Some years were corrected (based on the publisher's info), or volumes/numbers/pages added. The .bib file was normalized (re-saved) with JabRef, which sorts these by last name.

### DIFF
--- a/docs/source/about.rst
+++ b/docs/source/about.rst
@@ -23,22 +23,9 @@ datums using all but the most obscure geodetic techniques.
 Citation
 -------------------------------------------------------------------------------
 
-To cite PROJ in publications use:
 
-  PROJ contributors (2018). PROJ coordinate transformation software
-  library. Open Source Geospatial Foundation. URL https://proj4.org/.
+.. include:: ../../CITATION
 
-A BibTeX entry for LaTeX users is
-
-.. code-block:: latex
-
-  @Manual{,
-    title = {{PROJ} coordinate transformation software library},
-    author = {{PROJ contributors}},
-    organization = {Open Source Geospatial Foundation},
-    year = {2018},
-    url = {https://proj4.org/},
-  }
 
 License
 -------------------------------------------------------------------------------
--- a/docs/source/references.bib
+++ b/docs/source/references.bib
@@ -3,7 +3,7 @@
 
 
 @Article{Altamimi2002,
-  Title                    = {ITRF2000: A new release of the International Terrestrial Reference Frame for earth science applications},
+  Title                    = {{ITRF2000}: A new release of the {International Terrestrial Reference Frame} for earth science applications},
   Author                   = {Z. Altamimi and P. Sillard and C. Boucher},
   Journal                  = {Journal of Geophysical Research: Solid Earth},
   Year                     = {2002},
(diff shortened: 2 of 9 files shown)


docs/source/about.rst
```diff
@@ -25,7 +25,7 @@ Citation
 
 To cite PROJ in publications use:
```
kbevers: This section would actually be better to just include from `CITATION`:

https://github.com/OSGeo/proj.4/blob/0a78cc67fc06d2fe1769c5582ce5b8b92b1ef4a3/CITATION#L1-L14

Maintaining this in two places is error-prone.

mwtoews: Done, and I simplified one step of HOW-TO-RELEASE

docs/source/references.bib
```diff
@@ -140,19 +132,24 @@ @Book{Helmert1880
   Doi                      = {10.5281/zenodo.32050}
 }
 
-@Article{Hensley2002,
-  Title                    = {Improved Processing of AIRSAR Data Based on the GeoSAR Processor},
-  Author                   = {Scott Hensley and Elaine Chapin and Adam Freedman and Thierry Michel},
-  Journal                  = {Airsar earth science and applications workshop},
+@InProceedings{Hensley2002,
+  Title                    = {Improved Processing of {AIRSAR} Data Based on the {GeoSAR} Processor},
+  Author                   = {S. Hensley and E. Chapin and A. Freedman and T. Michel},
```
QuLogic, line 137: I'm not sure there's any benefit to reducing the information here. Whether or not the full name appears is up to the bibliography style in the end, but I don't think there's any reason there should _not_ be full first names.

man/man1/projinfo.1
```diff
@@ -89,7 +89,7 @@ The default is crs.
 .TP
 .B \-\-summary
 When listing coordinate operations available between 2 CRS, return the
-result in a summary format, mentionning only the name of the coordinate
+result in a summary format, mentioning only the name of the coordinate
```
kbevers: No need to make changes here. The man pages are auto generated from sphinx at release time.

docs/source/references.bib
mwtoews: In regards to reducing first names to initials, I see your point. I mainly did this for consistency with the initial bibliography that just used initials. If we kept some entries with first names, it may look a bit more chaotic.

man/man1/projinfo.1
mwtoews: I figured so. Did you want this change backed out, or is it fine to leave in for now?

docs/source/references.bib
kbevers: Consistency is better than inconsistency. If reducing a few first names to initials is the simpest way to achieve consistency that is what should be done.

man/man1/projinfo.1
kbevers: If it's not too much trouble, yes. I don't like to introduce unnecessary clutter to the git history. At some point I will se if we can remove these files altogether and generate them as part of "make dist" when preparing releases.

mwtoews: Done. And I've squashed these commits into one.

docs/source/references.bib
busstoptaktik: I **strongly** disagree here: The bib-file is a database of available information, and should primarily register information **as published**. 

The bibliography style a different thing: Removing information from the database is patching the problem at the wrong end. Better, for some time have a mix of initials and full names - if we want initials only, the bib-style should enforce that.

mwtoews: Yeah, I was split on this one. The idea from ~8 months ago was to make the references section Harvard-like, which uses initials. Usually it should be up to the bib style to decide how to format this, but the current structure doesn't seem to support this. I believe this support could eventually happen, by extending parts of `docs/source/bibstyle.py`.

So I suppose the options are to either keep with hard-coded initials (for now), or mix the citation style (initials/full names). And if I do eventually tidy the bib style to format initials, then the .bib file can be revised to reflect names as published.

mwtoews: Ah, I seem to have missed that the [API does have a `abbreviate_names` option with default of False](https://pythonhosted.org/pybtex/api/styles.html). So this won't be too difficult to do after all. However, could this wait until after the release (i.e. next week)? Basically, I'd fix the citation style and go thru all the references to check the consistency with publications (where feasible).

kbevers: > However, could this wait until after the release (i.e. next week)?

Of course. Also note that the documentation isn't part of the release tarball so the deadline for inclusion of contributions in 6.0 is more fluid than for code contributions. We can merge this now and you can fix the names later, or we can wait a bit until you are ready. Whatever you prefer.

mwtoews: @kbevers yes, this is ready to merge. I'll revise names as published for the .bib file later.

rouault:
Merged into 6.0 as well

PR #1302, opened 2019-03-01
Doc: rename to ISO-19111:2019

rouault:
And publish link to corresponding promoted and public OGC doc:
http://docs.opengeospatial.org/as/18-005r4/18-005r4.html


rouault:
Merged into 6.0 as well

PR #1309, opened 2019-03-06
Doc updates, most notably rewrite of FAQ

kbevers:
Some minor doc updates and a work-in-progress rewrite of the FAQ.

Suggestions for more frequently asked questions are welcome. The once I have added so far are mostly in response to questions I often get asked in my job at the Danish geodetic office. I have removed basically all the existing questions as they were from a time where there wasn't much other documentation. Most of them were covered better in other places or was simply not applicable any more. At least in my opinion, if I have removed something you think is important please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The following command line utilities are included in the PROJ package:
 - geod, for geodesic (great circle) computations.
 - cct, for generic Coordinate Conversions and Transformations.
 - gie, the Geospatial Integrity Investigation Environment.
-
+- projinfo, for geodetic object and coordinate operation queries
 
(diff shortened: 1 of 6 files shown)


snowman2:
I think I remember something about `+init` style epsg projection support being deprecated. If this is true, documentation about that would be helpful.

snowman2:
It would also be useful to have this documented: https://github.com/pyproj4/pyproj/issues/201

kbevers:
Good suggestions. I will try to include those later today.

snowman2:
Perfect, thanks!

Issue #1311, opened 2019-03-06
gie should exit with non-zero code when it can't open a file


kbevers:
Case in point:

```
$ ./bin/gie unknown_file
-------------------------------------------------------------------------------
Cannot open spec'd input file 'unknown_file' - bye!
-------------------------------------------------------------------------------
$ echo $?
0
```

This behaviour is unfortunate if a new gie file is added to the PROJ test-suite with a faulty file name. It will just display the test as passed.

PR #1312, opened 2019-03-06
Make gie return non-zero exit code when file can't be opened

kbevers:
Simple fix to #1311

### DIFF
--- a/src/apps/gie.cpp
+++ b/src/apps/gie.cpp
@@ -300,6 +300,20 @@ int main (int argc, char **argv) {
         return 1;
     }
 
+    for (i = 0; i < o->fargc; i++ ) {
+        FILE* f = fopen (o->fargv[i], "rt");
+        if (f == nullptr) {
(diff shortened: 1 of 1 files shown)


kbevers:
Downside to this fix is that it still runs the tests for files that can be found:

```
$ ./bin/gie ../repo/test/gigs/5101.4-jhs-etmerc.gie unknown
Reading file '../repo/test/gigs/5101.4-jhs-etmerc.gie'
total: 69 tests succeeded,  0 tests skipped,  0 tests failed.
Cannot open spec'd input file 'unknown' - bye!
```

Should the input files be tested for existence before running the tests? It goes a bit against the architecture of the application (one pass does everything) but it can prevent that says that everything is okay before giving a error that seems insignificant at a glance.

kbevers:
> Should the input files be tested for existence before running the tests?

I decided to do this. Seems better to me.

Issue #1310, opened 2019-03-06
Let gie use proj_create_crs_to_crs


kbevers:
Currently `gie` can only instantiate transformations via `proj_create`. It would be nice to be ably to also use `proj_create_crs_to_crs`. My main reason for this wish is that it would allow us to properly run the GIGS tests.

There are two ways to achieve this in the `gie` command langue:

1. A command per CRS: `crs_src` and `crs_dst`
2. One command per CRS: `from_to`

Additionally an optional `area_of_use` should be possible to specify.

Option 1 would result in `gie` files looking like:

```
<gie>
crs_src	EPSG:22175
crs_dst EPSG:4190
area_of_use Argentina - 61.5°W to 58.5°W onshore

tolerance  0.03 m
accept    -63.9993433 80.0002644
expect    5422500.0 18889800.0
</gie>
```

and option 2 would look like:

```
<gie>
from_to	EPSG:22175 EPSG:4190
area_of_use Argentina - 61.5°W to 58.5°W onshore

tolerance  0.03 m
accept    -63.9993433 80.0002644
expect    5422500.0 18889800.0
</gie>
```

Option 1 has the advantage that it will be easier to accept WKT(2) CRS definitions since it is not necessary to figure out where one starts and the other ends which would be necessary with option 1.

Discuss :-)

kbevers:
This is actually fixed in #1314 but it appears that both documentation and a example gie-file is missing. I'll get that sorted before 7.0.0

PR #1314, opened 2019-03-06
Allow gie to use proj_create_crs_to_crs

kbevers:
In response to #1310 

`test/gie/crs_to_crs.gie` is added as a demonstration for now. I will add some actual tests later. The GIGS tests are a good candidate for using this new functionality.


busstoptaktik (approved):
LGTM, although funny results may happen if inadvertently specifying crs_src without crs_dst.

I suppose this is "by design", and something that may be bitched over in similar terms as when discussing the merits (or mis-merits) of the fallthrough mechanism of the C `switch/case` construct

PR #1316, opened 2019-03-11
Fix cross-compilation detection

seanwarren:
Currently, cross compilation detection requires that the variable `CMAKE_CROSSCOMPILING` in the consuming project be an exact string match for that during compilation. This can cause a problem if there is a mismatch (for example if `CMAKE_CROSSCOMPILING` is set to `FALSE` in the consuming project rather than `OFF`). 

This can cause a `find_project(PROJ4 ...)` call to fail erroneously when using the config file.

This PR fixes this issue by converting a boolean evaluation to `ON` or `OFF`. 

While this seems like a bit of an estoric edge case, this situation occurs when using `vcpkg` to build and consume proj4.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -4,6 +4,12 @@
 # be relocated.)
 file(RELATIVE_PATH PROJECT_ROOT_DIR ${CMAKE_INSTALL_PREFIX}/${CMAKECONFIGDIR} ${CMAKE_INSTALL_PREFIX})
 
+if (CMAKE_CROSSCOMPILING)
+  set (CMAKE_CROSSCOMPILING_STR "ON")
+else ()
+  set (CMAKE_CROSSCOMPILING_STR "OFF")
+endif ()
+
 string(TOLOWER "${PROJECT_NAME}" PROJECT_NAME_LOWER)
 configure_file (project-config.cmake.in project-config.cmake @ONLY)
 configure_file (project-config-version.cmake.in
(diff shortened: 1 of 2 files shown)


kbevers:
I am not a CMake expert in any way, so please bear with me. Isn't it the consuming projects problem if it isn't setting CMake variables properly? Or is this just another one of CMake's quirks that makes things harder than they need to be?

In any case I guess we better merge this if problems can be fixed by it. @mloskot do you have an opinion on this?

seanwarren:
@kbevers, I think the issue is that `CMAKE_CROSSCOMPILING` should be treated as a boolean and not string value so `FALSE` is equivalent to `OFF` (so, yes, it is one of those CMake quirks!). In fact some of CMakes internal tests set this variable to TRUE/FALSE rather than ON/OFF.  

However as this is pretty obscure case I can equally submit this as a patch to vcpkg if you'd prefer.

kbevers:
> In fact some of CMakes internal tests set this variable to TRUE/FALSE rather than ON/OFF.

Well, if CMake itself can't even make up its mind I think we have to use your work around

> However as this is pretty obscure case I can equally submit this as a patch to vcpkg if you'd prefer.

That seems like the correct thing to do, but it will take a few release cycles before that can be relied on, so applying the fix in both places is probably the safer option for now.

kbevers:
Thanks @seanwarren!

seanwarren:
Great, thanks for the quick merge

PR #1318, opened 2019-03-14
Reject eccentricity values larger than one

rouault:
Valid eccentricity should be between 0 (included) or 1 (excluded)

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13665
Credit to OSS Fuzz


kbevers (approved):
Looks good to me

PR #1319, opened 2019-03-14
Molodensky: avoid division by zero

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13069
Credit to OSS Fuzz

### DIFF
--- a/src/transformations/molodensky.cpp
+++ b/src/transformations/molodensky.cpp
@@ -146,10 +146,20 @@ static PJ_LPZ calc_standard_params(PJ_LPZ lpz, PJ *P) {
     dphi  = (-dx*sphi*clam) - (dy*sphi*slam) + (dz*cphi)
             + ((nu * P->es * sphi * cphi * da) / a)
             + (sphi*cphi * ( rho/(1-f) + nu*(1-f))*df);
-    dphi /= (rho + lpz.z);
+    const double dphi_denom = rho + lpz.z;
+    if( dphi_denom == 0.0 ) {
+        lpz.lam = HUGE_VAL;
+        return lpz;
+    }
+    dphi /= dphi_denom;
 
     /* delta lambda */
-    dlam = (-dx*slam + dy*clam) / ((nu+lpz.z)*cphi);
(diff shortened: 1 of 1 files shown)


src/transformations/molodensky.cpp
```diff
@@ -146,10 +146,20 @@ static PJ_LPZ calc_standard_params(PJ_LPZ lpz, PJ *P) {
...
+        lpz.phi = HUGE_VAL;
+        return lpz;
+    }
+    dphi /= dphi_denom;
 
     /* delta lambda */
-    dlam = (-dx*slam + dy*clam) / ((nu+lpz.z)*cphi);
+    const double dlam_denom = (nu+lpz.z)*cphi;
+    if( dlam_denom == 0.0 ) {
+        lpz.phi = HUGE_VAL;
```
kbevers: It probably doesn't matter in the end, but shouldn't it rather be `lpz.lam = HUGE_VAL` here?

src/transformations/molodensky.cpp
```diff
@@ -230,6 +245,13 @@ static PJ_XYZ forward_3d(PJ_LPZ lpz, PJ *P) {
     } else {
         lpz = calc_standard_params(lpz, P);
     }
+    if( lpz.phi == HUGE_VAL ) {
+        proj_errno_set(P, PJD_ERR_TOLERANCE_CONDITION);
+        point.xyz.x = HUGE_VAL;
+        point.xyz.y = HUGE_VAL;
+        point.xyz.z = HUGE_VAL;
+        return point.xyz;
```
kbevers: You could simplify this a bit by just going `return proj_coord_error()`. The same below.

For reference: https://github.com/OSGeo/proj.4/blob/ea2d6723856036060b067d7b43fbec23003da64c/src/internal.cpp#L65-L69

src/transformations/molodensky.cpp
rouault: Done

src/transformations/molodensky.cpp
rouault: Done

PR #1322, opened 2019-03-16
Various ossfuzz fixes

rouault:


src/strerrno.cpp
```diff
@@ -30,7 +30,7 @@ pj_err_list[] = {
     "acos/asin: |arg| >1.+1e-14",                                      /* -19 */
     "tolerance condition error",                                       /* -20 */
     "conic lat_1 = -lat_2",                                            /* -21 */
-    "lat_1 >= 90",                                                     /* -22 */
+    "lat >= 90",                                                       /* -22 */
```
kbevers: Maybe this would be better as "lat_1 or lat_2 >= 90"? "lat" can be mistaken for the input latitude.

src/transformations/helmert.cpp
```diff
@@ -349,6 +349,13 @@ static PJ_LP helmert_reverse (PJ_XY xy, PJ *P) {
     double x, y, sr, cr;
     point.xy = xy;
 
+    if( Q->scale == 0.0 ) {
+        /* In practice can only happen if calling the 2d interface on a */
```
kbevers: I have run into problems with this elsewhere as well. I think a better option to avoid issues caused by the 2D functions when using what is clearly a 3D helmert, is to only enable the 2D-functions when those are needed. Something like turning this

https://github.com/OSGeo/proj.4/blob/3cce54767afe77412b850d2d88ee168370570b6f/src/transformations/helmert.cpp#L566-L577

into this

```
    /* In the 2D case, the coordinates are projected */
    if (pj_param_exists (P->params, "theta")) {
        P->left  = PJ_IO_UNITS_PROJECTED;
        P->fwd    = helmert_forward;
   	    P->inv    = helmert_reverse;
    }

    P->fwd4d  = helmert_forward_4d;
    P->inv4d  = helmert_reverse_4d;
    P->fwd3d  = helmert_forward_3d;
    P->inv3d  = helmert_reverse_3d;
```

It may be necessary to add "fake" 2D-functions for the 3D helmert as well:

```
static PJ_XY helmert_fake_forward_2d(PJ_LP lp, PJ *P) {
	PJ_COORD coord;
	coord.lp = lp;
	coord.xyz = helmert_forward_3d(coord.lpz, P);
	return coord.xy;
}
```

kbevers (commented):
Nice to get these bugs out of the way! I've added a couple of inline comments with suggestions for improvement.

src/strerrno.cpp
rouault: I wanted this to be generic, and potentially be used for lat_0 as well. So 'lat_0, lat_1 or lat_2 >= 90' ?

kbevers:
> (I'm not sure why Travis-CI and AppVeyor builds aren't launched on this pull request. Seems also the case of #1309. Probably a github issue)

#1309 sneakily includes a `[skip ci]` in the commit message of the last commit :-)

But #1324 also hasn't started CI builds.

src/transformations/helmert.cpp
rouault: I just tried moving helmert_forward and helmert_reverse to the `if (pj_param_exists (P->params, "theta"))` case and this causes a tail failure in 4D-API_cs2cs-style.gie line 189, which is a pipeline with a cart + helmert + inv cart, with 2D input coordinates, which causes the pj_approx_2D_trans() function to be used, and ignore the Z component. I'm not sure how it could work before but dropping the Z in cart/inv_cart by doing 2D only operations is absolutely non-sense ! I'm not feeling like I want to do that helmert_fake_forward_2d() since 7- or 15- Helmert really operate in the cartesian space. So I'm adjusting the gie test to use 3D coordinates

src/strerrno.cpp
rouault: agreed

src/transformations/helmert.cpp
kbevers: It was also in gie I ran into this problem and like you I found `pj_approx_2D_trans()` to be the culprit.

My issue is with GIGS test that wants to use 3D helmert on 2D input and output which gie has problems with currently. I believe cct and cs2cs is doing the correct thing. I didn't find a good solution at the time but thought that it might work with the fake 2D operations but yeah I can see how that wouldn't work. I guess the work around for now is to add zeros as the z component. I think the correct fix is to make all operations be 4D. It will make many things simpler but it is daunting task!

kbevers:
Note to self when writing release notes: These are six individual bug fixes and should be documented as such.

PR #1328, opened 2019-03-17
Fix some issues raised by latest cppcheck

rouault:
- coordinateoperation_internal.hpp: missing 'explicit' keyword
- proj.cpp: unused 'generic' member in enumeration
- init.cpp: useless assignment to a_orig and es_orig, because done again a
            few lines below.
- crs.cpp: unused variable
- datum.cpp: inefficient use of find() function
- io.cpp:
  * missing 'static' qualifier for method
  * useles ternary test (left and right have same value)
- aeqd.cpp: useless assignment of inv and fwd, snice done again a few lines
            below
- isea.cpp: useless assignment of resolution and aperture since done again a
            few lines below, and with default values when params are absent
- mod_ster.cpp: useless assignment of lp.lam, overriden in below code paths.
- stere.cpp: false positive, but better not modify another variable than the
             iterator in a for() loop.

### DIFF
--- a/include/proj/internal/coordinateoperation_internal.hpp
+++ b/include/proj/internal/coordinateoperation_internal.hpp
@@ -253,7 +253,7 @@ class PROJBasedOperation : public SingleOperation {
     gridsNeeded(const io::DatabaseContextPtr &databaseContext) const override;
 
   protected:
-    PROJBasedOperation(const OperationMethodNNPtr &methodIn);
+    explicit PROJBasedOperation(const OperationMethodNNPtr &methodIn);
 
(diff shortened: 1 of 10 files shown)


kbevers:
Should I regard this as a bugfix that needs backporting?

rouault:
> Should I regard this as a bugfix that needs backporting?

not necessarily, that shouldn't change behaviour. This is mostly cleanup

kbevers:
Okay, let's just bring it into master then. Just looking for a chance to use the new backporting facility :-)

PR #1333, opened 2019-03-19
proj_create_crs_to_crs: better deal with coordinates outside of bbox (fixes #1329)

### LINKED ISSUE #1329: cs2cs fails to transform coordinates outside area of use
Sorry for the bad title, I haven't been able to narrow the problem down to something more specific.

Let's have a look at transformations between EPSG:4275 and EPSG:4807:

```
$ ./projinfo -s EPSG:4275 -t EPSG:4807 -o PROJ
Candidate operations found: 2
Operation n°1:

INVERSE(EPSG):1763, Inverse of NTF (Paris) to NTF (1), 0 m, France - onshore - mainland and Corsica

PROJ string:
+proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=longlat +ellps=clrk80ign +pm=paris +step +proj=unitconvert +xy_in=rad +xy_out=grad +step +proj=axisswap +order=2,1

Operation n°2:

INVERSE(EPSG):1764, Inverse of NTF (Paris) to NTF (2), 0 m, France - onshore - mainland and Corsica

PROJ string:
+proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +inv +proj=longlat +ellps=clrk80ign +pm=-2.33720833333333 +step +proj=unitconvert +xy_in=rad +xy_out=grad +step +proj=axisswap +order=2,1
```

Two possible transformations, all good. Let's try:

```
$ echo 58 5 0 | cs2cs EPSG:4275 EPSG:4807
*	* inf
```

Whoops. But cct to the rescue:

```
$ echo 58 5 0 0 |../../../cmake/bin/cct +proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=longlat +ellps=clrk80ign +pm=paris +step +proj=unitconvert +xy_in=rad +xy_out=grad +step +proj=axisswap +order=2,1
      64.4444         2.9586        0.0000        0.0000
```

I don't really understand what is going on here. The accuracy of the transformation is "0", maybe that has something to do with this?

I see something similar with transformations between EPSG:4267 and EPSG:4269 but there's a lot more going on in that case so probably not the best to use for debugging.

rouault:
In case several coordinate operations are returned for a CRS to CRS transformation,
we currently determine the one to use by selecting the first operation whose
bounding box contains the input point.

This commit adds a fallback case where after doing that first iteration and finding
no appropriate candidate, we try again by selecting the first operation available
that does not involve grid based transformations.

### DIFF
--- a/src/4D_api.cpp
+++ b/src/4D_api.cpp
@@ -51,6 +51,7 @@
 #include "proj/common.hpp"
 #include "proj/coordinateoperation.hpp"
 #include "proj/internal/internal.hpp"
+#include "proj/internal/io_internal.hpp"
 
 using namespace NS_PROJ::internal;
 
@@ -192,6 +193,8 @@ similarly, but prefers the 2D resp. 3D interfaces if available.
         direction = opposite_direction(direction);
 
     if( !P->alternativeCoordinateOperations.empty() ) {
+        // Do a first pass and select the first coordinate operation whose area
+        // of use is compatible with the input coordinate
         int i = 0;
         for( const auto &alt: P->alternativeCoordinateOperations ) {
             if( direction == PJ_FWD ) {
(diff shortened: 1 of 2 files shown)


kbevers:
This seems like the most sensible thing to do. Thanks for taking care of this.

PR #1336, opened 2019-03-20
Build: automatically enable system error messages

cjmayo:
Define HAVE_STRERROR from configure.

Before:
```
$ cs2cs +proj=latlong +to +proj=latlong dummy
<cs2cs>:
Sys errno: 2: <system mess. texts unavail.>
dummy
```
After:
```
$ cs2cs +proj=latlong +to +proj=latlong dummy
<cs2cs>:
Sys errno: 2: No such file or directory
dummy
```
---

Tested against 6.0.0, and 5.2.0.


kbevers:
Nice catch. Please apply the same fix to the CMake setup as well, so that build with either system works the same.

cjmayo:
Added CMake support.

kbevers:
> Added CMake support.

Thank you very much!

PR #1337, opened 2019-03-20
Port custom ellipsoid definitions from QGIS

nyalldawson:
As discussed on the mailing list, this PR introduces a number of custom ellipsoid definitions under the PROJ authority. Most of these match the current output from proj -le, but the Earth2000 ellipsoid is ported from QGIS and of unknown origin.

Sponsored by ICSM

### DIFF
--- a/data/sql/customizations.sql
+++ b/data/sql/customizations.sql
@@ -36,3 +36,19 @@ INSERT INTO authority_to_authority_preference(source_auth_name, target_auth_name
 
 INSERT INTO authority_to_authority_preference(source_auth_name, target_auth_name, allowed_authorities) VALUES
     ('ESRI', 'EPSG', 'PROJ,ESRI,EPSG' );
+
+-- Custom ellipsoids (from proj -le)
+
+INSERT INTO "ellipsoid" VALUES('PROJ','ANDREA','Andrae 1876 (Den., Iclnd.)',NULL,'PROJ','EARTH',6377104.43,'EPSG','9001',300.0,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','CPM','Comm. des Poids et Mesures 1799',NULL,'PROJ','EARTH',6375738.7,'EPSG','9001',334.29,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','DELMBR','Delambre 1810 (Belgium)',NULL,'PROJ','EARTH',6376428.0,'EPSG','9001',311.5,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','KAULA','Kaula 1961',NULL,'PROJ','EARTH',6378163.0,'EPSG','9001',298.24,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','LERCH','Lerch 1979',NULL,'PROJ','EARTH',6378139.0,'EPSG','9001',298.257,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','MERIT','MERIT 1983',NULL,'PROJ','EARTH',6378137.0,'EPSG','9001',298.257,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','MPRTS','Maupertius 1738',NULL,'PROJ','EARTH',6397300.0,'EPSG','9001',191.0,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','NEW_INTL','New International 1967',NULL,'PROJ','EARTH',6378157.5,'EPSG','9001',NULL,6356772.2,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','WGS60','WGS 60',NULL,'PROJ','EARTH',6378165.0,'EPSG','9001',298.3,NULL,0);
+
+-- Custom ellipsoids (ported from QGIS - original source unknown)
+
+INSERT INTO "ellipsoid" VALUES('PROJ','EARTH2000','Earth2000',NULL,'PROJ','EARTH',6378140.0,'EPSG','9001',NULL,6356750.0,0);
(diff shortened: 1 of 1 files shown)


data/sql/customizations.sql
```diff
@@ -35,4 +35,20 @@ INSERT INTO authority_to_authority_preference(source_auth_name, target_auth_name
...
+INSERT INTO "ellipsoid" VALUES('PROJ','CPM','Comm. des Poids et Mesures 1799',NULL,'PROJ','EARTH',6375738.7,'EPSG','9001',334.29,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','DELMBR','Delambre 1810 (Belgium)',NULL,'PROJ','EARTH',6376428.0,'EPSG','9001',311.5,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','KAULA','Kaula 1961',NULL,'PROJ','EARTH',6378163.0,'EPSG','9001',298.24,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','LERCH','Lerch 1979',NULL,'PROJ','EARTH',6378139.0,'EPSG','9001',298.257,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','MERIT','MERIT 1983',NULL,'PROJ','EARTH',6378137.0,'EPSG','9001',298.257,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','MPRTS','Maupertius 1738',NULL,'PROJ','EARTH',6397300.0,'EPSG','9001',191.0,NULL,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','NEW_INTL','New International 1967',NULL,'PROJ','EARTH',6378157.5,'EPSG','9001',NULL,6356772.2,0);
+INSERT INTO "ellipsoid" VALUES('PROJ','WGS60','WGS 60',NULL,'PROJ','EARTH',6378165.0,'EPSG','9001',298.3,NULL,0);
+
+-- Custom ellipsoids (ported from QGIS - original source unknown)
```
rouault: According to https://github.com/qgis/QGIS/blame/master/resources/iau2000.sql , that would come from the IAU2000 dictionary.
@thareUSGS I guess IAU2000 will be injected some time into the PROJ database ?
In the mean time, we can probably keep it under the PROJ authority, and remove it once IAU2000 is integrated.

data/sql/customizations.sql
```diff
@@ -35,4 +35,20 @@ INSERT INTO authority_to_authority_preference(source_auth_name, target_auth_name
     ('IGNF', 'EPSG', 'PROJ,IGNF,EPSG' );
 
 INSERT INTO authority_to_authority_preference(source_auth_name, target_auth_name, allowed_authorities) VALUES
-    ('ESRI', 'EPSG', 'PROJ,ESRI,EPSG' );
\ No newline at end of file
+    ('ESRI', 'EPSG', 'PROJ,ESRI,EPSG' );
+
+-- Custom ellipsoids (from proj -le)
+
+INSERT INTO "ellipsoid" VALUES('PROJ','ANDREA','Andrae 1876 (Den., Iclnd.)',NULL,'PROJ','EARTH',6377104.43,'EPSG','9001',300.0,NULL,0);
```
busstoptaktik: ANDREA should be ANDRAE (or, in the ideal Unicode-everywhere world, "ANDRÆ")

rouault: For the sake of simplicity, let's stick to ASCII character set for identifiers...

data/sql/customizations.sql
rouault: @nyalldawson Just point that it comes from the IAU2000 dictionary. (you may also entirely remove that entry since I doubt anyone would use such a generic ellipsoid for Earth)

data/sql/customizations.sql
kbevers: While we're at it, why not expand the countries names fully? `Denmark, Iceland`. We can afford the few extra characters :-)

data/sql/customizations.sql
```diff
@@ -35,4 +35,20 @@ INSERT INTO authority_to_authority_preference(source_auth_name, target_auth_name
...
+INSERT INTO "ellipsoid" VALUES('PROJ','CPM','Comm. des Poids et Mesures 1799',NULL,'PROJ','EARTH',6375738.7,'EPSG','9001',334.29,NULL,0);
```
kbevers: Similar to above: `Commission des Poids et Mesures`. Or maybe `Comité international des poids et mesures` as [French wikipedia suggests](https://fr.wikipedia.org/wiki/Comit%C3%A9_international_des_poids_et_mesures)? Final say goes to our resident Frenchman @rouault :-)

rouault: I don't care honestly and don't want to harass Nyall with nitpicking. If it was only me, I wouldn't have included at all those esoteric and probably unused/deprecated ellipsoids...

busstoptaktik: > probably unused/deprecated ellipsoids...

The Andræ ellipsoid is still in use for interoperability with historical registers especially for a few Icelandic systems. So not entirely unused, but certainly esoteric.

data/sql/customizations.sql
busstoptaktik: Also, I think the modified Andræ (a.k.a. "Danish") is missing here:

    danish a=6377019.2563   rf=300.0         Andrae 1876 (Denmark, Iceland)

Included since PROJ 5.(0?)

Based on a revised conversion factor between fathoms and meters, I believe - I'll have to look it up, but cannot remember from where I got that material.

nyalldawson: > While we're at it, why not expand the countries names fully? `Denmark, Iceland`. We can afford the few extra characters :-)

Done -- to be honest, I didn't even realise that was what Den. represented !

nyalldawson: > Also, I think the modified Andræ (a.k.a. "Danish") is missing here:
> 
> ```
> danish a=6377019.2563   rf=300.0         Andrae 1876 (Denmark, Iceland)

It's already present in the database, from the EPSG db -- EPSG:7051, "Danish 1876"

nyalldawson:
Ok, comments addressed

kbevers:
> Ok, comments addressed

Thanks, Nyall!

PR #1338, opened 2019-03-20
Add -k ellipsoid option to projinfo

nyalldawson:
Allows querying of ellipsoid definitions by auth:code lookup

Without this option it is not possible to query ellipsoids by auth:code: e.g, `projinfo  EPSG:7001` fails. With this change it's possible to lookup via `projinfo -k ellipsoid EPSG:7001`

Sponsored by ICSM

### DIFF
--- a/docs/source/apps/projinfo.rst
+++ b/docs/source/apps/projinfo.rst
@@ -16,7 +16,7 @@ Synopsis
 ********
 
     | **projinfo**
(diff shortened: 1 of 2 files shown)


rouault:
docs/source/apps/projinfo.rst should also be updated

nyalldawson:
Done

kbevers:
Thanks, Nyall! Congrats on your first PROJ contribution. I hope to see many more in the future :-)

PR #1349, opened 2019-03-22
Doc: Be consistent in use of +opts

cjmayo:
Apps use the form [ +opts[=arg] ] in the help they print to stderr, some
man pages do so in the summary but not in the text. This renames the
instances of +args to +opts.

---

Being consistent seems important but I don't have an opinion on which should be used (easier to change docs than code) - please check.

### DIFF
--- a/docs/source/apps/cct.rst
+++ b/docs/source/apps/cct.rst
@@ -13,7 +13,7 @@ cct
 Synopsis
 ********
 
-    **cct** [ **-cIostvz** [ args ] ] *+opts[=arg]* file[s]
+    **cct** [**-cIostvz** [args]] *+opt[=arg]* ... file ...
 
 Description
 ***********
@@ -70,7 +70,7 @@ The following control parameters can appear in any order:
 
     Print version number.
 
-The *+args* arguments are associated with coordinate operation parameters.
+The *+opt* arguments are associated with coordinate operation parameters.
 Usage varies with operation.
 
 .. only:: html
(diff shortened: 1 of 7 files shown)


kbevers:
@cjmayo Thanks for this, consistency is definitely important.

The man pages are auto-generated from the Sphinx docs so no need to update those (they will be overwritten at release time anyway).

cjmayo:
Think I've got there after a couple of attempts. Please do check, especially  the stderr of cs2cs where there seemed to be a stray bracket that I have removed - I'm assuming the `+to` is not optional if you specify a target string.
I've taken onboard the ellipsis idea @busstoptaktik (from the comment that is now hidden https://github.com/OSGeo/proj.4/commit/45b88c0014facd60a57b1be7088c2f44102789ca#commitcomment-32880983) and also used it with `file`. As in the updated commit comment I've used man(1) as a reference.

kbevers:
Looks good to me. `+to` is a bit special but in reality you can get away with almost anything. Here's a few calls that are all valid:

```
# 1.
$ ./cs2cs EPSG:25832 EPSG:4326

# 2.
$ ./cs2cs EPSG:25832 +to EPSG:4326

# 3.
$ ./cs2cs "+proj=utm +zone=32 +ellps=GRS80" "+proj=latlong +datum=WGS84"

# 4. 
$ ./cs2cs +proj=utm +zone=32 +ellps=GRS80 +to +proj=latlong +datum=WGS84

# 5.
$ ./cs2cs "+proj=utm +zone=32 +ellps=GRS80" +to "+proj=latlong +datum=WGS84"

# 6.
$ ./cs2cs +proj=utm +zone=32 +ellps=GRS80 +to EPSG:4326
```

It's only in example 4 and 6 that `+to` is needed for `cs2cs` to understand what you mean.

PR #1351, opened 2019-03-23
[WIP] Pass coordinates by reference in 3D and 4D functions

rouault:
This fixes cppcheck master warnings.
This could have a (likely tiny) performance advantage by saving useless
copies of the xyz/xyzt argument.

I should probably look at the differences in machine code in optimized builds. Passing by value as done currently might not perhaps be that bad on x86_64 if, as I presume values, might be stored in SSE registers and not passed on the stack.

### DIFF
--- a/src/apps/gie.cpp
+++ b/src/apps/gie.cpp
@@ -874,7 +874,7 @@ static int expect_failure_with_errno_message (int expected, int got) {
 
 /* For test purposes, we want to call a transformation of the same */
 /* dimensionality as the number of dimensions given in accept */
-static PJ_COORD expect_trans_n_dim (PJ_COORD ci) {
+static PJ_COORD expect_trans_n_dim (const PJ_COORD& ci) {
     if (4==T.dimensions_given_at_last_accept)
         return proj_trans (T.P, T.dir, ci);
 
--- a/src/conversions/axisswap.cpp
+++ b/src/conversions/axisswap.cpp
@@ -104,7 +104,7 @@ static PJ_LP reverse_2d(PJ_XY xy, PJ *P) {
 }
 
 
-static PJ_XYZ forward_3d(PJ_LPZ lpz, PJ *P) {
+static PJ_XYZ forward_3d(const PJ_LPZ& lpz, PJ *P) {
     struct pj_opaque *Q = (struct pj_opaque *) P->opaque;
     unsigned int i;
     PJ_COORD out, in;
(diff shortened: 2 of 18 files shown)


kbevers:
Is this still work-in-progress?

I don't think I understand the reason for this. What are the warnings that are avoided by this fix? It somehow seems a bit unnatural to me to pass by reference here, but I can't really explain why. Just one of those things, I guess :-) 
Also, is it not necessary to give the 2D-functions the same treatment?

rouault:
> Is this still work-in-progress?

yes, until I manage find the motivation to look at the disassembly to see if that doesn't worsen things at least.

> I don't think I understand the reason for this. What are the warnings that are avoided by this fix?

cppcheck (more recent versions than the one used in CI) warns as soon as a C++ struct/class instance is passed by value instead of reference. The rationale is that it saves copying of objects (passing by reference is under-the-hood similar as passing a pointer to the structure), and thus can help performance.

>  is it not necessary to give the 2D-functions the same treatment?

in theory yes, but cppcheck doesn't detect it (it is an inperfect tool). In the case of the 2D functions, the size of the XY/LP structure is just 16 bytes. Passing by reference on 64bit builds would be 8 bytes, so the gain would not be that much and basically all files under src/projections/ should be changed, so a lot of effort.

kbevers:
> yes, until I manage find the motivation to look at the disassembly to see if that doesn't worsen things at least.

Right, doesn't sound like the most exciting thing to do :-)

> The rationale is that it saves copying of objects (passing by reference is under-the-hood similar as passing a pointer to the structure), and thus can help performance.

Yes, I got the part about the performance, just not that that was the problem. It just sounded like a nice side effect. We have been getting a bunch of performance complaints lately (valid or not) so it may be a good idea to focus on that for a bit.

> basically all files under src/projections/ should be changed, so a lot of effort.

This will have to be done sooner or later anyway[0], so we might as well coordinate the effort so that it only requires one pass. 

[0] In the not-so-distant future all operations should be 4D, and 4D only.

rouault:
> so it may be a good idea to focus on that for a bit.

That would be just a micro-optimization. Nothing that I'd expect to have measurable consequences in performance profiles... The main focus was to have clean cppcheck output.

busstoptaktik:
> That would be just a micro-optimization. Nothing that I'd expect to have measurable consequences in performance profiles

I did some timing checks some years ago, before introducing `proj_trans()`.

Passing the PJ_COORD by reference made absolutely no difference on my setup at that time (I do not remember the details, but I am pretty sure that even on a PDP-11 the difference would be immaterial).

Also, the return value is returned by value, so some struct-sized copying is done anyway.

CherryKitty (changes requested):
Please fix to original content. I am not a coder. This is taking me litterally forever in a day to try and figure out. And preventing me from making art.

rouault:
Dropping this PR. scripts/cppcheck.sh has been adapted to ignore the related warnings

PR #1356, opened 2019-03-24
pj_strerrno: enable system error messages

cjmayo:
HAVE_STRERROR is defined in proj_config.h.

---

Missed this one last time. Don't have a simple example, caught it from pyproj.


cjmayo:
Updated to fix CMake test failure - unit test needed to be given the directory where proj_config.h can be found.

kbevers:
Looks good to me. Thanks :-)

PR #1368, opened 2019-03-25
Database: update to EPSG v9.6.1, IGNF v3.0.3, ESRI 10.7.0 and add operation_version column

rouault:
Only the 4 last commits are specific to this PR, the other ones come from https://github.com/OSGeo/proj.4/pull/1366

I'm not sure what our policy regarding backport is for that case. There's no API change, but we change database content and structure.


kbevers:
> I'm not sure what our policy regarding backport is for that case. There's no API change, but we change database content and structure.

Maybe a topic for discussion on the mailing list? My immediate thoughts are that it is probably safe to do a backport right now but it may cause problems for people in the future. In the future we may very well see people that are interfacing with the database directly.

rouault:
Discussion started in https://lists.osgeo.org/pipermail/proj/2019-March/008376.html

hobu:
👍 to this being done at this time. It is still quite early in the 6.0 cycle and people are just getting acclimated.

PR #1380, opened 2019-03-27
ocea: fix behaviour when +alpha is specified

### LINKED ISSUE #1379: ocea: division by zero if alpha is specified and alpha=0 or lat_0=0
Triggered by https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13930

Can be be easily reproduced with: `proj +proj=ocea +alpha=0 +lat_0=1` or `proj +proj=ocea +alpha=1 +lat_0=0`
Both lead to finite results (since atan() has always finite output), but a division by zero still occurs at https://github.com/OSGeo/proj.4/blob/15a9adb5ff112184ea5abfe083ea280b7bef7dd4/src/projections/ocea.cpp#L69

The formulas come directly from Snyder but the later doesn't say anything about the validity range of the parameters.

So should alpha=0 and/or lat_0=0 be allowed ?

If we don't know and want to keep current behaviour unchanged, then a potential solution is to detect the div by zero before it occurs, and return the result of atan(+infinity)=pi/2 or atan(-infinity)=-pi/2 directly.

rouault:
The one-point case was completely broken with lat_0 being ignored.
I've fixed that, and also modified the alpha orientation, so that the angle
corresponds to the one of the 2 point case when going from point 1 to point 2,
similarly to what omerc does.
This was found rather experimentaly with the added test cases that try to
find equivalence between 1-point and 2-point cases.

Fixes #1379
and adresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13930


kbevers (approved):
Looks alright to me.

src/projections/ocea.cpp
```diff
@@ -63,12 +63,17 @@ PJ *PROJECTION(ocea) {
...
+        // ERO: I've added M_PI so that the alpha is the angle from point 1 to point 2
+        // from the North in a clockwise direction
+        // (to be consistent with omerc behaviour)
+        alpha   = M_PI + pj_param(P->ctx, P->params, "ralpha").f;
         lonz = pj_param(P->ctx, P->params, "rlonc").f;
         /*Equation 9-8 page 80 (http://pubs.usgs.gov/pp/1395/report.pdf)*/
-        lam_p = atan(-cos(alpha)/(-sin(phi_0) * sin(alpha))) + lonz;
+        // Actually slightliy modified to use atan2(), as it is suggested by
+        // Snyder for equation 9-1, but this is not mentionned here
+        lam_p = atan2(-cos(alpha) , (-sin(P->phi0) * sin(alpha))) + lonz;
```
kbevers: You could remove the extra parenthesis in the last part of the `atan2` call: `atan2(-cos(alpha) , -sin(P->phi0) * sin(alpha))`

PR #1381, opened 2019-03-28
Objet -> Object

nyalldawson:


kbevers:
Looks like someone's been doing some cut'n'pasting :-). Thanks for the fix!

PR #1387, opened 2019-03-28
Add proj_normalize_for_visualization()

### LINKED ISSUE #1301: C API: add a function to tell if the axis order of a CRS is "swapped"
@pramsey created a `proj_crs_is_swapped` function in  https://github.com/pramsey/postgis/blob/7ecf6839c57a838e2c8540001a3cd35b78a730db/liblwgeom/lwgeom_transform.c#L299 to tell if the axis order is lat/lon or northing/easting, instead of the natural GIS order, and be able to swap if needed. This might be of interest to a number of projects already using PROJ and wanting to keep the 'sane' GIS axis order rather than the EPSG one.
GDAL has also simillar code in the refreshAxisMapping() method ( https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrspatialreference.cpp#L287 ), except that it explicitly tests for "north" + "east", so as to avoid doing axis inversion for south-pole projections where the 2 axis are pointing to the north...
I'm not sure about the naming of the function however. Perhaps this should `proj_crs_axis_order_is_north_east` to be really explicit

rouault:
Fixes #1301

This function takes the output PJ from proj_create_crs_to_crs(),
and add (or undo) the needed axis swap operations so that the
object returned by proj_normalize_for_visualization() has the usual
GIS axis order.
In this implementation, this does something only if the coordinate
 system of the source or target CRS, geographic or projected, has
NORTH, EAST ordering.
CompoundCRS wrapping those objects are also handled.

Question: should this be backported to 6.0.x ?

### DIFF
--- a/include/proj/coordinateoperation.hpp
+++ b/include/proj/coordinateoperation.hpp
@@ -152,6 +152,8 @@ class PROJ_GCC_DLL CoordinateOperation : public common::ObjectUsage,
 
     PROJ_DLL static const std::string OPERATION_VERSION_KEY;
 
+    PROJ_DLL CoordinateOperationNNPtr normalizeForVisualization() const;
+
   protected:
     PROJ_INTERNAL CoordinateOperation();
     PROJ_INTERNAL CoordinateOperation(const CoordinateOperation &other);
--- a/include/proj/crs.hpp
+++ b/include/proj/crs.hpp
@@ -126,6 +126,10 @@ class PROJ_GCC_DLL CRS : public common::ObjectUsage {
     PROJ_FOR_TEST CRSNNPtr
     alterCSLinearUnit(const common::UnitOfMeasure &unit) const;
 
+    PROJ_INTERNAL bool mustAxisOrderBeSwitchedForVisualization() const;
+
+    PROJ_INTERNAL CRSNNPtr normalizeForVisualization() const;
+
     //! @endcond
 
   protected:
--- a/src/iso19111/c_api.cpp
+++ b/src/iso19111/c_api.cpp
@@ -6767,3 +6767,35 @@ int proj_cs_get_axis_info(PJ_CONTEXT *ctx, const PJ *cs, int index,
     }
     return true;
 }
+
+// ---------------------------------------------------------------------------
+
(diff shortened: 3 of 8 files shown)


kbevers:
> In this implementation, this does something only if the source or
target CRS have latitude,longitude ordering.

But it is not limited to only coordinate systems with geodetic coordinates, right? The code seems to be checking projected coordinates as well but from the documentation it seems as if it only applies when latitudes and longitudes (i.e. angular coordinates) are involved.

> Question: should this be backported to 6.0.x ?

No. This adds a new function so should not be included in a bug-fix release.

rouault:
> But it is not limited to only coordinate systems with geodetic coordinates, right? The code seems to be checking projected coordinates as well but from the documentation it seems as if it only applies when latitudes and longitudes (i.e. angular coordinates) are involved.

ah sorry, the commit message is wrong. I'm going to fix it. It does the inversion for geographic or projected CRS with axis order direction NORTH, EAST.

nyalldawson:
I'd like to humbly request that this IS backported to 6.0 -- without it you're requiring all clients to implement their own version of https://github.com/pramsey/postgis/blob/7ecf6839c57a838e2c8540001a3cd35b78a730db/liblwgeom/lwgeom_transform.c#L299 in order to correctly project x/y coordinates. I think it's MUCH safer to have this sitting in the underlying proj library then have multiple varying versions of this test floating out their in the client applications.

kbevers:
> I think it's MUCH safer to have this sitting in the underlying proj library then have multiple varying versions of this test floating out their in the client applications.

I absolutely agree. I am just not very keen on deviating from [semantic versioning](https://semver.org/) rules. The simple solution is to release 6.1.0 right away instead of releasing a few patch releases first. Thoughts?

rouault:
> The simple solution is to release 6.1.0 right away instead of releasing a few patch releases first. Thoughts?

So the 6.0.1 release of May 1st would become a 6.1.0 ? I'm fine with that.

kbevers:
> So the 6.0.1 release of May 1st would become a 6.1.0 ? I'm fine with that.

Yes. It would also fit well with the changes to the database. Are there any other functions that would be beneficial to add before May 1st?

PR #1391, opened 2019-03-28
Add no-op operation. It does nothing.

### LINKED ISSUE #1185: Create a 'noop' operation ?
The new C++ code that creates operations can return an empty PROJ string when the transformation is a no opeation. Currently it instanciates a dummy +proj=affine for that purpose, but probably a 'noop' operation would be better

kbevers:
Closes #1185


rouault:
noop.cpp lacking in the commit

The following line could be upgraded to use noop
```
../src/iso19111/c_api.cpp:                ctx, projString.empty() ? "+proj=affine" : projString.c_str());
```

And, a bit harder, `projinfo -s EPSG:4326 -t EPSG:4326 -o PROJ` currently returns an empty string. I think that the fix would be to return "+proj=noop" at the end of PROJStringFormatter::toString() in io.cpp if the string is empty. And in that case the above test of the empty string in c_api.cpp could be just removed.

kbevers:
> noop.cpp lacking in the commit

🤦

> The following line could be upgraded to use noop

I was looking for that but couldn't locate it in my first attempt. I had hunch you would remind me where it was :-) I will see if I can figure out how to do it the hard way. If not, I'm taking the easy path...

kbevers:
I've added a very naive adoption of noop in iso19111 code. I saw some test failures locally so this probably won't work here either. It's a starting point for further discussion :)

rouault:
> I saw some test failures locally so this probably won't work here either

from Travis output, at least the first failures are no real issues. testprojinfo_out just needs to be updated with the new output

kbevers:
> from Travis output, at least the first failures are no real issues. testprojinfo_out just needs to be updated with the new output

Yeah, it does seem like that. I had a problem locally that I couldn't immediately determine if was from this change or something else. I've updated the test output for `testprojinfo`.

PR #1406, opened 2019-04-02
CMake: Set -fvisibility=hidden using global variable

mwtoews:
The CMake way to use the compile flag `-fvisibility=hidden` is to set the property:
```
set(CMAKE_C_VISIBILITY_PRESET hidden)
set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
It's appropriate to set this globally, right? Otherwise, any exceptions for other targets can use `set_target_properties(MyTarget PROPERTIES CXX_VISIBILITY_PRESET default)`

See also:
- https://cmake.org/cmake/help/v3.5/variable/CMAKE_LANG_VISIBILITY_PRESET.html
- https://stackoverflow.com/a/31157258


kbevers:
Seems like the right thing to do but I'll keep this one open for a bit so that others with more CMake insight than me can object.

Issue #1263, opened 2019-02-13
WIP: Modernize CMake configuration


mloskot:
**Status:** _Request For Comments_

Current configuration pre-dates CMake 3. Since the minimum required version of CMake bumped to 3.5 (#1262), it is possible to modernize the configuration following the **Modern CMake** approach and conventions in future versions of PROJ.4

Below, discussed are issues to consider. 

## BUILD_SHARED_LIBS

Remove custom `BUILD_LIBPROJ_SHARED` and rely on CMake native [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html).

## Compiler flags management

It would be good to consider CMake modernisation to avoid manual fiddling with `CMAKE_CXX_FLAGS`.

Instead, there should be `add_library(proj4_compiler_flags INTERFACE)` followed by `target_compile_features`, `target_compile_options` etc. calls setting any generic features and compiler specific flags (using [generator expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html)).

The `proj4_compiler_flags` target could also be accompanied with `proj4_compiler_flags_devel` extending the set of flags (e.g. with stricter warnings) for building PROJ.4 during development, CI testing, etc.

The `INTERFACE` targets approach is nowadays a recommended CMake way and https://gitlab.kitware.com/vtk/vtk-m or https://github.com/libgeos/geos/pull/130 can serve as working examples.

## C++ version requirement as usage requirement

Possible future refinements are discussed below as continuation of https://github.com/OSGeo/proj.4/issues/1118#issuecomment-463194388

> TBH, I'm not sure about setting `CXX_STANDARD` on any public targets, in the Config-files, etc.

The required C++ version required can also be imposed as usage requirement by setting compilation flags using CMake's [C++ features](https://cmake.org/cmake/help/v3.1/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html) requirement on the library target, in `src/lib_proj.cmake` file:

```
if(CMAKE_VERSION VERSION_LESS 3.8)
  target_compile_features(${PROJ_CORE_TARGET} INTERFACE cxx_nullptr)
else()
  target_compile_features(${PROJ_CORE_TARGET} INTERFACE cxx_std_11)
endif()
```

With that in place, installation will generate the following in both, `proj4-targets.cmake`  and `proj4-namespace-targets.cmake`:

```
[...]

PR #1409, opened 2019-04-04
CMake: Remove need to fiddle with CMAKE_C_FLAGS / CMAKE_CXX_FLAGS

mwtoews:
This PR is to help modernize the CMake setup (#1263)

There a are likely several ways to accomplish this, but this is the cleanest way that I can see. Most of the custom flags added to date are warning flags. There are now cached into `PROJ_C_WARN_FLAGS`/`PROJ_CXX_WARN_FLAGS`, then applied to the relevant targets.

Currently, these target compile options `PRIVATE`, since warning flags should not change the behavior of the library/utilities. But perhaps these should be `INTERFACE`?

Also, `/D_CRT_SECURE_NO_WARNINGS` is not exactly a warning flag, so it's been bumped into a global `add_definitions`.

The only remaining `CMAKE_C_FLAGS` modification is related to an Intel flag, which I may address another day.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,30 +34,45 @@ message(STATUS "Requiring C++${CMAKE_CXX_STANDARD} - done")
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 
-# Set warnings
-if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
-  # Suppress warning 4706 about assignment within conditional expression
-  # Suppress warning 4996 about sprintf, etc., being unsafe
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} \
-/W4 /wd4706 /wd4996 /D_CRT_SECURE_NO_WARNINGS")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} \
-/EHsc /W4 /wd4706 /wd4996 /D_CRT_SECURE_NO_WARNINGS")
-elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} \
--Wall -Wextra -Wswitch -Wshadow -Wunused-parameter \
--Wmissing-prototypes -Wmissing-declarations -Wformat -Wformat-security")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} \
--Wall -Wextra -Wswitch -Wshadow -Wunused-parameter \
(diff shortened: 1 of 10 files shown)


mwtoews:
Oh, nearly forgot to mention moving the interprocedural optimization flag `-flto` as a target compile option. It is currently set as `PRIVATE`, but should it be `PUBLIC` or `INTERFACE` instead? From what I understand, this option is both a compile and linker flag.

The two failures are related to implicit function declaration, which changed between C89 and C99 Standards. PR #1410 makes it bit more clear to see the compile command to help sort out this type of situation.

mwtoews:
The two C89 failures are now sorted, which needed "all warnings" added to `CMAKE_REQUIRED_FLAGS` (I'm nearly certain there is a CMake bug with `check_c_source_compiles`, as the C99 example clearly fails with `-std=c89`; also, CMake has [>2k open issues](https://gitlab.kitware.com/cmake/cmake/issues), so I'd rather not dig any deeper).

This PR is ready for review. I.e., are there any opinions on the use of `PRIVATE` vs. `PUBLIC` or `INTERFACE` as mentioned above for the scope of warning flags or IPO?

kbevers:
> This PR is ready for review. I.e., are there any opinions on the use of PRIVATE vs. PUBLIC or INTERFACE as mentioned above for the scope of warning flags or IPO?

Overall I think this is great. I have no idea what is the right thing to do regarding the PRIVATE/PUBLIC thing. What will be different if PUBLIC is chosen instead of PRIVATE?

mwtoews:
PRIVATE is the easiest option to go with, as these flags only go to those targets, and don't get imported or passed around to others.

Warning flags are generally added to targets with PRIVATE (e.g. [here](https://foonathan.net/blog/2018/10/17/cmake-warnings.html), [here](https://stackoverflow.com/a/50882216)). However, I'm aware the folks in the libgeos camp (e.g. [dbaston's modern-cmake](https://github.com/dbaston/libgeos/blob/modern-cmake/CMakeLists.txt)) have chosen to go with INTERFACE for warnings, and generally have a different approach to applying the flags to targets (for instance, I generally find if-conditions more readable than generator expressions).

As for IPO/LTO, PRIVATE produces the same compile commands as existing, so this selection is probably fine. But in the CMake realm, I'm not sure if those properties need to be retained if imported/used by other targets, since it's a compiler and linker flag.

kbevers:
After reading [this](https://cmake.org/pipermail/cmake/2016-May/063400.html) I think that PRIVATE is the right option here. It would be interesting to know why GEOS chose to use INTERFACE over PRIVATE though.

mwtoews:
Yup, that makes sense. Nevertheless, the scope keyword is simple to toggle, if there is a good argument to be made.

kbevers:
Right, let's just go with what you've made so far. We'll change it if is needed at some point in the future.

PR #1410, opened 2019-04-05
Adjust test outputs

mwtoews:
- Travis CI only has two cores, so `-j2` is sufficent
 - Tests should use unified diff with expected results
 - Show CMake build compile commands (`VERBOSE=1`)


mwtoews:
I've just cobbled a few other Travis CI items into this PR:
 - Simply .travis.yml to specify one language at the top. The second test won't assume Ruby (the default language).
 - Remove progress bars from `pip install`, which clutter the log files.

PR #1413, opened 2019-04-07
Add Cirrus CI with FreeBSD

rouault:


kbevers:
What is the purpose of running these tests? Just to check that they compile on FreeBSD? Or do you have something more involved in mind for the future? The testing is less rigorous than what is done on Travis for instance. That's totally fine, just trying to get a sense of the perspectives of this setup.

rouault:
> What is the purpose of running these tests? Just to check that they compile on FreeBSD?

yes, compile and make check as well. That should be fine to check the sanity of the build on BSD (there were a few issues with PROJ 6.0.0 on netBSD found by Greg Troxel, so I guess that FreeBSD would be close enough to have caught them too). The more complete tests we do in travis/install.sh on other configs could be done just on one actually.

kbevers:
> That should be fine to check the sanity of the build on BSD (there were a few issues with PROJ 6.0.0 on netBSD found by Greg Troxel, so I guess that FreeBSD would be close enough to have caught them too)

Let's hope so! In any case there's a better chance of finding problems if tested on more systems. Thanks for setting this up.

> The more complete tests we do in travis/install.sh on other configs could be done just on one actually.

That would actually be nice, yes. There's no need to hog ressources if it can be avoided.

PR #1414, opened 2019-04-08
Docs: fix typo in vertical grid shifts

andrew-hardin:
The vertical grid shift examples have a small typo: they ought to be using `egm96_15.gtx` - not `egm96_16.gtx`.


kbevers:
Nice catch, thanks for the fix!

PR #1417, opened 2019-04-09
DOC: update line numbers in dev quickstart example

jorisvandenbossche:
While reading the docs, I noticed that the included example lines were no longer correct.

I built the docs locally to verify it is now correct (I can show a screenshot if needed).

### DIFF
--- a/docs/source/development/quickstart.rst
+++ b/docs/source/development/quickstart.rst
@@ -35,7 +35,7 @@ this in detail.
 
 .. literalinclude:: ../../../examples/pj_obs_api_mini_demo.c
   :language: c
-  :lines: 48
+  :lines: 49
   :dedent: 4
 
 Next we create the ``PJ`` transformation object ``P`` with the function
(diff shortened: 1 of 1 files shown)


kbevers:
Good find! Thanks for the fix.

docs/source/development/quickstart.rst
```diff
@@ -62,32 +62,32 @@ for further details.
 
 .. literalinclude:: ../../../examples/pj_obs_api_mini_demo.c
   :language: c
-  :lines: 56
+  :lines: 57
   :dedent: 4
 
-The coordinate defined above is transformed with ``proj_trans_coord``. For this
+The coordinate defined above is transformed with ``proj_trans``. For this
```
jorisvandenbossche, line 68: Another question: I could replace things like ``` ``proj_trans`` ``` in this file with `` :c:func:`proj_trans` `` so it automatically creates a link to the reference page. Would there be interest in a PR for that?

kbevers: Yes, definitely. There are many of such occurrences. I usually fix them when I stumble upon them. Any help is greatly appreciated.

jorisvandenbossche: OK, then I will do a PR for that. Perfect combination while reading through the docs to directly do some fixes :)

kbevers: Thanks. I'm sure there's many things that can be improved. It's been a long time since I've read the docs from start to finish so I've lost track of how consistent it is. All improvements are welcome :)

PR #1420, opened 2019-04-10
DOC: add sphinx links to reference in dev quickstart

jorisvandenbossche:
Question before merging this: I now only made the first occurrence of a type of function into a sphinx link, and left the subsequent as normal sphinx code format to check this is working. But you might prefer to convert all occurrences of linkable objects? (I do I think)


kbevers:
>  But you might prefer to convert all occurrences of linkable objects? (I do I think)

If are willing to do that, definitely yes.

PR #1425, opened 2019-04-16
Remove `PROJ_COMPILATION=1` definition

mwtoews:
These do not appear to be used, and I can't see what they are for. They were added from the d928db15d53805d9b728b440079756081961c536 mega commit for the "GDAL SRS barn" work.

@rouault is it OK to remove these?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,8 +19,6 @@ if(NOT CMAKE_VERSION VERSION_LESS 3.1)
   cmake_policy(SET CMP0054 NEW)
 endif()
 
-add_definitions(-DPROJ_COMPILATION=1)
-
 # Set C++ version
 # Make CMAKE_CXX_STANDARD available as cache option overridable by user
 set(CMAKE_CXX_STANDARD 11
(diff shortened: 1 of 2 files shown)


rouault:
> is it OK to remove these?

yes, can't remember why this was added and doesn't appear to be used anywhere

cffk:
I can see a use for PROJ_COMPILATION...

I'd like to have geodesic.[ch] usable independently of PROJ (it's
currently a separate mini-project within GeographicLib).  In that case
it would be nice if the compiler could know if geodesic.c were being
compiled as part of PROJ (in which case proj_math.h can be included) or
not.  Currently this uses the PJ_LIB__ macro, but I don't think this
ever worked.  However PROJ_COMPILATION would fill the bill.

mwtoews:
@cffk sure, we can keep it.

Are there any other use-cases for this define? If not, then with the CMake setup, I may just reduce the scope of this define to `lib_proj.cmake` via:
```cmake
# Use PROJ_COMPILATION to distinguish code compiled as part of PROJ 
set_source_files_properties(
  geodesic.c geodesic.h
  PROPERTIES COMPILE_DEFINITIONS PROJ_COMPILATION
)
```
And with the autotools setup, just set it in `AM_CPPFLAGS`, but it'd still be global.

cffk:
Since it'll be a few months before I'm in a position to update the geodesic
routines, why don't you remove the definition for now.  I'll add it back in
when I need it (narrowing the scope if I can).

mwtoews:
@cffk great to know about new development with the C geodesic routines! And yup, it should be simple to re-apply this macro to autoconf/cmake when the time comes.

PR #1427, opened 2019-04-16
createOperations(): do not attempt using a unrelated datum intermediate when doing geog2D<-->geog3D conversions of same datum

rouault:
Seen when testing transformations between "CR 05" (EPSG:5365) and "CR-SIRGAS" (EPSG:8907)
which require going through their corresponding 3D GeogCRS to find a Helmert
transformation.


rouault:
Actually it turns that this case shouldn't happen in practice as the current entry in the EPSG database is bogus ( https://lists.osgeo.org/pipermail/proj/2019-April/008459.html ), but it is good that PROJ is able to be tolerant about such cases

edurojasr:
Mailing lists!

PR #1441, opened 2019-04-22
proj_create(): add support for compoundCRS and concatenatedOperation named from their components

rouault:
Support following syntaxes:
- OGC URN combining references for compoundCRS:
  e.g. "urn:ogc:def:crs,crs:EPSG::2393,crs:EPSG::5717"
- its GDAL shortcut: e.g. "EPSG:2393+5717"
- OGC URN combining references for concatenated operations:
  e.g. "urn:ogc:def:coordinateOperation,coordinateOperation:EPSG::3895,coordinateOperation:EPSG::1618"


busstoptaktik:
Awesome!

PR #1454, opened 2019-05-06
createOperations(): fix case of ETRS89 3D to proj string with nadgrids and geoidgrids

rouault:
Fixes https://lists.osgeo.org/pipermail/proj/2019-May/008519.html


data/sql/grid_alternatives.sql
```diff
@@ -798,3 +798,41 @@ INSERT INTO grid_alternatives(original_grid_name,
...
+                              inverse_direction,
+                              package_name,
+                              url, direct_download, open_license, directory)
+                      VALUES ('naptrans2008.gtx',
+                              'naptrans2008.gtx',
+                              'GTX',
+                              'vgridshift',
+                              0,
+                              NULL, -- package name
+                              'https://salsa.debian.org/debian-gis-team/proj-rdnap/raw/master/naptrans2008.gtx',
```
sebastic: The URL on the master branch may not be very stable, using the upstream tag for the 2008 release is probably better:

 https://salsa.debian.org/debian-gis-team/proj-rdnap/raw/upstream/2008/naptrans2008.gtx
 https://salsa.debian.org/debian-gis-team/proj-rdnap/raw/upstream/2008/rdtrans2008.gsb

As mentioned by Jochem Lesparre [on the proj list](https://lists.osgeo.org/pipermail/proj/2019-March/008402.html), Kadaster will release new NTv2 and Vdatum grids soon. He also asked me if the CC-BY license would make it possible to include the grids in PROJ, in that case we can include the grids in proj-datumgrid. If they keep the same license, I'll update proj-rdnap package to include the new grids.

rouault: Thanks, I've updated the commit with the new URLs

Yes, if they release new gridss with CC-BY, then that could make it proj-datumgrid-europe. I've added recently a kind of tutorial for all what is involved in adding new grids: https://github.com/OSGeo/proj-datumgrid/blob/master/CONTRIBUTING.md

PR #1462, opened 2019-05-08
Update to EPSG 9.6.2

rouault:


kbevers:
Are you confident that this will not introduce new problems downstream?

rouault:
> Are you confident that this will not introduce new problems downstream?

yes, at 99%. The updates are very minor: 2 grid names changed for Australia (none of which we actually ship in proj-datumgrid-oceania), and a fix for the oddity I noticed around Costa Rica Helmert transformations. 
The biggest change is in scripts/build_db.py which is only used when refreshing the database manually like I did (that change should have been done previously, and is independent, but needed, to do the update)
So, nothing critical and we could as well restrict that to master if you prefer to exercise your right in saying "no" :-)

kbevers:
> So, nothing critical and we could as well restrict that to master if you prefer to exercise your right in saying "no" :-)

It did indeed cross my mind :) I think we will be okay to include this on.

PR #1473, opened 2019-05-15
update doc building Docker container to Python3 for Sphinx 2.0 support

hobu:


hobu:
This container is also being used for upcoming GDAL Sphinx document refresh.

kbevers:
Should this also be added to the 6.1 maintenance branch?

hobu:
Yes, this should be backported.

PR #1494, opened 2019-05-31
Add proj_grid_get_info_from_database 

nyalldawson:
Allows retrieval of grid metadata from a grid filename. Currently the only way to retrieve this is via proj_coordoperation_get_grid_used, but that requires an existing coordinate operation object (which can be impossible to create if a particular grid is not locally available).

### DIFF
--- a/src/iso19111/c_api.cpp
+++ b/src/iso19111/c_api.cpp
@@ -626,6 +626,65 @@ int proj_uom_get_info_from_database(PJ_CONTEXT *ctx, const char *auth_name,
 
 // ---------------------------------------------------------------------------
 
+/** \brief Get information for a grid from a database lookup.
+ *
+ * @param ctx Context, or NULL for default context.
+ * @param grid_name Grid name (must not be NULL)
+ * @param out_full_name Pointer to a string value to store the grid full
+ * filename. or NULL
+ * @param out_package_name Pointer to a string value to store the package name
+ * where
+ * the grid might be found. or NULL
+ * @param out_url Pointer to a string value to store the grid URL or the
+ * package URL where the grid might be found. or NULL
+ * @param out_direct_download Pointer to a int (boolean) value to store whether
+ * *out_url can be downloaded directly. or NULL
+ * @param out_open_license Pointer to a int (boolean) value to store whether
+ * the grid is released with an open license. or NULL
+ * @param out_available Pointer to a int (boolean) value to store whether the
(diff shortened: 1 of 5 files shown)


src/iso19111/c_api.cpp
```diff
@@ -624,6 +624,69 @@ int proj_uom_get_info_from_database(PJ_CONTEXT *ctx, const char *auth_name,
...
+ */
+int PROJ_DLL proj_grid_get_info_from_database(PJ_CONTEXT *ctx,
+                               const char *grid_name,
+                               const char **out_full_name,
+                               const char **out_package_name,
+                               const char **out_url,
+                               int *out_direct_download,
+                               int *out_open_license,
+                               int *out_available) {
+    assert(grid_name);
```
rouault, line 652: Could you run scripts/reformat_cpp.sh (with clang-format pointing to clang-format-3.8) ?
(CI doesn't do that check currently)

src/iso19111/c_api.cpp
```diff
@@ -624,6 +624,69 @@ int proj_uom_get_info_from_database(PJ_CONTEXT *ctx, const char *auth_name,
...
+        std::string package_name_str;
+        std::string url_str;
+        bool direct_download;
+        bool open_license;
+        bool available;
+        if ( !db_context->lookForGridInfo( grid_name, full_name_str, package_name_str, url_str, direct_download, open_license, available ) )
+            return false;
+
+        if ( out_full_name )
+            *out_full_name = full_name_str.c_str();
```
rouault: unfortunately the C pointer will no longer be valid once that function returns, since full_name_str is a temporary object. Same for package_name and url. You probably need to store those objects in the ctx class.

nyalldawson:
@rouault done!

kbevers:
Complaint from Travis:

```
proj_get_source_crs
 proj_get_target_crs
 proj_get_type
+proj_grid_get_info_from_database
 proj_grid_info
 proj_identify
 proj_info
Difference(s) found in exported symbols. If intended, refresh scripts/reference_exported_symbols.txt with 'scripts/dump_exported_symbols.sh src/.libs/libproj.so > scripts/reference_exported_symbols.txt'
The command "./travis/${BUILD_NAME}/install.sh" failed and exited with 1 during .
```

Error message explains what to do :)

nyalldawson:
@kbevers thanks, fixed

PR #1496, opened 2019-06-01
Update website: proj4.org -> proj.org

mwtoews:
Most of the changes are fairly obvious, and a few links were fixed.

@hobu does `echo "proj.org" > CNAME` make sense for the [CNAME record](https://en.wikipedia.org/wiki/CNAME_record)?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,7 +18,7 @@ Project Steering Committee
 --------------------------------------------------------------------------------
 
 Process and membership can be found at:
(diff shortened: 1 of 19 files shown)


rouault:
> does `echo "proj.org" > CNAME` make sense for the [CNAME record](https://en.wikipedia.org/wiki/CNAME_record)?

yes, @hobu did it manually for the test.

I've merged and backported this to 6.1

PR #1497, opened 2019-06-02
Fix incorrect delimator ";" used to separate proj_info().searchpath entries

nyalldawson:
... on non-windows platforms, should be ":", not ";"

### DIFF
--- a/src/4D_api.cpp
+++ b/src/4D_api.cpp
@@ -1367,9 +1367,9 @@ static char *path_append (char *buf, const char *app, size_t *buf_size) {
     }
     assert(buf);
 
-    /* Only append a semicolon if something's already there */
+    /* Only append a delimiter if something's already there */
     if (0 != buflen)
-        strcat (buf, ";");
+        strcat (buf, delim);
     strcat (buf, app);
     return buf;
 }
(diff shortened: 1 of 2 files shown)


nyalldawson:
Refs https://github.com/qgis/QGIS/pull/30048

src/4D_api.cpp
```diff
@@ -1369,7 +1369,7 @@ static char *path_append (char *buf, const char *app, size_t *buf_size) {
 
     /* Only append a semicolon if something's already there */
```
QuLogic:
```suggestion
    /* Only append a delimiter if something's already there */
```

nyalldawson: Good catch, fixed

rouault:
@kbevers Do you think we should change the behaviour or keep as it ? After all, the implementation and documentation are consistent currently.
It is true that if we do the analogy with PATH, semi-colon for Windows and colon for Unix are more common.
When looking at the code, I was confused by the fact that path_append() has a #ifdef WIN32 / #else to have a different separator, but it isn't actually used

kbevers:
> Do you think we should change the behaviour or keep as it ? After all, the implementation and documentation are consistent currently.
It is true that if we do the analogy with PATH, semi-colon for Windows and colon for Unix are more common.

I think we should accept this PR. As you say, it conforms to normal practice. I believe this is also how we deal with delimeters when parsing `PROJ_LIB`, right?

> When looking at the code, I was confused by the fact that path_append() has a #ifdef WIN32 / #else to have a different separator, but it isn't actually used

That is clearly a bug - `delim` should have been used in the first place.

PR #1498, opened 2019-06-02
proj_info().searchpath should not include PROJ_LIB path

nyalldawson:
... if the context search_paths is not empty

The logic used when actually searching for files is that if
the context search_paths are set, these completely override
the PROJ_LIB variable. So we should make proj_info().searchpath
correctly reflect this, and not include the PROJ_LIB path
in searchpath if the context has search_paths manually set.

Refs https://github.com/qgis/QGIS/pull/30048/


kbevers:
Thanks for noticing and fixing this :)

PR #1505, opened 2019-06-05
createFromUserInput(): support OGC URN to create projectedCRS, for example to instanciate a projected 3D CRS

rouault:


docs/source/apps/projinfo.rst
```diff
@@ -38,6 +38,9 @@ Synopsis
     - a OGC URN combining references for compound coordinate reference systems
       (e.g "urn:ogc:def:crs,crs:EPSG::2393,crs:EPSG::5717" or custom abbreviated
       syntax "EPSG:2393+5717"),
+    - a OGC URN combining references for references for projected or derived CRSs
```
kbevers, line 41: Since this is new functionality it should be mentioned when it was introduced. Maybe in a parenthesis? The normal way of adding a `.. versionadded::` tag doesn't seem as the best option here...

rouault: will that be 7.0 ?

kbevers: Nope, 6.2.0 as per https://github.com/OSGeo/PROJ/milestones?direction=asc&sort=due_date :)

rouault: ok, precision added

kbevers: Thanks

PR #1517, opened 2019-06-17
Use relative directory to locate data

abellgithub:


abellgithub:
Hi,

This fails to pass because the relative locations of the lib/binary directory and share directory are different for a cmake build and an autoconf build.  I'm thinking it would be nice to make things consistent and always have the binaries located in a sibling directory with the data.  I don't know much about autoconf or what is intended in the build process.

Feedback/recommendations appreciated.

kbevers:
This is in response to https://github.com/OSGeo/PROJ/issues/1490, right?

I haven't had time to look this over thoroughly, but one immediate question I have is why the `data` dir should be changed to `share`?

abellgithub:
Because it makes it consistent with the standard install location.  It makes the build/dev layout look like the install layout.  It allows development without installation or setting of PROJ_LIB.

kbevers:
I see. I am not particularly happy with that change, although it may seem more consistent with the install. I have to reasons:

1. Not everything from `data/` is moved to `share/` during installation. 
2. We recently created `data/` from the old `nad/` folder. Every file/path rename in git disrupts continuity in the history. I am not keen on making more disruptions like that if it can be avoided.

However, good arguments in favour of `share/` can change my mind.

abellgithub:
I don't have any arguments other than what I've provided.  I'm not sure what people do now for development, but for me, being able to run straight from the build directory is valuable.  Other than churn (which only impacts those doing development), I don't see any downside.  I suppose you could split the installed files from those not being installed, but that seems like a hassle.  You could also have the build make a link from share -> data, but I've had people complain of symbolic links before.

There's no requirement at all to move things, but it simplifies things, which is IMO good.  I think more references to PROJ_LIB can be removed, but I didn't want to make assumptions on this since I don't know how things are used during development/test.

rouault:
Note: for development, not all files are always in data/ (or share/) if you do a out-of-source build, which is quite the norm for cmake development, or even for autoconf builds (and this is what is tested by CI). For such out-of-source builds, the proj.db file for example will not be in the $(top_dir)/data, but in  $(build_dir)/data

abellgithub:
Yes, this isn't important.  What matters is the relative locations of the resource files to the library.  I made an autoconf build and the issue was that the libraries were in \<build root\>/src/.libs while the resources were in \<build root\>/share.  If the libraries were in a sibling directory to the share directory, all would be fine.

As another option, we could also go looking for, say, proj.db, in either ../share or ../data.

kbevers:
I am not sure what we should do with this PR. I like the idea of just finding the correct resource files without setting `PROJ_LIB`. Not super happy about rename `data/` to `share/` so if that can be worked around by looking in both `../share/` and `../data/` I think that is a good solution.

And then there's the differences between autconf and CMake. Can you make it work without changing the output dirs for resource files and binaries? I don't want to change that, there's too many people using what we have already.

abellgithub:
I don't understand what you're talking about WRT output directories.  That's a build detail, right?

kbevers:
Sorry, maybe output dirs is not the right way to put it. My understanding was that it is problematic that CMake and autoconf behave differently. That's how I interpreted

> This fails to pass because the relative locations of the lib/binary directory and share directory are different for a cmake build and an autoconf build

at least. I understand that it is not optimal, but what I was trying to ask was, can you make it work with that obstacle in the way or is it necessary to change the  either autoconf or CMake setup?

hobu:
> However, good arguments in favour of `share/`can change my mind.

Let's finish ripping off the band-aid and align the development base layout of our resources with our installation base layout. I always found this dissonnent in this codebase. I suppose the history of the `nad/` layout is @warmerdam didn't want to disrupt Gerry's code when pulling in any new PROJ baselines, but we're well past that now.

If we catch up the autoconf builds to the new layout, I do not see any reason why the arrangement of our internal development layout to align with how we deploy to `share/` should have any external impact.

PR #1531, opened 2019-06-25
grid_alternatives.sql: add the two Swisstopo grids (refs https://github.com/OSGeo/proj-datumgrid/pull/53)

rouault:


kbevers:
I haven't really understood this yet. Is the reason we add the grids to the alternative grids table to provide a download link? As far as I can tell the rest of the information provided in the table should also be available from the EPSG registry, no?

rouault:
>  Is the reason we add the grids to the alternative grids table to provide a download link?

yes, and also to provide an alternate name when the name in EPSG and the name in our packages don't match.

PR #1544, opened 2019-07-02
Create a '.deb' package builder with Docker

woshilapin:
I recently had to build a '.deb' package of PROJ. Debian provide old version of PROJ (version 4) and I needed a version 6 in order to use it in Rust (see crate [proj](https://crates.io/crates/proj)). I couldn't find anything in the project to do that.

I documented in the `README.md` how to use and get the '.deb' package.

I'm more than open to discuss and/or refactor totally this solution. I'm aware the solution is not ideal.

### DIFF
--- a/Dockerfile.deb-packager
+++ b/Dockerfile.deb-packager
@@ -1,0 +1,37 @@
+ARG DEBIAN_VERSION="stretch"
+FROM debian:${DEBIAN_VERSION}-slim
+ARG PROJ_VERSION="6.1.0"
+RUN apt update && apt install -y wget build-essential pkg-config sqlite3 libsqlite3-dev libssl-dev clang
+
+# Download and unzip PROJ
+WORKDIR /tmp
+RUN wget https://github.com/OSGeo/PROJ/releases/download/${PROJ_VERSION}/proj-${PROJ_VERSION}.tar.gz
+RUN tar -xzvf proj-${PROJ_VERSION}.tar.gz
+
+# Compile PROJ and install in /usr
+WORKDIR proj-${PROJ_VERSION}
+RUN ./configure --prefix=/usr
+RUN make
+RUN touch /tmp/timestamp
+RUN make install
+
+# Prepare the folder for building the .deb package
+ENV DEB_NAME "proj_${PROJ_VERSION}_amd64"
+ENV DEB_PATH "/tmp/${DEB_NAME}"
+WORKDIR ${DEB_PATH}
+RUN find /usr ! -type d -cnewer /tmp/timestamp | while read FILE; \
+	do \
+		mkdir -p "${DEB_PATH}$( dirname "${FILE}" )"; \
+		cp "${FILE}" "${DEB_PATH}${FILE}" ; \
+	done
+RUN mkdir DEBIAN/
+RUN echo "2.0\n" > debian-binary
+COPY DEBIAN/control.template DEBIAN/control
+RUN sed --in-place "s/{PROJ_VERSION}/${PROJ_VERSION}/" DEBIAN/control
+RUN sed --in-place "s/{INSTALLED_SIZE}/$( du --summarize --bytes --exclude DEBIAN ${DEB_PATH} | cut --field=1 )/" DEBIAN/control
+RUN find . -type f ! -regex '.*?DEBIAN.*' -printf '%P ' | xargs md5sum > DEBIAN/md5sums
+
+# Build the .deb package
+WORKDIR /
+RUN mkdir /deb/
+RUN dpkg --build ${DEB_PATH} /deb/${DEB_NAME}.deb
--- a/DEBIAN/control.template
+++ b/DEBIAN/control.template
@@ -1,0 +1,11 @@
+Package: proj
+Package-Type: deb
+Version: {PROJ_VERSION}
+Maintainer: Jean SIMARD <woshilapin@tuziwo.info>
+Description: Cartographic Projections and Coordinate Transformations Library
+Section: utils
+Priority: optional
+Installed-Size: {INSTALLED_SIZE}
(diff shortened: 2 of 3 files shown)


Dockerfile.deb-packager
```diff
@@ -0,0 +1,36 @@
...
+RUN apt update && apt install -y wget build-essential pkg-config sqlite3 libsqlite3-dev libssl-dev clang
+
+# Download and unzip PROJ
+WORKDIR /tmp
+RUN wget https://github.com/OSGeo/PROJ/releases/download/${PROJ_VERSION}/proj-${PROJ_VERSION}.tar.gz
+RUN tar -xzvf proj-${PROJ_VERSION}.tar.gz
+
+# Compile PROJ and install in /usr
+WORKDIR proj-${PROJ_VERSION}
+RUN ./configure --prefix=/usr
```
rouault, line 13: shouldn't we use a /opt/proj-${PROJ_VERSION} prefix, so that this package doesn't conflict with a regular proj package ? I can imagine that people might want to still use their GDAL, QGIS, etc stack from the packaging system. And have newer PROJ alongside

Dockerfile.deb-packager
```diff
@@ -0,0 +1,36 @@
+FROM debian:stretch-slim
```
rouault: make the base distro as an ARG (that can default todebian:stretch-slim) so that people have a chance to use an alternative ?

rouault (commented):
Just a few remarks. I don't take position myself on the appropriateness for the project to provide this.

Dockerfile.deb-packager
woshilapin: First of all, note that all of this build happen in a Docker container so it won't pollute the host.

In the hypothesis that they install it through the Debian package manager, I think they should expect only one version (unless `apt-get` allow to manage multiple versions).

I could see a rational to do what you propose if the package was manually installed with `dpkg -i proj_6.1.0_amd64.deb`. But ideally, this file would be hosted on some Debian package repository and it would be installed through `apt-get` (which means only one version installed on the machine as far as I know).

Dockerfile.deb-packager
woshilapin: Good idea, here is the change 19b5c0d17e894b8244960aecab0532ca1fe7388e.

kbevers:
> I don't take position myself on the appropriateness for the project to provide this

I will do that then. While I can see the value in having a recent PROJ version readily available for Debian users, ultimately it is not the responsibility of the PROJ project to make that happen. It is a maintainance hell for the PROJ project to take on contributions like this. We cannot reliably provide updates and fixes to all the various packaging systems out there, so I am rejecting this pull request.

We have a good collaboration with the maintainer of the PROJ debian package and we are working on solving the breaking dependencies that are the cause of the outdated PROJ version on non-experimental releases of Debian. That is the price you pay when running a operating system like Debian - dependency management it excellent but from time to time you are stuck with an older version of a package until it meshes well with the rest of the ecosystem.

I suggest you keep this Docker package alive in a separate repository so that others may benefit from your work. Your Docker image seems to be in excellent shape and can very likely provide value for many other Debian users in a similar situation as yours, this is just not the correct place to maintain it.

woshilapin:
Sounds fair, thank you for the detailed answer :-)

For now, I think I'll let the branch live on my repository so those who look at this issue, you can still access the [`Dockerfile`](https://github.com/woshilapin/proj/blob/deb-builder/Dockerfile.deb-packager) and some help about how to use it in the [`README.md`](https://github.com/woshilapin/proj/blob/deb-builder/README.md#build-the-deb-debian-package).

I might create a repository later if I need to maintain this file (and hopefully not forget to put back a link here about the new repository!).

woshilapin:
Just as an information, can you tell who is the DEB maintainer? We might be able to collaborate on that topic.

kbevers:
ping @sebastic

sebastic:
Not using Debian packaging tools to build `.deb` packages makes Debian people sad.

If you want to provide a proper package, you would build the [proj source package](https://salsa.debian.org/debian-gis-team/proj) with git-buildpackage like the package maintainers do (as documented in the [Debian GIS team policy](https://debian-gis-team.pages.debian.net/policy/packaging.html#git-packaging)), or build the [.dsc](http://deb.debian.org/debian/pool/main/p/proj/proj_6.1.1-1~exp1.dsc) with dpkg-buildpackage which all proper Debian packaging tools use under the hood, and install the resulting package in your container.

For added value to your users also install the regional datumgrid packages in the container which are not packaged for Debian.

woshilapin:
It was my first DEB package building so it obviously is not the best first shot :-)

That said, I'm using `dpkg` to build the package. The only reason I used Docker to build it (note that I'm not using Docker to install the package, just for building it), is that I was not able to install `dpkg` from the package manager of my Linux Distribution (which is ArchLinux). Docker was a way to bypass that problem.

I'll take a look in all the documentation later when I come back to it. For now, the package of version 6.1.0 I built for my project is sufficient and I hosted it on an Artifactory instance for my needs.

One last question, I'd like to know what is blocking the update of the package on Debian ?  Because I'm worry that whatever I did is causing me problem at some point. Thank you.

sebastic:
Not all reverse dependencies build successfully with PROJ 6, see:

 https://lists.debian.org/debian-gis/2019/04/msg00004.html

And the follow-ups in that thread.

Because of the freeze in preparation for the Debian buster stable release, not much work has been done in preparation of the proj transition in the past few months. Some packages got fixes in the mean time, but the old vtk packages are still an issue. With the release of buster last weekend I'll do a new round of rebuilds soon to evaluate the current state of the reverse dependencies, and request a transition slot (if the impact is manageable) after filing bugreports for the packages that still fail to build.

PR #1558, opened 2019-07-25
Database: update to EPSG 9.7.0

rouault:


nyalldawson:
Very quick turnaround, lovely stuff!

Do these upgrades get backported to stable releases?

rouault:
@kbevers This would go to master, which will be released as PROJ 6.2 right ? Anyway this update itself could probably be candidate for a backport to a maintenance branch if there was an active one currently.

snowman2:
Maybe this could be something where the PROJ database and the PROJ source code could be separate? Similar to `proj-datumgrids`, you could also distribute `proj-database`?

hobu:
> Maybe this could be something where the PROJ database and the PROJ source code could be separate? 

I lean toward no on that because the db potentially changes the behavior and results of the software. I think we want releases pinned to known versions rather than letting users manage it. Also, releasing the db as its own entity brings up a number of challenging packaging issues (licensing and such) that we'd rather not have to revisit now that we have seem to have overcome them.

rouault:
> I lean toward no on that because the db potentially changes the behavior

The model of the DB might also change, and will certainly change in the coming months because EPSG is revising their data model to align more closeling with ISO19111:2019 and this will certainly require non trivial changes to the structure of the PROJ db and the code as well.

snowman2:
Okay, sounds like it definitely should remain as is then.

kbevers:
> @kbevers This would go to master, which will be released as PROJ 6.2 right ? Anyway this update itself could probably be candidate for a backport to a maintenance branch if there was an active one currently.

Yes and yes. 6.1.1 included an update to the EPSG database as well so that practice has already been established.

PR #1559, opened 2019-08-05
Fix tests still running despite being disabled in Release 5.2

HeroCC:
This is currently in master / release 6.0, however in the lab I work at we are currently on Proj 5.2, so it would be great to backport this fix despite it being an older release. CMake 2.8.12.2 (testing on CentOS7) fails to download the GoogleTest files, and without this patch there is no way to disable testing.

I'm not sure what your policy is on backporting fixes to older versions / releasing them, but I figured it may help someone to have this here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,5 +164,6 @@ add_subdirectory(nad)
 add_subdirectory(src)
 add_subdirectory(man)
 add_subdirectory(cmake)
-add_subdirectory(test)
-
+if(PROJ_TESTS)
+  add_subdirectory(test)
+endif(PROJ_TESTS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@
 cmake_minimum_required(VERSION 2.6.0 FATAL_ERROR)
 
 # For historic reasons, the CMake PROJECT-NAME is PROJ4
-project(PROJ4 LANGUAGES C CXX)
+project(PROJ4 C CXX)
 set(PROJECT_INTERN_NAME PROJ)
 
 if (NOT CMAKE_VERSION VERSION_LESS 3.1)


kbevers:
> I'm not sure what your policy is on backporting fixes to older versions / releasing them, but I figured it may help someone to have this here.

We are not going to make any more releases based on the 5.x branches, so adding patches to those is not going to have an effect unless someone deliberately builds from the 5.2 branch.

HeroCC:
Ah OK, up to you if you want to merge then. The other fix I needed to get a minimal build working was removing the "LANGUAGES" keyword from https://github.com/OSGeo/PROJ/blob/5.2/CMakeLists.txt#L15, I can make a PR for that too if you'd like.

mwtoews:
There have been ~17 commits on the [5.2 branch](https://github.com/OSGeo/PROJ/commits/5.2) since the [5.2.0 tag](https://github.com/OSGeo/PROJ/commits/5.2.0), so I don't see any reason not to add more to this branch, without a formal release.

And correct, LANGUAGES is a CMake3 feature that should not have been in the old CMake2 setup configured at the time.

kbevers:
> There have been ~17 commits on the 5.2 branch since the 5.2.0 tag, so I don't see any reason not to add more to this branch, without a formal release.

Those are all documentation updates as far as I can tell and are thus slightly different since we build the website based on the maintenance branch for the active release. But sure, we can merge it, I just wanted to stress the fact that this is never going into an actual release.

HeroCC:
> But sure, we can merge it, I just wanted to stress the fact that this is never going into an actual release.

Fine with me! When you merge it it may be good to add a note to the commit details pretty much saying that. I'll throw the "LANGUAGES" patch in this PR if it's alright with you, then up to you when to merge.

PR #1566, opened 2019-08-19
 C API: add proj_context_set_autoclose_database() to automatically close database (fixes #1565)

### LINKED ISSUE #1565: PJ_CONTEXT not releasing file handle for sqlite until destroyed
Related to:
- https://github.com/pyproj4/pyproj/issues/374

The solution for pyproj was to use a global context so only one file handle is added. This is more efficient as there is some overhead when creating a new PJ_CONTEXT. 

However, I have noticed this impacting the use of rasterio recently when upgrading to use PROJ 6 where I had too many open file handles. I expect that this may impact other applications.

Additionally. this could impact applications that are multithreaded with one PJ_CONTEXT per thread as well (https://github.com/pyproj4/pyproj/issues/386)

What are the reasonings behind the current implementation?
Are there alternative options? If so, what are the upside/downsides of switching to one of them?

rouault:


snowman2:
Thanks for this addition! This works well for the tests I ran on my machine. No issues found. Apparently the global context idea I had does not work on Windows for some reason (issue referenced in this PR), but this should fix the pyproj issues after removing the global context.

snowman2:
Is this going to make it into 6.2.0?

rouault:
Maybe @kbevers want to have a look at that ?

kbevers:
If everyone is happy with the PR as it is now, sure, let's get in 6.2.0 :-)

PR #1567, opened 2019-08-22
Add proj_create_crs_to_crs_from_pj()

rouault:
I've been frustrated a number of times with proj_create_crs_to_crs()
not accepting a PJ* object for the source and target CRS.
And thus constraining to go back to WKT2 in a artificial way.

### DIFF
--- a/docs/source/development/reference/functions.rst
+++ b/docs/source/development/reference/functions.rst
@@ -153,6 +153,18 @@ paragraph for more details.
     :type `area`: PJ_AREA
     :returns: :c:type:`PJ*`
 
+.. c:function:: PJ* proj_create_crs_to_crs_from_pj(PJ_CONTEXT *ctx, PJ *source_crs, PJ *target_crs, PJ_AREA *area, const char* const *options)
+
+    .. versionadded:: 6.2.0
+
+    Create a transformation object that is a pipeline between two known
+    coordinate reference systems.
+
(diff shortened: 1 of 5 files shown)


docs/source/development/reference/functions.rst
```diff
@@ -153,6 +153,18 @@ paragraph for more details.
...
 
+.. c:function:: PJ* proj_create_crs_to_crs_from_pj(PJ_CONTEXT *ctx, PJ *source_crs, PJ *target_crs, PJ_AREA *area, const char* const *options)
+
+    .. versionadded:: 6.2.0
+
+    Create a transformation object that is a pipeline between two known
+    coordinate reference systems.
+
+    This is the same as :c:func:`proj_create_crs_to_crs` except that the source and
+    target CRS are passed as PJ* objects.
```
kbevers: The `PJ` objects has to be the kind where `proj_is_crs()` returns true, right?

rouault: yes, I hope this is clear enough from the argument name and above text.

kbevers: It is reasonable clear, but it can't hurt to explicitly state that the PJ has to be of the CRS variety.

rouault: OK, precision just added

kbevers (approved):

PR #1570, opened 2019-08-27
Doc: add link to final WKT2:2019 document

rouault:
This PR is boring by itself, but didn't want to commit that in master as we're in RC mode and wasn't sure what our policy is.

Another more important question (but boring itself) is that due to the lennnnnngthy pubication process what I thought was going to be WKT2:2018 ended up being WKT2:2019 / ISO-19162:2019. All references in PROJ documentation, code... and API are about WKT2:2018. Should we do something about that ? Like switching to 2019, and having alias enumeration values for 2018 to go on 2019 ?

### DIFF
--- a/src/general_doc.dox
+++ b/src/general_doc.dox
@@ -102,11 +102,10 @@ definitions, and coordinate operations. This is an implementation of
 
 PROJ implements the two following revisions of the standard:
 
-\subsubsection WKT2_2018 WKT2:2018
+\subsubsection WKT2_2018 WKT2:2018 / WKT2:2019
 
-[OGC 18-010r1, 2018-06-08, WKT2-2018]
-(https://portal.opengeospatial.org/files/?artifact_id=79826)
-(not yet adopted, at time of writing)
+[OGC 18-010r7, 2019-06-24, WKT2-2019]
+(http://docs.opengeospatial.org/is/18-010r7/18-010r7.html)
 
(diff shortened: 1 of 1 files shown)


kbevers:
> This PR is boring by itself, but didn't want to commit that in master as we're in RC mode and wasn't sure what our policy is.

In the end it doesn't really matter since this only affect the website and PDF. Never the less it is probably best to merge this after 6.2.0 is out. And of course bring it into the 6.2 maintenance branch once that is set up.

> Should we do something about that ? Like switching to 2019, and having alias enumeration values for 2018 to go on 2019 ?

That would be nice, yes. Seems like a thing for PROJ 7 :-)

rouault:
> Seems like a thing for PROJ 7 :-)

Could be done in a completely backward compatible way.

kbevers:
> Could be done in a completely backward compatible way.

So you would put it in 6.2.1?

rouault:
> So you would put it in 6.2.1?

Hum, not really something for a pure bugfix version. Was more thinking about a 6.3.0 if there's ever one.

kbevers:
> Hum, not really something for a pure bugfix version. Was more thinking about a 6.3.0 if there's ever one.

Yeah, my thinking was the same. I just happen that the next scheduled release where it would fit is 7.0. But I'll soon start a discussion on the next version on the mailing list so it may change depending on how the discussion pans out. Anyway, let's get this merge for now :)

PR #1571, opened 2019-08-29
grid_alternatives.sql: Add DLX_ETRS89_geo.gsb and D73_ETRS89_geo.gsb …

PedroVenancio:
…grids in grid_alternatives (added in proj-datumgrid-europe package per https://github.com/OSGeo/proj-datumgrid/pull/54)


PedroVenancio:
Travis and Cirrus checks are failing with

`Error: near line 34951: insert on grid_alternatives violates constraint: original_grid_name must be referenced in grid_transformation.grid_name`

but both original_grid_name I'm inserting are present in grid_transformation.grid_name:

https://github.com/OSGeo/PROJ/blob/master/data/sql/grid_transformation.sql#L193

https://github.com/OSGeo/PROJ/blob/master/data/sql/grid_transformation.sql#L194

Any hint about what can be causing the fails?

rouault:
> Any hint about what can be causing the fails?

Yes, subtle difference in the capitalization : `DLx_ETRS89_geo.gsb` (in grid_transformation.sql) vs `DLX_ETRS89_geo.gsb` in grid_alternatives.sql. But the better would perhaps be that you fix https://github.com/OSGeo/proj-datumgrid/pull/54 to use the exact case used by EPSG ?

PedroVenancio:
Hi @rouault 

Thank you very much! I think it is ready now. And also https://github.com/OSGeo/proj-datumgrid/pull/54

Issue #1479, opened 2019-05-23
Windows proj.lib instead of/in additon to proj_<major>_<minor>.lib


snowman2:
Thoughts on this? 
https://github.com/conda-forge/proj.4-feedstock/pull/46

Are there reasons that this has not been done already?

kbevers:
For the lazy that can't be bothered to click the link (myself including, hence the long answer time):

The linked issue adresses that on Windows the CMake build will output `proj_6_1.lib` and not `proj.lib`. This causes problems since packages depending on PROJ need to be updated every time a new version of PROJ is released.

What is common practice on Windows regarding this? Is it always just `<name>.lib` with no mentioning of the version number? If so, how is compatibility between versions evaluated?

Would it be sufficient (and possible?) to create a symbolic link to the versioned lib-file?

busstoptaktik:
> If so, how is compatibility between versions evaluated?

@kbevers since this is for static linking, at least the case of trying to call new API entries on an old `.lib` will be diagnosed by the linker at link time. And once the material is statically linked it doesn't matter what you do the the installed `.lib`, since its relevant parts will have been statically included in the `.exe`

kbevers:
Right, so no real consequences of switching to producing a `proj.lib` file? If so, we should make that change with version 7, as this is somewhat a breaking change.

JackBoosY:
Hi @kbevers @busstoptaktik , I'm recently upgrading the proj version in vcpkg(PR [#7192](https://github.com/microsoft/vcpkg/pull/7192)), and then I found that the name generated by proj in Windows and Unix is **inconsistent**. Some proj-dependent libraries (such as gdal and liblas) use proj's default library name in windows as proj.lib, which is obviously inconsistent with the generated name.

To solve this problem, we have two ways:
1. Fix the library name generated by proj as proj, but this obviously violates your original intention (refer to _cmake/ProjUtilities.cmake_: 105).
2. Fix the name of the libraries that depends on proj in each of the erroneous ports.

So, I want to hear from your authors.

Thanks.

kbevers:
I think we should change the CMake setup so that `proj.lib` is always created. As far as I can tell it is what most people expect to get.

[...]

PR #1581, opened 2019-09-03
stick with proj.lib on windows

jef-n:
I'd suggest to keep using proj.lib as library name - even if the dll is called differently.  Had to adapt the lib name twice in the last few days for QGIS and GDAL…


rouault:
There's an issue with proj.lib coming from the library, and proj.lib coming from the src/apps/proj.cpp utility. This shows up only in the x86 builds which use -DBUILD_LIBPROJ_SHARED=OFF
Probably something should be tweaked in src/bin_proj.cmake to make it use a non conflicting name

jef-n:
> 
> 
> There's an issue with proj.lib coming from the library, and proj.lib coming from the src/apps/proj.cpp utility. This shows up only in the x86 builds which use -DBUILD_LIBPROJ_SHARED=OFF
> Probably something should be tweaked in src/bin_proj.cmake to make it use a non conflicting name

on it

kbevers:
Resolves https://github.com/OSGeo/PROJ/issues/1479

Issue #1579, opened 2019-09-03
Error in Bertin1953?


rschmunk:
The final "north warp" at lines 69-71 in the Bertin1953 projection code has that

```
    if (xy.y > 0.) {
      xy.x *= 1. + d / 1.5 * xy.x * xy.x;
    }
```

However, in attempting to duplicate the rendering of the Bertin 1953 implemented in D3 and shown in the gallery at https://github.com/d3/d3-geo-projection , I found that the northern hemisphere wasn't coming out right. Double-checking the code for that warp, I found that the Javascript code on https://github.com/d3/d3-geo-projection/blob/master/src/bertin.js alters the value for xy.y (or r[1] in the d3-geo repo) rather than xy.x. Thus, if the JS code in the d3-geo repo is correct, then the code here in PROJ should read

```
    if (xy.y > 0.) {
      xy.y *= 1. + d / 1.5 * xy.x * xy.x;
    }
```

@Fil, can you confirm?

Fil:
Looks like you're right — how did I not see that mistake? 😱https://github.com/OSGeo/PROJ/commit/6b0567b3c01ada8412a446310af9784124453099#diff-31a3d34b36547f14e56143393932f3efR71

Here's a drawing that shows what the correct version should be:

![correct](https://user-images.githubusercontent.com/7001/64147662-4ba72a80-ce21-11e9-8c8d-e033628d5649.png)


The following shows what you would get with the incorrect code:
![incorrect](https://user-images.githubusercontent.com/7001/64147675-5c57a080-ce21-11e9-9dde-4e10758dd379.png)


(I think my problem when I ported the JS code to CPP is that I am struggling to draw with PROJ.)

How can we fix — and test — this?

kbevers:
> How can we fix — and test — this?

Seems like you've already made fix. Submit it in a pull request. As part of the pull request it would be great to have the figure in the docs updated also, so that the projection is displayed correctedly. Here is the current version: https://proj.org/operations/projections/bertin1953.html

You can update the figures by running the `plot.py` script in `docs/source/plot`. Something like `python3 plot.py plotdefs.json  ../source/operations/projections/images/ bertin1953`

There's a conda environment file in the plot folder which can be used to set up a suitable environment for building the plots. It is a bit old so some packages may need to be updated but it shouldn't be too difficult to figure out.

[...]

PR #1582, opened 2019-09-03
Debug Bertin1953

rschmunk:
Northern hemisphere "warp" implemented incorrectly. See #1579


kbevers:
Tests fail. You need to adjust the test case in `test\gie\more_builtins.gie` as per:

```
     -----
     FAILURE in more_builtins.gie(112):
     expected: 0.937431733950 0.700084268525
     got:      0.920230566844   0.713170409026
```

rschmunk:
Thanks, @kbevers. I was writing up the projection in another language, so do not have PROJ installed and could not generate that test value.

PR #1583, opened 2019-09-04
proj_create_crs_to_crs_from_pj(): make the PJ* arguments const PJ*

rouault:
This will make the life of C++ users using the C API slightly easier.
There's no ABI or API backward compatibility issue in doing that change as
constness of arguments is not retained in the C ABI, and any non-const
object passed as argument is implicitly converted as a const object if
that is what is mentionned in the argument list.

### DIFF
--- a/src/4D_api.cpp
+++ b/src/4D_api.cpp
@@ -946,7 +946,7 @@ static PJ* add_coord_op_to_list(PJ* op,
 }
 
 /*****************************************************************************/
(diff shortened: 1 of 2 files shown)


kbevers:
> /bin/bash: q : commande introuvable

is this a copy'n'paste gone wrong?

rouault:
> is this a copy'n'paste gone wrong?

yep, just fixed in commit message and PR text

PR #1584, opened 2019-09-04
C API: add proj_create_ellipsoidal_3D_cs()

rouault:


kbevers:
In which cases would this be used, that can't be done with the current API functions? This just for my understanding, I am not against having the function.

rouault:
> In which cases would this be used, that can't be done with the current API functions?

typically people starting with a geocentric or geographic 2D CRS and wanting to create a geographic 3D CRS from it, after having fetched the datum. This could be done with proj_create_cs() but this is a bit more work. This is quite "symmetrical" to proj_create_ellipsoidal_2D_cs()

PR #1589, opened 2019-09-05
PROJStringParser::createFromPROJString(): avoid potential infinite recursion (fixes #1574)

### LINKED ISSUE #1574: PROJ core dumps when setting data directory on specific context and it is not set on global context
https://github.com/pyproj4/pyproj/pull/419

This is a wierd edge case that has taken me some time to track down as the cause has been elusive for quite some time due to the core dumping. However, I believe the symptoms and the solution should help find a solution in PROJ I believe.

In the case where the data directory is not able to be determined by PROJ in the default locations on the global context, it will core dump when attempting to use `proj_create` with a perfectly valid custom context that has the proper data directory set. However, this does not happen in all instances. It only happens when using the `+init=` syntax.

For example:
```python
>>> from pyproj import CRS
>>> cc = CRS("+proj=utm +zone=33 +ellps=WGS84 +datum=WGS84 +units=m +no_defs")
>>> cc = CRS(init="epsg:4326")
Segmentation fault (core dumped)
```
My guess is it has to do with:
https://github.com/OSGeo/PROJ/blob/f059b90ab4c52a67344886a76deecaeb53299ebd/src/iso19111/c_api.cpp#L437-L440

The fix I applied in pyproj was to set the data directory for the global PROJ context along with the custom context.

rouault:
The exact circumstances are a bit difficult to explain, but they involve
using a non-default context, enabling proj_context_use_proj4_init_rules() on it,
using proj_create(ctxt, "+init=epsg:XXXX +type=crs"), whereas PROJ_LIB is
defined to a directory that has a 'epsg' file in it.

### DIFF
--- a/src/iso19111/io.cpp
+++ b/src/iso19111/io.cpp
@@ -8638,6 +8638,21 @@ static const metadata::ExtentPtr &getExtent(const crs::CRS *crs) {
 
 //! @endcond
 
+namespace {
+struct PJContextHolder {
+    PJ_CONTEXT *ctx_;
+    bool bFree_;
+
+    PJContextHolder(PJ_CONTEXT *ctx, bool bFree) : ctx_(ctx), bFree_(bFree) {}
+    ~PJContextHolder() {
+        if (bFree_)
+            proj_context_destroy(ctx_);
+    }
+    PJContextHolder(const PJContextHolder &) = delete;
+    PJContextHolder &operator=(const PJContextHolder &) = delete;
+};
+} // namespace
+
 // ---------------------------------------------------------------------------
 
 /** \brief Instantiate a sub-class of BaseObject from a PROJ string.
@@ -8652,8 +8667,10 @@ PROJStringParser::createFromPROJString(const std::string &projString) {
 
     // In some abnormal situations involving init=epsg:XXXX syntax, we could
     // have infinite loop
-    if (d->ctx_ && d->ctx_->curStringInCreateFromPROJString == projString) {
-        throw ParsingException("invalid PROJ string");
+    if (d->ctx_ &&
+        d->ctx_->projStringParserCreateFromPROJStringRecursionCounter == 2) {
+        throw ParsingException(
+            "Infinite recursion in PROJStringParser::createFromPROJString()");
     }
 
     d->steps_.clear();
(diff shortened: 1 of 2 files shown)


rouault:
@snowman2 Would you mind testing this ? Creating a uni ttest for that case is a bit painful, so I didn't do it. Hopefully this fix should be quite robust.

src/iso19111/io.cpp
```diff
@@ -8701,27 +8718,38 @@ PROJStringParser::createFromPROJString(const std::string &projString) {
         const std::string &stepName = d->steps_[0].name;
         if (ci_starts_with(stepName, "epsg:") ||
             ci_starts_with(stepName, "IGNF:")) {
+
+            /* We create a new context so as to avoid messing up with the */
+            /* errorno of the main context, when trying to find the likely */
+            /* missing epsg file */
+            auto ctx = proj_context_create();
```
snowman2, line 8725: Is there a way to pass the context passed in from `proj_create` here? That would definitely help from the perspective of the developer who passed in a context with specific settings configured & search paths set.

snowman2: Ah, I see. Nevermind. Looks like it copies those settings.

snowman2:
> @snowman2 Would you mind testing this ? Creating a uni ttest for that case is a bit painful, so I didn't do it. Hopefully this fix should be quite robust.

Sure, it may be a while as I need to double check that it actually worked and am not getting false positives.

snowman2:
Thanks for the fix! Seems to work: 

Steps:

1. `mv share/proj share/proj.inactive`
2. Verify PROJ cannot find DB
```
$ ./bin/projinfo "epsg:4326"
WARNING: Cannot create database connection: Cannot find proj.db
input string: parsing of user string failed: no database context specified
```
3. Test pyproj
```python
>>> from pyproj.datadir import get_data_dir
>>> get_data_dir()
  File "<stdin>", line 1, in <module>
  File "/home/snowal/miniconda3/envs/pyproj/lib/python3.7/site-packages/pyproj/datadir.py", line 99, in get_data_dir
    "Valid PROJ data directory not found. "
pyproj.exceptions.DataDirError: Valid PROJ data directory not found. Either set the path using the environmental variable PROJ_LIB or with `pyproj.datadir.set_data_dir`.
>>> from pyproj.datadir import get_data_dir, set_data_dir
>>> set_data_dir("/home/snowal/scripts/proj.4/install_dir/share/proj.inactive")
>>> get_data_dir()
'/home/snowal/scripts/proj.4/install_dir/share/proj.inactive'
>>> from pyproj import CRS
>>> cc = CRS(init="epsg:4326")
>>> cc
<Geographic 2D CRS: +init=epsg:4326 +type=crs>
Name: WGS 84
Axis Info [ellipsoidal]:
- lon[east]: Longitude (degree)
- lat[north]: Latitude (degree)
Area of Use:
- name: World
- bounds: (-180.0, -90.0, 180.0, 90.0)
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich
```
4. Set PROJ_LIB to 5.2.0 data dir with `epsg` file
```
export PROJ_LIB=/home/snowal/miniconda3/envs/geocube/share/proj
```
5. Verify PROJ cannot find `proj.db`
```
$ ./bin/projinfo "epsg:4326"
WARNING: Cannot create database connection: Cannot find proj.db
input string: parsing of user string failed: no database context specified
```
6. Verify pyproj still works:
```python
>>> from pyproj.datadir import get_data_dir
>>> get_data_dir()
  File "<stdin>", line 1, in <module>
  File "/home/snowal/miniconda3/envs/pyproj/lib/python3.7/site-packages/pyproj/datadir.py", line 99, in get_data_dir
    "Valid PROJ data directory not found. "
pyproj.exceptions.DataDirError: Valid PROJ data directory not found. Either set the path using the environmental variable PROJ_LIB or with `pyproj.datadir.set_data_dir`.
>>> from pyproj.datadir import get_data_dir, set_data_dir
>>> set_data_dir("/home/snowal/scripts/proj.4/install_dir/share/proj.inactive")
>>> get_data_dir()
'/home/snowal/scripts/proj.4/install_dir/share/proj.inactive'
>>> from pyproj import CRS
>>> cc = CRS(init="epsg:4326")
>>> cc
<Geographic 2D CRS: +init=epsg:4326 +type=crs>
Name: WGS 84
Axis Info [ellipsoidal]:
- lon[east]: Longitude (degree)
- lat[north]: Latitude (degree)
Area of Use:
- name: World
- bounds: (-180.0, -90.0, 180.0, 90.0)
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich
```

PR #1590, opened 2019-09-06
BUG: add SANITIZE_CTX for coordoperation is_instantiable and has_ballpark_transformation

snowman2:
This one was causing core dumping for me - had me digging through my code to find it and eventually I pinpointed the issue to these functions.

### DIFF
--- a/src/iso19111/c_api.cpp
+++ b/src/iso19111/c_api.cpp
@@ -6091,6 +6091,7 @@ PJ *proj_create_conversion_equal_earth(PJ_CONTEXT *ctx, double center_long,
 
 int proj_coordoperation_is_instantiable(PJ_CONTEXT *ctx,
                                         const PJ *coordoperation) {
+    SANITIZE_CTX(ctx);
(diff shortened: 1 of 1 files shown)


kbevers:
Can you explain the circumstances of the core dump?

rouault:
> Can you explain the circumstances of the core dump?

Pretty simple I think. If you passed ctx == NULL to this function, it failed to use the default global context, and thus a null pointer dereferece would occur.

@snowman2 So you use an implicit context in some places of your code ?

kbevers:
> Pretty simple I think. If you passed ctx == NULL to this function, it failed to use the default global context, and thus a null pointer dereferece would occur.

Thanks. I am just trying to get enough info in here so I don't have to spend too much time writing release notes in two months time. This helps a lot!

snowman2:
> So you use an implicit context in some places of your code ?

Not currently. Just testing it out for a potential solution to #1591 when I ran into this.

PR #1595, opened 2019-09-06
Quick proof-of-concept: building PROJ without sqlite dependencies

funchal:


kbevers:
Let me prefix this by saying that I am not very well-versed in these parts of the code-base.   @rouault would be able to a better review of this.

I get the general idea. Your approach seems reasonable to me and so far it can mostly be described as refactoring, to be able to add more types of database-classes. So far so good. I wonder what the impact would be of turning the SQlite dependency off? I reckon that most of the tests will fail. That of course needs to be dealt with somehow.

include/proj/coordinateoperation.hpp
```diff
@@ -155,10 +155,19 @@ class PROJ_GCC_DLL CoordinateOperation : public common::ObjectUsage,
...
-    PROJ_PRIVATE :
-        //! @cond Doxygen_Suppress
-        PROJ_FOR_TEST CoordinateOperationNNPtr
-        shallowClone() const;
+  PROJ_PRIVATE :
+    //! @cond Doxygen_Suppress
+    PROJ_FOR_TEST CoordinateOperationNNPtr
+    shallowClone() const;
+
+    PROJ_INTERNAL void setCRSs(const crs::CRSNNPtr &sourceCRSIn,
```
rouault, line 163: what justifies setCRS and setProperties to be public/PROJ_PRIVATE and no longer protected ?

include/proj/io.hpp
```diff
@@ -787,14 +807,77 @@ class PROJ_GCC_DLL PROJStringParser {
 
 // ---------------------------------------------------------------------------
 
-/** \brief Database context.
+/** \brief Database context interface.
  *
  * A database context should be used only by one thread at a time.
  */
-class PROJ_GCC_DLL DatabaseContext {
+class PROJ_GCC_DLL DatabaseContext :
```
rouault, line 814: I'd have expected this interface to be at the SQL level. Here you have a few "high level" methods in the interface

include/proj/io.hpp
```diff
@@ -803,145 +886,143 @@ class PROJ_GCC_DLL DatabaseContext {
...
  *
  * A AuthorityFactory should be used only by one thread at a time.
  *
  * \remark Implements [AuthorityFactory]
  * (http://www.geoapi.org/3.0/javadoc/org/opengis/referencing/AuthorityFactory.html)
  * from \ref GeoAPI
  */
-class PROJ_GCC_DLL AuthorityFactory {
+class PROJ_GCC_DLL AuthorityFactory  :
+        public util::nn_enable_shared_from_this<AuthorityFactory> {
```
rouault, line 962: why this derivation for util::nn_enable_shared_from_this<AuthorityFactory> ? I had avoided deriving from std::enable_shared_from_this<> to avoid potential issues with MSVC and exporting interfaces of std:: namespace which used to cause issues in the past

rouault: Same remark as a previous one, but with an increased criticality. You put in the interface really high level functionnalities. Implementing correctly some of those methods is really involved and requires good knowledge of the database model. It might - and will - change in a near future as well, so maintaining that for external implementations of this class will be a lot of work

rouault (commented):
Besides my inline remarks,
- you've prototyped changes only at the C++ level. Do you intend ultimately to be able to use the C API with alternate DB implementations ? Open question
- if you go to having AuthorityFactory as an implementation, which I'm not a big fan of, you'll probably need to be able to run the regression test suite with your implementation as well, so that might need some changes in the unit tests to centralize the instantiation of the right factory.
- I think it will be reasonable to merge proposed changes only once you get a working alternate implementation of the DB

rouault:
Suggesting to close this pull request as it is not in a mergeable state.

kbevers:
> Suggesting to close this pull request as it is not in a mergeable state.

+1. This seem to have lost momentum. If momentum is regained at a later stage it can be re-opened.

Issue #1597, opened 2019-09-11
PROJ 6 issues with +init=epsg:2056 (CH1903+/LV95)


dg0yt:
Using the old API with PROJ 6 instead of 4.9.3, we noticed wrong transformation results between EPSG:4326 and EPSG:2056 (CH1903+/LV95) (https://github.com/OpenOrienteering/mapper/issues/1325). Even after switching to the new API (with legacy proj4 init rules enabled), results are correct only if the `no_defs` parameter is not used. Despite all the legacy stuff: This is a clear regression which should not happen at least when using the old API.

I was now able to create a test executable which uses the [EPSG:2056 center of projection](https://map.geo.admin.ch/?lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&E=2600000&N=1200000&zoom=13&crosshair=marker) for demonstration. Note that only the last line is matching the expected results.

```
expected:    	2600000.00	1200000.00
pj_transform:	2599927.44	1199852.66	(+init=epsg:2056 +no_defs)
pj_transform:	2599927.44	1199852.66	(+init=epsg:2056)
pj_trans:    	2599927.44	1199852.66	(+init=epsg:2056 +no_defs)
pj_trans:    	2599999.82	1199999.69	(+init=epsg:2056)
```
[proj-test.zip](https://github.com/OSGeo/PROJ/files/3599466/proj-test.zip) (updated)

rouault:
@dg0yt  Potential fixes in https://github.com/OSGeo/PROJ/pull/1600 . Note however as indicated in the PR that the commit really fixing your issue could be controversial. Basically one cannot expect PROJ6 to be perfectly backward compatible with previous PROJ versions. Hence the deprecation for the +init=epsg:XXXX syntax. Using proj_create_crs_to_crs(ctx, "EPSG:4326", "EPSG:2056", NULL) is the strongly recommended way of proceeding. Or fully expand the +init=epsg:XXXX to their "+proj= ...." content including +towgs84 clauses.
[...]

PR #1600, opened 2019-09-11
Fixes related to #1597

rouault:
@kbevers The first 2 and the last commits in this PR are non controversial, but the third one (https://github.com/OSGeo/PROJ/commit/18c553a45de0d68e08f6588395a9b5a044b44272 : PROJ string ingestion: adds +towgs84= when expanding +init=epsg:XXXX) is more arguable.  It is really hard to perfectly emulate PROJ4/PROJ5 behaviour. Opinion welcome


kbevers:
> It is really hard to perfectly emulate PROJ4/PROJ5 behaviour. Opinion welcome

I think it is best to leave it as is. This is one of those cases where PROJ6 is almost incompatible with previous versions. Effort of emulating the behaviour of older versions is not worth the trouble.

rouault:
Reworked version of this PR in https://github.com/OSGeo/PROJ/pull/1614

PR #1618, opened 2019-09-17
Remove unneeded C99 compatibility functions from math.cpp and proj_math.h

cffk:
The supported C99 math functions provided by math.cpp are thus

   hypot
   log1p
   asinh
   atanh
   copysign
   cbrt
   remainder
   remquo
   round
   lround

Make compiler checks in CMakeLists.txt and configure.ac consistent with
this set.

Make geodesic.c use the math.cpp defined (instead of the internally
defined) versions of

   hypot
   atanh
   copysign
   cbrt

This is keyed off the presence of the PROJ_LIB macro.  I had at one
time #1425
suggested supplying an additional macro PROJ_COMPILATION when compiling
geodesic.c.  However, PROJ_LIB seems to fill the bill OK.

The *next* version of geodesic.c (due out in a few weeks) will also use

   remainder
   remquo

All of this is only of concern for C compilers without C99 support.  So
this may become an historical footnote at some point.


rouault:
Do we really need manually redefinition of all those standard functions in maths.cpp ? Couldn't we just use the C++11 std::XXXX functions() ?
That is 
```
double  pj_remquo ( double x, double y, int* quo )
{
  return std::remquo(x,y,quo):
}
```

Another thing to study would be to completely drop proj_math.h & math.cpp and use plain std:XXXX() C++11 functions. If we want geodesic.c to remaing a C file if it is shared elsewhere, we might try to not compile it directly, but have a dummy geodesic_wrapper.cpp that would #include "geodesic.c" and do necessary magic before.

cffk:
I am fine with math.cpp merely calling the C++11 functions provided I am guaranteed that the C++ compiler will support C++11.  Is this the case?  (By the way, the code in math.cpp is only active if the C compiler doesn't support C99, so for many platforms the question is moot.)

rouault:
> I am fine with math.cpp merely calling the C++11 functions provided I am guaranteed that the C++ compiler will support C++11

C++11 is a requirement to build PROJ

> the code in math.cpp is only active if the C compiler doesn't support C99

The issue might have been with Visual Studio, but since we require VS 2015 and according to https://docs.microsoft.com/en-us/previous-versions/hh409293(v=vs.140)?redirectedfrom=MSDN it is supposed to handle C99, we might probably drop math.cpp  @kbevers  ?

cffk:
Well it's not completely straightforward invoking std::hypot, etc. in math.cpp.  The cmath header needs to be included and then there are complaints because proj_math.h defines a macro called hypot.  I'm sure all this could be sorted out easily enough.

However, perhaps it's simpler just to wait until such time as we're willing to make C99 a requirement.  Presumably this can't be too far off.  (However several of the test machines flunk the C99_MATH test.) Then proj_math.h and math.cpp can be gracefully retired.

kbevers:
> we might probably drop math.cpp @kbevers ?

I am happy to get rid of that file if we can.

> perhaps it's simpler just to wait until such time as we're willing to make C99 a requirement.

According to [RFC3](https://proj.org/community/rfc/rfc-3.html#bootstrapping) C99 is already a requirement. It is somewhat soft since all the C-code is C89 compatible but you are allowed to use C99 features.

cffk:
Enforcing C99 compliance would be simple enough.  However, I know that this will break a number of builds.  So I recommend holding off on this until someone can inventory what system are affected.  Perhaps all that needs to be done on such systems is to supply -std=c99 to the compiler.

However, I'd prefer this be decoupled from my updating geodesic.c (which shouldn't involve wide ranging changes).  So unless I hear differently, I'll merge this pull request.

kbevers:
> However, I know that this will break a number of builds

Which builds are those?

> So unless I hear differently, I'll merge this pull request.

Let's just wait a minute and investigate the consequences of removing `math.cpp`/`proj_math.h` a bit more. We can justify (small) breaking changes for the next release, so I don't think that it is a big problem to use C99 math functions without wrapping them in `math.cpp`/`proj_math.h`.

busstoptaktik:
> Which builds are those?

@kbevers Probably all MSVC versions except the latest. MS was extremely feet dragging wrt supporting anything but C89.

rouault:
> Probably all MSVC versions except the latest

We only support MSVC >= 2015 now, and according to some Microsoft docs I've seen, it is supposed to be C99 ready finally (well "mostly compliant, except for the parts where it is not")

kbevers:
> well "mostly compliant, except for the parts where it is not"

and since we build and test with both VS2015 and VS2017 we will discover soon enough if we are using one of those parts where it is not

kbevers:
@cffk you mention in the commit message that 

> All that is left is the handling of isnan (and I've this because math.cpp notes that gie.c uses pj_isnan).

That note in `math.cpp` must be outdated because I can't find any use of either `isnan` or `pj_isnan` in `gie.cpp`. I think that it once was part of the gie app as part of the self-test that it originally had. The tests has been moved elsewhere now and `std::isnan()` is used instead. All this to say that you might as well get rid of the last bits of `math.cpp` and `proj_math.h`.

kbevers:
Travis fails because `pj_isnan` needs to be removed from `scripts/reference_exported_symbols.txt`

rouault:
Looks good !

Probably that in configure.ac we should remove the part
https://github.com/OSGeo/PROJ/blob/21c6d9cccde9318dd38c60b8bc559508bcee1a91/configure.ac#L85:L97 that tries to add -std=gnu89 for gcc ?

And in travis/linux_gcc/install.sh and travis/linux_gcc7/install.sh, I see we have manual override of CFLAGS to add -std=c89 which no longer makes sense. Actually we should probably replace it by -std=c99, to make sure we don't use C11 features

On Travis with gcc, the compilation lines are currently -std=gnu89 -std=c89. I presume that the gnu89 took actually precedence over c89, and gnu89 presumably allows the use of those new C99 math functions. Otherwise I don't understand why the compiler doesn't complain...

PR #1620, opened 2019-09-17
Assorted set of fixes for boring oss-fuzz related errors (mostly divisions by zero in odd situations)

rouault:


kbevers:
Should these be back-ported to 6.2?

rouault:
> Should these be back-ported to 6.2?

None of them are likely to be hit in normal use cases, nor cause crashes or other memory related errors. Mostly to keep up with those super pedantic oss-fuzz issues and clean the backlog

kbevers:
Although not likely to be reached by end-users I still consider them bugs. I'm including these fixes in 6.2.1, we may as well provide them as quickly as possible!

Issue #1621, opened 2019-09-17
Enforce C99 standard for the C compiler


cffk:
This grows out of pull request #1618 which provided some updates to math routines in math.cpp.  However math.cpp should really go away because C99 provides the needed routines.  The job then

* if possible, supply flags to turn on C99 support
* die if C99 support can't be provided

kbevers:
> if possible, supply flags to turn on C99 support

Previously we have enforced C89 which turned out to create problems for a number of users. So that was turned of again, and only applied in our own CI tests where it makes sense (e.g. test that the `proj.h` interface is usable by a C89 only application). Users may have good reasons for enforcing a later standard and we shouldn't stand in the way of that.

> die if C99 support can't be provided

If I may, I'd like to ammend that a bit: "die *gracefully* if...". If we use C99 or C++11 features and a compiler can't use them it will die anyway. Probably with an error message that is not easily understood.

cffk:
Completed by #1624

PR #1624, opened 2019-09-19
Require C99 compiler see #1621

cffk:
Require C99 compiler see #1621

* cmake and autoconf now stipulate C99
* change c89 to c99 in travis jobs
* remove HAVE_C99_MATH checks

(unrelated) relax Visual Studio compatibility check in cmake/project-config-version.cmake.in (VS 2019 can use a VS 2015 library but not vice versa).


cffk:
Well, that went pretty smoothly.  I'll wait a day or so before merging to let others look over the changes.

rouault (approved):

PR #1625, opened 2019-09-19
set PROJ_LIB on test targets for CMake

hobu:
Set PROJ_LIB so test script invocation does not have to manually set the directory.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -50,6 +50,8 @@ build_script:
 
 test_script:
   - echo test_script
+  - set PROJ_LIB=%PROJ_DIR%\share\proj
+  - cd %PROJ_LIB%
   - curl -O https://download.osgeo.org/proj/proj-datumgrid-1.8.zip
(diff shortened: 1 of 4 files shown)


appveyor.yml
```diff
@@ -50,8 +50,6 @@ build_script:
 
 test_script:
   - echo test_script
-  - set PROJ_LIB=%PROJ_DIR%\share\proj
-  - cd %PROJ_LIB%
```
rouault: this cd is currently needed so that the grids are decompressed in the right directory

travis/install.sh
```diff
@@ -86,7 +86,7 @@ VERBOSE=1 make -j${NPROC}
 make install
 # The cmake build is not able to generate the null file, so copy it at hand
 cp /tmp/proj_autoconf_install_from_dist_all/share/proj/null /tmp/proj_cmake_install/share/proj
```
rouault: I believe that this cp and the above comment are no longer valid since data/null is now in git, and no longer generated

PR #1626, opened 2019-09-21
Improve Visual Studio compatibility check in CMake.

cffk:
Make the setting of major toolset version variables parallel that for
cross-compilation.  Add comments so it's easier for the next person
looking at this code to figure what's going on.  Note: no need to unset
variables at the end of the version check file (the file is read in a
nested scope).

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -5,11 +5,24 @@
 file(RELATIVE_PATH PROJECT_ROOT_DIR
   ${CMAKE_INSTALL_PREFIX}/${CMAKECONFIGDIR} ${CMAKE_INSTALL_PREFIX})
 
-if(CMAKE_CROSSCOMPILING)
-  set(CMAKE_CROSSCOMPILING_STR "ON")
-else()
-  set(CMAKE_CROSSCOMPILING_STR "OFF")
-endif()
+# Variables needed by ${PROJECT_NAME_LOWER}-config-version.cmake
+if (MSVC)
+  # For checking the compatibility of MSVC_TOOLSET_VERSION; see
+  # https://docs.microsoft.com/en-us/cpp/porting/overview-of-potential-upgrade-issues-visual-cpp
+  # Assume major version number is obtained by dropping the last decimal
+  # digit.
+  math (EXPR MSVC_TOOLSET_MAJOR "${MSVC_TOOLSET_VERSION}/10")
+else ()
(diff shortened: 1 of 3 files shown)


cffk:
I assume the build failure is some transient problem.  Is there some way (other that forcing a new commit) to trigger the checks again?

snowman2:
> I assume the build failure is some transient problem. Is there some way (other that forcing a new commit) to trigger the checks again?

You can close and reopen the PR.

cffk:
Try checks again.

rouault:
I've manually restarted the job

rouault:
It seems there has been an upgrade in the version of sphinxcontrib-bibtex available in pip, which is installed per travis/linux_gcc/before_install.sh.
The last successful build https://api.travis-ci.com/v3/job/237477314/log.txt used sphinxcontrib_bibtex-0.4.2-py2.py3-none-any.whl which installed Sphinx-1.8.5-py2.py3-none-any.whl.
The failed build installs sphinxcontrib_bibtex-1.0.0-py3-none-any.whl which requires Sphinx >= 2.0, but the later requires Python 3.5, whereas Trusty has only Python 3.4
So the easiest solution I think would be to modify travis/linux_gcc/before_install.sh to do `pip3 install --user sphinxcontrib-bibtex==0.4.2` . Or find a way of using a newer Python version

rouault:
@cffk  Yes it didn't work because the commit added a new command 2 lines after the faulty line at line 18, instead of just modifying the faulty one. Sorry for not having been clearer.

rouault:
I'll try the fix in another pull request, since that will need to be backported to 6.2

rouault:
Potential fix for sphinx issue in https://github.com/OSGeo/PROJ/pull/1627

cffk:
@rouault Ah, sorry, I just blindly added the extra line.  I'll wait for you pull request to be merged and then resubmit.

rouault:
> I'll wait for you pull request to be merged

Done. You can rebase on top of updated master

rouault:
@cffk In those circumstances with a git history with trials & errors, I would suggest a git rebase -i to improve a bit the commit history.

cffk:
OK, thanks for the tip.  I got into the habit of making pedestrian updates to git after being burned in my early days of git usage by incompetently backing out of changes.

rouault:
you can always do a "git checkout -b mybackup_branch" before doing potential destructive changes

PR #1631, opened 2019-09-23
Document the autopromotion to 3D when mix of 2D/3D CRS in createOperations(), and test it (fixes #1623)

### LINKED ISSUE #1623: Accidental autopromotion to 3D for vertical transforms
@kbevers Remember the discussion in #1587 about transformations between a 2D CRS and a 3D one not leading to a transformation term for the vertical part, which lead to the introduction of --3d / promoteTo3D() ?  It turns out that commit eed28e51 accidentally causes a kind of autopromotion (that is to say 'projinfo -s EPSG:4326 -t  EPSG:4326+3855' will propose the egm08_25.gtx grid). I didn't intend that side effect and I'm not sure what needs to be done.

Two possibilities:
- either keep that behaviour. In which case, projinfo should be modified to no longer suggest "Warning: mix of 2D and 3D CRS. Vertical transformations, if available, will not be applied. Consider using 3D version of the CRS, or the --3d switch"
- or amend eed28e51 to be more restrictive, and check that when considering a tuple (vertical CRS, Geographic CRS), the GeogCRS is a 3D one

In any case, the 2D to 3D case should be tested somewhere, so that any behaviour change doesn't get unnoticed :-)

rouault:


kbevers:
Looks good. Have you removed the `--3d` option in `projinfo` or is that still useful?

rouault:
> or is that still useful?

I've kept it. It could be useful if you need to get the WKT of a proper 3D CRS from a 2D one (projinfo --3d EPSG:32631). When using -s + -t, it can also be used, but should normally give the same result as not using it.

kbevers:
Right, that makes sense!

PR #1634, opened 2019-09-24
Upgrade Travis CI distribution

mwtoews:
This PR updates the Travis CI distribution from [`trusty`](https://docs.travis-ci.com/user/reference/trusty/) to [`xenial`](https://docs.travis-ci.com/user/reference/xenial/). The motivation behind this is that Ubuntu Trusty 14.04 is no longer maintained, and therefore an unlikely system in widespread use today. Issues like #1627 are symptoms of using outdated test VMs.

The first commit rearranges the testing scripts, but does not upgrade anything. And the second commit upgrades the test components. The upgrades include:
* From openjdk-7 via apt to openjdk-11 via Travis CI xenial's $JAVA_HOME
* From mingw 4.8 -> 5.3, as supported by xenial
* clang+llvm for csa kept at version 6.0.0, but for ubuntu-16.04
* Furthermore xenial implicitly has many upgrades, specifically:
  - gcc 4.8.4 -> 5.4.0
  - clang 4.0.0 -> 5.0.0
  - cmake 3.9.2 -> 3.12.4
  - python 3.4.3 -> 3.5.2

A few remarks:
* A few upgrades included here may need approval from PSC (e.g. [RFC 2](https://github.com/OSGeo/PROJ/blob/master/docs/source/community/rfc/rfc-2.rst) describes test environment components and their version numbers); let me know and I'll start that discussion on the email list.
* Cppcheck from the `linux_gcc` test was effectively disabled due to too many false positives. This is a temporary measure to take a look at what is going on, for instance [scripts/cppcheck.sh](https://github.com/OSGeo/PROJ/blob/master/scripts/cppcheck.sh) is tested with cppcheck 1.72 as shipped with Ubuntu 16.04 (xenial), but several of the uninitvar errors seem strange.

### DIFF
--- a/docs/source/community/code_contributions.rst
+++ b/docs/source/community/code_contributions.rst
@@ -117,18 +117,19 @@ Preliminary step: install clang. For example:
 
     wget http://releases.llvm.org/6.0.0/clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-14.04.tar.xz
     tar xJf clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-14.04.tar.xz
+    mv clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-14.04 clang+llvm-6
 
 Run configure under the scan-build utility of clang:
 
 ::
 
-    ./clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-14.04/bin/scan-build ./configure
+    ./clang+llvm-6/bin/scan-build ./configure
 
 Build under scan-build:
 
 ::
 
-    ./clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-14.04/bin/scan-build make [-j8]
+    ./clang+llvm-6/bin/scan-build make [-j8]
 
 If CSA finds errors, they will be emitted during the build. And in which case,
 at the end of the build process, scan-build will emit a warning message
@@ -137,7 +138,7 @@ is with someling like
 
 ::
 
-    ./clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-14.04/bin/scan-view /tmp/scan-build-2018-03-15-121416-17476-1
+    ./clang+llvm-6/bin/scan-view /tmp/scan-build-2018-03-15-121416-17476-1
 
 
 This will open a web browser with the interactive report.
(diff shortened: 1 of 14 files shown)


kbevers:
This is great! Thanks. Do you think you can get cppcheck up and running again?

> A few upgrades included here may need approval from PSC (e.g. RFC 2 describes test environment components and their version numbers); let me know and I'll start that discussion on the email list.

That shouldn't be necessary. RFC2 describes the situation at the time it was introduced. It is not a governing document in the same way as RFC3 is.

rouault:
Great work!

>     * for instance [scripts/cppcheck.sh](https://github.com/OSGeo/PROJ/blob/master/scripts/cppcheck.sh) is tested with cppcheck 1.72 as shipped with Ubuntu 16.04 (xenial)

ah, this script initially comes from GDAL, and the above statement of working with cppcheck 1.72 holds for GDAL, but actually not for PROJ... There are indeed a number of errors with it

mwtoews:
I've undone the previous commit using snap, and will set that up as a task to complete later, as more work would be required to revise the PROJ source code to determine whether the errors are useful or not.

For now, what I've done is kept cppcheck at version 1.61, installed using some .deb packages from Launchpad, rather than installing using xenial's version via apt.

This PR is ready for review!

docs/source/community/code_contributions.rst
```diff
@@ -91,9 +91,7 @@ You can run locally ``scripts/cppcheck.sh`` that is a wrapper script around the
 cppcheck utility. It is known to work with cppcheck 1.61 of Ubuntu Trusty 14.0,
 since this is what is currently used on Travis-CI
 (``travis/linux_gcc/before_install.sh``).
-At the time of writing, this also works with cppcheck 1.72 of Ubuntu Xenial
-16.04, and latest cppcheck
-master.
+At the time of writing, this works with the latest cppcheck master.
```
kbevers: In `scripts/cppcheck.sh` you mention that we are using version 1.61 - which is true?

kbevers (approved):
I still think this is great. I like that you've separated repetitive things into environment variables and splitting the before_install scripts into several files is a neat solution.

docs/source/community/code_contributions.rst
mwtoews: hmm, perhaps this last phrase should be taken out, as I certainly saw some new errors with cppcheck 1.89.

kbevers: Or possibly just refrased to say that we use cppcheck as part of the quality control of the code. It is futile to try to remember to update this every time a new version of the software is used.

PR #1637, opened 2019-09-24
io.cpp: fix build on Solaris

rouault:
Rename CS template argument, to avoid conflict with macro in Solaris system headers.

Similar to 2f8bd934860b135044c5122e3272f7cc41ba81e7

### DIFF
--- a/src/iso19111/io.cpp
+++ b/src/iso19111/io.cpp
@@ -4444,12 +4444,12 @@ class JSONParser {
     }
 
     template <class TargetCRS, class DatumBuilderType,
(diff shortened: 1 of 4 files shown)


mloskot:
For archive, related thread is [[PROJ] build failure with GCC 9.1.0](https://lists.osgeo.org/pipermail/proj/2019-September/008900.html)

mloskot:
if the macro is it a function-like macro, then possible workaround is to replace `io::WKTConstants::CS` with  `(io::WKTConstants::CS)`, etc.. That should prevent the macro expansion.

PR #1638, opened 2019-09-25
Add rotation support to the HEALPix projection

Schneegans:
# Motivation

At the German Aerospace Center (DLR) we are developing a VR-Framework called [CosmoScout VR](https://github.com/cosmoscout/cosmoscout-vr). This software can render highly detailed level-of-detail planetary bodies in real-time. The required surface tiles are downloaded from a [MapServer](https://mapserver.org/) instance and are provided in the HEALPix projection. The advantage is, that the data density is equal all around the globe without any singularities at the poles.

However, in order to retrieve the HEALPix base patches as image tiles, the map has to be rotated by 45°. Here is an image from our documentation which illustrates this with the figures in the center row:

image

# Implemented Changes

The file `src/projections/healpix.cpp` has been changed to accept an optional `+rot=` parameter, which has to be given in degrees and defaults to zero.

# Usage

We use this in a custom projection like this and it works very well:

```
<900914> +proj=healpix +lon_0=0 +x_0=2.5 +y_0=2.5 +a=0.900316316157106 +rot=45 +no_defs <>
```


src/projections/healpix.cpp
```diff
@@ -78,6 +79,12 @@ static double sign (double v) {
     return v > 0 ? 1 : (v < 0 ? -1 : 0);
 }
 
+static PJ_XY rotate(PJ_XY p, double a) {
```
kbevers: I suggest you use `angle` instead of `a` here. Makes it immediately clear what's going on

src/projections/healpix.cpp
```diff
@@ -622,6 +637,8 @@ PJ *PROJECTION(healpix) {
     P->opaque = Q;
     P->destructor = destructor;
 
+    Q->rot = pj_param(P->ctx, P->params,"drot").f * M_PI / 180.0;
```
kbevers: Improve readability a bit:

```suggestion
	double angle = pj_param(P->ctx, P->params,"drot").f;
    Q->rot = PJ_TORAD(angle);
```

kbevers (changes requested):
Very interesting. I can't see why we shouldn't include this. Before we do that though you need to update the documentation. Start here `https://github.com/OSGeo/PROJ/blob/master/docs/source/operations/projections/healpix.rst`.

Also, does it make sense to add the rotation parameter to the rHEALPix projection as well?

docs/source/operations/projections/healpix.rst
```diff
@@ -51,6 +51,12 @@ Parameters
 
 .. note:: All parameters for the projection are optional.
 
+.. option:: +rot
+
+    Rotation of the HEALPix map in degrees.
```
rouault: I believe it should be explained that the rotation is done in the cartesian / projected coordinate space, and what is the convention for the orientation (trigonometric ? clockwise ?)

rouault:
One thought I add is that such a +rot parameter could be something generic to all projections, such as +x_0, +y_0. But we could potentially move that to generic code at a later stage if we find that several projections need this.
Was wondering if rot_xy might not be more expressive about in which coordinate space the rotation is done ?

rouault:
Extending heapix tests in test/gie/builtins.gie to test this new parameter would also be welcome

kbevers:
> One thought I add is that such a +rot parameter could be something generic to all projections, such as +x_0, +y_0. But we could potentially move that to generic code at a later stage if we find that several projections need this.

Like a rotation operation? Something like
```
+proj=pipeline +step +proj=healpix ... +step +proj=rot +angle=45.0
```
perhaps?

Schneegans:
Thanks for the feedback! I'll update the PR as soon as possible.

> Also, does it make sense to add the rotation parameter to the rHEALPix projection as well?

I like the idea of supporting this parameter for all projections, however for the rHEALPix projection I don't see any particular use case.

> Extending heapix tests in test/gie/builtins.gie to test this new parameter would also be welcome

I will do this. Are these tests auto-generated? Or are the numbers somehow computed by hand? Or shall I just use the current implementation and assume that the values are correct?

kbevers:
> Or shall I just use the current implementation and assume that the values are correct?

Presumably you have tested elsewhere that your implementation actually works. If that is the case, then yes, just use values generated by the new code in the tests. They mainly exist as regression tests. When test material from a difference source is available, like test coordinates from a scientific paper, we like to use those as well but that is not possible in this case.

rouault:
> Like a rotation operation? Something like

yeah, I was thinking to that. Could be useful as an operation, but the issue is as soon as you define a pipeline, this is no longer a CRS definition, and you'll struggle using that in a Mapserver context where you define a CRS for your project or layer

>  Are these tests auto-generated? Or are the numbers somehow computed by hand? Or shall I just use the current implementation and assume that the values are correct?

manually. Just add a new entry and run "make check". It will give you the output result it found if not consistent with the one you put, so you can bootstrap with dummy output values, and use the ones it will get to you. But you're supposed to do some validation of your input/output tuples ;-)

kbevers:
> yeah, I was thinking to that. Could be useful as an operation, but the issue is as soon as you define a pipeline, this is no longer a CRS definition, and you'll struggle using that in a Mapserver context where you define a CRS for your project or layer

I see that as a Mapserver problem :-) But seriously, this could be dealt with by setting up a WKT string that includes a rotation step, no? Provided that the rotation operation is put in the WKT machinery of course.

rouault:
> I see that as a Mapserver problem :-)

yeah, which means my problem at some point (I've a port of MapServer to PROJ 6 API almost ready, and this wasn't an easy ride. And MapServer will still continue to use heavily PROJ strings for CRS in a foreseable future)

> this could be dealt with by setting up a WKT string that includes a rotation step, no?

Yes, likely a DerivedProjectedCRS with a deriving conversion being a rotation. They will have derived projected CRS in EPSG v10 (see https://www.iogp.org/blog/epsg/upgrade-of-epsg-dataset-data-model/), but I haven't looked if the EPSG guidance note 7-2 describes a rotation method. Probably. Actually the +proj=affine ( https://proj.org/operations/transformations/affine.html ) could already be used for that (not completely sure to remember why it's classified as a transformation rather than a conversion, but whatever)

---------------

To come back to this pull request, what is the rotation center: (0, 0) or (x_0, y_0) ?

busstoptaktik:
@kbevers said:
> Like a rotation operation? Something like

We have that already, I believe, as a special case of the 4 parameter Helmert

kbevers:
> We have that already, I believe, as a special case of the 4 parameter Helmert

Correct. Although you have to give the rotation in arc seconds

Schneegans:
> To come back to this pull request, what is the rotation center: (0, 0) or (x_0, y_0) ?

I think the correct answer to this question is `(x_0, y_0)`.

The first figure in the second row of the picture I included above shows a result of `+x_0=2.5 +y_0=2.5 +rot_xy=45` (plus some scaling).

test/gie/builtins.gie
```diff
@@ -2021,6 +2021,29 @@ expect  -0.001790493 0.000759909
 accept  -200 -100
 expect  -0.001790493 -0.000759909
 
+-------------------------------------------------------------------------------
+operation +proj=healpix   +R=6400000   +lat_1=0.5 +lat_2=2  +rot_xy=42
+-------------------------------------------------------------------------------
+tolerance 0.1 mm
+accept  2 1
```
rouault, line 2028: to test just that parameter, one expect/accept tuple should be enough.
But I find it more convincing to use the expect value as the acept value of the `direction inverse` case. This shows correct implementation of the roundtripping

docs/source/operations/projections/healpix.rst
```diff
@@ -51,6 +51,13 @@ Parameters
 
 .. note:: All parameters for the projection are optional.
 
+.. option:: +rot_xy
```
kbevers, line 54: It should be mentioned when this was introduced:

```suggestion
.. option:: +rot_xy

  .. versionadded:: 7.0.0
```

Schneegans:
Are there any updates for this one? I hope that I resolved all suggested changes :smile:

rouault:
Looks good to me. I let @kbevers do his check. (we should use "squash and merge" to fold the commits)

kbevers:
Looks good to me

PR #1641, opened 2019-09-26
proj_create_crs_to_crs(): fix when there are only transformations with ballpark steps

rouault:
Currently we would discard all operations, resulting in a PJ object with
zero candidates. Better use those operations if nothing better is available.

Was seen on transforming from ETRS89 / UTM zone 31N + EGM96 height to WGS 84 (G1762).
The horizontal transformation from ETRS89 to WGS 84 (G1762) is a ballpark one.


rouault:
Backport PR in https://github.com/OSGeo/PROJ/pull/1643

PR #1653, opened 2019-10-03
ob_tran doc: fix/clarify semantics of o_lat_p/o_lon_p

rouault:
This is a super confusing topic, but from experiments, and in particular with
the example of Fairgrieve projection, o_lat_p and o_lon_p are the coordinates
of the North pole of the unrotated geographic CRS expressed in the rotated
geographic CRS.
This is different from defining a 'new pole', which would be the coordinates
of the rotated North pole expressed in the unrotated CRS.

@kbevers I'd appreciate your review of this. This is a topic I'm stumbling upon when dealing with raster formats such as GRIB, netCDF or PDS/ISIS3 that have different conventions than PROJ ob_tran & Snyder formulas. The term "new pole" is confusing/wrong to me. "New pole" to me would sound like the North pole of the rotated grid, which to the best of my understanding&experiments, is not the case.


kbevers:
Sorry for not getting round to this sooner. It is indeed difficult to interpret what the different parameters mean. I *think* you've got it right in your updated doc descriptions.

PR #1659, opened 2019-10-04
Add logo to website

### LINKED ISSUE #1647: Doc: put PROJ logo on website
Could possibly be done like https://gdal.org on the top left corner. Relevant code: https://github.com/OSGeo/gdal/blob/master/gdal/doc/source/conf.py (look for logo)

kbevers:
Adds various logo files to `media/`. Thanks to Sylvain Beorchia for supplying the logo.

I have changed the color of the navigation header to the same as the actual navigation bar instead of the usual blue color which looks weird behind the logo. Example below. I think it looks good but I'll leave this PR open for a bit in case there are better suggestions.

Screen Shot 2019-10-04 at 17 44 48

Closes #1647


rouault:
Did we get an explicit licensing statement from Sylvain ? In which case it might be appropriate to add it as a XML comment in the SVG, or even better in the rdf:RDF metadata as it is likely possible

kbevers:
Yes, CC-0. It should be in the SVG. At least it was in the original I received from Sylvain (I split it into two, color and B/W). In Inkscape I did add the license again before saving but it may have disappeared without me knowing...

kbevers:
> I did add the license again before saving but it may have disappeared without me knowing...

Yeah, it did. Added in cc993ad.

PR #1664, opened 2019-10-09
proj.pc.in: add Libs.Private

ffontaine:
Add -lstdc++ to Libs.Private to allow packages such as libgeotiff to
retrieve this dependency when linking statically with proj

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/proj.pc.in
+++ b/proj.pc.in
@@ -9,4 +9,5 @@ Description: Cartographic Projections Library.
 Requires:
 Version: @VERSION@
 Libs: -L${libdir} -lproj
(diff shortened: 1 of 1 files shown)


hobu:
> `-lstdc++`

What about the `libc++` scenario?

ffontaine:
Indeed, this PR does not cover this use case. I sent this patch to fix static building failures on buildroot however an other option would be to disable the possibility of building statically proj and libgeotiff.

PR #1669, opened 2019-10-13
Remove sentence saying that coordinate system might by null

desruisseaux:
The sentence "This might be null, in which case `datumEnsemble()` return will not be
null" It is true for `datum()`, but does not apply to `coordinateSystem()`.


rouault:
Thanks

PR #1673, opened 2019-10-15
Database: update to EPSG v9.8.3

rouault:
Note: a hack has been added into customizations.sql to cancel deprecatation
of USA geoid2012 grids by geoid2018 grids, as the later are not yet available
in proj-datumgrid-northamerica (https://github.com/OSGeo/proj-datumgrid/issues/55)


kbevers:
Looks good to me.

And I guess the Americans will have to update the datumgrid package if they want their new grids to work in PROJ.

rouault:
> And I guess the Americans will have to update the datumgrid package if they want their new grids to work in PROJ.

@hobu "convinced" me :-) to deal with interesting issues with NA geoid grids and LAS WKT1 hacks, so in the coming weeks I'll work on adding various versions of those geoid grids (some people might want to use old versions for reproducibility of results)

nyalldawson:
Looking over this PR, I feel compelled once again to extend my thanks to the PROJ team for the brilliant work they've done in taking ownership over these updates and removing all downstream ~~hacks~~ handling. Thank you! :tada:

kbevers:
> @hobu "convinced" me :-) 

Haha, have fun :-)

rouault:
@kbevers I'm not entirely convinced this is material for 6.2.1

kbevers:
@rouault alright. We have previously included EPSG-updates in patch-version releases. The EPSG update is also a patch. What's the downside to including it in 6.2.1?

rouault:
> What's the downside to including it in 6.2.1?

This addition of the geoid2018 grids. Should hopefully be mitigated with the manual removal of the supersession entries, but coordinate operation computation research is super sensitive and that could possibly cause unexpected changes. I don't see anything super critical in v9.8.3 changes that calls for a backport

kbevers:
> I don't see anything super critical in v9.8.3 changes that calls for a backport

Okay, we can save it for later. I'll change the milestone to 7.0 for now.

PR #1679, opened 2019-10-17
Fix segfaults in case of out-of-memory situations

### LINKED ISSUE #1678: potential segfault in out of memory situations
I found two placed that did not handle out-of-memory situations correctly.
Problematic functions:
- pj_datum_set(...) in src/datum_set.cpp
- ellps_ellps(...) in src/ell_set.cpp

I'll provide a pull request for a fix.

yonarw:
This fixes #1678

### DIFF
--- a/src/ell_set.cpp
+++ b/src/ell_set.cpp
@@ -160,7 +160,7 @@ static int ellps_ellps (PJ *P) {
     if (nullptr == new_params)
         return proj_errno_set (P, ENOMEM);
     new_params->next = pj_mkparam (ellps->ell);
-    if (nullptr == new_params)
+    if (nullptr == new_params->next)
     {
         pj_dealloc(new_params);
         return proj_errno_set (P, ENOMEM);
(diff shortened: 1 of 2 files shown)


src/ell_set.cpp
```diff
@@ -156,7 +157,14 @@ static int ellps_ellps (PJ *P) {
     err = proj_errno_reset (P);
 
     paralist* new_params = pj_mkparam (ellps->major);
+    if (nullptr == new_params)
+        return proj_errno_set (P, ENOMEM);
     new_params->next = pj_mkparam (ellps->ell);
+    if (nullptr == new_params)
```
rouault: cppcheck complains:
```
/home/travis/build/OSGeo/PROJ/src/ell_set.cpp:163,warning,nullPointer,Possible null pointer dereference: new_params - otherwise it is redundant to check it against null.
```
Indeed the check should be against new_params->next

PR #1688, opened 2019-10-23
Fix errors running gie-based tests in Debug mode on Windows

apacker1:
In Debug mode on Windows, the gie executable is actually built as "gie_d", so all the gie tests fail because they can't find the executable.

### DIFF
--- a/cmake/ProjTest.cmake
+++ b/cmake/ProjTest.cmake
@@ -36,7 +36,7 @@ endfunction()
 
 function(proj_add_gie_test TESTNAME TESTCASE)
 
-    set(GIE_BIN "gie")
+    set(GIE_BIN $<TARGET_FILE_NAME:gie>)
     set(TESTFILE ${CMAKE_SOURCE_DIR}/test/${TESTCASE})
     add_test(NAME ${TESTNAME}
       WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/test


kbevers:
Is gie the only app that get's the `_d` prefix or would the same fix also be applicable to `proj`, `cs2cs`, `cct`, `projinfo` and `geod`? It's been a while since I build on Windows but I do seem to remember seeing this before.

mwtoews:
See https://github.com/OSGeo/PROJ/blob/master/CMakeLists.txt#L169-L173 which does `set(CMAKE_DEBUG_POSTFIX _d)` for MSVC multi-config builds.

apacker1:
@kbevers 

> Is gie the only app that get's the `_d` prefix or would the same fix also be applicable to `proj`, `cs2cs`, `cct`, `projinfo` and `geod`? It's been a while since I build on Windows but I do seem to remember seeing this before.

Yes, all those executables are built with "_d" suffix.  None of those seem to cause any failures when I run ctest though.

kbevers:
> Yes, all those executables are built with "_d" suffix. None of those seem to cause any failures when I run ctest though.

Still kinda annoying, I think.

@mwtoews is there a good reason to keep the `_d` suffixes?

apacker1:
The dll also has the suffix, ie, "proj_6_2_d.dll"

mwtoews:
I was investigating this issue sometime last year (sorry, I didn't post this as an issue earlier), but I can't recall if I came to any conclusions on the matter except that there was an inconstancy that needed fixing...

PR #1701, opened 2019-10-30
Document PROJJSON

rouault:
More could probably be written, but at least this can serve as a
landing/reference page for other documents/specifications to point to.

Triggered by https://twitter.com/matthmohr/status/1189545919710138369 (CC @m-mohr )

@kbevers I midly satisfied with placing it under the Usage category, but couldn't see a better fit

### DIFF
--- a/docs/source/_extensions/redirects.py
+++ b/docs/source/_extensions/redirects.py
@@ -1,0 +1,39 @@
+import os
+
+# https://tech.signavio.com/2017/managing-sphinx-redirects
+
+
+template="""<html>
+  <head>
+    <meta http-equiv="refresh" content="1; url=%s" />
+    <script>
+      window.location.href = "%s"
+    </script>
+  </head>
+</html>"""
+
+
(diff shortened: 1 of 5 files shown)


rouault:
I've added an extra commit with a Sphinx extension (as done on GDAL) to create redirect HTML pages. That way once this is merged https://proj.org/projjson.html can be used as a permalink and we might decide to move the location of the page elsewhere in our structure if needed.

kbevers:
> I've added an extra commit with a Sphinx extension (as done on GDAL) to create redirect HTML pages. That way once this is merged proj.org/projjson.html can be used as a permalink and we might decide to move the location of the page elsewhere in our structure if needed.

That is awesome! I am sure we can use that facility for other pages in the future, for instance moving the API reference pages to a more top-level location. Thanks for adding that.

As for the PROJJSON description, I think it looks good. If you are satisfied with it yourself I think you should just go on and merge it.

Issue #1718, opened 2019-11-07
proj 6.2.1 is very slow with sqlite 3.7.17-8 on RHEL 7 


akrherz:
I've discovered that Proj 6.2.1 (packaged by yum.postgresql.org) runs very slow on Red Hat Enterprise 7.  For example:

```
echo "0 0" | time /usr/proj62/bin/cs2cs +proj=longlat +datum=WGS84 +to +init=epsg:2163
9473741.42	1181205.06 0.00
3.04user 0.11system 0:03.15elapsed 99%CPU (0avgtext+0avgdata 14744maxresident)k
0inputs+0outputs (0major+92566minor)pagefaults 0swaps
```

So 3.04 seconds!  On RHEL8, that command finishes in 0.06 seconds. 

RHEL7 ships with sqlite version 3.7.17. Is this version too old?

As a test, I built (./configure && make install) proj 6.2.1 manually myself on RHEL7 too and got a similar result
```
echo "0 0" | time /usr/local/bin/cs2cs +proj=longlat +datum=WGS84 +to +init=epsg:2163
9473741.42	1181205.06 0.00
2.92user 0.10system 0:03.03elapsed 99%CPU (0avgtext+0avgdata 14748maxresident)k
0inputs+0outputs (0major+92567minor)pagefaults 0swaps
```

Thank you

rouault:
> Is this version too old?

Likely. Indeed I can reproduce slowness with old SQLite versions. This seems related to doing queries in the coordinate_operation_with_conversion_view view. I could reproduce also with 3.8.2. With 3.11.0 that comes with Ubuntu 16.04 the timings are OK. Looks like an optimization was added in SQLite for such queries between those two points.

kbevers:
@rouault Is this grounds for changing the minimum requirements for SQLite? Or perhaps just state somewhere that we recommend using a newer version? Currently we require version 3.7.

rouault:
> Is this grounds for changing the minimum requirements for SQLite?

yes: https://github.com/OSGeo/PROJ/pull/1721

mwtoews:
For the record, I've [found](https://lists.osgeo.org/pipermail/proj/2019-November/008999.html) the exact minimum release of SQLite should be 3.8.11

rouault:
Resolution is to require SQLite 3.11 at the minimum per #1721

devrimgunduz:
Are you all aware that you'll kick out RHEL 7 users after this?

[...]

PR #1721, opened 2019-11-07
Build: require SQLite 3.11 (refs #1718)

rouault:
PROJ can build and run against older version (3.7 for example), but it has
been found that performance is horrible. With 3.11 (Ubuntu 16.04), it is
fine, so sets this as the minimum version.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -119,14 +119,19 @@ Compilation and installation from source code
 The classical way of installing PROJ is via the source code distribution. The
 most recent version is available from the :ref:`download page<current_release>`.
 
-PROJ requires C and C++11 compilers. It also requires SQLite3 (headers,
-library and executable).
-
 You will need that and at least the standard *proj-datumgrid* package for a
 successful installation.
 
 The following guides show how to compile and install the software using the
 Autotools and CMake build systems.
+
+Build requirements
+--------------------------------------------------------------------------------
+
+- C99 compiler
+- C++11 compiler
+- SQLite3 >= 3.11 (headers, library and executable)
+- GNU make for autotools build or CMake >= 3.5
 
 Autotools
 --------------------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,12 @@ if(NOT SQLITE3_FOUND)
   message(SEND_ERROR "sqlite3 dependency not found!")
 endif()
 
+# Would build and run with older versons, but with horrible performance
+# See https://github.com/OSGeo/PROJ/issues/1718
+IF("${SQLITE3_VERSION}" VERSION_LESS "3.11")
+  message(SEND_ERROR "sqlite3 >= 3.11 required!")
+ENDIF()
+
 ################################################################################
 # threading configuration
 ################################################################################
--- a/cmake/FindSqlite3.cmake
+++ b/cmake/FindSqlite3.cmake
@@ -10,6 +10,7 @@
 # and following variables are set:
 #    SQLITE3_INCLUDE_DIR
 #    SQLITE3_LIBRARY
+#    SQLITE3_VERSION
 
 
 # find_path and find_library normally search standard locations
@@ -53,9 +54,21 @@ if(SQLITE3_INCLUDE_DIR AND SQLITE3_LIBRARY)
   set(SQLITE3_FOUND TRUE)
 endif()
 
+# Extract version information from the header file
+if(SQLITE3_INCLUDE_DIR)
+    file(STRINGS ${SQLITE3_INCLUDE_DIR}/sqlite3.h _ver_line
+         REGEX "^#define SQLITE_VERSION  *\"[0-9]+\\.[0-9]+\\.[0-9]+\""
+         LIMIT_COUNT 1)
+    string(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]+"
+           SQLITE3_VERSION "${_ver_line}")
+    unset(_ver_line)
+endif()
+
+
 if(SQLITE3_FOUND)
   if(NOT SQLITE3_FIND_QUIETLY)
     message(STATUS "Found Sqlite3: ${SQLITE3_LIBRARY}")
+    message(STATUS "Sqlite3 version: ${SQLITE3_VERSION}")
   endif()
 
 else()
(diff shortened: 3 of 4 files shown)


kbevers:
Thanks. Just a reminder from RFC3:

```
A change in programming language standard can only be introduced with a new major version release of PROJ. Changes for software package dependencies can be introduced with minor version releases of PROJ. Changing the version requirements for a dependency needs to be approved by the PSC.
```

So, we can introduce this with PROJ 6.3 *provided* that the PSC accepts this change. SQLite 3.11 was released in 2016 so we are clear of the older-than-two-years requirement for package dependencies.

I would suggest that RFC3 be amended with the new requirements of SQLite.

rouault:
>  RFC3 be amended with the new requirements of SQLite.

As I read it, RFC3 is general policy and bootstrapping. Does it need to be a living document ? Probably not the best place to capture requirements that might be per PROJ version

rouault:
Motion sent to the list

kbevers:
> Does it need to be a living document ?

It doesn't have to but I figure that it might be useful. Maybe it's too much admin work to keep up to date, I dont know... I was just thinking a history section that keeps track of when new requirements are introduced.

Where do we keep the list of requirements now, that isn't RFC3? The README? Somewhere on the website? I think this is a place where we can improve things a bit at least.

rouault:
> Where do we keep the list of requirements now

In the checks of build systems :-)

kbevers:
> In the checks of build systems :-)

I'll go out on a limb here and say that that isn't good enough :)

rouault:
> I'll go out on a limb here and say that that isn't good enough :)

it turns out I discovered there was already more or less something for that. I've tweaked it in https://github.com/OSGeo/PROJ/pull/1721/commits/b2873d7a788ec8ee7233a492696c7f6e048ac2e0

docs/source/install.rst
```diff
@@ -119,15 +119,23 @@ Compilation and installation from source code
...
 You will need that and at least the standard *proj-datumgrid* package for a
 successful installation.
 
 The following guides show how to compile and install the software using the
 Autotools and CMake build systems.
 
+Build requirements
+--------------------------------------------------------------------------------
+
+PROJ 6.2 or later
```
kbevers: I suggest leaving this out and just have this section state what is necessary for whatever is the current version. It is made clear in the docs which PROJ version they refer to

rouault: done

rouault:
Motion passed

devrimgunduz:
Can you please give me a hint about compiling Proj with a non-standard sqlite version (I mean, which is outside $PATH, etc?)

I'm having trouble doing it. Proj successfully finds my new sqlite, but then I get:

```
libtool: link: g++ -std=c++11 -Wall -Wextra -Winit-self -Wunused-parameter -Wformat -Werror=format-security -Wno-format-nonliteral -Wlogical-op -Wshadow -Werror=vla -Wmissing-declarations -Wnon-virtual-dtor -Woverloaded-virtual -Wdeprecated -Wzero-as-null-pointer-constant -g -O2 -fvisibility=hidden -Wl,-rpath -Wl,/usr/proj62/lib64 -o .libs/geod apps/geod.o apps/geod_set.o apps/geod_interface.o apps/emess.o  ./.libs/libproj.so -lpthread -lm -Wl,-rpath -Wl,/usr/proj62/lib
./.libs/libproj.so: undefined reference to `sqlite3_column_text'
./.libs/libproj.so: undefined reference to `sqlite3_close'
```
...(and this goes on a bit more)

Any ideas? Thanks!

rouault:
> Any ideas? Thanks!

How do you invoke ./configure ? Do you use pkg-config or explicitly set SQLITE3_CFLAGS and SQLITE3_LIBS ?

devrimgunduz:
Argh, thanks! I was using them, but apparently in a wrong way. Here is what I did, which worked. %{sqlite33dir } is an RPM macro, and it's replaced by /usr/sqlite330 in my case:

```
export SQLITE3_LIBS="-L%{sqlite33dir}/lib -lsqlite3"
export SQLITE3_INCLUDE_DIR='%{sqlite33dir}/include'
export SQLITE3_CFLAGS="$CFLAGS -I%{sqlite33dir}/include"
export PATH=%{sqlite33dir}/bin/:$PATH

```

Thanks! Just a FYI: I'm working on custom sqlite3 packages on RHEL7, and Proj packages in the PostgreSQL RPM repository will depend of this sqlite version, which will solve the issues.

rouault:
> Just a FYI: I'm working on custom sqlite3 packages on RHEL7, and Proj packages in the PostgreSQL RPM repository will depend of this sqlite version, which will solve the issues.

Yes, @hobu made me aware of that. Much appreciated ! I've posted about your work on the PROJ mailing list too in https://lists.osgeo.org/pipermail/proj/2019-November/009002.html

So, you're no replacing the system libsqlite3 ? How does that work with packages such as GDAL which themselves link to sqlite3 (do you build it againt that /usr/sqlite330 too ?). Otherwise that might potentially cause runtime clashes

devrimgunduz:
Hi @rouault ,

I'm not replacing  the system sqlite. The package I created install things under /usr/sqlite330 directory. I rebuilt proj, and with rpath magic, made sure that libproj.so is linked to the new sqlite:
-bash-4.2$ ldd /usr/proj62/lib/libproj.so|grep sqlite
	libsqlite3.so.0 => /usr/sqlite330/lib/libsqlite3.so.0 (0x00007f6c46200000)
I'll do the same for GDAL,  libspatialite, libgeotiff, and then rebuild PostGIS against these.

Do you see any potential problems with this plan?

Thanks!

Regards, Devrim

rouault:
> Do you see any potential problems with this plan?

No, that sounds good. Downstream users of those packages having an explicit sqlite3 dependency should make sure to link against sqlite330 too.

Before you attempted this, I tried a quick & dirty approach by having an internal copy of sqlite inside PROJ, but I also tried to avoid any potential symbol clash by having a #define sqlite330_sqlite3_XXXX sqlite3_XXXX for each sqlite3 symbol. That way you can link both against PROJ and the system sqlite3 without risks
A solution to do that easily is from the amalgation sqlite3.c to
```bash
gcc -shared -fPIC sqlite3.c -shared -o temp.so
(for i in `objdump -T temp.so | grep sqlite3 | awk '{print $7}'`; do echo "#define $i sqlite330_$i"; done) > rename.h
cat rename.h sqlite3.h > sqlite3_with_rename.h
mv sqlite3_with_rename.h sqlite3.h
```
That way only the PROJ package would have to been built against this hacked sqlite3 lib, and all other packages could link against the regular sqlite3 without risk of conflict

PR #1726, opened 2019-11-12
Fix proj_assign_context()/pj_set_ctx() with pipelines and alternative coord operations

### LINKED ISSUE #1989: gdalwarp with multithreading return corrupted data (netcdf + geos projection)
Hi, following some experiment mentioned in #1985, I encountered another strange behaviour when doing warping with multithreading enabled.

## Steps to reproduce the problem.

```bash
$ aws s3 cp s3://noaa-goes16/ABI-L1b-RadF/2019/295/14/OR_ABI-L1b-RadF-M6C01_G16_s20192951400357_e20192951410065_c20192951410137.nc .

$ gdalwarp netcdf:OR_ABI-L1b-RadF-M6C01_G16_s20192951400357_e20192951410065_c20192951410137.nc:Rad rad_multi.tif -wo NUM_THREADS=8 -t_srs EPSG:4326
```
See the image above 👇 . Using multithreading creates lines with `nodata` values where it should be valid values.

result

When reducing the size of the chunk (reducing memory) fixes the lines (but print some error message, which from my understanding, are expected because some points are in space.)

```
$ gdalwarp netcdf:OR_ABI-L1b-RadF-M6C01_G16_s20192951400357_e20192951410065_c20192951410137.nc:Rad rad_multi_oneMb.tif -wm 1 -wo NUM_THREADS=8 -t_srs EPSG:4326
....
ERROR 1: Too many points (529 out of 529) failed to transform, unable to compute output bounds.
Warning 1: Unable to compute source region for output window 10474,9855,699,658, skipping.
....
```

Note: Not sure if this a bug report or more an issue to serve as documentation for people doing such work.

## GDAL version and provenance

gdal 3.0.1 (source)

rouault:
Fixes https://github.com/OSGeo/gdal/issues/1989

pj_set_ctx() only changes the context to the main object. It should also
recurse down to the steps of the pipeline and the alternative coordinate
operations hold in alternativeCoordinateOperations

In the GDAL use case with multithreaded reprojection, and objects being transferred
between thread, this would cause a failed coordinate transformation to affect
an unrelated transformation of another thread...

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -60,6 +60,14 @@ void pj_set_ctx( projPJ pj, projCtx ctx )
     if (pj==nullptr)
         return;
     pj->ctx = ctx;
+    if( pj->is_pipeline )
+    {
+        pj_pipeline_assign_context_to_steps(pj, ctx);
+    }
+    for( const auto &alt: pj->alternativeCoordinateOperations )
+    {
+        pj_set_ctx(alt.pj, ctx);
+    }
 }
 
 /************************************************************************/
--- a/src/pipeline.cpp
+++ b/src/pipeline.cpp
@@ -136,7 +136,11 @@ static PJ_LPZ    pipeline_reverse_3d (PJ_XYZ xyz, PJ *P);
 static PJ_XY     pipeline_forward (PJ_LP lp, PJ *P);
 static PJ_LP     pipeline_reverse (PJ_XY xy, PJ *P);
 
-
+void pj_pipeline_assign_context_to_steps( PJ* P, PJ_CONTEXT* ctx )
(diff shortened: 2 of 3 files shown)


kbevers:
@rouault the next version is 6.3.0 so it is strictly not necessary to backport to the 6.2 branch. No harm in doing it, just want to make sure that we are on the same page regarding the release schedule.

As for the fix, it seems like the right way to deal with this problem.

rouault:
> No harm in doing it, just want to make sure that we are on the same page regarding the release schedule.

Agreed. Just in case people would pull from the 6.2 branch for such narrow scope fixes

PR #1733, opened 2019-11-17
createOperations(): remove the concept of geodetic_datum_preferred_hub

rouault:
Was introduced in master in September, but with a fix done in existing findsOpsInRegistryWithIntermediate() logic, it does not seem to be necessary


rouault:
Please do not merge yet. Waiting for feedback from a beta tester

PR #1734, opened 2019-11-19
createOperations(): chain operations whose middle CRSs are not identical but have the same datum

rouault:
in some situations, consider when going from A to D intermediates B and C, such there's a A->B operation and C->D operation, and A and C are not exactly the same CRS but use the same geodetic datum


kbevers:
Trying to figure out if this deserves mentioning in release notes. How should this be categorized? Is it a new feature? A optimization? Internal house-keeping that doesn't concern users?

rouault:
I'd say it is a functional improvement in the computation of concatenated operation pipelines for some coordinate transformations between CRS that have no direct transformation between them

kbevers:
Thanks

PR #1742, opened 2019-11-24
Add scripts/grid_checks.py

rouault:
Little script used lately to check consistency of the database
regarding the grid_transformation vs grid_alternatives tables.
And also check the content of the database vs proj-datumgrid
to identify gaps.

### DIFF
--- a/scripts/grid_checks.py
+++ b/scripts/grid_checks.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+###############################################################################
+# $Id$
+#
+#  Project:  PROJ
+#  Purpose:  Tool to check consistency of database regarding grids and against
+#            what is available in proj-datumgrid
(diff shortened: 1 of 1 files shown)


scripts/grid_checks.py
```diff
@@ -0,0 +1,93 @@
...
+#  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#  DEALINGS IN THE SOFTWARE.
+###############################################################################
+
+import fnmatch
+import os
+import sqlite3
+import sys
+
+if len(sys.argv) != 4:
```
QuLogic: Might want to consider using `argparse` for this.

rouault: Thanks for the suggestion. Applied

PR #1744, opened 2019-11-24
Optimize pipelines involving horizontal shift grid, vertical shift grid, inverse horizontal shift grid

rouault:
Add a +only_on_forward flag to mark that a step must only be applied in forward path

This is used to optimize constructs like

+step +proj=hgridshift +grids=foo
+step +proj=vgridshift +grids=bar
+step +inv +proj=hgridshift +grids=foo

Such steps are used for CRS to CRS transformations where applying the vertical grid
requires to do a transformation to an interpolating CRS. One can notice that
in the last step will just restore the horizontal coordinates before the first step, so
doing an inverse hgridshift is overkill.

So this is optimized as:

+step +proj=push +v_1 +v_2
+step +proj=hgridshift +grids=foo +only_on_forward
+step +proj=vgridshift +grids=bar
+step +inv +proj=hgridshift +grids=foo +only_on_forward
+step +proj=pop +v_1 +v_2

In the forward path, this will be equivalent to:
+step +proj=push +v_1 +v_2
+step +proj=hgridshift +grids=foo
+step +proj=vgridshift +grids=bar
+step +proj=pop +v_1 +v_2

And similarly in the reverse path, this will be quivalent to:
+step +proj=push +v_1 +v_2
+step +proj=hgridshift +grids=foo
+step +inv +proj=vgridshift +grids=bar
+step +proj=pop +v_1 +v_2

Following +inv +only_on_forward in the forward path means that
the step will not be executed
Follwing +inv +only_on_forward in the reverse path means that the
step will be execute (as inverse operation)

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -837,6 +837,8 @@ pj_init_ctx_with_allow_init_epsg(projCtx ctx, int argc, char **argv, int allow_i
     else
         PIN->from_greenwich = 0.0;
 
+    PIN->only_on_forward = pj_param(ctx, start, "bonly_on_forward").i != 0;
+
     /* Private object for the geodesic functions */
     PIN->geod = static_cast<struct geod_geodesic*>(pj_calloc (1, sizeof (struct geod_geodesic)));
     if (nullptr==PIN->geod)
--- a/src/inv.cpp
+++ b/src/inv.cpp
@@ -145,6 +145,9 @@ PJ_LP pj_inv(PJ_XY xy, PJ *P) {
     PJ_COORD coo = {{0,0,0,0}};
     coo.xy = xy;
 
+    if( P->only_on_forward )
+        return coo.lp;
+
     last_errno = proj_errno_reset(P);
 
     if (!P->skip_inv_prepare)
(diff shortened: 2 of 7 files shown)


rouault:
I can imagine that this could be later extended with a +only_on_inverse for example if you do horizontal grid shift, but have a grid for the forward path and another path for the reverse

+step +proj=hgridshift +grids=forward.gsb +only_on_forward
+step +proj=hgridshift +grids=inverse.gsb +only_on_inverse

At some point PROJ pipelines will be Turing complete :-)

kbevers (commented):
Fun fact: We actually implemented this before as `+omit_fwd` and `+omit_inv` in the original incarnations of the pipeline. See  https://github.com/OSGeo/PROJ/pull/453/files (look at `src/pipeline.c`). I believe it was removed again because of a lack of use cases at the time. But now that you have demonstrated a need for it is time for a reintroduction :-)

Looking at the two different implementations, I think the original one is a bit cleaner in that it is more local to `src/pipeline.cpp`. From a functional standpoint the difference is that +only_on_forward is applied on *any* `PJ`, pipeline or not. Where as the `+omit_inv` implementation on takes the keyword into consideration when used in a pipeline.

I think that it is better to keep the functionality contained to the pipeline operation because 1) it avoids use of the keyword in non-pipeline operations and 2) it keeps the code where you would intuitively look for it.

The `+omit_inv` syntax is a bit shorter so perhaps that is better? Not a huge deal, just something to consider.

rouault:
@kbevers Thanks for the hints ! This PR is now superseded by https://github.com/OSGeo/PROJ/pull/1748 using +omit_fwd and +omit_inv (and actually in a less convoluted way than this first attempt !)

PR #1748, opened 2019-11-25
Optimize pipelines involving horizontal shift grid, vertical shift grid, inverse horizontal shift grid (take 2)

rouault:
Cancels and replaces https://github.com/OSGeo/PROJ/pull/1744

3 commits:
- first commit is mostly a code cleanup/improvement in the pipeline implementation so as to be able to implement easily the next commit. Should have no user visible effect
- second commit adds +omit_fwd and +omit_inv keywords (resurecting an idea of the initial implementation of https://github.com/OSGeo/PROJ/pull/453/files)
- third commit uses that capability to optimize  constructs like
      +step +proj=hgridshift +grids=foo
      +step +proj=vgridshift +grids=bar
     +step +inv +proj=hgridshift +grids=foo

Such steps are used for CRS to CRS transformations where applying the vertical grid
requires to do a transformation to an interpolating CRS. One can notice that
in the last step will just restore the horizontal coordinates before the first step, so
doing an inverse hgridshift is overkill.

So this is optimized as:

+step +proj=push +v_1 +v_2
+step +proj=hgridshift +grids=foo +omit_inv
+step +proj=vgridshift +grids=bar
+step +inv +proj=hgridshift +grids=foo +omit_fwd
+step +proj=pop +v_1 +v_2

In the forward path, this will be equivalent to:
+step +proj=push +v_1 +v_2
+step +proj=hgridshift +grids=foo
+step +proj=vgridshift +grids=bar
+step +proj=pop +v_1 +v_2

And similarly in the reverse path, this will be quivalent to:
+step +proj=push +v_1 +v_2
+step +proj=hgridshift +grids=foo
+step +inv +proj=vgridshift +grids=bar
+step +proj=pop +v_1 +v_2

### DIFF
--- a/src/pipeline.cpp
+++ b/src/pipeline.cpp
@@ -101,6 +101,7 @@ Thomas Knudsen, thokn@sdfe.dk, 2016-05-20
 #include <stddef.h>
 #include <string.h>
 #include <stack>
+#include <vector>
 
 #include "geodesic.h"
 #include "proj.h"
@@ -112,12 +113,25 @@ PROJ_HEAD(push, "Save coordinate value on pipeline stack");
 
 /* Projection specific elements for the PJ object */
 namespace { // anonymous namespace
+
+struct Step {
+    PJ* pj = nullptr;
+
+    Step(PJ* pjIn): pj(pjIn) {}
+    Step(Step&& other): pj(std::move(other.pj)) { other.pj = nullptr; }
+    Step(const Step&) = delete;
+    Step& operator=(const Step&) = delete;
+
+    ~Step() {
+        proj_destroy(pj);
+    }
+};
+
 struct pj_opaque {
-    int steps;
-    char **argv;
-    char **current_argv;
-    PJ **pipeline;
-    std::stack<double> *stack[4];
+    char **argv = nullptr;
+    char **current_argv = nullptr;
+    std::vector<Step> steps{};
+    std::stack<double> stack[4];
 };
 
 struct pj_opaque_pushpop {
@@ -138,32 +152,27 @@ static PJ_LP     pipeline_reverse (PJ_XY xy, PJ *P);
 
 void pj_pipeline_assign_context_to_steps( PJ* P, PJ_CONTEXT* ctx )
 {
-    for (int i = 1; i <= static_cast<struct pj_opaque*>(P->opaque)->steps; i++)
-        proj_assign_context(static_cast<struct pj_opaque*>(P->opaque)->pipeline[i], ctx);
+    auto opaque = static_cast<struct pj_opaque*>(P->opaque);
+    for( auto& step: opaque->steps )
+        proj_assign_context(step.pj, ctx);
 }
 
 
 static PJ_COORD pipeline_forward_4d (PJ_COORD point, PJ *P) {
-    int i, first_step, last_step;
-
-    first_step = 1;
-    last_step  = static_cast<struct pj_opaque*>(P->opaque)->steps + 1;
-
-    for (i = first_step;  i != last_step;  i++)
-        point = proj_trans (static_cast<struct pj_opaque*>(P->opaque)->pipeline[i], PJ_FWD, point);
+    auto opaque = static_cast<struct pj_opaque*>(P->opaque);
+    for( auto& step: opaque->steps )
+         point = proj_trans (step.pj, PJ_FWD, point);
 
     return point;
 }
 
 
 static PJ_COORD pipeline_reverse_4d (PJ_COORD point, PJ *P) {
-    int i, first_step, last_step;
-
-    first_step = static_cast<struct pj_opaque*>(P->opaque)->steps;
-    last_step  =  0;
-
-    for (i = first_step;  i != last_step;  i--)
-        point = proj_trans (static_cast<struct pj_opaque*>(P->opaque)->pipeline[i], PJ_INV, point);
+    auto opaque = static_cast<struct pj_opaque*>(P->opaque);
+    for( auto it = opaque->steps.rbegin(); it != opaque->steps.rend(); ++it )
+    {
+        point = proj_trans (it->pj, PJ_INV, point);
+    }
 
     return point;
 }
@@ -173,11 +182,10 @@ static PJ_COORD pipeline_reverse_4d (PJ_COORD point, PJ *P) {
 
 static PJ_XYZ pipeline_forward_3d (PJ_LPZ lpz, PJ *P) {
     PJ_COORD point = {{0,0,0,0}};
-    int i;
     point.lpz = lpz;
-
-    for (i = 1;  i <= static_cast<struct pj_opaque*>(P->opaque)->steps;  i++)
-        point = pj_approx_3D_trans (static_cast<struct pj_opaque*>(P->opaque)->pipeline[i], PJ_FWD, point);
+    auto opaque = static_cast<struct pj_opaque*>(P->opaque);
+    for( auto& step: opaque->steps )
+         point = pj_approx_3D_trans (step.pj, PJ_FWD, point);
 
     return point.xyz;
 }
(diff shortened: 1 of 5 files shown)


src/pipeline.cpp
```diff
@@ -138,32 +152,27 @@ static PJ_LP     pipeline_reverse (PJ_XY xy, PJ *P);
...
 
 static PJ_COORD pipeline_forward_4d (PJ_COORD point, PJ *P) {
-    int i, first_step, last_step;
-
-    first_step = 1;
-    last_step  = static_cast<struct pj_opaque*>(P->opaque)->steps + 1;
-
-    for (i = first_step;  i != last_step;  i++)
-        point = proj_trans (static_cast<struct pj_opaque*>(P->opaque)->pipeline[i], PJ_FWD, point);
+    auto opaque = static_cast<struct pj_opaque*>(P->opaque);
```
kbevers: I think this would be even more readable if you call the variable "pipeline" instead of "opaque"

Applicable several places below.

src/pipeline.cpp
```diff
@@ -138,32 +152,27 @@ static PJ_LP     pipeline_reverse (PJ_XY xy, PJ *P);
...
 static PJ_COORD pipeline_reverse_4d (PJ_COORD point, PJ *P) {
-    int i, first_step, last_step;
-
-    first_step = static_cast<struct pj_opaque*>(P->opaque)->steps;
-    last_step  =  0;
-
-    for (i = first_step;  i != last_step;  i--)
-        point = proj_trans (static_cast<struct pj_opaque*>(P->opaque)->pipeline[i], PJ_INV, point);
+    auto opaque = static_cast<struct pj_opaque*>(P->opaque);
+    for( auto it = opaque->steps.rbegin(); it != opaque->steps.rend(); ++it )
```
kbevers: I think it would read better if `it` was replaced by `step` here and in other similar occurences below

kbevers (approved):
This is great! I like the C++'ing of the code. It improves readability a lot. I've put in a few suggestions for easier reading but that's it.

src/pipeline.cpp
rouault: Well, the type of `it` is a `std::vector<Step>::iterator`, not a Step... Unfortunately in C++11 you can't use range-based for loop for reverse iteration (you need a later C++ standard), so we have to use "old school" reverse iterators. The value of *it is a Step object

kbevers: I wasn't thinking so much about the types (but maybe I should) but more about what done and how to best communicate that. This would be a lot easier to express in Python :-)

```
for step in pipeline:
    ...
```

I was just thinking that it would be possible to achieve something similar by renaming `opaque` and `it` to `pipeline` and `step`. Anyway, it's your call. There's nothing wrong with the code you've written.

src/pipeline.cpp
rouault: done

src/pipeline.cpp
rouault: it renamed to iterStep to better reflect its nature (iterator) and what the iterator points to (Step)

kbevers:
Whhoops, I almost forgot! `+omit_fwd` and `+omit_inv` should be added to the docs of the pipeline operator.

rouault:
> `+omit_fwd` and `+omit_inv` should be added to the docs of the pipeline operator.

done

src/pipeline.cpp
busstoptaktik: > it renamed to iterStep to better reflect its nature (iterator) and what the iterator points to (Step)

Oh boy - you have gotten me started ranting about the virtues of "Apps Hungarian" vs. the evils of "Systems Hungarian" [1].

What @rouault  calls "nature" is "Systems Hungarian nature", i.e. **type** - a kind of Hungarian we should have left behind in the dust with the arrival of rich and strongly typed languages.

When skimming through code, I don't care much about the type of a variable (the compiler will tell me if my guess is wrong - and C++11 `auto myvar = yourvar` will make sure that the "difference is identical, only the colour has a different sound").

I acknowledge the usefulness of using `it` as the name of a generic iterator in C++ docs. But that really is just another a case of "Apps Hungarian": Telling the reader that we are dealing with a generic iterator in a demo case where the actual type doesn't matter. In all other cases, I would prefer the identifiers to tell me about the **semantics** (i.e. "Apps Hungarian"), rather than the **type** of the variable. Encoding aspects of the type into the identifier name is just adding additional overhead for the reader to comprehend.

So I really believe that @kbevers's Python example is way easier to read and comprehend, even though you may formally say that `step` is just an iterator pointing to a `*step` in a pipeline. But who cares (if you were not brought up with type obsession - well, I was, but I have learned to enjoy the joy of type ignorance): The pipeline is not a sequence of `*step`s either (but of `step`s).

Nevertheless, avoiding to encode type information into the identifier, really clarifies the code. So imho, the name should be **either** `it` to indicate a generic iterator of the context relevant type, **or** `step`, but not a combination of those.

However, this is a generic rant about a pet peeve of mine, not necessarily useful in this specific case, so @rouault in the very likely case that this rant is not useful (or in the even more likely case that it is actually counterproductive), please ignore: This is your code and I do not want to force you to obfuscate it wrt. your mental model of the subject matter - just want to express that modern C++ really is a wonderful combination of Python expressiveness and Ada type safety, and that the `auto` keyword is a gift of productive ignorance from greater spirits than most of us.

I do, however, admit that most of the time I consider @rouault more of a member of the "greater spirits" set, than the "most of us" set. So please do as you please, and thank you for letting me rant in your thread.

[1] https://en.wikipedia.org/wiki/Hungarian_notation: 

> [Charles Simonyi's] proposal was largely concerned with decorating identifier names based upon the semantic information of what they store (in other words, the variable's purpose), consistent with Apps Hungarian. However, his suggestions were not entirely distinct from what became known as Systems Hungarian, as some of his suggested prefixes contain little or no semantic information

rouault: My final word on that topic:
I do think that a distinct name is needed for that reverse iterator. If you look closely in the forward code path, we can use a range-based for loop: "for( const auto& step: pipelines->steps )". The step here is a Step object. In the reverse code path, due to C++11 inability to have range-based for loop with a reverse iterator, I must use old-school (reverse) iterators. Calling that iterator `step` would be confusing regarding the nearby forward code path.

src/pipeline.cpp
```diff
@@ -223,44 +233,24 @@ static PJ_LP pipeline_reverse (PJ_XY xy, PJ *P) {
...
+    auto opaque = static_cast<struct pj_opaque*>(P->opaque);
 
-    pj_dealloc (static_cast<struct pj_opaque*>(P->opaque)->argv);
-    pj_dealloc (static_cast<struct pj_opaque*>(P->opaque)->current_argv);
+    pj_dealloc (opaque->argv);
+    pj_dealloc (opaque->current_argv);
 
-    for (i=0; i<4; i++)
-        delete static_cast<struct pj_opaque*>(P->opaque)->stack[i];
+    delete opaque;
```
busstoptaktik: Is this safe? I admit that I do not have the full overview of this PR, but shouldn't the `*P` destructor handle this, for cases where additional memory was allocated at construction time?

rouault: yes, this is safe because I take care just the line below to nullify P->opaque, so that generic code doesn't try to use free() on it. And use of delete is safe (and needed!) since opaque was allocated by new at https://github.com/OSGeo/PROJ/pull/1748/files#diff-ca6b9218666b6a49063bd0c51b946e35R426 . Allocating with new is needed since the Pipeline struct holds C++ objects (std::vector<Step> steps and std::stack<double> stack[4]) thus those for proper construction and destruction of them
Admitdely the current situation of PROJ operator implementation coming from C, and with some C++ bits added there and there, can be a bit challenging :-). But here I found that using C++ facilities made the code much easier to follow despite that subtelty in the destructor function.

busstoptaktik: @rouault thanks for the explanation. A full C++ core certainly would improve the general readability of the code, but that is probably many years away :-)

src/pipeline.cpp
busstoptaktik: @rouault thanks for giving a final comment (even though I indicated that I did not expect, and much less deserve one). As I also indicated with my @rouault ∈ {greater spirits} comment, I am confident that your way of doing things is most likely the right way, and my comment really was more of a generic rant, inspired by your renaming of `it` to `iterStep`, than specific to this PR.

PR #1753, opened 2019-11-27
Find proj.db under a bin\..\share\proj at runtime

joa-quim:
This PR is Windows only. It adds the directory .....\bin\..\share\proj, where ....\bin is the directory hosting proj.dll, to the list of places where proj.db is searched. In addition to what happens at build time, this PR ads also the ability to do that at run-time. This means that a structure like

....\bin\proj.exe, proj.dll, ...
....\share\proj\proj.db, ...

will still find proj.db without needing to set the PROJ_LIB environment variable

### DIFF
--- a/src/open_lib.cpp
+++ b/src/open_lib.cpp
@@ -151,6 +151,59 @@ void pj_set_searchpath ( int count, const char **path )
     proj_context_set_search_paths( nullptr, count, const_cast<const char* const*>(path) );
 }
 
+
+#ifdef WIN32
+#include <windows.h>
+char *get_projlib_dir(const char *name) {
+    /* Check if proj.db lieves in a share/proj dir parallel to bin/proj.dll */
+    /* Based in https://stackoverflow.com/questions/9112893/how-to-get-path-to-executable-in-c-running-on-windows */
+    size_t k;
+    DWORD last_error;
+    DWORD result;
+    DWORD path_size = 1024;
+    char* path      = (char *)malloc(1024);
+    struct stat fileInfo; 
+
+    for (;;) {
+        memset(path, 0, path_size);
+        result     = GetModuleFileNameA(NULL, path, path_size - 1);
+        last_error = GetLastError();
+
+        if (result == 0) {
+            free(path);
+            path = nullptr;
+            break;
+        }
+        else if (result == path_size - 1) {
+            free(path);
+            if (ERROR_INSUFFICIENT_BUFFER != last_error) {
+                path = nullptr;
+                break;
+            }
+            path_size = path_size * 2;
+            path = (char *)malloc(path_size);
+        }
+        else {
+            break;
+        }
+    }
+    // Now remove the program's name. It was (example) "C:\programs\gmt6\bin\gdal_translate.exe"
+    k = strlen(path);
+    while (path[k] != '\\') k--;
+    path[k] = '\0';
+
+    strcat(path, "/../share/proj/");
+    strcat(path, name);
+
+    if (stat(path, &fileInfo) == 0)	// Check if file exists (probably there are simpler ways)
+        return path;
+    else {
+        free(path);
+        return nullptr;
+    }
+}
+#endif
+
 /************************************************************************/
 /*                          pj_open_lib_ex()                            */
 /************************************************************************/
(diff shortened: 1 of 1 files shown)


src/open_lib.cpp
```diff
@@ -229,6 +282,10 @@ pj_open_lib_ex(projCtx ctx, const char *name, const char *mode,
                 if( fid )
                     break;
             }
+#ifdef _WIN32
+        /* check if it lieves in a ../share/proj dir of the proj dll */
```
rouault:
```suggestion
        /* check if it lives in a ../share/proj dir of the proj dll */
```

src/open_lib.cpp
```diff
@@ -151,6 +151,59 @@ void pj_set_searchpath ( int count, const char **path )
     proj_context_set_search_paths( nullptr, count, const_cast<const char* const*>(path) );
 }
 
+
+#ifdef WIN32
+#include <windows.h>
+char *get_projlib_dir(const char *name) {
```
rouault, line 157: the returned value will not be freed later. This is a C++ file so you could return a std::string. That said some tweaking will still be needed on the call site to make sure that this string is stored in a function local storage. I mean that sys_name = get_projlib_dir(...).c_str() wouldn't work

rouault: actually... the easir would be that you add an extra std::string& returnedString  parameter, that you pass fname as the value of it, and then the return of your function can be const char* which is returnedString.c_str(). Not particularly sexy either, but can't think to anything better yet. And then in get_projlib_dir() you no longer need to do malloc() / free() but resize the string where appropriate.

joa-quim: Even, can I ask if you can do that? You know that me and ++ are not the best couple.

rouault: ok, I'll have a try. But I'll let you test that it works properly at runtime :-)

joa-quim: Sure, thanks.

rouault: can you merge the get_path_from_win32_projlib branch of https://github.com/rouault/PROJ ? This is completely untested. Even not compiled.

rouault: the changeset is there : https://github.com/rouault/PROJ/commit/a7fc1952f1ae3d894a7acf0b0830b50967a6c025

joa-quim: OK, will test.
(yes, never sure when WIN32 or _WIN32. I know that latter is for MSVC but never know what MinGW does with them)

rouault: I too :-) I think not too antiquitated MinGW support _WIN32 as well

joa-quim: It compiled fine, but I'm now having this error

surface depth_points.dat -R443231/448359/6803096/6808575 -I10 -Dlakepoly.gmt -Lu0 -Gdem_with_d_and_lu.tif=gd:GTiff -JEPSG:3067

ERROR 1: PROJ: proj_create_from_database: cannot build projectedCRS 3067: cannot build geodeticCRS 4258: SQLite error on SELECT name, ellipsoid_auth_name, ellipsoid_code, prime_meridian_auth_name, prime_meridian_code, area_of_use_auth_name, area_of_use_code, publication_date, deprecated FROM geodetic_datum WHERE auth_name = ? AND code = ?: no such column: publication_date

What the GMT command does is to interpolate (the Jukka's lake) and assign the EPSG:3067 (converted to it's proj4 representation) to the GeoTIFF file. This was working before.

joa-quim: I'm afraid something else is going on. I reverted the open_lib.cpp to my original version and I still get the error above. So, it isn't caused by your changes.

rouault: > no such column: publication_date

The database structure in master has changed a few weeks ago. You need to make sure the proj.db you use is the one installed by current master

joa-quim: Ghrrr, it took a long time to find this. My installing scripts worked fine (they also update the proj data dir), but I had meanwhile declared %PROJ_LIB% in another program pointing to a place with an older proj.db

Once that solved I could test your version and it works fine.

rouault: > Once that solved I could test your version and it works fine.

Great. Can you cherry-pick / merge my branch into yours and push to update this PR and triggers continuous integration ?

joa-quim: Trying, but not sure it will end up in this PR. Still fighting with Github for commits in repos that I've not commit rights.

rouault:
Superseded by https://github.com/OSGeo/PROJ/pull/1755

PR #1754, opened 2019-11-28
EPSG codes for YEAR and SECOND are interchanged

desruisseaux:
EPSG registry gives EPSG::1040 for second and EPSG::1029 for year.


rouault:
Thanks!

PR #1755, opened 2019-11-28
Add projlib runtime

joa-quim:


src/open_lib.cpp
```diff
@@ -151,6 +151,50 @@ void pj_set_searchpath ( int count, const char **path )
...
+static const char *get_path_from_win32_projlib(const char *name, std::string& out) {
+    /* Check if proj.db lieves in a share/proj dir parallel to bin/proj.dll */
+    /* Based in https://stackoverflow.com/questions/9112893/how-to-get-path-to-executable-in-c-running-on-windows */
+
+    DWORD path_size = 1024;
+
+    for (;;) {
+        out.resize(path_size);
+        memset(&out[0], 0, path_size);
+        DWORD result     = GetModuleFileNameA(NULL, &out[0], path_size - 1);
```
rouault:
```suggestion
        DWORD result     = GetModuleFileNameA(nullptr, &out[0], path_size - 1);
```

src/open_lib.cpp
```diff
@@ -151,6 +151,50 @@ void pj_set_searchpath ( int count, const char **path )
     proj_context_set_search_paths( nullptr, count, const_cast<const char* const*>(path) );
 }
 
+#ifdef _WIN32
+#include <windows.h>
```
rouault, line 155: also add
```
#include <sys/stat.h>
```

rouault (commented):
A few suggestions to hopefully unbreak mingw32 builds

src/open_lib.cpp
```diff
@@ -229,6 +274,10 @@ pj_open_lib_ex(projCtx ctx, const char *name, const char *mode,
                 if( fid )
                     break;
             }
+#ifdef _WIN32
+        /* check if it lieves in a ../share/proj dir of the proj dll */
```
rouault:
```suggestion
        /* check if it lives in a ../share/proj dir of the proj dll */
```

rouault:
OK, so we are not all good w.r.t CI. Maybe @kbevers want to confirm that the idea seams wise to him ?

docs/source/faq.rst has a question about why proj.db cannot be found. Perhaps we should extend it a bit to reflect this change and also the use of proj_context_set_search_paths()

kbevers:
It seems fine to me. I would appreciate a description of the logic that finds `proj.db`. Possibly as part of `docs/source/resource_files.rst`. The FAQ and API docs can then link to that description.

joa-quim:
Sorry, it's not clear to me. Do you want me to add something to the docs?
The logic of this PR is simple. Like what happens with cmake that sets an hard-wired search path to ``../share/proj`` relatively to the ``bin`` directory at build time, this PR does similar but on run-time. Means, if the installation is such it exists a ``......\bin`` and a ``......\share\proj`` then ``proj.db`` is found automatically without the need to set a ``%PROJ_LIB% environment variable. But if that var exists, then it takes precedence over this automatic mechanism.

rouault:
> Do you want me to add something to the docs?

That would be ideal. Otherwise we'll do that as a follow-up

joa-quim:
Ok but where, in ``docs/source/resource_files.rst``? But it doesn't mention that ``proj.db`` lives in ``share/proj`` so I fear that information is described somewhere else.

rouault:
I believe @kbevers meant we should create a new paragraph to describe where resource files are looked up. Probably before https://proj.org/resource_files.html#proj-db . We can reuse the current content of the FAQ to bootstrap that paragraph and improves it.

joa-quim:
OK, I made a blend of both and added the part respecting this PR. See what you think.

```
proj.db

A proj installation includes a SQLite database of transformation information
that must be accessible for the library to work properly.  The library will
print an error if the database can't be found. This file :ref:`proj.db<proj-db>`,
must be readable for the library to function properly. Like other
:doc:`resource files<../resource_files>`, it is located using a set of search
paths. In most cases, the following paths are checked in order:

    - A path provided by the environment variable :envvar:`PROJ_LIB`.
    - A path built into PROJ as its resource installation directory
      (typically *../share/proj* relative to the PROJ library). Note, however,
      that since this is a hard-wired path setting it only works if the whole
      PROJ installation is not moved somewhere else ... except on Windows.
    - On Windows, the *../share/proj/* and its contents are found automatically
      at run-time if the installation respects the build structure. That is, the
      binaries and proj.dll are installed under *....\bin\*, and :ref:`proj.db<proj-db>`
      lives in *.....\share\proj\* 
    - The current directory.
```

rouault:
> what you think.

I think the logic to locate resources should be in its own paragraph, before proj.db. It applies to proj.db and grids. I'm not completely sure about all the markup you've put. Once it is in the .rst file, Travis-CI tries to build the doc. You can also build it locally with a "make html" from docs/ after doing a `pip3 install --user sphinxcontrib-bibtex` and installing  doxygen andd graphviz (I'm just blindly hoping that works like that on cygwin on Windows. never tried)

joa-quim:
I'm sure I can build the docs on Windows without any cygwin, but I'm quite busy this week and next one goto [AGU](https://www.agu.org/fall-meeting). Perhaps you could break/rephrase the text to your liking?

rouault:
> Perhaps you could break/rephrase the text to your liking?

OK, I'll merge this at is, and follow up with doc improvements. Thanks for your contribution!

PR #1756, opened 2019-11-28
Doc: add a operations_computation.rst describe the magic behind createOperations()

rouault:
Long overdue. All you wanted to know, or maybe not :-)

### DIFF
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -215,7 +215,7 @@ and will add at the end a conversion between the second intermediate CRS and the
 target CRS. Those conversions are conversion between geographic 2D and geographic 3D
 CRS or geographic 2D/3D and geocentric CRS.
 
-This is done by the createOperationsWithDatumPivot() method.
+This is done by the :c:func:`createOperationsWithDatumPivot()` method.
 
 So if transforming between EPSG:7042, RGF93 (lon-lat), geographic 3D and
 EPSG:4936, ETRS89, geocentric, one get the following concatenated operation,
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -449,7 +449,7 @@ If one forces the datum hub to be considered to be EPSG:4326, ones gets:
 
 Those operations are less accurate, since WGS 84 is assumed to be equivalent to
 GDA2020 with an accuracy of 4 metre. This is an instance demonstrating that using
-systematically the WGS 84 hub can be sub-optimal.
+WGS 84 as a hub systematically can be sub-optimal.
 
 There are still situations where the attempt to find a hub CRS does not work,
 because there is no such hub. This can occur for example when transforming from
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -116,7 +116,7 @@ not attempt any datum shift, hence the "ballpark" qualifier in its name.
 Filtering and sorting of coordinate operations
 ----------------------------------------------
 
-The last step is to filter and sort results in a relevance order.
+The last step is to filter and sort results in order of relevance.
 
 The filtering
 takes into account a minimum accuracy that the user might have expressed, an
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -69,7 +69,7 @@ Geographic CRS to Geographic CRS, with known identifiers
 
 With the above example of two geographic CRS, that have an identified identifier,
 (:program:`projinfo` internally resolves NAD27 to EPSG:4267 and NAD83 to EPSG:4269)
-the algorithm will first research
+the algorithm will first search
 in the coordinate operation related tables of the proj.db if there are records
 that list direct transformations between the source and the target CRS. The
 transformations typically involve Helmert-style operations or datum shift based on
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -207,7 +207,7 @@ So the 3 typical EPSG entries, 3 equivalent (with long, lat ordering for the
 geographic CRS) and one from the ESRI registry;
 
 PROJ can now test 10 x 7 different combinations of source x target CRSs, using
-the database research explained in the previous section. As soon as
+the database searching method explained in the previous section. As soon as
 one of this combination returns at least one non-ballpark combination, the result
 set coming from that combination is used. PROJ will then add before that
 transformation a conversion between the source CRS and the first intermediate CRS,
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -72,7 +72,7 @@ With the above example of two geographic CRS, that have an identified identifier
 the algorithm will first search
 in the coordinate operation related tables of the ``proj.db`` if there are records
 that list direct transformations between the source and the target CRS. The
-transformations typically involve Helmert-style operations or datum shift based on
+transformations typically involve :ref:`Helmert <helmert>-style operations or datum shift based on
 grids (more esoteric operations are possible).
 
 A request similar to the following will be emitted:
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -505,7 +505,7 @@ This is implemented by the ``createOperationsDerivedTo`` method
 
 For the symetric case, source CRS to a derived CRS, the above algorithm is applied
 by switching the source and target CRS, and then inverting the resulting operation(s).
-This is mostly a matter of avoiding to write twice very similar code. This logic
+This is mostly a matter of avoiding to write very similar code twice. This logic
 is also applied to all below cases when considering the transformation between 2 different
 types of objects.
 
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -70,7 +70,7 @@ Geographic CRS to Geographic CRS, with known identifiers
 With the above example of two geographic CRS, that have an identified identifier,
 (:program:`projinfo` internally resolves NAD27 to EPSG:4267 and NAD83 to EPSG:4269)
 the algorithm will first search
-in the coordinate operation related tables of the proj.db if there are records
+in the coordinate operation related tables of the ``proj.db`` if there are records
 that list direct transformations between the source and the target CRS. The
 transformations typically involve Helmert-style operations or datum shift based on
 grids (more esoteric operations are possible).
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -71,7 +71,7 @@ With the above example of two geographic CRS, that have an identified identifier
 the algorithm will first search
 in the coordinate operation related tables of the ``proj.db`` if there are records
 that list direct transformations between the source and the target CRS. The
-transformations typically involve :ref:`Helmert <helmert>-style operations or datum shift based on
+transformations typically involve :ref:`Helmert <helmert>`-style operations or datum shift based on
 grids (more esoteric operations are possible).
 
 A request similar to the following will be emitted:
@@ -405,10 +405,9 @@ Actually, WGS 84 has been considered during the above lookup, because there are
 transformations between AGD84 and WGS 84 and WGS 84 and GDA2020. However those
 have been discarded in a step which we did not mention previously: just after
 the initial filtering of results and their sorting, there is a final filtering
-that is done. In the list of sorted results, if a less prioritary result than
-its previous one has the same area of use, but a lesser accuracy and that the
-more accurace results does not use grids, or the grids are available, then the
-less accurate result is discarded.
+that is done. In the list of sorted results, given two operations A and B that
+have the same area of use, if B has an accuracy lower than A, and A does not use
+grids, or all the needed grids are available, then B is discarded.
 
 If one forces the datum hub to be considered to be EPSG:4326, ones gets:
 
@@ -593,7 +592,7 @@ the transformation from a compound CRS made of a geographic CRS to another geogr
 
 It first starts by the vertical transformations from the vertical CRS of the
 source compound CRS to the target geographic CRS, using the strategy detailed
-in verttogeog_
+in :ref:`Vertical CRS to a Geographic CRS <verttogeog>`
 
 What we did not mention is that when there is not a transformation registered
 between the vertical CRS and the target geographic CRS, PROJ attempts to find
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -161,7 +161,7 @@ the CRS to identify and whose name exactly matches the one provided to the
 
 If this search did not succeed, or if the previous case with known CRS identifiers
 did not result in matches in the database, the search will be based on the
-datums. That is: a list of geographic CRS whose datum matches the datum of the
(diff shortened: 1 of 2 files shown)


nyalldawson:
You should probably submit this to Cartographic Perspectives or some other journal and get some nice academic rep points 😜

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+* put at top operations whose area of use is larger (note: the computation of the
+  are of use is approximate, based on a bounding box)
+* put at top operations that have a better accuracy
+* in case of same accuracy, put at top operations that do not use grids (operations
+  that use only parameters will be faster)
+* put at top operations that involve less transformation steps
+* and for completness, if two operations are comparable given all the above criteria,
+  put at first the one that has the shorter name, and if they have the same lengt, sort
+  by lexicographic order (obviously completely arbitrary, but a sorting
+  algorithm must be able to compare all entries)
```
kbevers: I am confused by this. I know this is not easy to express this simply! You are building the list from the bottom, yes? The second and third step feels wrong to me, that you put operations without ballpark operations almost at the bottom at the list. I would intiutively assume that operations like that would be more direct and thus better suited at the top of the list?

At the moment I don't have a good suggestion for how to improve this.

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+
+Geodetic/geographic CRS to Geodetic/geographic CRS, without known identifiers
+-----------------------------------------------------------------------------
+
+In a number of situations, the source and/or target CRS do not have an identifier
+(WKT without identifier, PROJ string, ..)
+The first step is to try to find in the proj.db a CRS of the same nature of
```
kbevers:
```suggestion
The first step is to try to find in the ``proj.db`` a CRS of the same nature of
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+the CRS to identify and whose name exactly matches the one provided to the
+createOperations method. If there is exactly one match and that the CRS are
```
kbevers:
```suggestion
:c:func:`createOperations` method. If there is exactly one match and that the CRS are
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+- IGNF:ETRS89, ETRS89 cartesiennes geocentriques, geocentric
+- IGNF:ETRS89G, ETRS89 geographiques (dms), geographic 2D
+- IGNF:ETRS89GEO, ETRS89 geographiques (dms), geographic 3D
+- ESRI:104129, GCS_EUREF_FIN, geographic 2D
+
+So the 3 typical EPSG entries, 3 equivalent (with long, lat ordering for the
+geographic CRS) and one from the ESRI registry;
+
+PROJ can now test 10 x 7 different combinations of source x target CRSs, using
+the database research explained in the previous section. As soon as
```
kbevers:
```suggestion
the database searching method explained in the previous section. As soon as
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+one of this combination returns at least one non-ballpark combination, the result
+set coming from that combination is used. PROJ will then add before that
+transformation a conversion between the source CRS and the first intermediate CRS,
+and will add at the end a conversion between the second intermediate CRS and the
+target CRS. Those conversions are conversion between geographic 2D and geographic 3D
+CRS or geographic 2D/3D and geocentric CRS.
+
+This is done by the createOperationsWithDatumPivot() method.
```
kbevers:
```suggestion
This is done by the :c:func:`createOperationsWithDatumPivot()` method.
```

kbevers (changes requested):
I've made it half-way through so far. I'll continue my review later. So far I've put in a few re-wordings and suggestions for improved markup.

docs/source/operations/operations_computation.rst
rouault: > You are building the list from the bottom, yes?

not sure to understand what you mean...

> The second and third step feels wrong to me, that you put operations without ballpark operations almost at the bottom at the list

So I wrote "put at top operations that do not have a synthetic ballpark vertical transformation" . Maybe you missed the negation ? Or "put at top" is not clear ? "Put at top" = "more relevant" (the one you see first in the output of projinfo)

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+    EPSG:1573, NAD27 to NAD83 (6), 1.5 m, Canada - Quebec, at least one grid missing
+    EPSG:9111, NAD27 to NAD83 (9), 1.5 m, Canada - Saskatchewan, at least one grid missing
+    unknown id, Ballpark geographic offset from NAD27 to NAD83, unknown accuracy, World, has ballpark transformation
+    EPSG:8555, NAD27 to NAD83 (7), 0.15 m, USA - CONUS and GoM, at least one grid missing
+    EPSG:8549, NAD27 to NAD83 (8), 0.5 m, USA - Alaska, at least one grid missing
+
+The algorithm involves many cases, so we will progress in the explanation from
+the most simple case to more complex ones. We document here the working of this
+algorithm as implemented in PROJ 6.3dev, november 2019,
+revision cef7a8a209717938a08e909dcc03c05b09c5cd77
```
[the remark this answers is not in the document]
rouault: done

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+    PROJ string:
+    +proj=pipeline +step +proj=axisswap +order=2,1 \
+                   +step +proj=unitconvert +xy_in=deg +xy_out=rad \
+                   +step +proj=hgridshift +grids=National_84_02_07_01.gsb \
+                   +step +proj=hgridshift +grids=GDA94_GDA2020_conformal.gsb \
+                   +step +proj=unitconvert +xy_in=rad +xy_out=deg \
+                   +step +proj=axisswap +order=2,1
+
+One can see that the selected intermediate CRS that has been used is GDA94.
+This is a completely novel behaviour of PROJ 6 regarding to the PROJ.4 logic
```
kbevers:
```suggestion
This is a completely novel behaviour of PROJ 6 as opposed to the logic of PROJ.4 
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+datum hub. PROJ 6 no longer hardcodes it as the mandatory datum hub, and relies
+on the database to find the appropriate hub(s).
+Actually, WGS 84 has been considered during the above lookup, because there are
+transformations between AGD84 and WGS 84 and WGS 84 and GDA2020. However those
+have been discarded in a step which we did not mention previously: just after
+the initial filtering of results and their sorting, there is a final filtering
+that is done. In the list of sorted results, if a less prioritary result than
+its previous one has the same area of use, but a lesser accuracy and that the
+more accurace results does not use grids, or the grids are available, then the
+less accurate result is discarded.
```
kbevers: Try to rephrase this sentence. I struggle to figure out what you are saying in the beginning. If all other things are equal, the less accurate hub datum is discarded, I guess?

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+    PROJ string:
+    +proj=pipeline +step +proj=axisswap +order=2,1 \
+                   +step +proj=unitconvert +xy_in=deg +xy_out=rad \
+                   +step +proj=hgridshift +grids=National_84_02_07_01.gsb \
+                   +step +proj=unitconvert +xy_in=rad +xy_out=deg \
+                   +step +proj=axisswap +order=2,1
+
+Those operations are less accurate, since WGS 84 is assumed to be equivalent to
+GDA2020 with an accuracy of 4 metre. This is an instance demonstrating that using
+systematically the WGS 84 hub can be sub-optimal.
```
kbevers:
```suggestion
WGS 84 as a hub systematically can be sub-optimal.
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+because there is no such hub. This can occur for example when transforming from
+GDA94 to the latest realization at time of writing of WGS 84, WGS 84 (G1762).
+There are transformations between WGS 84 (G1762). Using the above described
+techniques, we would only find one non-ballpark operation taking the route:
+1. Conversion from GDA94 (geog2D) to GDA94 (geocentric): synthetized by PROJ
+2. Inverse of ITRF2008 to GDA94 (1): from EPSG
+3. Inverse of WGS 84 (G1762) to ITRF2008 (1): from EPSG
+4. Conversion from WGS 84 (G1762) (geocentric) to WGS 84 (G1762): synthetized by PROJ
+
+This is not bad, but the global validity area of is "Australia - onshore and EEZ",
```
kbevers:
```suggestion
This is not bad, but the global validity area of use is "Australia - onshore and EEZ",
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+whereas GDA94 has a larger area of use.
+There is another road that can be taken by going throug GDA2020 instead of ITRF2008.
+The GDA94 to GDA2020 transformations operate on the respective geographic CRS,
+whereas GDA2020 to WGS 84 (G1762) operate on the geocentric CRS. Consequently,
+GDA2020 cannot be identifier as a hub by a "simple" self-join SQL request on
+the coordinate operation table. This requires to do the join based on the datum
+referenced by the source and target CRS of each operation rather than the
+source and target CRS themselves. Then we there is a match, PROJ inserts the required
```
kbevers:
```suggestion
source and target CRS themselves. When there is a match, PROJ inserts the required
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+3. GDA2020 to WGS 84 (G1762) (1): frmo EPSG
+4. Conversion from WGS 84 (G1762) (geocentric) to WGS 84 (G1762) (geog2D): synthetized by PROJ
+
+Projected CRS to any target CRS
+---------------------------------------------------------------------------------
+
+This actually extends to any Derived CRS, whose Projected CRS is a well-known
+particular case. Such transformations are done in 2 steps:
+
+1. Use the inverse conversion of the derived CRS to its base CRS, typically a
```
kbevers:
```suggestion
1. Use the inverse conversion of the derived CRS to its base CRS, typically an
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+    unknown id, Inverse of UTM zone 31N + Inverse of RGF93 to WGS 84 (1), 1 m, France
+
+    PROJ string:
+    +proj=pipeline +step +inv +proj=utm +zone=31 +ellps=WGS84 +step +proj=unitconvert +xy_in=rad +xy_out=deg +step +proj=axisswap +order=2,1
+
+This is implemented by the ``createOperationsDerivedTo`` method
+
+For the symetric case, source CRS to a derived CRS, the above algorithm is applied
+by switching the source and target CRS, and then inverting the resulting operation(s).
+This is mostly a matter of avoiding to write twice very similar code. This logic
```
kbevers:
```suggestion
This is mostly a matter of avoiding to write very similar code twice. This logic
```

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+When the horizontal component of the compound source CRS is a projected CRS, we
+first look for the operation from this source CRS to another compound CRS made
+of the geographic CRS base of the projected CRS,
+like "NAD83 / California zone 1 (ftUS) + NAVD88 height" to "NAD83 + NAVD88 height",
+which ultimately goes to one of the above described case. Then we can consider
+the transformation from a compound CRS made of a geographic CRS to another geographic CRS.
+
+It first starts by the vertical transformations from the vertical CRS of the
+source compound CRS to the target geographic CRS, using the strategy detailed
+in verttogeog_
```
kbevers:
```suggestion
in :ref:`verttogeog_`
```

I guess?

docs/source/operations/operations_computation.rst
```diff
@@ -0,0 +1,741 @@
...
+the CRS to identify and whose name exactly matches the one provided to the
+createOperations method. If there is exactly one match and that the CRS are
+"computationnaly" equivalent, then use the code of the CRS for further computations.
+
+If this research did not succeed, or if the previous case with known CRS identifiers
+did not result in matches in the database, the research will be based on the
+datums. That is: a list of geographic CRS whose datum matches the datum of the
+source and target CRS is researched in the database (by querying the `geodetic_crs`
+database table). If the datum has a known
+identifier, we will use it, otherwise we will research an equivalent datum in the
```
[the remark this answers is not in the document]
rouault: sorry for that repeated frenchism ! I'll try to remember :-)

docs/source/operations/operations_computation.rst
rouault: Both are possible I think (see http://docutils.sourceforge.net/docs/user/rst/quickref.html#internal-hyperlink-targets). I'm always confused with rst vs sphinx linking mechanisms. At least, the syntax I used works

rouault: except that the title wasn't expanded. So fixing that

docs/source/operations/operations_computation.rst
rouault: rephrased as "In the list of sorted results, given two operations A and B that
have the same area of use, if B has an accuracy lower than A, and A does not use
grids, or all the needed grids are available, then B is discarded."

docs/source/operations/operations_computation.rst
kbevers: Yes, I think it is the "put a top" that is putting me off, so to speak. As well as there being many steps of which many are not trivial. I feel like I need a figure with a decision tree to decipher this algortithm :-)

If I understand it correctly, you start by making a list of all possible operations and then narrowing it down according to various "test" of subsequent priority. The final list is sorted by the number "tests" that the operations passes, with the first operation in the list being the one with the most passed "tests". Have I understood that somewhat correctly?

rouault: Not sure I understand your questions... lol
Let me try to put it otherwise. So this long enumeration is about the transcription of https://github.com/OSGeo/PROJ/blob/master/src/iso19111/coordinateoperation.cpp#L10777 . This is just a sorting function. It doesn't remove operations (operation filtering happens mostly before sorting, and also a bit after, but I didn't even get into that detail). The sort function takes 2 operations, a and b, and determines which one must be considered as "less" than the other. If it returns true, which means a < b, then a will be at the top of the list, and at the top of the output of projinfo, that is the more relevant.

rouault: @kbevers I've tried in fa7633e to clarify a bit this sorting stuff

kbevers: Thanks, this is a much better description of the process. I had to read the code to understand the other, I think this one can stand by itself. Good job!

kbevers (approved):

kbevers:
I think this is ready to be merged. Since it states in the text that the algorithms described matches those of PROJ 6.3 we should probably not backport to the 6.2 branch?

rouault:
yes, fine in master. Most of the principles remain true for 6.2, but anyway this will be live on proj.org in one month when 6.3 is out, so it's fine

PR #1763, opened 2019-12-01
Update grid_alternatives.sql for Icelandic grids

dalia2:
Updated the grid_alternatives.sql with ISN93_ISN2016.gsb and ISN2004_ISN2016.gsb


data/sql/grid_alternatives.sql
```diff
@@ -1468,3 +1468,34 @@ INSERT INTO grid_alternatives(original_grid_name,
...
+                              proj_grid_format,
+                              proj_method,
+                              inverse_direction,
+                              package_name,
+                              url, direct_download, open_license, directory)
+                      VALUES ('ISN93_ISN2016.gsb',
+                              'ISN93_ISN2016.gsb',
+                              'NTv2',
+                              'hgridshift',
+                              1,
```
rouault:
```suggestion
                              0,
```
I believe this should be 0, to mean that there is no inversion of direction.

data/sql/grid_alternatives.sql
```diff
@@ -1468,3 +1468,34 @@ INSERT INTO grid_alternatives(original_grid_name,
...
+                              proj_grid_format,
+                              proj_method,
+                              inverse_direction,
+                              package_name,
+                              url, direct_download, open_license, directory)
+                      VALUES ('ISN2004_ISN2016.gsb',
+                              'ISN2004_ISN2016.gsb',
+                              'NTv2',
+                              'hgridshift',
+                              1,
```
rouault:
```suggestion
                              0,
```
Same here

kbevers:
The conflicts in this PR needs to be fixed before it can be merged. I suggest rebasing on top of master. Will you please do that @dalia2

kbevers:
@rouault are all the necessary grids from the datumgrid PR registered here?

data/sql/grid_transformation_custom.sql
```diff
@@ -55,3 +55,38 @@ INSERT INTO "grid_transformation" VALUES(
...
+
+INSERT INTO "grid_transformation" VALUES(
+    'PROJ','EPSG_5323_TO_EPSG_8089','ISN2004 height to vertical datum ISH2004',
+    NULL,NULL,
+    'PROJ','HEIGHT_TO_GEOGRAPHIC3D','GravityRelatedHeight to Geographic3D',
+    'EPSG','5323', -- source CRS (ISN2004 height)
+    'EPSG','8089', -- target CRS (ISH2004)
+    'EPSG','1120', -- area of use: Iceland - onshore and offshore
+    NULL,
+    'EPSG','8666','Geoid (height correction) model file','Icegeoid_ISN2004.gtx',
```
rouault, line 69: there should be 3 entries in grid_alternatives for Icegeoid_ISN2004.gtx, Icegeoid_ISN93.gtx and Icegeoid_ISN2016.gtx. They should be similar to the one of SWEN17_RH2000.gtx (the important thing is to have inverse_direction = 0 for that particular case as you use a 'GravityRelatedHeight to Geographic3D' method)

rouault:
> are all the necessary grids from the datumgrid PR registered here?

Except the remark I just added regardign missing entries in grid_alternatives, it looks complete to me.

What is potential lacking is the registration of the velocity models, but I suggested previously to  let that for later as it would require putting a PROJ string pipeline in the other_transformation table, which is theoretically possible, but not exercised beyond the unit tests.

dalia2:
All grids are registered in PR and now I will fill the filelist.csv

dalia2:
I did not add only these two ISN2004_ISN2016.gtx and ISN93_ISN2016.gtx  into filelist.csv, because these two gtx should go together with ISN2004_ISN2016.gsb and ISN93_ISN2016.gsb like deformation model taht would require putting a PROJ string pipeline, so w skipped that idea.

PR #1766, opened 2019-12-03
Adding New Zealand grid alternative names

ccrook:
This adds the current EPSG grid names to grid_alternatives.sql.  This may be a short term fix as it is hoped to get these names updated in EPSG.

### DIFF
--- a/data/sql/grid_alternatives.sql
+++ b/data/sql/grid_alternatives.sql
@@ -1713,3 +1713,231 @@ INSERT INTO grid_alternatives(original_grid_name,
                               0,
                               'proj-datumgrid-europe',
                               NULL, NULL, NULL, NULL);
+
+-- New Zealand grid shift models.  EPSG names are not for GTX files (at time of writing)
+INSERT INTO grid_alternatives(original_grid_name,
+                              proj_grid_name,
+                              proj_grid_format,
+                              proj_method,
+                              inverse_direction,
+                              package_name,
+                              url, direct_download, open_license, directory)
+                      VALUES ('auckland-1946-to-nzvd2016-conversion.csv',
+                              'auckht1946-nzvd2016.gtx',
+                              'GTX',
+                              'vgridshift',
+                              0,
+                              'proj-datumgrid-oceania',
+                              NULL, NULL, NULL, NULL);
+
+INSERT INTO grid_alternatives(original_grid_name,
+                              proj_grid_name,
+                              proj_grid_format,
+                              proj_method,
+                              inverse_direction,
+                              package_name,
+                              url, direct_download, open_license, directory)
+                      VALUES ('bluff-1955-to-nzvd2016-conversion.csv',
+                              'blufht1955-nzvd2016.gtx',
+                              'GTX',
+                              'vgridshift',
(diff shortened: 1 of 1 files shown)


rouault:
Can you rebase that on top of current master ?

rouault:
Can you:
- rebase this on top of latest master ? (I believe your added test might need to be adjusted)
- run "python scripts/grid_checks.py --not-in-proj-datumgrid /path/to/data/proj.db /path/to/proj-datumgrid" (it will verify that the alias name you put match files that do exist in proj-datumgrid). No output = everything OK

ccrook:
@rouault The grid_checks.py ran with no output, so I think it is Ok.  

I'm a little unclear as to whether the test on the NZLVD method should have failed.  Currently it is passing without modification, using the original name for the auck grid (ie auckland-1946-to-nzvd2016-conversion.csv), and although grid_alternatives does contain the correct grid name:

```
$ sqlite3 data/proj.db
SQLite version 3.22.0 2018-01-22 18:45:57
Enter ".help" for usage hints.
sqlite> select * from grid_alternatives where original_grid_name like '%auck%';
auckland-1946-to-nzvd2016-conversion.csv|auckht1946-nzvd2016.gtx|GTX|vgridshift|0|proj-datumgrid-oceania||||
sqlite>
```
So it looks like the alternative name isn't being used.  On the other hand if I use projinfo the new grid is found:

```
$ PROJ_LIB=/home/ccrook/proj/build/data bin/projinfo -s EPSG:4167+5759 -t EPSG:4167
Candidate operations found: 2
Operation n°1:

unknown id, Inverse of NZVD2016 height to Auckland 1946 height (1) + Inverse of NZGD2000 to NZVD2016 height (1), 0.12 m, New Zealand - North Island - Auckland vcrs, at least one grid missing

PROJ string:
+proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +inv +proj=vgridshift +grids=auckht1946-nzvd2016.gtx +multiplier=1 +step +proj=vgridshift +grids=nzgeoid2016.gtx +multiplier=1 +step +proj=unitconvert +xy_in=rad +xy_out=deg +step +proj=axisswap +order=2,1

...

Grid auckht1946-nzvd2016.gtx needed but not found on the system. Can be obtained from the proj-datumgrid-oceania package at https://download.osgeo.org/proj/proj-datumgrid-oceania-1.1.zip
Grid nzgeoid2016.gtx needed but not found on the system. Can be obtained from the proj-datumgrid-oceania package at https://download.osgeo.org/proj/proj-datumgrid-oceania-1.1.zip

Operation n°2:
```

So it all appears to be working as it should, just the test isn't picking the renamed file.

rouault:
Merged

> So it looks like the alternative name isn't being used.

Yes, I finally figured out why. The PROJStringFormatter of the unit test didn't have a database context available, so couldn't make the substitution. Fixed per 5f36c9ebde150e7ece7d153bf1dc7bd213de6016

ccrook:
Brilliant! Thanks for all your help with this one.

PR #1770, opened 2019-12-03
Adding support for NZLVD vertical height transformation

ccrook:
@rouault Here are the changes you outlined to enable support for the  NZLVD transformation grid (EPSG:1071).  

I think this needs more work but thought creating a pull request would be the easiest way to find out what!  Two issues I know of:

1. I haven't created a unit test for the method. Looking at the VERTCON method I think this would entail building a createNZLVD method, and then adding a unit test that creates it and checks the resultant transformation string?  

2. The EPSG:1071 method also includes a parameter for the interpolation CRS.  I'm not sure what to do with this!  At the moment I believe I am ignoring it, basically assuming that the horizontal CRS is correct without checking.  Alternatives would be to fail if it doesn't match, or to convert the horizontal coordinates to match before interpolating.

The good thing (for me) is that with this change the New Zealand local vertical datum coordinate systems can be used from EPSG codes, and give the correct results.


rouault:
>* I think this would entail building a createNZLVD method, and then adding a unit test that creates it and checks the resultant transformation string?

Hum not sure we want to extend the API with createNZLVD() (the createVERTCON() wasn't probably a so great idea). What could be done instead is to instantiate one of the NZLVD transformations from the database by its EPSG code (with AuthorityFactory::createCoordinateOperation()) and export it to a PROJ string and chech the value of that PROJ string which is what matters ultimately

Something along
```
auto factory = AuthorityFactory::create(DatabaseContext::create(), "EPSG");
auto op = factory->createCoordinateOperation("7860", false);
EXPECT_EQ(op->exportToPROJString(PROJStringFormatter::create().get()), "put here proj pipeline");
```

> The EPSG:1071 method also includes a parameter for the interpolation CRS. I'm not sure what to do with this! 

Nothing to do normally. The interpolation CRS is already registered in the interpolation_crs_auth_name and interpolation_crs_code columns of grid_transformation, 
See 'EPSG','4167' at the end of
```
INSERT INTO "grid_transformation" VALUES('EPSG','7860','NZVD2016 height to Auckland 1946 height (1)','Derived at 260 control points. Mean offset 0.292m, standard deviation 0.029m, maximum difference from mean 0.075m.','Transformation between national and local height systems.','EPSG','1071','Vertical Offset by Grid Interpolation (NZLVD)','EPSG','7839','EPSG','5759','EPSG','3764',0.02,'EPSG','8732','Vertical offset file','auckland-1946-to-nzvd2016-conversion.csv',NULL,NULL,NULL,NULL,'EPSG','4167','LINZ-NZ AUCK',0);
```
and it will be set to the corresponding member of the C++ Transformation object (returned by the Transformation::interpolationCRS() method). And the logic that creates global pipelines should hopefully use it correctly to do horizontal conversions to it before applying the vertical transform.
You could check that it works by transforming for example from WGS84+NZVD2016 height to WGS84+Auckland 1946 height with projinfo and look at the steps in the WKT output. This will have probably have no effect on the resulting PROJ pipeline as NZGD2000 to WGS84 is a null shift

src/proj_constants.h
```diff
@@ -513,6 +513,9 @@
 #define EPSG_CODE_METHOD_VERTCON 9658
 #define EPSG_NAME_METHOD_VERTCON "VERTCON"
 
+#define EPSG_CODE_METHOD_VERTICALGRID_NZLVD 1071
+#define EPSG_NAME_METHOD_VERTICALGRID_NZLVD "NZLVD"
```
rouault: The value should be the exact name in EPSG, so "Vertical Offset by Grid Interpolation (NZLVD)"
```suggestion
#define EPSG_NAME_METHOD_VERTICALGRID_NZLVD "Vertical Offset by Grid Interpolation (NZLVD)"
```

ccrook:
I am expecting the method code and name to change due to EPSG change request 2019.092.  Apparently this will be discussed at a meeting tomorrow.  So will amend if possible following that to match future method name/code.

ccrook:
@rouault  - for debugging you suggested using something like:

```
auto factory = AuthorityFactory::create(DatabaseContext::create(), "EPSG");
auto op = factory->createCoordinateOperation("7860", false);
EXPECT_EQ(op->exportToPROJString(PROJStringFormatter::create().get()), 
```

I think this will make the test dependent on data in EPSG, as well as being able to access EPSG.  Are we assuming that EPSG method cannot change (though presumably could be made invalid)?  Is this Ok for testing the operation implementation?  I guess so as you are suggesting it!  But just want to confirm.

rouault:
> I think this will make the test dependent on data in EPSG

That's a good point. Yes, the test suite is highly dependent of the actual EPSG content, and has to be adapted when we refresh it. Both a weakness and a strength. The thing is that as users will do use the actual proj.db and that the algorithms are sometimes quite sensitive to its content, this helps identifying if an update breaks real use cases (either because of a bug/weakness in the software or in the database itself)

ccrook:
True.. It's just particularly an issue for me as there is a a pull request for updating grid alternatives, this request, the proj-datumgrid update that will mean grids become available, and the proposed EPSG change all in flight at once.  In particular if I get the new EPSG code then I won't have a testable method until it is published (maybe that is instantaneous - I don't know).  Maybe I just need to be a bit patient!

rouault:
>  if I get the new EPSG code then I won't have a testable method until it is published (maybe that is instantaneous - I don't know)

for now, I'd say: ignore the fact that EPSG is going to make some later changes, and just stick with the existing code & name. One problem at a time :-)

ccrook:
Test built on current EPSG and grid_alternatives.
@rouault Thanks for suggesting the test code.  That made it very easy  :-)

rouault:
Merging

PR #1773, opened 2019-12-04
Build: Only export symbols if building DLL

orudge:
If you build Proj as a static library (e.g., using vcpkg), then subsequently build a DLL linking to the Proj libraries, all the Proj symbols get exported in your DLL.

### DIFF
--- a/src/geodesic.h
+++ b/src/geodesic.h
@@ -158,7 +158,7 @@
                       GEODESIC_VERSION_PATCH)
 
 #if !defined(GEOD_DLL)
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) && defined(PROJ_MSVC_DLL_EXPORT)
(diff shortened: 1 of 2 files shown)


rouault:
ping @cffk : could you review the change in geodesic.h ?

cffk:
> ping @cffk : could you review the change in geodesic.h ?

This looks fine...

cffk:
Sorry to second guess myself.  But I don't see where geodesic.h has a
dllimport option when used by a program that links to proj.  Also
instead of the cmake files for the utilities setting
PROJ_MSVC_DLL_IMPORT=1, wouldn't it make more sense for the *library* to
make this an INTERFACE definition?

I'm not familiar enough with the intricacies of Windows dlls to know how
this might need to be fixed without actually trying it -- and
unfortunately, I'm off on vacation soon and so won't be able to look
into this until the beginning of Jan.

PR #1777, opened 2019-12-06
[RFC4_dev] Grid loading refactoring

rouault:
Pull request against the rfc4_dev branch. This pull request features:
- removal of the obsolete/undocumented/unfinished grid catalog functionality
- reworking of horizontal and vertical grid loading so that they are no longer ingested in memory. 

Incidentally I removed support for the old CTable format (CTable2 remains), that I think was broken for 64-bit due to the CTABLE structure having a pointer whose size is different in 32 & 64 bit builds, and we don't have any grids in that format.


src/apply_gridshift.cpp
```diff
@@ -334,33 +289,92 @@ PJ_LP proj_hgrid_value(PJ *P, PJ_LP lp) {
...
+    out = nad_cvt(lp, inverse, grid);
 
     if (out.lam == HUGE_VAL || out.phi == HUGE_VAL)
-        pj_ctx_set_errno(P->ctx, PJD_ERR_GRID_AREA);
+        pj_ctx_set_errno(ctx, PJD_ERR_GRID_AREA);
 
     return out;
+}
+
+PJ_LP proj_hgrid_apply(PJ *P, const ListOfHGrids& grids, PJ_LP lp, PJ_DIRECTION direction) {
```
kbevers, line 322: `proj_hgrid_apply` is only used internally, so you are allowed to change its function parameters etc. Seems better than to add the `_internal` function (unless there's something I haven't understood).

The `proj_` prefix is misleading. Should be changed to `pj_` here as well.

src/apply_vgridshift.cpp
```diff
@@ -314,33 +160,43 @@ int proj_vgrid_init(PJ* P, const char *grids) {
...
-    }
-
-    pj_dealloc(sgrids);
-    return P->vgridlist_geoid_count;
+    return grids;
 }
 
 /***********************************************/
-double proj_vgrid_value(PJ *P, PJ_LP lp, double vmultiplier){
+double proj_vgrid_value(PJ *P, const ListOfVGrids& grids, PJ_LP lp, double vmultiplier){
```
kbevers, line 199: Another case of bad prefix: `proj_` -> `pj_`.

kbevers (commented):
Overall I think this is great. I like the new classes for reading grids, a good example of C++ having advantages over C I think. I am also happy to see the grid catalog code and CTable support removed. 

The grid code has needed a good clean up for a while. This is takes care of most of it. I have a few comments that might improve things further:

- `grids.hpp`/`grids.cpp`: Great addition to the code base! I think it would be better to go full monty and place all grid related functions in those two files. It improves the overall structure of the code and makes it easier to find what you are looking for. 

- nad_* functions: If I understand correctly some of the `nad_` functions has been removed. I think it would be an improvement to rename the remaining ones to something more easily understood. These days the functions are more generic than "North American Datum".

- `proj_` prefixes on internal functions: These should really be prefixed with `pj_`. I am talking about the ´proj_hgrid_*` and `proj_vgrid_` functions. I initally created them I think and should have given them another prefix from the start. Since they (at least some) are being modified now they may as well also get the proper prefix.

rouault:
> I have a few comments that might improve things further:

I agree with your comments. I tried to do the minimal amount of changes to accomplish my goal and limit the amount of code churn, but I'm happy to go further along your suggestions. If you don't mind, I'll queue the changes as extra commits at the end of the latest PR of RFC4 (https://github.com/OSGeo/PROJ/pull/1817) (or perhaps a rfc4_code_review branch based on it), so as to avoid potentially complex rebase/merge among all those branches.

kbevers:
> I tried to do the minimal amount of changes to accomplish my goal and limit the amount of code churn, but I'm happy to go further along your suggestions

Sure, I can understand that reasoning. But since you are already touching more or less everything related to grids it is a good time to get things sorted once and for all. 

> If you don't mind, I'll queue the changes as extra commits at the end of the latest PR of RFC4 (#1817) (or perhaps a rfc4_code_review branch based on it), so as to avoid potentially complex rebase/merge among all those branches.

Of course. Rebasing this could be a difficult task, no need to waste too much time on that.

kbevers (approved):
With my comments adressed in #1826 I am happy with this pull request.

Issue #1525, opened 2019-06-22
pipeline parsing issues with `+proj=ob_tran +o_proj=longlat`


snowman2:
This is based on a question I noticed: https://gis.stackexchange.com/questions/312084/transform-coordinate-from-rotated-lat-lon-to-normal-lat-lon-wgs84

Version info:
```
$ projinfo
Rel. 6.1.0, May 15th, 2019
...
```
Initially I did attempted the transformation and got inf:
```
$ echo '46 -122' | cs2cs -v 'epsg:4326' '+proj=ob_tran +o_proj=longlat +lon_0=-40 +o_lat_p=22 +R=6.371e+06 +type=crs'
# ---- From Coordinate System ----
epsg:4326
# ---- To Coordinate System ----
+proj=ob_tran +o_proj=longlat +lon_0=-40 +o_lat_p=22 +R=6.371e+06 +type=crs
*	* inf

```
Then I did `projinfo` on the transformation:
```
$ projinfo -s 'epsg:4326' -t '+proj=ob_tran +o_proj=longlat +lon_0=-40 +o_lat_p=22 +R=6.371e+06 +type=crs'
Candidate operations found: 1
-------------------------------------
Operation n°1:

unknown id, Ballpark geographic offset from WGS 84 to unknown + unknown, unknown accuracy, World, has ballpark transformation

PROJ string:
+proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=ob_tran +o_proj=longlat +lon_0=-40 +o_lat_p=22 +R=6371000 +step +proj=unitconvert +xy_in=rad +xy_out=deg
...
```
And then I tried `projinfo` on just the PROJ string:
```
$ projinfo '+proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=ob_tran +o_proj=longlat +lon_0=-40 +o_lat_p=22 +R=6371000 +step +proj=unitconvert +xy_in=rad +xy_out=deg'
input string: parsing of user string failed: Error -50: malformed pipeline
```

Is there something I am doing incorrectly? Thanks!

snowman2:
Thanks for the fix!

PR #1783, opened 2019-12-10
ob_tran: restore traditional handling of +to_meter with pj_transform() and proj utility (fixes #1782)

### LINKED ISSUE #1782: 'to_meter' on 'proj=ob_tran o_proj=latlon' broke in 6.1.1
In Cartopy, there is a `RotatedGeodetic` CRS which uses 'ob_tran' to define a rotated coordinate system, using degrees with 'o_proj=latlon'. It specifies `to_meter=math.radians(1)` to produce results in degrees as well. Cartopy has not been updated to use any new API, so a transform from `PlateCarree` to `RotatedGeodetic` is something like:

```
pj_transform("+ellps=WGS84 +a=57.29577951308232 +proj=eqc +lon_0=0.0 +no_defs"
             "+ellps=WGS84 +proj=ob_tran +o_proj=latlon +o_lon_p=0.0 +o_lat_p=90.0 +lon_0=360.0 +to_meter=0.0174532925199433 +no_defs",
             nx, ny, &x, &y, NULL);
```
(I am skipping all the init stuff here, of course.)

The pole location and rotation angle _should_ result in no change to the result. Starting with 6.1.1, the `to_meter` has no effect and results are `to_meter` times smaller than they should be (or the inverse, if transforming in reverse.) For example, with 6.1.0, some test points are:
```
(-179, -89) -> (-179.00000000000003, -89.00000000000007)
(-179, 0) -> (-179.00000000000003, 3.507820309861257e-15)
(-179, 89) -> (-179.00000000000003, 89.00000000000007)
(0, -89) -> (0.0, -89.00000000000007)
(0, 0) -> (0.0, -3.508354649267437e-15)
(0, 89) -> (0.0, 89.00000000000007)
(179, -89) -> (178.99999999999997, -89.00000000000007)
(179, 0) -> (178.99999999999997, 3.507820309861257e-15)
(179, 89) -> (178.99999999999997, 89.00000000000007)
```
but with 6.1.1, the results are:
```
(-179, -89) -> (-3.124139361069851, -1.553343034274955)
(-179, 0) -> (-3.124139361069851, 6.122301397540666e-17)
(-179, 89) -> (-3.124139361069851, 1.553343034274955)
(0, -89) -> (0.0, -1.553343034274955)
(0, 0) -> (0.0, -6.123233995736766e-17)
(0, 89) -> (0.0, 1.553343034274955)
(179, -89) -> (3.12413936106985, -1.553343034274955)
(179, 0) -> (3.12413936106985, 6.122301397540666e-17)
(179, 89) -> (3.12413936106985, 1.553343034274955)
```

I can also reproduce this directly with `proj`, 6.1.0:
```
for lon in -179 0 179; do
 for lat in -89 0 89; do
  echo $lon $lat
 done
done | proj -E +ellps=WGS84 +proj=ob_tran +o_proj=latlon +o_lon_p=0.0 +o_lat_p=90.0 +lon_0=360.0 +to_meter=0.0174532925199433 +no_defs

-179 -89        -179.00 -89.00
-179 0  -179.00 0.00
-179 89 -179.00 89.00
0 -89   0.00    -89.00
0 0     0.00    -0.00
0 89    0.00    89.00
179 -89 179.00  -89.00
179 0   179.00  0.00
179 89  179.00  89.00
```
vs 6.1.1:
```
for lon in -179 0 179; do
 for lat in -89 0 89; do
  echo $lon $lat
 done
done | proj -E +ellps=WGS84 +proj=ob_tran +o_proj=latlon +o_lon_p=0.0 +o_lat_p=90.0 +lon_0=360.0 +to_meter=0.0174532925199433 +no_defs

-179 -89        -3.12   -1.55
-179 0  -3.12   0.00
-179 89 -3.12   1.55
0 -89   0.00    -1.55
0 0     0.00    -0.00
0 89    0.00    1.55
179 -89 3.12    -1.55
179 0   3.12    0.00
179 89  3.12    1.55
```

Bisect says this started with 11a53b7c0255087c2b324b36147deb837902dfb9.

rouault:
Fixes side-effect of https://github.com/OSGeo/PROJ/issues/1525 that went in 6.1.1

The correction is horribly hacky. Sorry...


kbevers (changes requested):
> The correction is horribly hacky. Sorry...

Too hacky in my opinion. I don't think there's a way around a hacky solution but at least it should be localized to `ob_tran`. I think that should be possible by doing a checking what type of projection is used in `o_forward()` and then scaling the values so they come out as degrees.

rouault:
>  I think that should be possible by doing a checking what type of projection is used in `o_forward()` and then scaling the values so they come out as degrees.

That's pretty much what we already do in ob_tran.cpp with
```
    /* Support some rather speculative test cases, where the rotated projection */
    /* is actually latlong. We do not want scaling in that case... */
    if (Q->link->right==PJ_IO_UNITS_RADIANS)
        P->right = PJ_IO_UNITS_WHATEVER;
```

The reallity is that ob_tran has always been hacky. With the new interface of proj_create_crs_to_crs(), it is somewhat less hacky since we don't require any more (and no longer honour) the to_meter=0.0174532925199433 . Hence adding back the hack only when invoked from the ancient API.

I'm happy to let you find a better solution if you can think of one that doesn't make the fix of #1525 regress...

kbevers:
> That's pretty much what we already do in ob_tran.cpp with

Except that it is done in the setup and not on the individual coordinate. I would put in a check similar to 

```
    if( Proj->right == PJ_IO_UNITS_WHATEVER && Proj->descr &&
        strncmp(Proj->descr, "General Oblique Transformation",
                strlen("General Oblique Transformation")) == 0 )
    {
        Proj->right = PJ_IO_UNITS_PROJECTED;
    }
```

in `o_forward()`. 

> I'm happy to let you find a better solution if you can think of one that doesn't make the fix of #1525 regress...

I'd like to give it a go. I'm a bit overloaded at the moment but I think I can manage to squeeze in some time in the weekend.

rouault:
One thing I considered to make it slightly less awkward was to add a "int is_ob_trans" member in struct PJconsts, to avoid the string comparison.

> Except that it is done in the setup and not on the individual coordinate

I don't think this will have measurable runtime impact as it is done outside of the loop that transforms several coordinates at once and the Proj->right == PJ_IO_UNITS_WHATEVER  comparison is going to be false for pretty much all projection methods, so the string comparison should be done only in a few cases

> I would put in a check similar to in o_forward()

We do want a different behaviour depending if we use it through proj_create() or pj_transform(), so I'm not sure the solution can come from a change in ob_tran.cpp itself. Happy to be be proven wrong ! Or perhaps through all the magic that occurs in fwd.cpp / inv.cpp with the pre and post processing stages.

kbevers:
> We do want a different behaviour depending if we use it through proj_create() or pj_transform(), so I'm not sure the solution can come from a change in ob_tran.cpp itself. Happy to be be proven wrong ! Or perhaps through all the magic that occurs in fwd.cpp / inv.cpp with the pre and post processing stages.

Good point about the differing behaviour in pj_transform() and proj_create(). It could be that the pre- and post-processing stages are the correct place to fix this.

kbevers (approved):
I didn't find the time to look at this. The hacky fix it is!

PR #1784, opened 2019-12-10
support EPSG:102100 (fixes #1730)

### LINKED ISSUE #1730: Support EPSG:102100 (ESRI legacy) as alias for EPSG:3857
Feel free to reject this RFE without hurting my feelings :)

Much like how `EPSG:900913` is aliased to `EPSG:3857`, it would be nice if `EPSG:102100` was aliased to `EPSG:3857` as well.  As I review my various WMS/TMS services using Proj, I still see a fair number of ESRI clients out in the wild still using that legacy code.

A couple years back, I attempted to remove `EPSG:102100` from my manually augmented Proj-epsg database and I got downstream users that had no available workaround from ESRI within their mapping apps.  Maybe I should try to pull the bandaid off again and see what breaks :)

At least this issue is documented (hopefully I did not miss this issue elsewhere) here now for reference and thanks for the consideration.

akrherz:
Untested by me other than it appears to build :/

The motivation here is that I continue to see (right or wrong) lots of in-the-wild usage of `EPSG:102100` against my WMS (mapserver).  Ditto for `EPSG:900913`, which is already aliased like this in PROJ.

### DIFF
--- a/data/other.extra
+++ b/data/other.extra
@@ -51,3 +51,5 @@
 #  Funny epsgish code for google mercator - you should really use EPSG:3857
 #
 <900913> +proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext  +no_defs <>
+# ESRI code sometimes applied to EPSG, again use EPSG:3857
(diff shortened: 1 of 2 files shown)


rouault:
I've submitted an alternative resolution in https://github.com/OSGeo/PROJ/pull/1786

akrherz:
Thank you very much @rouault

PR #1790, opened 2019-12-11
[RFC4_dev] Add GeoTIFF grids support

rouault:
This is a continuation of https://github.com/OSGeo/PROJ/pull/1777. Only the commits starting with "Build and CI: add libtiff dependency" are specific to this pull request.

As a bonus point, I've added a +proj=xyzgridshift method to perform geocentric translation by grid, demonstrating the use of grids that have 3 samples (dx, dy, dz) per grid point. It is the official method for French NTF to RGF93 transformation using [gr3df97a.tif](https://github.com/rouault/sample_proj_gtiff_grids/blob/master/gr3df97a.tif) grid (the difference with the classic NTv2 approach of ntf_r93.gsb is within one millimetre...), converted with [convert_gr3df97a.py](https://github.com/rouault/sample_proj_gtiff_grids/blob/master/convert_gr3df97a.py)


kbevers:
> Only the commits starting with "Build and CI: add libtiff dependency" are specific to this pull request.

I am a bit unsure of how much this PR covers. A bunch of commits has been made to the branch since this was opened so I is not as clear anymore. Is it *only* the commit 1b702a8 or is it everything between 1b702a8 and 8063c0b ?

rouault:
> Is it _only_ the commit [1b702a8](https://github.com/OSGeo/PROJ/commit/1b702a82f5e009745f5ca8bbed92579e33f598d4) or is it everything between [1b702a8](https://github.com/OSGeo/PROJ/commit/1b702a82f5e009745f5ca8bbed92579e33f598d4) and [8063c0b](https://github.com/OSGeo/PROJ/commit/8063c0b992af82154b3d9b637c78aa769a482f8a) ?

No, all commits starting from https://github.com/OSGeo/PROJ/commit/1b702a82f5e009745f5ca8bbed92579e33f598d4 down to https://github.com/OSGeo/PROJ/pull/1790/commits/f36d2374268608d276da45c96a7b4792ab8ffdb5

docs/source/operations/transformations/xyzgridshift.rst
```diff
@@ -0,0 +1,91 @@
...
++-----------------+-------------------------------------------------------------------+
+| **Output type** | Cartesian coordinates                                             |
++-----------------+-------------------------------------------------------------------+
+
+Perform a geocentric translation by bilinear interpolation of dx, dy, dz
+translation values from a grid. The grid is referenced against either the
+2D geographic CRS corresponding to the input (or sometimes output) CRS.
+
+This method is described (in French) in :cite:`NTF_88`
+and as EPSG operation methode code 9655 in :cite:`IOGP2018` (§2.4.4.1.1
```
kbevers, line 24: methode -> method

src/transformations/xyzgridshift.cpp
```diff
@@ -0,0 +1,329 @@
+/******************************************************************************
+ * Project:  PROJ
+ * Purpose:  Geocentric translation using a grid
```
kbevers, line 3: So, this one is quite close to +proj=deformation as far as I can tell. Is the main difference that the grids used here have adjustments in geocentric space rather than ENU-space as the deformation operation?

CMakeLists.txt
```diff
@@ -137,9 +137,14 @@ ENDIF()
 # Check for libtiff
 ################################################################################
 
-find_package(TIFF REQUIRED)
-if(NOT TIFF_FOUND)
-  message(SEND_ERROR "libtiff dependency not found!")
+option(DISABLE_TIFF_IS_STRONGLY_DISCOURAGED "Disable TIFF support (strongly discouraged !)" OFF)
```
kbevers, line 140: Is this the actual flag for disabling tiff support? So you would run cmake with `cmake -DDISABLE_TIFF_IS_STRONGLY_DISCOURAGED` or something like that?

kbevers (commented):
Generally this looks good to me. I have not read all the new code additions in great detail but my overall impression is good. I have put a few comments in-line which may lead to some discussion (mostly the xyzgridshift one).

test/gie/geotiff_grids.gie
```diff
@@ -0,0 +1,283 @@
+
+-------------------------------------------------------------------------------
+===============================================================================
+Test GeoTIFF grids
+===============================================================================
+
+<gie>
```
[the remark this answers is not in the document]
rouault: >     1. You generally do not set the tolerance - I take it you are happy with the default of 0.5 mm?

yes. Those files contain dummy values by gap of 1 meter: 1, 2, 3, 4, etc.

> 2. You only test the forward case,

yes, testing forward case only is sufficient. We want to test here the ability to read a value in a GeoTIFF grid. Inverse case if about the specifics of the hgridshift/vgridshift/etc method and is already tested in many other places. The roundrip 1 is for the specific testing of the new +proj=xyzgridshift which indeed deserves testing both directions

src/transformations/xyzgridshift.cpp
rouault: > Is the main difference that the grids used here have adjustments in geocentric space rather than ENU-space as the deformation operation?

Those are also conceptually different operations. deformation is about velocities and imply to specify a time difference. xyzgridshift is for static transformations.

CMakeLists.txt
rouault: > Is this the actual flag for disabling tiff support?

yes, this is intended to be cumbersome so that only people realize they probably do something they shouldn't. You need to specify -DDISABLE_TIFF_IS_STRONGLY_DISCOURAGED=ON
I'd be open to use a more normal ENABLE_TIFF=ON approach if you think we're going too far

docs/source/operations/transformations/xyzgridshift.rst
rouault: fixed

test/gie/geotiff_grids.gie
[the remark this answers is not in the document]
kbevers: Right, got it. You could mention at the top that the purpose of the gie tests in that file is to check that the grid values are read correctly. I think that would be helpful.

src/transformations/xyzgridshift.cpp
kbevers: > Those are also conceptually different operations. deformation is about velocities and imply to specify a time difference. xyzgridshift is for static transformations.

Right, yes, there's that of course :-) What I wanted to figure out, is if there's is any sense in sharing code between these two operations. My initial read through made me think of that but perhaps that is not ideal.

test/gie/geotiff_grids.gie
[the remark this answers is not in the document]
rouault: comment added

CMakeLists.txt
kbevers: I think it is fine to have TIFF enabled by default. I would go for a more neutral wording in the off-switch. "disable tiff is strongly discouraged" doesn't really sound like an on/off switch to me.  `DISABLE_TIFF` would be more suitable in my opinion. I would however encourage plastering the screen with warnings when that switch is turned on.

rouault: done

src/transformations/xyzgridshift.cpp
rouault: Some code sharing in 408cb1e

kbevers: Awesome, thanks!

rouault:
Merged through https://github.com/OSGeo/PROJ/pull/1813

PR #1810, opened 2019-12-25
projinfo: no longer call createBoundCRSToWGS84IfPossible() for WKT1:GDAL

rouault:
To align with GDAL 3.0.3 behaviour, no longer automatically try to create
a boundCRS to WGS84 when exporting to WKT1:GDAL. The user has to
explicitly specify --boundcrs-to-wgs84 if he wishes this behaviour.


kbevers:
Maybe also add a note to https://proj.org/usage/differences.html ?

rouault:
> Maybe also add a note to https://proj.org/usage/differences.html ?

done

Issue #866, opened 2018-03-17
New file access API


paxswill:
With proj_api.h being removed (#837) in the future, are there plans for an alternative file access API to replace `projFileAPI`, `pj_ctx_set_fileapi` and `pj_ctx_get_fileapi`? For my particular use case, I'm bundling the normal data files into a framework for iOS/macOS, and providing an alternative open file function that searches in the bundle resources for those files.

kbevers:
Nothing is decided yet but I would say that it is likely to happen, yes. It has been very difficult to determine if anyone was actually using that part of the code. Several enquiries has been made on the mailing list with no positive feedback, so it has not been a priority to build that into the new API so far. Since then a few cases of it being used has surfaced (including this) so at least there is now some evidence that it is used and not just a few thousand lines of unused code. 

It would be fairly easy to wrap the existing code into new functions in the `proj.h` API but that is not necessarily the best way to go. I know that @busstoptaktik has been missing a `fgetc` function in the current fileapi. It has not been possible to build that in without breaking the existing API as far as I understand. So that could be a place where we want to change things a little bit for the new API. If we want to include the fileapi in `proj.h` for version 6 where `proj_api.h` and `proj.h` will co-exists it might be necessary to break the existing API. Would that be acceptable to you?

Also, if you could design the fileapi form scratch, how would you go about doing it?

Finally, Is the code you are using it for open source? I'd like to see use of this in the wild if possible.

paxswill:
I have no problem with breaking the fileapi, my application is really just using the fopen function to search in some extra areas at the moment.

One thing I was toying with for later was adding a sort of “in-memory init file” using the fileapi, but I’m not sure at the moment how I’d redesign the API to make that easier. I’ll have to think a bit more about this.

I’ve really just wrote the framework this week, so not much external use yet, but [here](https://github.com/paxswill/SwiftProjection) it is, specifically [bundle-reader.c](https://github.com/paxswill/SwiftProjection/blob/master/Classes/bundle-reader.c)

kbevers:
I see you have written this comment below. So really exposing `pj_set_searchpath` and `pj_set_finder` the new API might be better for you?

```
/*
 In truth, using pj_set_searchpath would be easier, but it (and pj_set_finder,
 which would also be easier) has been deprecated and is due to be removed in a
 future release of PROJ. Instead we're doing some shenanigans with testing to
 see if a file with a given name exists in the Framework resources, and reading
[...]

PR #1813, opened 2019-12-26
[RFC4_dev] Add networking capabilities

rouault:
This is a continuation of #1790. Only the commits starting with "Rename PJ_CONTEXT::fileapi member as fileapi_legacy" are specific to this pull request.

Summary:
- Add a optional libcurl dependency (enabled on all CI targets, except the mingw32 one)
- Allow alternate network callbacks to be plugged in.
- Networking only enabled if allowed with PROJ_NETWORK=ON, proj_context_set_enable_network() or through the proj.ini configuration file
- Network access is done by chunks of multiples of 16 KB
- In-memory cache of downloaded chunks
- Default network endpoint is https://cdn.proj.org (defined in proj.ini). Can also be configured with PROJ_NETWORK_ENDPOINT or proj_context_set_url_endpoint()
- When network is enabled, grid filenames starting with http[s]:// are allowed
- When network is enabled, pj_create_crs_to_crs[_from_pj] will consider that grids in `grid_alternatives` can be used even if not available locally.  With lower level API, can be done with `proj_operation_factory_context_set_grid_availability_use(..., PROJ_GRID_AVAILABILITY_KNOWN_AVAILABLE)`
- When network is enabled, grids referenced by a short name will be looked up remotely if not available locally. e.g grids=egm08_25.gtx will be substituted by grids=https://cdn.proj.org/egm08_25.tif
- Adds a test/unit/test_network.cpp file

Not yet implemented: on-disk cache of downloaded chunks


docs/source/apps/cs2cs.rst
```diff
@@ -175,6 +175,14 @@ normally be in DMS format (use ``-f %.12f`` for decimal degrees with 12 decimal
 places), while projected (cartesian) coordinates will be in linear
 (meter, feet) units.
 
+Use of remote grids
```
kbevers, line 178: Perhaps this should also be added to `docs/source/apps/cct.rst`. I assume the same applies.

kbevers (approved):
I am not in a position to thoroughly review all of the technical bits in this PR but the overall structure looks good to me. The reworking of the filemanager looks like a big improvement. I wonder if it can easily be exposed via `proj.h`? If so, we would have a replacement for the old file accessing API and  #866 can be dealt with properly.

docs/source/apps/cs2cs.rst
rouault: done in https://github.com/OSGeo/PROJ/pull/1839/commits/4d21a242a10cf0e3c5b9503a9d0e8013ed4afd78

rouault:
>  I wonder if it can easily be exposed via `proj.h`

Done per https://github.com/OSGeo/PROJ/pull/1839/commits/55625528a17f7fe36d4100f23497ea77b95d39a6

kbevers:
> Done per 5562552

Thank you. The public functions needs documentation. I can do it if you want, I just need to know which branch to commit on.

rouault:
> The public functions needs documentation

As they already have Doxygen doc, this is addressed by https://github.com/OSGeo/PROJ/pull/1839/commits/c40ee9dfbbac32ebf3daa1b9edca64a247d4ad53 that just uses Breathe to include it in the appropriate sections

kbevers:
> As they already have Doxygen doc, this is addressed by c40ee9d that just uses Breathe to include it in the appropriate sections

Perfect!

PR #1817, opened 2019-12-31
[RFC4_dev] Add a SQLite3 local cache of downloaded chunks

rouault:
Final piece of RFC4 ! Only commits starting at "Add a SQLite3 local cache of downloaded chunks" are specific to this PR


kbevers:
This is too far away from my area of expertise to be able to give a decent review.

@hobu, I've requested a review from you. I hope you can find the time to work your way through this PR.

src/filemanager.cpp
```diff
@@ -0,0 +1,2301 @@
...
+}
+
+// ---------------------------------------------------------------------------
+
+DiskChunkCache::DiskChunkCache(PJ_CONTEXT *ctx, const std::string &path)
+    : ctx_(ctx), path_(path), vfs_(SQLite3VFS::create(true, false, false)) {
+    if (vfs_ == nullptr) {
+        return;
+    }
+    sqlite3_open_v2(path.c_str(), &hDB_,
```
hobu, line 404: It might be better to move all of this into an `initialize()` method instead of having it in the ctor. As it is now, if sqlite3_open_v2 fails to open the db, are we partially constructed? Is ever method of DiskChunkCache supposed to check if hDB_ is good?

src/filemanager.cpp
```diff
@@ -0,0 +1,2301 @@
...
+            pj_log(ctx_, PJ_LOG_ERROR, "Cannot take exclusive lock on %s",
+                   path.c_str());
+            sqlite3_close(hDB_);
+            hDB_ = nullptr;
+            return;
+        }
+        pj_log(ctx, PJ_LOG_TRACE, "Lock taken on cache. Waiting a bit...");
+        // Retry every 5 ms for 50 ms, then every 10 ms for 100 ms, then
+        // every 100 ms
+        proj_sleep_ms(i < 10 ? 5 : i < 20 ? 10 : 100);
```
hobu, line 433: Can this get in a situation where it never exits?

src/filemanager.cpp
```diff
@@ -0,0 +1,2301 @@
...
+        SQLITE_OK) {
+        pj_log(ctx_, PJ_LOG_ERROR, "%s", sqlite3_errmsg(hDB_));
+        return false;
+    }
+    return true;
+}
+
+// ---------------------------------------------------------------------------
+
+#define INVALIDATED_SQL_LITERAL "'invalidated'"
```
hobu, line 510: Why is this up here?

src/filemanager.cpp
```diff
@@ -0,0 +1,2301 @@
...
+
+    fprintf(stderr, "check ok\n");
+    return true;
+}
+
+// ---------------------------------------------------------------------------
+
+DiskChunkCache::~DiskChunkCache() {
+    if (hDB_) {
+        sqlite3_exec(hDB_, "COMMIT", nullptr, nullptr, nullptr);
```
hobu, line 664: what if this fails?

src/filemanager.cpp
```diff
@@ -0,0 +1,2301 @@
...
+    cache_.insert(Key(url, chunkIdx), dataPtr);
+
+    auto diskCache = DiskChunkCache::open(ctx);
+    if (!diskCache)
+        return;
+    auto hDB = diskCache->handle();
+
+    // Always insert DOWNLOAD_CHUNK_SIZE bytes to avoid fragmentation
+    std::vector<unsigned char> blob(*dataPtr);
+    assert(blob.size() <= DOWNLOAD_CHUNK_SIZE);
```
hobu, line 907: Should DOWNLOAD_CHUNK_SIZE be configurable?

src/filemanager.cpp
```diff
@@ -0,0 +1,2301 @@
...
+    delete reinterpret_cast<CurlFileHandle *>(handle);
+}
+
+// ---------------------------------------------------------------------------
+
+static size_t pj_curl_read_range(PJ_CONTEXT *ctx,
+                                 PROJ_NETWORK_HANDLE *raw_handle,
+                                 unsigned long long offset, size_t size_to_read,
+                                 void *buffer, size_t error_string_max_size,
+
```
hobu, line 2301: In my opinion, filemanager.cpp is too long and has too many classes in it. You're going to have to search hard to find specific things unless you know what you're looking for. I prefer things broken up a little bit more, but that's a preference thing.

src/grids.hpp
```diff
@@ -0,0 +1,240 @@
...
+NS_PROJ_START
+
+struct ExtentAndRes {
+    double westLon;  // in radian
+    double southLat; // in radian
+    double eastLon;  // in radian
+    double northLat; // in radian
+    double resLon;   // in radian
+    double resLat;   // in radian
+
```
hobu, line 46: Would this ever be agnostic to angle units? ie., add `bool IS_RADIANS` or some such?

src/iso19111/factory.cpp
```diff
@@ -45,6 +45,8 @@
 #include "proj/internal/lru_cache.hpp"
 #include "proj/internal/tracing.hpp"
 
+#include "sqlite3.hpp"
```
hobu, line 48: should this be a system include?

src/lib_proj.cmake
```diff
@@ -443,6 +443,16 @@ endif()
 include_directories(${SQLITE3_INCLUDE_DIR})
 target_link_libraries(${PROJ_CORE_TARGET} ${SQLITE3_LIBRARY})
 
+if(NOT DISABLE_TIFF_IS_STRONGLY_DISCOURAGED)
+  include_directories(${TIFF_INCLUDE_DIR})
+  target_link_libraries(${PROJ_CORE_TARGET} ${TIFF_LIBRARY})
+endif()
+
+if(CURL_FOUND)
+  include_directories(${CURL_INCLUDE_DIR})
```
hobu, line 452: this should be target_include_directories

src/open_lib.cpp
```diff
@@ -197,24 +200,39 @@ static const char *get_path_from_win32_projlib(const char *name, std::string& ou
...
-pj_open_lib_ex(projCtx ctx, const char *name, const char *mode,
-               char* out_full_filename, size_t out_full_filename_size) {
-    try {
-        std::string fname;
-        const char *sysname = nullptr;
-        PAFile fid = nullptr;
 #ifdef WIN32
-        static const char dir_chars[] = "/\\";
-        const char dirSeparator = ';';
+static const char dir_chars[] = "/\\";
```
hobu, line 207: C++17 has std::experimental::filesystem::path::preferred_separator. it might be worthwhile to prepare for using that.

src/sqlite3.hpp
```diff
@@ -0,0 +1,127 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *****************************************************************************/
+
+#ifndef SQLITE3_HPP_INCLUDED
+#define SQLITE3_HPP_INCLUDED
+
+#include <memory>
+
+#include <sqlite3.h>
```
hobu, line 33: In relation to comment above, it is confusing to call this "sqlite3.hpp". I don't have a better name though...

CMakeLists.txt
```diff
@@ -83,6 +83,9 @@ elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "Intel")
     set(PROJ_CXX_WARN_FLAGS -Wall)
   endif()
 endif()
+
+add_definitions(-DNOMINMAX)
```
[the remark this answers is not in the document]
rouault: Makes sense for the guard.

hum are you suggesting to use (std::min)(x,y) everywhere ? Looks ugly... I'm not sure about the impacts on downstream users. This is just used in .cpp files

CMakeLists.txt
```diff
@@ -133,6 +136,35 @@ IF("${SQLITE3_VERSION}" VERSION_LESS "3.11")
...
+    message(SEND_ERROR "libtiff dependency not found!")
+  endif()
+  add_definitions(-DTIFF_ENABLED)
+endif()
+
+################################################################################
+# Check for curl
+################################################################################
+
+option(ENABLE_CURL "Enable Curl support" ON)
```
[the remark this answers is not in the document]
rouault: You could technically use +grids=http://example.com/foo.gtx

data/proj.ini
```diff
@@ -0,0 +1,16 @@
...
+; Network capabilities disabled by default.
+; Can be overriden with the PROJ_NETWORK=ON environment variable.
+; network = on
+
+; Can be overriden with the PROJ_NETWORK_ENDPOINT environment variable.
+cdn_endpoint = https://cdn.proj.org
+
+cache_enabled = on
+
+cache_size_MB = 100
```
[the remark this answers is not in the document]
rouault: What would you suggest ?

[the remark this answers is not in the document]
hobu: Like 300 or something? What would be a number that would fetch and keep almost all of the grids most users would ever be using? 750 is all of the grids, right?

CMakeLists.txt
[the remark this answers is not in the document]
hobu: That, or use target_compile_definitions explicitly on targets so NOMINMAX doesn't leak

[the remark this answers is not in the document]
rouault: Moving that to lib_proj.cmake with
```
if (WIN32)
    target_compile_definitions((${PROJ_CORE_TARGET} PRIVATE -DNONMINMAX)
endif()
```

configure.ac
```diff
@@ -166,8 +168,8 @@ AC_SUBST(C_WFLAGS,$C_WFLAGS)
 AC_SUBST(CXX_WFLAGS,$CXX_WFLAGS)
 AC_SUBST(NO_ZERO_AS_NULL_POINTER_CONSTANT_FLAG,$NO_ZERO_AS_NULL_POINTER_CONSTANT_FLAG)
 
-CFLAGS="${CFLAGS} -fvisibility=hidden"
-CXXFLAGS="${CXXFLAGS} -fvisibility=hidden"
+CFLAGS="${CFLAGS} -fvisibility=hidden -DNOMINMAX"
```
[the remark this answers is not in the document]
rouault: Restricting those defines to the case of a cygwin/mingw compiler

cmake/ProjTest.cmake
```diff
@@ -26,8 +26,13 @@ function(proj_add_test_script_sh SH_NAME BIN_USE)
         COMMAND ${PROJECT_SOURCE_DIR}/test/cli/${SH_NAME}
         ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${${BIN_USE}}
       )
-    set_tests_properties( ${testname}
-        PROPERTIES ENVIRONMENT "PROJ_LIB=${PROJECT_BINARY_DIR}/data")
+      if(MSVC)
```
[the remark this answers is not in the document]
rouault: function(proj_test_set_properties TESTNAME) added to avoid copy&paste

src/sqlite3.hpp
kbevers: `sqlite_utils.hpp` or `db_utils.hpp`?

src/filemanager.cpp
```diff
@@ -0,0 +1,2301 @@
...
+#define STR_HELPER(x) #x
+#define STR(x) STR_HELPER(x)
+
+using namespace NS_PROJ::internal;
+
+NS_PROJ_START
+
+// ---------------------------------------------------------------------------
+
+static void proj_sleep_ms(int ms) {
```
kbevers, line 101: It's confusing that a static function has the `proj_` prefix. Is `sleep_ms()` taken already?

src/filemanager.cpp
kbevers: Yeah, it is a big one. Perhaps the bits that has to do with networking can be moved to a separate file? `networkfilemanager.cpp` perhaps?

data/Makefile.am
```diff
@@ -38,12 +38,54 @@ SQL_ORDERED_LIST = sql/begin.sql \
...
 
-EXTRA_DIST = GL27 nad.lst nad27 nad83 \
+EXTRA_DIST = proj.ini GL27 nad.lst nad27 nad83 \
 		world other.extra \
 		CH \
 		ITRF2000 ITRF2008 ITRF2014 \
 		projjson.schema.json \
-		CMakeLists.txt tests/test_nodata.gtx null \
+		CMakeLists.txt \
+		tests/test_nodata.gtx \
```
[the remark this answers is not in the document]
rouault: Done

data/proj.ini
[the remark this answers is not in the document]
rouault: The current total size is 486 MB, so 300 would keep most. Using that

docs/source/resource_files.rst
```diff
@@ -49,6 +54,30 @@ A proj installation includes a SQLite database of transformation information
 that must be accessible for the library to work properly.  The library will
 print an error if the database can't be found.
 
+proj.ini
```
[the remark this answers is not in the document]
rouault: done

docs/source/usage/environmentvars.rst
```diff
@@ -56,3 +56,21 @@ done by setting the variable with no content::
     Set the debug level of PROJ. The default debug level is zero, which results
     in no debug output when using PROJ. A number from 1-3, whit 3 being the most
     verbose setting.
+
+.. envvar:: PROJ_NETWORK
+
+    .. versionadded:: 7.0.0
+
+    If set to ON, enable the capability to use remote grids stored on CDN
```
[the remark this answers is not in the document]
rouault: https://github.com/OSGeo/PROJ/issues/1852 created about that

src/filemanager.cpp
rouault: Moved to initialize(). The constructor being private and used by a static open() method, there is no risk that methods will be used if initialize() fails as it will return a nullptr object

src/filemanager.cpp
rouault: no, the loop is taken at most 30 times, which is a bit more than 1 second. Comment added before the apparently infinite for() to make that clearer

src/filemanager.cpp
rouault: it is used just in the below method (and other ones). Adding a comment

src/filemanager.cpp
rouault: nothing too bad normally: the content of the transaction will be lost (that is a chunk downloaded from the network that would normally have been cached won't), but hopefully integrity will be maintained. I've added a pj_log() trace in the unlikely event that would happen

src/filemanager.cpp
rouault: runtime configurable ? I don't see this as a urgent need. This is pretty much a development setting. We might reconsider that later i'd say

src/grids.hpp
rouault: it is much more convenient to have the unit normalized to radians, because all PROJ transformation methods will ask in radians.

src/iso19111/factory.cpp
rouault: renamed to sqlite3_utils.hpp to avoid any confusion

src/lib_proj.cmake
rouault: done

src/open_lib.cpp
rouault: defering on that, especially since it is only experimental on C++17... :-)

src/sqlite3.hpp
rouault: yep sqlite3_utils.hpp now

src/filemanager.cpp
rouault: Renamed to sleep_ms()

src/filemanager.cpp
rouault: All network related stuff moved to networkfilemanager.cpp

rouault:
Thanks for your reviews. Most of them are addressed in https://github.com/rouault/PROJ/commit/a958fe726b2f9e3df1ef683d5fa13856b769a952 (in https://github.com/OSGeo/PROJ/pull/1839)

PR #1822, opened 2020-01-05
Make sure tests pass if extra grids are present

rouault:
Should fix the issue reported in https://lists.osgeo.org/pipermail/proj/2020-January/009188.html
Some extra north-american grids present in data/ can affect the results of
some tests, so create a data/for_tests/ subdirectory in which we copy only
select grids.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,6 +41,9 @@ build_script:
   - if "%platform%" == "x86" SET BUILD_LIBPROJ_SHARED=OFF
   - echo "%VS_FULL%"
 #
+  - cd %APPVEYOR_BUILD_FOLDER%\data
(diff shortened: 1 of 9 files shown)


kbevers:
Looks good to me. Should this be back-ported to 6.3? And is it worth mentioning in release notes? It may be useful knowledge for packagers that things is simpler from now on.

rouault:
> Should this be back-ported to 6.3?

Probably more an improvement than a bug fix

>  is it worth mentioning in release notes?

up to you :-)

kbevers:
> up to you :-)

Not very helpful dammit! :-) Let's put this under "slight improvements that are probably not worth mentioning".

PR #1831, opened 2020-01-07
projinfo: use No. abbreviation instead of UTF-8 character (fixes #1828)

### LINKED ISSUE #1828: DEGREE SIGN character in projinfo output is incorrectly displayed on MS Windows console
On MS Windows 7 64 bit localized in Italian, the output of **projinfo** (6.2.1) tool can contain the **DEGREE SIGN** character (`U+00B0 | ° | \xc2\xb0`) that is incorrectly displayed in the command prompt console, e.g. the OSGeo4W Shell:

image

The default console codepage is 850 | ibm850 | OEM Multilingual Latin 1; Western European (DOS)

Is it possible to properly encode/decode the DEGREE SIGN character in the string "Operation n°" or to  avoid the use of such character (e.g. using "Operation n.")?
I would also add a space at the end of the string to separate it from the number that follows.

The problem also occurs changing the codepage to 65001 | utf-8 | Unicode (UTF-8) when the font used in the shell is non TrueType (the default font is Raster type).

https://github.com/OSGeo/PROJ/blob/529eb50f5d7e817dd27674984050a9bf1a6dce55/src/apps/projinfo.cpp#L689-L692

rouault:


kbevers (approved):

kbevers:
This should be back-ported to the 6.3 branch. I've added it to the 6.3.1 milestone but it is also a fix in 7.0.0. Unfortunately GitHub doesn't allow multiple milestones to be assigned to one PR.

rouault:
Backported in 6.3 branch per 91e360027bad225b7914b1fe7bdecfa73124a39c

Issue #1814, opened 2019-12-28
Automate creation of release notes


kbevers:
Preparing release notes for a new release requires significant effort, probably around 70-80 percent of the time spend on a new release has to do with release notes. This should be possible to automate using the [GitHub API](https://developer.github.com/v4/) in combination with milestones and tags on pull requests.

Automating the release notes requires a bit more discipline concerning pull request with new features and bug fixes. Anything that is worth mentioning in the release notes should be added to a specific milestone (e.g. 7.0.0) and tagged as either a bug fix or new feature. Additionally, the title of the PR should be worded in such a way that it can be used as a bullet point in the release notes (e.g. "Added projection xyz" or "Fix flipped axis order in EPSG:nnnn"). As far as I can tell it is possible to request all of the above using the GitHub API.

The release notes automator should be able to update both `/NEWS` and `/docs/source/news.rst`. For this to work as smoothly as possible the NEWS file should be converted to MarkDown or reStructuredText. 

It would be nice to add a template for pull request that gives instructions on how to make a good title for the PR so that we don't have to rename all incoming PR's for better use in release notes.

PR #1832, opened 2020-01-09
added issue & pull request templates

snowman2:
I noticed #1814 and thought the pull request template might be a helpful intermediate solution as you incrementally update the NEWS & the API in each PR as you go so it is not a big change in the end. Not sure how well this method would work for backporting :man_shrugging:. Also added issue templates while I was there. Here is an idea for how the issue templates would look: https://github.com/pyproj4/pyproj/issues/new/choose.

This probably will need tweaking based on how y'all do things. Definitely don't have to use it, but it is here if you would like to.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,7 @@ labels: bug
 <!-- Please search existing issues to avoid creating duplicates. -->
 
 
-#### Code Sample, a copy-pastable example if possible
+#### Example of problem
 Demonstrate the problem you have found, either by using PROJ tools like `cs2cs`, `cct` or `projinfo` **or** add a code snippet that highlights the problem using the PROJ C/C++ API.
 A "Minimal, Complete and Verifiable Example" will make it much easier for maintainers to help you:
 http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,31 @@
+---
+name: Bug report
+about: Create a report to help us improve
+labels: bug
+---
+
+<!-- Please search existing issues to avoid creating duplicates. -->
+
+
+#### Code Sample, a copy-pastable example if possible
+
+A "Minimal, Complete and Verifiable Example" will make it much easier for maintainers to help you:
+http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports
+
+```c
+# Your code here
+
+```
+#### Problem description
+
+[this should explain **why** the current behavior is a problem and why the expected output is a better solution.]
+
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
...
+#### Problem description
+
+[this should explain **why** the current behavior is a problem and why the expected output is a better solution.]
+
+#### Expected Output
+
+
+#### Environment Information
+ - PROJ version (`proj`)
+ - Operation System Information (`python -c "import platform; print(platform.platform())"`)
```
kbevers: Python can't be expected to be present on Windows - better leave platform agnostic, people will know what to do.
```suggestion
 - Operation System Information
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
+---
+name: Bug report
+about: Create a report to help us improve
+labels: bug
+---
+
+<!-- Please search existing issues to avoid creating duplicates. -->
+
+
+#### Code Sample, a copy-pastable example if possible
```
kbevers:
```suggestion
#### Example of problem
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
...
+
```
kbevers:
```suggestion
Demonstrate the problem you have found, either by using PROJ tools like `cs2cs`, `cct` or `projinfo` **or** add a code snippet that highlights the problem using the PROJ C/C++ API.
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,7 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+labels: proposal
```
kbevers:
```suggestion
labels: feature request
```

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,9 @@
+---
+name: Question
```
kbevers: I think this should be left out - we prefer to keep discussions and questions at the mailing list

kbevers (changes requested):
Thanks, I've been meaning to add these for a while but haven't gotten around to do it.

Regarding updating the NEWS files, I think it is better if we don't ask contributors to edit those files. It would however be nice to request a clear and concise description of the new feature or bug fix, which can then be put into NEWS.

.github/ISSUE_TEMPLATE/question.md
snowman2: Makes sense.

snowman2: Removed.

kbevers (approved):
Thanks for the updates!

PR #1836, opened 2020-01-10
createOperations(): fix dealing with projected 3D CRS whose Z units != metre

rouault:


rouault:
Backported in 6.3 per 6a2ee6fbc8a9ce915df8da36df8b041ee6573e05

PR #1837, opened 2020-01-10
CompoundCRS::identify(): avoid exception when horiz/vertical part is a BoundCRS

rouault:
The exception only affects C++ users. It was caught by the C layer.


rouault:
Backported in 6.3 per 7d46771382e394586477b0edd7833a89369bb248

PR #1838, opened 2020-01-10
WKT1_GDAL export: limit datum name massaging to names matching EPSG (fixes #1835)

### LINKED ISSUE #1835: ESRI morphing of datum names even when not asking for ESRI dialect WKT
https://github.com/OSGeo/PROJ/blob/7de0f0931c43056782c73837503f06e5e4e24d7b/src/iso19111/datum.cpp#L1242

Is it correct to ask for "ESRI morphing" of datum names in the location identified above?  (Still looking for a concrete case of harm, discovered due to us defining our own coordinate systems via GDAL.)

rouault:
CC @ravenAtSafe @TheDorkKnight


rouault:
Backported in 6.3 per e18e4823ac91b1395cc0b3f30539340bc10ef38d

PR #1840, opened 2020-01-11
Fix missing word in cct.rst and cct.1

anbj:


kbevers:
Thanks for this. `cct.1` is generated at release time from `cct.rst` using Sphinx so technically it doesn't need to be changed. I'll just merge this as is, since it does no harm :-)

anbj:
Great! :)

PR #1843, opened 2020-01-13
unitconvert with mjd time format: avoid potential integer overflow

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20072


kbevers:
cherry-picked into 6.3 branch

PR #1848, opened 2020-01-15
Reintroduce backport bot, now using a dedicated PROJ bot

### LINKED ISSUE #1741: Find alternative to backport bot
The backporting bot stopped working a week or two ago. This is not ideal since it has been a valuable tool since we installed it. After recent changes the bot now works as a GitHub Action. According to https://github.com/tibdex/backport/issues/33 this has the unfortunate side effect that the bot can only backport PR's from brances on the main repository and *not from forks*. Judging by the comments of the developer of the bot this doesn't seem like something that can be fixed (limitations in GitHub functionality I believe). We need to find an alternative.

A bit of googling reveals a few possible options:

1. https://github.com/MeeseeksBox/MeeseeksDev (suggested by @QuLogic in https://github.com/OSGeo/PROJ/pull/1334)
2. https://github.com/electron/trop
3. https://github.com/chainer/backport

Option 1 seems rather complicated and can do more than just backport PRs. Options 2 and 3 are more specific in their use but unfortunately doesn't offer much help in terms of setup. Both include an install section but doesn't go in details with *where* to install. I assume in a CI script but I have not looked further into that yet.

kbevers:
I've set up a dedicated GitHub user, PROJ-BOT, that can functions as a backporting bot. I've copied the setup from QGIS which seems to finally have cracked the nut. Hopefully it works as the previous bot.

 - [x] Closes #1741 

Not applicable:
 - [ ] Tests added

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,20 @@
+name: Backport
+on:
+  pull_request:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    runs-on: ubuntu-18.04
+    name: Backport
+    steps:
+      - name: Backport Bot
+        if: contains( join(github.event.pull_request.labels.*.name), 'backport')
+        uses: Gaurav0/backport@v1.0.24
+        with:
+          bot_username: PROJ-BOT
+          bot_token: c1c49a5799d0a3a3885db8a4792880cbe509c1d6
+          bot_token_key: b01bdebf64595ed77207019258316df1b05ff6ac
+          github_token: ${{ secrets.GITHUB_TOKEN }}


3nids:
One remark, this bot doesn't work on rebased PRs (only squashed or merged).
I might be working on an updated version to fix this. And the token issue might be solved by using a cron job on master instead of running the workflow on the PR.
But nothing strictly planned yet.

kbevers:
Thanks for the heads up. We don't do rebase-merges that often so hopefully it will not be a problem. I'll keep an eye on what you are doing with the QGIS repository. I'm glad you've managed to get this working so far. I tried myself and got thoroughly stuck!

PR #1853, opened 2020-01-16
Explain why sweep axis affects geolocation

gerritholl:
In the note on the sweep angle, add a line explaining why the sweep angle ultimately affects the geolocation, as it is not immediately or trivially obvious to see why it should matter what the outer or inner gimbal is.  To this end, link to a post on Mathematics Stack Exchange that explains this very well.


kbevers:
Thanks for adding this!

PR #1855, opened 2020-01-17
[RFC4_dev] Merge master back to rfc4 latest branch

rouault:
rfc4_dev was branched off from master 1 month 1/2 ago. Time to resynchronize.... 

The last commit "Merge branch 'master' into rfc4_merge_back_master" is the one of interest, but I'm afraid that, due to how git handle merges, it's impossible to review. Most merge conflicts were expected to me due to what I had in mind from the diverging history & mostly mechanical to fix.

One thing however that was unexpected and found at runtime was related to error handling. In 6.3, I added a behaviour to proj_trans() (when used through proj_create_crs_to_crs() that would select several candidate transformations) so that when an error occurs when running a selected pipeline, we fallback to the next candidate. This was driven by corner cases with the NTv2_0.GSB canadian grid made of several subgrids that don't fully cover the declared bounding box in EPSG. So based on the bbox of the transformation, one might select it, but not be able to use it for some points (in the US for example), hence the fallback that will try with US grids.

However, when network functionality is enabled, we need to be more careful if accessing a remote file fails. This is a case where we want to just error out without trying the next candidate. There was a number of places in the grid loading, and pipeline execution that reset the error flag. I've thus had to add at a few strategic points in grids.cpp special handling for a new PJD_ERR_NETWORK_ERROR errno, so that the error is preserved, and that proj_trans() exits without attempting using the next candidate transformation. I've also modified pipeline.cpp so that when a step returns HUGE_VAL for .x, we early exit instead of running the next step that would reset the error state due to  `int last_errno = proj_errno_reset(P)` being used in fwd.cpp & inv.cpp, for a unclear reason to me...

All of this was found by a unit test of test_network.cpp that simulates a network error.

Error handling in PROJ looks really super fragile & hard to comprehend to me. The concept of persistent vs transient errors, when the error state  is supposed to be reset or not.. Not sure I've a good idea how to make it simpler...


kbevers:
> but I'm afraid that, due to how git handle merges, it's impossible to review.

Luckily we've had the chance to review the individual parts in their own pull requests so I think we are fine here.

> Error handling in PROJ looks really super fragile & hard to comprehend to me. The concept of persistent vs transient errors, when the error state is supposed to be reset or not.. Not sure I've a good idea how to make it simpler...

I agree. This is something that would be nice to get sorted once and for all. I haven't got good ideas on top of my head. This could be a nice topic for a workshop at a code-sprint.

PR #1861, opened 2020-01-21
Document need for quotes when a pipeline grid path contains spaces

ravenAtSafe:
Not certain whether this is appropriate when you are dealing with lists of grids, but it definitely helps for a single grid.

### DIFF
--- a/docs/source/operations/transformations/deformation.rst
+++ b/docs/source/operations/transformations/deformation.rst
@@ -93,6 +93,9 @@ Parameters
     grid is considered optional and PROJ will the not complain if the grid is
     not available.
 
+    If a grid is specified with a path that contains spaces, the value should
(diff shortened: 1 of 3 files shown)


kbevers:
Thanks, this is good to clarify. I think it just one of those things that is second nature to some of us.
I think for a list of grids you would need to put each individual path in its own set of quotes.

Do you mind adding the same clarification to `vgridshift` and `deformation`?

ravenAtSafe:
vgridshift and deformation changes made.

Yup, the quoting solution was my first guess as well, but I am highly motivated to reduce the number of questions I have to field about the use of PROJ pipelines in FME!

kbevers:
@PROJ-BOT is not doing it's backporting job properly so I've manually cherrypicked the changes into master.

Thanks @ravenAtSafe !

PR #1864, opened 2020-01-22
appveyor.yml: tune PATH order to fix error when running 'proj' on x64 builds

rouault:


rouault:
I'm not sure what has changed on appveyor side. And the issue didn't affect all branches (or forks). I suspect there might be some caching effects

snowman2:
Yeah - it is strange that the order of PATH mattered.

PR #1868, opened 2020-01-22
Add RFC5 text: Adopt GeoTIFF-based grids for grids delivered with PROJ

rouault:
RST rendering at:
https://github.com/rouault/PROJ/blob/rfc5_text/docs/source/community/rfc/rfc-5.rst


ccrook:
Looks excellent! 

A couple of questions on naming:
* how do you see auto-promotion of names working if the naming convention agency_filename.tif is used?  Where will the agency come from?
* do you see a need for unique agency names, and who will look after that.  I'm hoping these will be abbreviated names, but I can imagine a lot of takers for "nga", for example.

Also could there be some mechanism for handling grid_alternatives through CDN or the like instead of (or optional alternative to) grid_alternatives in proj repository?

rouault:
>     * how do you see auto-promotion of names working if the naming convention agency_filename.tif is used?  Where will the agency come from?

The agency name is the one you can see from the directory names at https://github.com/OSGeo/proj-datumgrid-geotiff/. 
Regarding auto-promotion of names, the datbase will store both "ntf_r93.gsb" and "fr_ign_ntf_r93.tif" (if that was what you meant)

 > do you see a need for unique agency names, and who will look after that. I'm hoping these will be abbreviated names, but I can imagine a lot of takers for "nga", for example.

The general rule for the agency name is {two_letter_country_code_of_agency_nationality}_{some_abbreviation}. That should hopefully lead to unambiguous name.  An exception to this, but following the scheme, is eur_nkg, for the Nordic Geodetic Commission which isn't affiliate to a single country but to some european countries.

I'll add precision in the RFC text regarding this

> Also could there be some mechanism for handling grid_alternatives through CDN or the like instead of (or optional alternative to) grid_alternatives in proj repository?

Not part of my plan. The content of the grid_alternatives table must be kept consistent with the one of the EPSG database that PROJ embeds.

kbevers:
> A couple of questions on naming:

Another from someone employed by an agency that has had three different names in the span of 10 years: 

* What is the strategy when agencies change their name? 

Ignore it? Add another alias? Rename all files? Only use the new name for new incoming files?
Sooner or later we will face this and may as well think about it ahead of time.

rouault:
>     * What is the strategy when agencies change their name?

I would probably stick with the old name, like EPSG being defunct as an organization (it is IOGP since more 10 years), but still used because of the long history. That way this will discourage them from changing name because PROJ is a foundation (joking ;-)). 
If they are really keen in having PROJ to change name, they will have to put the effort in doing the needed renaming in PROJ database and datumgrid repo.

ccrook:
> The agency name is the one you can see from the directory names at https://github.com/OSGeo/proj-datumgrid-geotiff/.

That's clear - thanks.  I really like the structure of this - is there any reason to prefer underscore to slash separate agency from filename?  It seems much clearer to me looking at the grids in this structure.  Is this an EPSG limitation on not including directory names, or something like that?

>> Also could there be some mechanism for handling grid_alternatives through CDN or the like instead of (or optional alternative to) grid_alternatives in proj repository?

> Not part of my plan. The content of the grid_alternatives table must be kept consistent with the one of the EPSG database that PROJ embeds.

Related question: is there any prospect of a more dynamic connection to EPSG, or equivalent.  A LINZ ambition is to be able to update our deformation model after an earthquake with minimal latency.  With INSAR etc it may be feasible to have a new interim model available within a few months.  The challenge will then be to propogate it out to users.

rouault:
>  is there any reason to prefer underscore to slash separate agency from filename?

That would be a non-flat directory structure, and that would cause implementation problems / complications. For example if the proj_download_file function is called on "http://example.com/foo/bar/baz.tif", would the fileto put in the file system be baz.tif or bar/baz.tif or foo/bar/baz.tif etc ? Deciding for a flat structure simplifies things a lot. See https://lists.osgeo.org/pipermail/proj/2020-January/009274.html + https://lists.osgeo.org/pipermail/proj/2020-January/009276.html

> Related question: is there any prospect of a more dynamic connection to EPSG, or equivalent

That's hard. As you can see in https://github.com/OSGeo/PROJ/pull/1869, almost any update of EPSG (here it is a micro update from 9.8.4 to 9.8.6) can cause new methods to appear, to be renamed, reveal algorithmic limitation of PROJ to fix, etc. All those reasons require code changes. Or new grids to supersede older ones, but the new grids are not yet available for PROJ.

PR #1869, opened 2020-01-22
Database: update to EPSG v9.8.6

### LINKED ISSUE #1867: Update to EPSG 9.8.6

rouault:
Fixes #1867


rouault:
@ccrook This update has the new .gtx file names for the LINZ transformations. I've done a few needed changes to please the regression, but I'll let you check more deeply.

@Jochem-L  This update brings the rdtrans2018.gsb and nlgeo2018.gtx grids. I let you submit them to proj-datumgrid (or perhaps wait for proj-datumgrid-geotiff / RFC5 to be adopted & implemented first) and submit the additions to grid_alternatives table

kbevers:
@rouault can this be backported to 6.3 or has the database schema diverged between the two branches?

rouault:
> @rouault can this be backported to 6.3 or has the database schema diverged between the two branches?

This could technically be backported, but it has impacts. For example the EPSG:7415 to EPSG:4937, "Amersfoort / RD New + NAP height" to ETRS89 now assumes that the 2018 grids are available, since the 2008 ones are marked superseded (and even when asking PROJ to use superseded results, the results are not the same than with v9.8.4 due to some optimization/deficiency of createOperations() when dealing with supersession+horizontal+vertical grids ). Admitedly the impact for that case is minimum since we have never shipped the 2008 grids. Looking at the changes (and hoping I got the NZ grids OK), this is probably the only potentially disruptive change due to that update

ccrook:
I've looked at the NZ grids etc and it all looks good to me.  I'm not sure how the various SQL tables interact, but I'm assuming the the supersession table will ensure that PROJ tries to use the correct version of the vertical transformations and doesn't end up trying to use a .csv file and failing. Also not sure why the grid_alternatives is used where the filename is unchanged - is this to associate with the proj-datumgrid-xxxx package/directory?  Thanks again for sorting NZ grids.

rouault:
> I'm assuming the the supersession table will ensure that PROJ tries to use the correct version of the vertical transformations and doesn't end up trying to use a .csv file and failing

Yes with supersession enabled, the .gtx based transformations will be presented first. And even if the old .csv based transformations were used, that should still work, given that we have mappings for both the .csv and .gtx

> Also not sure why the grid_alternatives is used where the filename is unchanged - is this to associate with the proj-datumgrid-xxxx package/directory?

Yes, this was the primary reason. With RFC-5, this will also point to the URL of the file in the CDN. But it is also used by PROJ as a hint that this grid is known of PROJ (and users can get this information through the API), which is a criterion used by the sorting algorithm of https://proj.org/operations/operations_computation.html#filtering-and-sorting-of-coordinate-operations when sorting candidate operations.

kbevers:
> This could technically be backported, but it has impacts. 

It sounds like it is low impact. I think we should backport and bring a bit more life to the 6.3 branch. Is that a simple backport (cherry-pick the commit into 6.3) or is it necessary to run various database-related scripts again? (not sure how much master has diverged in this regard)

rouault:
@kbevers I've pushed an extra commit to remove a hack with how PROJ dealt with the EPSG 'Geographic3D to GravityRelatedHeight' methods. Those methods have the annoying particularity that to go from the source (Geog3D CRS) to the target (height), you must subtract the value of the grid, contrary to all other EPSG methods. The solution I had until now was on grid name substitution to use a custom PROJ 'GravityRelatedHeight to Geographic3D' but this turned to be fragile & confusing for users, since you had to indicate a inverse_direction = 1 value in grid_alternatives. 00dd98457af1f32e7bb2b9677f6319f72fcfacf9 should hopefully be more straightforward & robust.

master and 6.3 are pretty close regarding database management. I'll cherry-pick those commits

kbevers:
> I've pushed an extra commit to remove a hack with how PROJ dealt with the EPSG 'Geographic3D to GravityRelatedHeight' methods

Nice. The sign on the grid adjustments can be confusing - good that it is now simpler to understand!

>  I'll cherry-pick those commits

Awesome

sebastic:
> @Jochem-L This update brings the rdtrans2018.gsb and nlgeo2018.gtx grids. I let you submit them to proj-datumgrid (or perhaps wait for proj-datumgrid-geotiff / RFC5 to be adopted & implemented first) and submit the additions to grid_alternatives table

There was a user meeting about the RDNAPTRANS 2018 grids today where the use of the new grids in EPSG::9282 among others was discussed.

I had promised some time ago that I would prepare a PR for proj-datumgrid with the 2018 grids when those were used in the EPSG dataset, I'm doing that now.

PR #1870, opened 2020-01-23
CMake: replace PROJ_TESTS with CTest's BUILD_TESTING option

mwtoews:
As mentioned in #1263 this PR replaces `PROJ_TESTS` with CTest's `BUILD_TESTING` option, which is more generic to CMake projects. CTest (when included) automatically creates a `BUILD_TESTING` option, with a default ON. Backwards compatibility to the older option `PROJ_TESTS` is supported, but shows a deprecation message.

For example, a speedy build that does not require tests, configure a build like this:
```
$ mkdir build && cd build
$ cmake -DBUILD_TESTING=OFF ..
...
$ make -j
...
$ ctest
Test project /path/to/PROJ/build
No tests were found!!!
```
(as expected)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,13 +212,6 @@ if(MSVC OR CMAKE_CONFIGURATION_TYPES)
   set(CMAKE_DEBUG_POSTFIX _d)
 endif()
 
-option(PROJ_TESTS "Enable build of collection of PROJ tests" ON)
-boost_report_value(PROJ_TESTS)
-if(PROJ_TESTS)
-  enable_testing()
-endif()
-include(ProjTest)
-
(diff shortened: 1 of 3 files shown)


kbevers:
Looks good @mwtoews !

It would be nice to mention in the docs that you can turn testing off, in the CMake section in https://github.com/OSGeo/PROJ/blob/master/docs/source/install.rst 

I take it that we will want to remove the `PROJ_TESTS` option at some point in the future. It would  be good to keep track of things like that somewhere. Perhaps issues set to a 8.0.0 milestone?

mwtoews:
I've updated `install.rst`, and put a comment to indicate that the old option will be remove by PROJ 8.0, and I'll add a task for that.

PR #1874, opened 2020-01-23
[6.3 backport] Database: update to EPSG v9.8.6

rouault:
In case we would need to issue a 6.3.x maintainance release


kbevers:
> In case we would need to issue a 6.3.x maintainance release

I'll merge this and your other 6.3 backport but I don't think I'll release 6.3.1 alongside 7.0.0 now that proj_api.h is staying for a while longer.

PR #1880, opened 2020-01-24
createObjectsFromName(): use alias as fallback only

rouault:
Fixes a regression of 6.3.0 found when creating with GDAL a TIFF with
a Geographic3D CRS. As TIFF must also encode the Geographic2D CRS,
the code of the Geographic2D CRS is searched from the name of the
Geographic3D CRS. When doing createObjectsFromName(
"ETRS89", {AuthorityFactory::ObjectType::GEOGRAPHIC_2D_CRS}, false, 1),
the result returned was not EPSG:4258 as expected, but EPSG:4173 IRENET95
which is registered as an alias of ETRS89. So sort results such
that non-alias results are returned first.

### DIFF
--- a/src/iso19111/factory.cpp
+++ b/src/iso19111/factory.cpp
@@ -5214,9 +5214,9 @@ AuthorityFactory::createObjectsFromName(
     }
 
     std::string sql(
-        "SELECT table_name, auth_name, code, name, deprecated FROM ("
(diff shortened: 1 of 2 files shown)


rouault:
Update: this is actually a master issue only. 6.3.0 doesn't use the alias table

PR #1887, opened 2020-01-26
add #proj IRC channel notifications

jmckenna:
- #proj channel exists on IRC freenode (https://webchat.freenode.net/#proj)
- this change adds TRAVIS.CI notifications into the #proj IRC channel
  - (in addition to the notifications into the #gdal channel)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,4 +92,5 @@ notifications:
   irc:
     channels:
       - "irc.freenode.org#gdal"
+      - "irc.freenode.org#proj"
     use_notice: true


jmckenna:
I've added @rouault as founder of #proj channel on freenode (so Markus Neteler, Even, myself have 'op' status there).  Easy now to add others etc.

PS. I like this channel mostly because I can follow PROJ commit notifications, whereas #gdal has notifications from both the PROJ and GDAL projects coming in at once (which I find confusing, but I realize that is only my brain thinking that way).

Sorry for this old noise, but I think it's important to share that info here.

jmckenna:
Writing that here I realized I forgot to add @hobu .  Added.  So there are now 4 with op status in #proj 

The channel has existed since 2006-04-27

PR #1892, opened 2020-01-27
projinfo: add --searchpaths switch

rouault:


hobu:
I think it would be helpful if this also stated whether or not curl and CDN are available for grids.

rouault:
What would be the desired output and switch name ?

hobu:
> What would be the desired output and switch name 

🤷‍♂ What about ``--remote-data`` or ``--cdn``

rouault:
Something like this ?
```
$ projinfo --cdn
Network access unavailable: build without Curl support
```
```
$ projinfo --cdn
Network access disabled: not enabled in proj.ini or PROJ_NETWORK=ON not specified
```
```
$ projinfo --cdn
https://cdn.proj.org
```

rouault:
Or mabye
```
$ projinfo --cdn
Status: disabled
Reason: build without Curl support
```

```
$ projinfo --cdn
Status: disabled
Reason: not enabled in proj.ini or PROJ_NETWORK=ON not specified
```

```
$ projinfo --cdn
Status: enabled
URL: https://cdn.proj.org
```

kbevers:
> 🤷‍♂ What about --remote-data or --cdn

I think ``--remote-data`` is better. I don't think that CDN is a term most people are familiar with. RFC4 was the first time I was made aware it. "Remote data" is more intuitively understood, I think.

PR #1897, opened 2020-01-29
Add EPSG records for 'Geocentric translation by Grid Interpolation (IGN)' (gr3df97a.txt) and map them to new +proj=xyzgridshift

rouault:


QuLogic:
@kbevers just a note that this PR also got renamed like the other one.

kbevers:
@QuLogic thanks, I've reverted the title here as well

PR #1898, opened 2020-01-29
Remove 'null' grid file as it is now a special hardcoded case in grid code

rouault:


QuLogic:
>  [kbevers](/kbevers) changed the title ~Remove 'null' grid file as it is now a special hardcoded case in grid code~ Merge pull request #1898 from rouault/delete_null [11 hours ago](#event-2992706548)

Looks like you also hit https://github.com/sindresorhus/refined-github/issues/2713

kbevers:
> Looks like you also hit sindresorhus/refined-github#2713

Indeed it does. I've removed that chrome extension now, that behaviour is not helpful at all. Reverting the title.

PR #1903, opened 2020-02-03
Add projsync utility

### LINKED ISSUE #1750: Add a projsync utility
Idea mentioned at bootoom of https://lists.osgeo.org/pipermail/proj/2019-November/009032.html

rouault:
Fixes #1750


hobu:
Why do we have `projinfo` and then `proj_download`? What discriminates the underscore separation? GDAL's behavior on this topic is so arbitrary that it requires lookup just about every time you use the tools. Let's not repeat that with PROJ.

hobu:
I wonder if this should be called some variant of `sync`. It isn't a generic download tool. We could expand it to also include fetching latest DBs too, which would provide a lot of convenience.

kbevers (commented):
Looks good to me. This is a nice utility to add the package. 

I wonder about the tests of the application, how resilient are they to changes in PROJ-data/the CDN? Do we need to update the tests everytime a change is made on the CDN or will this still work unchanged?

rouault:
@hobu I was indeed midly satisfied with the name. projdownload seemed a bit stodgy to me, but projsync would be good. I'll go for it, if you and @kbevers is happy with that

Regarding synchronizing proj.db, this is a hard topic as I discussed a number of times. PROJ is highly sensitive to the content of proj.db. So if we were to do it, we would likely to have proj_7.0.0.db files compatible of PROJ 7.0.0 binary.

rouault:
> how resilient are they to changes in PROJ-data/the CDN?

They should be pretty resilient, unless we remove some files / source agencies.

hobu:
> So if we were to do it, we would likely to have proj_7.0.0.db files compatible of PROJ 7.0.0 binary.

My thought was to update if a new version of EPSG came out. Maybe it isn't worth the trouble.

kbevers:
> I'll go for it, if you and @kbevers is happy with that

Sounds good to me.

> My thought was to update if a new version of EPSG came out. Maybe it isn't worth the trouble.

It seems that EPSG doesn't really follow semantic versioning, so an increase in the patch-number can add a new operation or some other thing that requires new code. #1869 is a good example of that. It would be great if we in the future could decouple the EPSG-registry from specific PROJ releases and update the database when new updates appear in the EPSG registry but we're not there yet.

docs/source/apps/projsync.rst
```diff
@@ -0,0 +1,158 @@
...
+projsync
+================================================================================
+
+.. Index:: projsync
+
+.. only:: html
+
+    .. versionadded:: 7.0.0
+
+    Tool for downloading resource files
```
hobu: Tool for synchronizing PROJ datum and transformation support data.

docs/source/apps/projsync.rst
```diff
@@ -0,0 +1,158 @@
...
+    Restrict resource files to be downloaded to those whose name property
+    contains the NAME value. Specifying ``?`` as NAME will list all possible values.
+
+.. option:: --all
+
+    Ask to download all files.
+
+.. option:: --exclude-world-coverage
+
+    Exclude files which have world coverage.
```
hobu, line 105: What box defines world coverage?

docs/source/apps/projsync.rst
```diff
@@ -0,0 +1,158 @@
...
+    See also
+    ********
+
+    **cs2cs(1)**, **cct(1)**, **geod(1)**, **gie(1)**, **proj(1)**, **projinfo(1)**
+
+    Bugs
+    ****
+
+    A list of know bugs can be found at https://github.com/OSGeo/PROJ/issues
+    where new bug reports can be submitted to.
```
hobu, line 153: It would be helpful but maybe confusing to also link to the data repo

src/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ option(BUILD_PROJ
   "Build proj (cartographic projection tool)" ON)
 option(BUILD_PROJINFO
   "Build projinfo (SRS and coordinate operation metadata/query tool)" ON)
+option(BUILD_PROJSYNC
+  "Build projsync (Data downloader tool)" ON)
```
hobu: Build projsync (synchronize transformation support data)

src/apps/projsync.cpp
```diff
@@ -0,0 +1,556 @@
...
+                                grid_bbox[1] = std::min(grid_bbox[1], bbox[1]);
+                                grid_bbox[2] = std::max(grid_bbox[2], bbox[2]);
+                                grid_bbox[3] = std::max(grid_bbox[3], bbox[3]);
+                            }
+                        }
+                    } else {
+                        grid_bbox = get_bbox(j_coordinates);
+                    }
+                    if (grid_bbox.size() != 4) {
+                        break;
```
hobu, line 440: Are there boxes with Zs too?

src/apps/projsync.cpp
```diff
@@ -0,0 +1,556 @@
...
+                        grid_e = std::numeric_limits<double>::max();
+                    } else if (grid_e > 180 && queried_west < -180) {
+                        grid_w -= 360;
+                        grid_e -= 360;
+                    }
+                    if (queryAll) {
+                        matchBbox = true;
+                        break;
+                    }
+                    // std::cout << name << " " << grid_w  << " " << grid_s << "
```
hobu: cruft

src/bin_projsync.cmake
```diff
@@ -0,0 +1,16 @@
...
+add_executable(bin_projsync ${PROJSYNC_SRC})
+set_target_properties(bin_projsync
+  PROPERTIES
+  OUTPUT_NAME projsync)
+target_link_libraries(bin_projsync ${PROJ_LIBRARIES})
+target_compile_options(bin_projsync PRIVATE ${PROJ_CXX_WARN_FLAGS})
+install(TARGETS bin_projsync
+  RUNTIME DESTINATION ${BINDIR})
+
+if(MSVC AND BUILD_LIBPROJ_SHARED)
```
hobu, line 14: Use the standard BUILD_SHARED_LIBS cmake variable instead

src/filemanager.cpp
```diff
@@ -1178,6 +1178,8 @@ static void CreateDirectoryRecursively(PJ_CONTEXT *ctx,
 
 std::string pj_context_get_user_writable_directory(PJ_CONTEXT *ctx,
                                                    bool create) {
+    if (!ctx)
```
hobu, line 1181: Was this missing before?

src/filemanager.hpp
```diff
@@ -53,8 +53,8 @@ class FileManager {
 
   public:
     // "Low-level" interface.
-    static std::unique_ptr<File> open(PJ_CONTEXT *ctx, const char *filename,
-                                      FileAccess access);
+    static PROJ_DLL std::unique_ptr<File>
```
hobu, line 56: thread synchronization concerns with this?

hobu (approved):

docs/source/apps/projsync.rst
rouault: done

docs/source/apps/projsync.rst
rouault: "grid_e - grid_w > 359 && grid_n - grid_s > 179", to exclude EGM files

docs/source/apps/projsync.rst
rouault: Adding "Bugs specific to resource files should be submitted to
    https://github.com/OSGeo/PROJ-data/issues"

src/CMakeLists.txt
rouault: done

src/apps/projsync.cpp
rouault: no. sanity check only

src/apps/projsync.cpp
rouault: removed

src/bin_projsync.cmake
rouault: All other similar files, for better or worse (likely the later), use BUILD_LIBPROJ_SHARED, so I'll keep that for consistency.

src/filemanager.cpp
rouault: likely

src/filemanager.hpp
rouault: I don't see any thread related issue here

PR #1905, opened 2020-02-04
tests: force use of bash for proj_add_test_script_sh

russkel:
I am trying to run the tests on an Alpine Linux system that uses ash shell by default. This causes all `proj_add_test_script_sh` tests to error with `system error -8 BAD COMMAND`. Installing bash is not a problem, but this commit is required to force it to use bash.

### DIFF
--- a/cmake/ProjTest.cmake
+++ b/cmake/ProjTest.cmake
@@ -23,7 +23,7 @@ function(proj_add_test_script_sh SH_NAME BIN_USE)
     if(${TEST_OK})
       add_test(NAME "${testname}"
         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/data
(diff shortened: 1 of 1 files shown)


mwtoews (approved):

kbevers:
Do the test scripts contain bash specifics or can we get by using `sh` instead? It should be available on all POSIX systems and it would be nice not to have a soft dependency on bash.

russkel:
I am not too familiar with shell so I can't really comment. I think the `if` statements would have to be changed. I do know what you're looking at removing is called 'bashisms'.

kbevers:
It is probably safest to go with bash for now. Thanks for fixing this problem.

russkel:
Pleasure, thanks for merging.

Issue #1484, opened 2019-05-24
OSX: brew doesn't install cmake scripts


Nitaym:
Installing proj4 using `brew install proj` on OSX doesn't install the cmake scripts

under /usr/local/Cellar/proj/6.1.0/lib there's no cmake directory.
On the cmake script, using `find_package(PROJ4 REQUIRED)` results in the following message:

```
By not providing "FindPROJ4.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "PROJ4", but
  CMake did not find one.

  Could not find a package configuration file provided by "PROJ4" with any of
  the following names:

    PROJ4Config.cmake
    proj4-config.cmake

  Add the installation prefix of "PROJ4" to CMAKE_PREFIX_PATH or set
  "PROJ4_DIR" to a directory containing one of the above files.  If "PROJ4"
  provides a separate development package or SDK, be sure it has been
  installed.

```

mwtoews:
This occurred to me the other day too, and it seems that `cmake/ProjConfig.cmake` should be moved back to `cmake/Proj4Config.cmake`.

Nitaym:
This works okay if you checkout the git repo and build with cmake. The brew part is the one that has the problem. Isn't using brew the same as doing `make install`? Not sure why this is hapenning

kbevers:
Not sure if this is a problem with PROJ or a brew. I haven't seen the above error when I build on OS X. @mwtoews is there any downside to changing the filename back to `Proj4Config.cmake`?

mwtoews:
@kbevers investigating more, it seems the file should be moved to either `PROJ4Config.cmake` or `proj4-config.cmake`, but not `Proj4Config.cmake` (i.e. this wouldn't have worked before anyways).

The CMake docs for this are for the [`find_package` command](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package), which "searches for a file called `<PackageName>Config.cmake` or `<lower-case-package-name>-config.cmake`".

There is probably more work to be done to make this work correctly, as here's what I can do:
```
$ mkdir /tmp/xyz && cd /tmp/xyz
$ cat > CMakeLists.txt
cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
project(xyz LANGUAGES C CXX)
find_package(PROJ4)
$ mkdir build && cd build
$ cmake -DCMAKE_PREFIX_PATH=/path/to/PROJ/cmake/ ..
[...]

PR #1910, opened 2020-02-04
Make PROJ the CMake project name addressing first stage of #1885

### LINKED ISSUE #1885: Make PROJ the CMake project name
Now that PROJ is well established as the name of this project, it's time
to retire the use of PROJ4 as the CMake package name.

I propose that, prior to the release of version 7, CMake be switched to
using PROJ as the project name.  We'll need to make provision to
minimize the impact on end users.  So

  * Need to support find_package(PROJ4) (for 2 years?).
  * Continue to use offer PROJ4::proj (in an addition to PROJ::proj) as
    a target name (for 2 years?).

When the 2 years is nearly up, find_package(PROJ4) could issue a WARNING
about its obsolescence.

I'm less clear whether this is necessary

  * PROJ_LIBRARIES points to PROJ4::proj (for 1 year?) before switching
    it to PROJ::proj.

This would only be necessary to support

  * library foo is built against PROJ version 7
  * program bar links to foo but bar's find_package(PROJ4) finds a pre-7
    version of PROJ

However, this wouldn't normally be expected to work anyway.

The necessary changes are fairly straightforward and I can do them in a
week or so.  I'd like to get some feedback on this before I proceed.

cffk:
Allow both find_package(PROJ) and find_package(PROJ4).  More details
are in cmake/CMakeLists.txt.

 - [x] Closes #1885 (first stage only)


hobu (approved):

rouault:
Slightly out of topic thoughts:
- Do we have testing of find_package(PROJ) in the CI ? Would be cool if there was some dummy cmake project that would try to find PROJ with the proj-config.cmake file we generate.
- By the way is it expected that this .cmake file is only generated with PROJ cmake builds ? That means that, in distributions building PROJ with autoconf, cmake projects cannot import PROJ.

mwtoews:
> * Do we have testing of find_package(PROJ) in the CI ?

I've been thinking the same. It would be easiest to test in the CI script after install, as I don't think it's possible to do this with CTest.

> * By the way is it expected that this .cmake file is only generated with PROJ cmake builds ? That means that, in distributions building PROJ with autoconf, cmake projects cannot import PROJ.

Having just looking at this in #1484 it seems that only CMake builds can use find_package, and that Debian and Homebrew distributions don't bundle these files as they are Autotools-built.

cffk:
Yes, some basic testing of find_package(PROJ) (+ PROJ4) should be tested in the CI.  A more complete job would be testing the transitive inheritance of the dependency on PROJ.

I don't see how a decent job can be done providing config-style find_package support with autoconf without a lot of messy coding.  I have done this with our builds of boost (before it got builtin support).  But my solution was rather fragile and needed nursing along from version to version.

Conceivably someone has solved this problem already and there's a template to follow?  I wish autoconf would fade into non-existence.  (CMake certainly has its warts.  But at least I can get most jobs done without a whole lot of fuss.)

> > * Do we have testing of find_package(PROJ) in the CI ?
> 
> I've been thinking the same. It would be easiest to test in the CI script after install, as I don't think it's possible to do this with CTest.
> 
> > * By the way is it expected that this .cmake file is only generated with PROJ cmake builds ? That means that, in distributions building PROJ with autoconf, cmake projects cannot import PROJ.
> 
> Having just looking at this in #1484 it seems that only CMake builds can use find_package, and that Debian and Homebrew distributions don't bundle these files as they are Autotools-built.

PR #1912, opened 2020-02-05
cart: Avoid discontinuity at poles in the inverse case (partial fix to #1906)

kbevers:
This  avoids issues with numerical stability as uncovered in
https://github.com/OSGeo/PROJ/issues/1906.

Practically speaking this change isn't going to affect real life
scenarios since the position of the center of the Earth is rarely
expressed in geodetic coordinates.

 - [x] Tests added
 - [x] Added clear title that can be used to generate release notes
 - [ ] Fully documented, including updating `docs/source/*.rst` for new API

Issue #1906, opened 2020-02-04
6.3.0 Tests fail on Alpine Linux x86 (musl libc)


russkel:
#### Example of problem

I am in the process of fixing the Alpine Linux proj package and have found that is solely fails on the x86 platform. 

```
The following tests FAILED:
	  2 - Builtins (Failed)
	  3 - Builtins2 (Failed)
	  6 - Ellipsoid (Failed)
	 31 - testvarious (Failed)
Errors while running CTest
```

```
test 2
      Start  2: Builtins

2: Test command: /builds/russkel/aports/community/proj/src/proj-6.3.0/bin/gie "/builds/russkel/aports/community/proj/src/proj-6.3.0/test/gie/builtins.gie"
2: Environment variables: 
2:  PROJ_LIB=/builds/russkel/aports/community/proj/src/proj-6.3.0/data
2: Test timeout computed to be: 10000000
2: -------------------------------------------------------------------------------
2: proj_create: Error -22: lat_0, lat_1 or lat_2 >= 90
2: proj_create: Error -22: lat_0, lat_1 or lat_2 >= 90
2: proj_create: Error -21: conic lat_1 = -lat_2
2: proj_create: Error -6: effective eccentricity < 0 or >= 1.
2: proj_create: Error -21: conic lat_1 = -lat_2
2: proj_create: Error -22: lat_0, lat_1 or lat_2 >= 90
2: proj_create: Error -22: lat_0, lat_1 or lat_2 >= 90
2: proj_create: Error -21: conic lat_1 = -lat_2
2: proj_create: Error -30: h <= 0 or h > 1e10 * a
2: proj_create: Error -30: h <= 0 or h > 1e10 * a
2: proj_create: Error -58: argument not numerical or out of range
2: proj_create: Error -55: lat_0 = 0
2: proj_create: Error -22: lat_0, lat_1 or lat_2 >= 90
2: proj_create: Error -27: W <= 0 or M <= 0
2: proj_create: Error -22: lat_0, lat_1 or lat_2 >= 90
2: proj_create: Error -6: effective eccentricity < 0 or >= 1.
2: proj_create: Error -6: effective eccentricity < 0 or >= 1.
2: proj_create: Error -32: lat_1=lat_2 or lat_1=0 or lat_2=90
2: proj_create: Error -32: lat_1=lat_2 or lat_1=0 or lat_2=90
2: proj_create: Error -32: lat_1=lat_2 or lat_1=0 or lat_2=90
[...]

PR #1917, opened 2020-02-06
Fix test issues on i386

rouault:
Fix a few issues of #1906 found when running the test suite
on Ubuntu 16.04 with gcc 5.5 -m32 -O2. When applied on top of
the fix of #1912, make check succeeds
Fix additonal issues found with gcc 8.2 -m32 -O2

### DIFF
--- a/src/projections/vandg.cpp
+++ b/src/projections/vandg.cpp
@@ -39,6 +39,14 @@ static PJ_XY vandg_s_forward (PJ_LP lp, PJ *P) {           /* Spheroidal, forwar
         g2 = g * g;
         p2 = g * (2. / p2 - 1.);
         p2 = p2 * p2;
+        {
+            // Force floating-point computations done on 80 bit on
+            // i386 to go back to 64 bit. This is to avoid the test failures
+            // of https://github.com/OSGeo/PROJ/issues/1906#issuecomment-583168348
+            // The choice of p2 is completely arbitrary.
+            volatile double p2_tmp = p2;
+            p2 = p2_tmp;
+        }
         xy.x = g - p2; g = p2 + al2;
         xy.x = M_PI * (al * xy.x + sqrt(al2 * xy.x * xy.x - g * (g2 - p2))) / g;
         if (lp.lam < 0.) xy.x = -xy.x;
@@ -81,7 +89,7 @@ static PJ_LP vandg_s_inverse (PJ_XY xy, PJ *P) {           /* Spheroidal, invers
     m = 2. * sqrt(-THIRD * al);
     d = C2_27 * c2 * c2 * c2 + (c0 * c0 - THIRD * c2 * c1) / c3;
     const double al_mul_m = al * m;
-    if( al_mul_m == 0 ) {
+    if( fabs(al_mul_m) < 1e-16 ) {
         proj_errno_set(P, PJD_ERR_TOLERANCE_CONDITION);
         return proj_coord_error().lp;
     }
--- a/src/projections/laea.cpp
+++ b/src/projections/laea.cpp
@@ -44,7 +44,8 @@ static PJ_XY laea_e_forward (PJ_LP lp, PJ *P) {          /* Ellipsoidal, forward
 
     if (Q->mode == OBLIQ || Q->mode == EQUIT) {
         sinb = q / Q->qp;
-        cosb = sqrt(1. - sinb * sinb);
+        const double cosb2 = 1. - sinb * sinb;
+        cosb = cosb2 > 0 ? sqrt(cosb2) : 0;
     }
 
     switch (Q->mode) {
--- a/src/projections/robin.cpp
+++ b/src/projections/robin.cpp
@@ -84,7 +84,7 @@ static PJ_XY robin_s_forward (PJ_LP lp, PJ *P) {           /* Spheroidal, forwar
     (void) P;
 
     dphi = fabs(lp.phi);
-    i = isnan(lp.phi) ? -1 : lround(floor(dphi * C1));
+    i = isnan(lp.phi) ? -1 : lround(floor(dphi * C1 + 1e-15));
     if( i < 0 ){
         proj_errno_set(P, PJD_ERR_TOLERANCE_CONDITION);
         return xy;
--- a/test/gie/builtins.gie
+++ b/test/gie/builtins.gie
@@ -1333,7 +1333,7 @@ accept  -200 -100
 expect  -0.001790220 -0.000895247
 
 operation +proj=eqdc +a=9999999 +b=.9 +lat_2=1
-expect    failure errno invalid_eccentricity
+expect    failure
 
 operation +proj=eqdc   +R=6400000    +lat_1=1 +lat_2=-1
 expect    failure errno conic_lat_equal
@@ -4087,7 +4087,8 @@ expect  failure errno lat_0_or_alpha_eq_90
 -------------------------------------------------------------------------------
 operation +proj=omerc   +lat_1=0.1 +a=6400000 +b=1
 -------------------------------------------------------------------------------
-expect  failure errno invalid_eccentricity
+# Disabled since fails on i386. Not so important. Edge condition found by ossfuzz
+#expect  failure errno invalid_eccentricity
 
 -------------------------------------------------------------------------------
 operation +proj=omerc    +lat_1=0.8 +a=6400000 +b=.4
--- a/test/gie/ellipsoid.gie
+++ b/test/gie/ellipsoid.gie
@@ -44,7 +44,7 @@ tolerance 10 nm
 accept    1 2
 expect    111319.4907932736  221194.0771604237
 
-accept    12 55s
+accept    12 55
 expect    1335833.8895192828  7326837.7148738774
 -------------------------------------------------------------------------------
 
@@ -117,7 +117,9 @@ expect    1338073.2696101593  7374207.4801437631
 -------------------------------------------------------------------------------
 
 operation proj=merc a=1E77 R_lat_a=90 b=1
-expect    failure errno invalid_eccentricity
+#  errno invalid_eccentricity on x86_64
+#  errno pjd_err_ref_rad_larger_than_90 on i386
+expect    failure
 
 -------------------------------------------------------------------------------
 This one from testvarious failed at first version of the pull request


src/projections/vandg.cpp
```diff
@@ -39,6 +39,11 @@ static PJ_XY vandg_s_forward (PJ_LP lp, PJ *P) {           /* Spheroidal, forwar
         g2 = g * g;
         p2 = g * (2. / p2 - 1.);
         p2 = p2 * p2;
+        {
+            // The 2 following lines are a nasty hack to avoid test failures on i386
```
kbevers: It's obvious to me that this is a hack but it is in no way clear to me what it does. Can you explain that in the comment?

rouault: I'm not even sure this is the correct fix myself. I remembered that geodesic.c has some uses a volatile, that I believe force 80bit intermediate results to be stored on 64bit only. I did that completely arbitrarily on p2 and found that it was sufficient to make the tests pass. Hard science :-) ! I'll amend my comment

kbevers: Maybe @cffk can elaborate on the hard science behind this magical hack :-)

rouault: I did some reading & experimentation about that... That's a complicated story. Basically on x86_64, all floating point operations are done using SSE registers, and double computations are always done on 64bit, and float ones on 32bit.
On i386, by default GCC won't use the SSE registers (unless specifying -msse -mfpmath=sse), so floating point operations are done on the x87 FPU, which operates on 80 bits (this is FLT_EVAL_METHOD==2, whereas x86_64 is FLT_EVAL_METHOD == 0: https://en.cppreference.com/w/c/types/limits/FLT_EVAL_METHOD). Except that when it goes short of internal registers, it has to use memory, and thus 64 bits. So you may have inconsistant precision in computations. With gcc, with C languague (not C++) AND when using -std=c99/c11 explictly, everytime you have a variable assignment, a store/load to memory is done, to emulate (but not completely be conformant) with FLT_EVAL_METHOD == 0

Given
```
double op(double a, double b, double c)
{
    return a * b + c;
}

double op2(double a, double b, double c)
{
    double tmp = a * b;
    return tmp + c;
}

double op3(double a, double b, double c)
{
    volatile double tmp = a * b;
    volatile double res = tmp + c;
    return res;
}
```

C mode, C99 compilation:
```
$ ~/install-gcc-8.2.0/bin/gcc -std=c99 -m32 -O2 -c float.c && objdump -d float.o

float.o:     format de fichier elf32-i386


Déassemblage de la section .text :

00000000 <op>:
   0:	83 ec 0c             	sub    $0xc,%esp
   3:	dd 44 24 18          	fldl   0x18(%esp)
   7:	dc 4c 24 10          	fmull  0x10(%esp)
   b:	dc 44 24 20          	faddl  0x20(%esp)
   f:	dd 1c 24             	fstpl  (%esp)
  12:	dd 04 24             	fldl   (%esp)
  15:	83 c4 0c             	add    $0xc,%esp
  18:	c3                   	ret    
  19:	8d b4 26 00 00 00 00 	lea    0x0(%esi,%eiz,1),%esi

00000020 <op2>:
  20:	83 ec 0c             	sub    $0xc,%esp
  23:	dd 44 24 18          	fldl   0x18(%esp)
  27:	dc 4c 24 10          	fmull  0x10(%esp)
  2b:	dd 1c 24             	fstpl  (%esp)
  2e:	dd 04 24             	fldl   (%esp)
  31:	dc 44 24 20          	faddl  0x20(%esp)
  35:	dd 1c 24             	fstpl  (%esp)
  38:	dd 04 24             	fldl   (%esp)
  3b:	83 c4 0c             	add    $0xc,%esp
  3e:	c3                   	ret    
  3f:	90                   	nop

00000040 <op3>:
  40:	83 ec 1c             	sub    $0x1c,%esp
  43:	dd 44 24 28          	fldl   0x28(%esp)
  47:	dc 4c 24 20          	fmull  0x20(%esp)
  4b:	dd 1c 24             	fstpl  (%esp)
  4e:	dd 04 24             	fldl   (%esp)
  51:	dd 5c 24 08          	fstpl  0x8(%esp)
  55:	dd 44 24 08          	fldl   0x8(%esp)
  59:	dc 44 24 30          	faddl  0x30(%esp)
  5d:	dd 1c 24             	fstpl  (%esp)
  60:	dd 04 24             	fldl   (%esp)
  63:	dd 5c 24 10          	fstpl  0x10(%esp)
  67:	dd 44 24 10          	fldl   0x10(%esp)
  6b:	83 c4 1c             	add    $0x1c,%esp
  6e:	c3                   	ret    

```

C++ mode (or same output for C without explicit -std=)
```
$ ~/install-gcc-8.2.0/bin/g++ -m32 -O2 -c float.c && objdump -d float.o

float.o:     format de fichier elf32-i386


Déassemblage de la section .text :

00000000 <_Z2opddd>:
   0:	dd 44 24 04          	fldl   0x4(%esp)
   4:	dc 4c 24 0c          	fmull  0xc(%esp)
   8:	dc 44 24 14          	faddl  0x14(%esp)
   c:	c3                   	ret    
   d:	8d 76 00             	lea    0x0(%esi),%esi

00000010 <_Z3op2ddd>:
  10:	dd 44 24 04          	fldl   0x4(%esp)
  14:	dc 4c 24 0c          	fmull  0xc(%esp)
  18:	dc 44 24 14          	faddl  0x14(%esp)
  1c:	c3                   	ret    
  1d:	8d 76 00             	lea    0x0(%esi),%esi

00000020 <_Z3op3ddd>:
  20:	83 ec 14             	sub    $0x14,%esp
  23:	dd 44 24 18          	fldl   0x18(%esp)
  27:	dc 4c 24 20          	fmull  0x20(%esp)
  2b:	dd 1c 24             	fstpl  (%esp)
  2e:	dd 04 24             	fldl   (%esp)
  31:	dc 44 24 28          	faddl  0x28(%esp)
  35:	dd 5c 24 08          	fstpl  0x8(%esp)
  39:	dd 44 24 08          	fldl   0x8(%esp)
  3d:	83 c4 14             	add    $0x14,%esp
  40:	c3                   	ret    
```

One can see that in C++ / non-standard C mode, add3() with the volatile keywords has equivalent code to add2() in C99 mode.

In C mode, there's even a -ffloat-store mode where all floating point operations involve a save &  restore from memory, This is then pretty much equivalent to x86_64 mode, and op() and op2() generate the same code, but this is excessively costly

Bottom-line: i386 x87 math is broken.

With clang, SSE is used by default for i386, unless you specify -mno-sse (and clang doesn't implement the -std=C99 special mode of gcc)

kbevers: > i386 x87 math is broken.

That's nice to know. I wonder how long it makes sense to still support those old 32-bit systems. If I understand correctly work-arounds like this one here is not necessary on 64-bit systems, so at some point it would be nice to just not support those old slightly-broken systems and have the code base a little bit cleaner.

kbevers:
I think these could be backported to 6.3.1

src/projections/vandg.cpp
busstoptaktik: The `volatile` keyword tells the compiler that "the memory at this address is really a (memory mapped) hardware register, which *may* be controlled by external hardware. Hence, while both readable and writable, you cannot trust that what you wrote is what you will get upon reading. 

So don't optimize away a store-read back cycle (e.g. if writing a command, then reading a response from the same address, then the read back can be optimized away if not declared volatile).

cffk: I don't regard holding of intermediate results in 80-bit registers to be
intrinsically evil.  It does lead to uncertainty in the final results.
However, other compilation optimizations do that too.  For example the
Intel compiler, by default, ignores parens in expressions like
`a + (b + c)`.

If the resulting variability is the final results is problematic (so
that some tests fail), then it points to some underlying fragility in
the numeric code.  (I'm disregarding for now the possiblility that some
tests are too sensitive.)

I started looking at the vandg code and it is ripe of some
restructuring.  It is evidently following Synder, but this is obscured
by the old Fortran habit of reusing a few variables for different
purposes.  I see one obvious place were numerical accuracy could be
improved, i.e., replacing `x*x - y*y` by `(x-y) * (x+y)`.  Going further
than this would involve understanding the ranges of the various terms in
the code and I'm not sure I'm up for that (probably this should be done
with lots of legacy PROJ code).  I will however check in my annotation
of vandg.cpp linking the forward code up to Synder's paper.

cffk: Are we overthinking this?  One way 80-bit registers mess up a numerical
calculation is that they defeat a carefully thought out method for
controlling numerical error.  However a far more common situation is
that they merely yield a more accurate result.

Is that what's going on here?  It's possible that the test data was
generated ages ago using pure 64-bit arithmetic and the deviations
allowed are just arbitrary guesses.  Now more accurate results are being
flagged as incorrect.

This is easy to check.  Just regenerate the test data with higher
precision arithmetic (e.g., using long double, quad precision, or
maxima's bfloats).  Then enlarge the allowed deviations to permit the
tests to succeed for 64-bit arithmetic and maybe the need to work around
the use of temporary 80-bit registers goes away.

The remaining task would be to investigate how to improve the accuracy
of the 64-bit results so that the allowed deviations could be narrowed
down.

rouault: @cffk I've experimented in https://gist.github.com/rouault/6dc1af2d47426c930e895f4859d1a820 a port to long double as you suggested (except that the constants like M_PI are probably only 64 bit exact). I get results within the 0.25 mm tolerance of the tests. However in my initial attempt where I just changed the type of intermediate variables to long double but didn't use the tanl / fabsl / sqrtl long double variants, I got results above the tolerance (values in between the initial one we got on i386 and the one we get on x86_64). So it seems the implementation is sensitive to have a consistent computation depth, or is just numerically unstable and results we get are perhaps "random"...

cffk: I coded up vandg forward in maxima and the "exact" result (bfloats with
`fpprec:100`) for the test case, `[lon,lat] = [2,1]`, is

    [x,y] = [223395.249543421139, 111704.596516966554]

which is about 0.1 mm from the value in `builtins.gie` of

    [x,y] = [223395.249543407, 111704.596633675]

This reduces the error before your volatile fix from 0.5 mm to 0.4 mm.
So you are right that incorrect test data does not explain the problem
in this instance (although the test data could be improved a little).

PR #1921, opened 2020-02-09
Use relative directory to locate PROJ resource files.

### LINKED ISSUE #1490: Use relative path for PROJ_LIB rather than env. var for conda
The conda `activate.sh` sets the location of `proj.db` via PROJ_LIB.  If you happen to have some other proj installation on your system, setting this variable can cause confusion or an inability to locate resource files.

As an alternative, one might use a location relative to the proj shared library to locate the resources otherwise described by PROJ_LIB.  (../share/proj).  This can be done on Linux with dladdr().  I suppose, but don't know, that there's a way to do this on windows too.

I'm happy to do the work if this seems like a good idea to others.

rouault:
Fixes #1490

This is an extension of the Window-specific logic added recently to
Unix builds. This reuses parts of proposed past commit
https://github.com/OSGeo/PROJ/pull/1517/commits/82a07e51c6e24ddb936d131ababe29f1ac36ef14
(credits to @abellgithub)


hobu (approved):

kbevers:
@rouault merging #1920 causes conflicts here. Do you mind fixing those?

rouault:
> Do you mind fixing those?

Done

kbevers:
> Done

Thank you!

PR #1923, opened 2020-02-10
CMake: simplify message functions

mwtoews:
### Remove `colormsg()`; just use `message()`
The rational is that this custom function sometimes writes garbage on some consoles, e.g. from `cmd.exe`:
```
-- [1;4;34mConfiguring PROJ:[0m
```
while it's possible there is a fix for this one (since I see `cmake --build .` produce colour in the same console), it's not this projects' priority to fix, so just use CMake built-in [`message()`](https://cmake.org/cmake/help/v3.9/command/message.html) function.

### Rename `boost_report_value()` with `print_variable()`
This function appears to be initially based on something for Boost from around 2007 (looking at the various Copyrights at the top), but is now a really basic printing utility that doesn't resemble the original function intent. Rename to something more basic, and change the left-side padding width from 40 to 30 characters.


kbevers:
This is certainly an improvement. It is nice to get those boost remnants replaced with more generic function calls!

PR #1929, opened 2020-02-12
appveyor: Use sqlite tool from vcpkg

kbevers:
Potential fix for the failing appveyor builds. The problems seems to be the sqlite3 tool not being properly found. This installs it using vcpkg, as detailed in the build guide in the docs. No idea if this actually works...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,16 +26,9 @@ build_script:
   - bootstrap-vcpkg.bat
   - set PATH=%CD%;%PATH%
   - cd ..
-  - vcpkg install sqlite3:"%platform%"-windows
+  - vcpkg install sqlite3[core,tool]:"%platform%"-windows
   - vcpkg install tiff:"%platform%"-windows
   - vcpkg install curl:"%platform%"-windows
-  - set SQLITE3_BIN=%APPVEYOR_BUILD_FOLDER%\sqlite3\bin
-  - mkdir %SQLITE3_BIN%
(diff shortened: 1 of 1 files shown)


rouault:
@kbevers No, the issue is vcpkg build itself which is broken on MSVC 2015 since a couple days. I've reported that to https://github.com/microsoft/vcpkg/issues/10011

kbevers:
Right okay, it did seem odd that it just suddenly stopped working. I wonder if my approach in this PR will work anyway (once vcpk is fixed)? It is a bit cleaner I think.

rouault:
yes, if it works when vcpkg is up again, it is definitely cleaner!

dan-shaw:
@rouault @kbevers vcpkg maintainer here. We're trying to get this fixed as soon as we can. I'm curious - is this the first time your appveyor builds have failed due to vcpkg?

rouault:
> is this the first time your appveyor builds have failed due to vcpkg?

No. Initially we didn't build vcpkg but used the vcpkg binary already available on appveyor. But later, vcpkg install started to fail for sqlite, and the fix we found was to build vcpkg master from git HEAD (was in 1794d3a3090c2466d6699c02d9a05996302d4224 ). I don't recall build failures since we did that

snowman2:
Build seems to be working now for pyproj.

kbevers:
> Build seems to be working now for pyproj.

Also here. That's great!

PR #1931, opened 2020-02-13
Don't print version when using PROJ4 from cmake

calumr:
This will declutter cmake output for PROJ4 users.

I use a few dozen libraries in my C++ project, and none of them print their version to stdout when I run cmake. I realise it may be useful when developing PROJ4 to have this message, but for users who may not even be depending on PROJ4 directly (e.g. indirectly via another library), it's a bit noisy.

### DIFF
--- a/cmake/project-config.cmake.in
+++ b/cmake/project-config.cmake.in
@@ -8,12 +8,6 @@
 #  @PROJECT_VARIANT_NAME@_BINARY_DIRS = /usr/local/bin
 #  @PROJECT_VARIANT_NAME@_VERSION = 4.9.1 (for example)
 
-message (STATUS "Reading ${CMAKE_CURRENT_LIST_FILE}")
-# @PROJECT_VARIANT_NAME@_VERSION is set by version file
-message (STATUS
-  "@PROJECT_VARIANT_NAME@ configuration, \
-version ${@PROJECT_VARIANT_NAME@_VERSION}")
-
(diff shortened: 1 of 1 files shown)


kbevers:
Can you show us examples output of before and after applying this patch?

calumr:
Before:
```
C:\dev\bin-master> cmake ..\discovery\
-- Reading C:/dev/Applications/vcpkg/discovery-7/installed/x86-windows/share/proj4/proj4-config.cmake
-- PROJ4 configuration, version 4.9.3
-- Configuring done
-- Generating done
-- Build files have been written to: C:/dev/bin-master
```
After:
```
C:\dev\bin-master> cmake ..\discovery\
-- Configuring done
-- Generating done
-- Build files have been written to: C:/dev/bin-master
```

kbevers:
Thinks, this makes sense. @mwtoews do you agree?

mwtoews (approved):
Agree, status messages are generally [not part of the guidelines](https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-to-create-a-ProjectConfig.cmake-file)

calumr:
Thanks so much for the very quick turnaround!

kbevers:
> Thanks so much for the very quick turnaround!

No problem. Thanks for your input. This change will be available in PROJ 7.0.0 which will be released March 1st.

PR #1934, opened 2020-02-18
Don't assume $HOME to be writable.

### LINKED ISSUE #1933: proj_context.read_grid_from_user_writable_directory test fails in Debian build chroot
The Debian package for PROJ 7.0.0-RC1 fails to build due to a test failure:

```
[ RUN      ] proj_context.read_grid_from_user_writable_directory
proj_context_test.cpp:160: Failure
Value of: createTmpFile(filename)
  Actual: false
Expected: true
proj_create: Cannot find proj.db
proj_create: cannot expand +init=temp_proj_dic3:MY_PIPELINE
proj_create: Cannot find proj.db
proj_create: cannot expand +init=temp_proj_dic3:MY_PIPELINE
proj_context_test.cpp:172: Failure
Expected: (P) != (nullptr), actual: NULL vs (nullptr)
[  FAILED  ] proj_context.read_grid_from_user_writable_directory (1 ms)
```

The build chroot sets `$HOME` to `/nonexistent` to prevent packages from writing to the user home per [Debian Policy 9.2.3](https://www.debian.org/doc/debian-policy/ch-opersys.html#non-existent-home-directories).

The directory for the test should probably be set to a temporary directory.

One option is to not assume that `$HOME` is writable, and test this explicitly in `pj_context_get_user_writable_directory` which will set the path to `/tmp` if it's not true.

sebastic:
The read_grid_from_user_writable_directory test fails otherwise.

Fixes: #1933

### DIFF
--- a/src/filemanager.cpp
+++ b/src/filemanager.cpp
@@ -1220,7 +1220,7 @@ std::string pj_context_get_user_writable_directory(PJ_CONTEXT *ctx,
             path = xdg_data_home;
         } else {
             const char *home = getenv("HOME");
-            if (home) {
+            if (home && access(home, W_OK) == 0) {
 #if defined(__MACH__) && defined(__APPLE__)
                 path = std::string(home) + "/Library/Application Support";
 #else


kbevers:
cherry-picked into 7.0. Thanks for the fix, Bas.

PR #1936, opened 2020-02-18
Fix static linking (due to typo in pc)

### LINKED ISSUE #32: Configure script not finding dependencies as static libraries
I'm building libgeotiff in an environment with only static libraries. Therefore, libproj needs to be linked with `-lproj -lsqlite3` and libtiff needs to be linked as follows:

```sh
pkg-config --libs-only-l --static libtiff-4
#  -ltiff -ljpeg -lz
```

When I run `./configure --enable-static --disable-shared --with-proj` it errors like so:

```
checking for PROJ >= 6 library... checking for proj_create_from_wkt in -lproj... no
checking for internal_proj_create_from_wkt in -lproj... no
checking for internal_proj_create_from_wkt in -linternalproj... no
configure: error: PROJ 6 symbols not found
```

From `configure.log` I see the configure test fails because:
 - It tries to link to `libproj.a` with the C linker instead of C++ (lots of undefined stdc++ errors)
 - It is not using -lsqlite3

I was able to work around it like so:

```sh
export LIBS="-lsqlite3 -lstdc++"
./configure --enable-static --disable-shared --with-proj
```

How the configure script succeeds, and I am able to build `libgeotiff.a` however the build fails when it needs to link to libtiff:

```sh
libtool: link: x86_64-w64-mingw32-gcc -I../../libgeotiff-1.5.1 -I../../libgeotiff-1.5.1/libxtiff -DHAVE_TIFF=1 -march=x86-64 -mtune=generic -O2 -pipe -O3 -DNDEBUG -pipe -o applygeo.exe applygeo.o -pipe  ../.libs/libgeotiff.a -lproj -lsqlite3 -ljpeg -lz -lstdc++ -ltiff
C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../lib/libtiff.a(tif_jpeg.o):(.text+0x244): undefined reference to `jpeg_std_error'
C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../lib/libtiff.a(tif_jpeg.o):(.text+0x2a3): undefined reference to `jpeg_CreateDecompress'
C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../lib/libtiff.a(tif_jpeg.o):(.text+0x375): undefined reference to `jpeg_abort'
... + 100+ more like this
```

The problem is that it need to link to libtiff using `-ltiff -ljpeg -lz` instead of `-ltiff`. However I cannot work around these by adding them to `LIBS` because it seems the configure script is explicitly stripping `-ltiff` from LIBS and then putting it back at the end (but `-ljpeg` and `-lz` need to appear _after_ `-ltiff` in the linker flags)

I was able to work around it like so:

```sh
sed -i.bak "s/-ltiff/$(pkg-config --libs-only-l --static libtiff-4)/g" configure
```

Then it builds 🎉  But it's all a bit hacky of course. The configure script could be improved by:
 - Trying to link libproj using `-lproj -lsqlite3` __using the c++ linker__
 - Trying to link libtiff using `pkg-config --libs --static libtiff-4` 

Thanks! Not super urgent (I have workarounds).

jeroen:
This fixes: https://github.com/OSGeo/libgeotiff/issues/32

Before:

```sh
$ pkg-config --libs --static proj
# -LC:/rtools40/mingw64/lib -lproj
```

After:

```sh
$ pkg-config --libs --static proj
# -LC:/rtools40/mingw64/lib -lproj -lsqlite3 -lz -lstdc++
```

### DIFF
--- a/proj.pc.in
+++ b/proj.pc.in
@@ -9,5 +9,5 @@ Description: Cartographic Projections Library.
 Requires:
 Version: @VERSION@
 Libs: -L${libdir} -lproj
-Libs.Private: @SQLITE3_LIBS@ -lstdc++
+Libs.private: @SQLITE3_LIBS@ -lstdc++
 Cflags: -I${includedir}


kbevers:
Thanks, this will be included in the upcoming 7.0.0 release.

PR #1939, opened 2020-02-18
Install null grid file if available.

sebastic:
It is one of the files included in proj-datumgrid.

It was added to the proj source tree in 6.3.0, and removed in 7.0.0.

When the content of proj-datumgrid in unpacked in the data directory the null grid should be installed along with the others grids like alaska, conus, etc.

### DIFF
--- a/data/CMakeLists.txt
+++ b/data/CMakeLists.txt
@@ -7,6 +7,7 @@ set(CONFIG_FILES
 )
 
 set(PROJ_DICTIONARY
+  null
   world
   other.extra
   nad27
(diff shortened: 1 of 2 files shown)


rouault:
this is not needed. in proj 7 the null file is no longer materialized but a special hardcoded case

sebastic:
> this is not needed. in proj 7 the null file is no longer materialized but a special hardcoded case

Where is this documented?

Does that also work for init files that use it in `+nadgrids`, and pipelines that use it in `+proj=hgridshift` or `+proj=vgridshift`?

AFAIK at least `+nadgrids` references an actual file.

rouault:
documented in a commit message probably. 

the virtual null will be used in all code paths where a grid can be used. nadgrids hgridshift vgridshift deformation xyzgridshift

sebastic:
NEWS references #1898 which doesn't reference where it's currently hardcoded.

`src/datum_set.cpp` references `pj_apply_gridshift` to handle the `nadgrids` option,
this in turn references `NS_PROJ::getListOfGridSets` to get the list of grids,
that uses `HorizontalShiftGridSet::open` which doesn't reference any hardcoded null.

From what I can see, not having the null grid will break `+nadgrids`.

rouault:
look at line 2287. i'm pretty sure there are tests triggering that

sebastic:
Thanks, that's what I was looking for.

PR #1941, opened 2020-02-18
Alternative grid for JTSK - JTSK03 transform (EPSG:8364)

wonder-sk:
The definition of EPSG:8364 uses NADCON method for horizontal grid (.las/.los files).
UGKK (Slovak Geodetic and Cartographic Institute) provides NADCON .las/.los files here:
https://www.geoportal.sk/files/gz/slovakia_jtsk03_to_jtsk.zip

Additionally UGKK also provides the same grid file in NTv2 format:
https://www.geoportal.sk/files/gz/slovakia_jtsk03_to_jtsk_ntv2.zip

Unfortunately PROJ cannot use the NADCON method files, so this pull request adds the NTv2 grid file to the grid_alternatives table so that PROJ can automatically pick it up...

Note: this addition does not specify any `package_name` for the grid because it is not included in the proj datumgrids package due to missing license. I have already contacted UGKK to resolve this issue so that the grid file may be included as well, but it may take them some time to sort it out.

Currently to get this transformation working with PROJ/QGIS, there is documentation which involves some fragile steps like patching proj.db and srs.db (qgis):
https://www.geoportal.sk/files/gz/s-jtsk_jtsk03_v_qgis.pdf (sorry only in Slovak)

Without this change:
```
$ projinfo --summary -s EPSG:5513 -t EPSG:8352
Candidate operations found: 4
Note: using '--spatial-test intersects' would bring more results (7)
unknown id, Inverse of Krovak (Greenwich) + S-JTSK to ETRS89 (3) + ETRS89 to S-JTSK [JTSK03] (1) + Krovak (Greenwich), 0.501 m, Slovakia
unknown id, Inverse of Krovak (Greenwich) + S-JTSK to ETRS89 (3) + Inverse of S-JTSK [JTSK03] to ETRS89 (1) + Krovak (Greenwich), 0.501 m, Slovakia
unknown id, Inverse of Krovak (Greenwich) + Ballpark geographic offset from S-JTSK to S-JTSK [JTSK03] + Krovak (Greenwich), unknown accuracy, World, has ballpark transformation
unknown id, Inverse of Krovak (Greenwich) + Inverse of S-JTSK [JTSK03] to S-JTSK (1) + Krovak (Greenwich), 0.05 m, Slovakia, at least one grid missing
```
(even when the .las/.los grid file is available, the last option will be unable to generate proj string)

With this change:
```
$ projinfo --summary -s EPSG:5513 -t EPSG:8352
Candidate operations found: 1
unknown id, Inverse of Krovak (Greenwich) + Inverse of S-JTSK [JTSK03] to S-JTSK (1) + Krovak (Greenwich), 0.05 m, Slovakia
```

### DIFF
--- a/data/sql/grid_alternatives.sql
+++ b/data/sql/grid_alternatives.sql
@@ -2110,3 +2110,19 @@ INSERT INTO grid_alternatives(original_grid_name,
                               0,
                               'proj-datumgrid-oceania',
                               NULL, NULL, NULL, NULL);
+
+-- Slovakia
+INSERT INTO grid_alternatives(original_grid_name,
+                              proj_grid_name,
+                              proj_grid_format,
+                              proj_method,
+                              inverse_direction,
+                              package_name,
+                              url, direct_download, open_license, directory)
+                      VALUES ('Slovakia_JTSK03_to_JTSK.LAS',
+                              'Slovakia_JTSK03_to_JTSK.gsb',
+                              'NTv2',
+                              'hgridshift',
+                              0,
+                              NULL,
+                              NULL, NULL, NULL, NULL);
--- a/data/sql/grid_alternatives.sql
+++ b/data/sql/grid_alternatives.sql
@@ -2112,6 +2112,12 @@ INSERT INTO grid_alternatives(original_grid_name,
                               NULL, NULL, NULL, NULL);
 
 -- Slovakia
+--
+-- The definition of EPSG:8364 (JTSK03 to JTSK) uses NADCON method which is not supported by PROJ.
+-- UGKK (Slovak Geodetic and Cartographic Institute) provides also NTv2 grid file in addition
+-- to NADCON .las/.los files, so we define the NTv2 file here as an alternative.
+-- The file is available at https://www.geoportal.sk/files/gz/slovakia_jtsk03_to_jtsk_ntv2.zip
+-- but it does not have a confirmed license yet, so it is not available in proj-datumgrid-europe for now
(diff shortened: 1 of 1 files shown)


rouault:
We would need an equivalent pull request against master (some changes will be needed). This PR could be enhanced by adding more comments in the SQL itself as it is a particular case. The URL https://www.geoportal.sk/files/gz/slovakia_jtsk03_to_jtsk_ntv2.zip could be added in the dedicated column, with open_license and direct_download to 0.

wonder-sk:
@rouault I have added extra comments to make the new entry more clear

I am happy to port this addition also to master, but I wanted to first test with a stable version and make sure this approach is fine.

rouault:
> make sure this approach is fine.

pending obtaining compatible licensing terms for the grids, that's probably OK. There's the precedent of the Dutch 2008 non-free grids.

What about my suggestions regarding improving the content of the INSERT itself ?

wonder-sk:
> What about my suggestions regarding improving the content of the INSERT itself ?

Ah sorry, I have just misunderstood and thought those updates for INSERT were only meant when updating master. Fixed now.

kbevers:
This now lives in 6.3 but a port to master/7.0 is also needed. It is not just a trivial cherry-pick but as far as I can tell the fix is fairly trivial.

> I am happy to port this addition also to master, but I wanted to first test with a stable version and make sure this approach is fine.

I will release 7.0.0RC2 on Monday - if you can make the port to master/7.0 be then I will include it in the release.

PR #1950, opened 2020-02-19
Add/update description for pkg-config and CMake, ignore proj.pc.in warning

mwtoews:
This PR adds or updates the description string based on the title in `CITATION`.

The [`PROJECT_DESCRIPTION` variable](https://cmake.org/cmake/help/v3.9/variable/PROJECT_DESCRIPTION.html) is new since CMake 3.9.

Also, clear up an annoyance with `./configure`, based on [autoconf's docs](https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/Changed-Directory-Variables.html), ignore:
> WARNING:  'proj.pc.in' seems to ignore the --datarootdir setting

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,10 @@
 ################################################################################
 cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
 
-project(PROJ LANGUAGES C CXX)
+project(PROJ
+  DESCRIPTION "PROJ coordinate transformation software library"
+  LANGUAGES C CXX
+)
 
 # Only interpret if() arguments as variables or keywords when unquoted
 cmake_policy(SET CMP0054 NEW)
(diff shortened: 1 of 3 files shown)


mwtoews:
A few other footnotes to mention:

* You can inspect `pkg-config --list-all` to see what's available on your system
* Name here is changed from `proj` to `PROJ`; according to [the guide](https://people.freedesktop.org/~dbn/pkg-config-guide.html):
> **Name**: A human-readable name for the library or package. This does not affect usage of the `pkg-config` tool, which uses the name of the `.pc` file.

So these changes get `pkg-config --list-all` to output:

    proj                        PROJ - Coordinate transformation software library

---

Also, should the description for this GitHub project also be changed to "Coordinate transformation software library"?

kbevers:
> Also, should the description for this GitHub project also be changed to "Coordinate transformation software library"?

We can do that if you think it is a better description. I think the "cartographic projections" part is in there since that was what it originally said. "Cartographic projections library" or something like that.

PR #1951, opened 2020-02-19
Switch build configuration logic from DISABLE_TIFF to ENABLE_TIFF

mwtoews:
This is subtle change to the build configuration options, as I found that the CMake `-DDISABLE_TIFF=YES` option to be a slightly awkward way to turn-off TIFF support, and `-DENABLE_TIFF=NO` more natural.

While I have also changed the Autotools setup here too, I can keep it as `--disable-tiff` to make it more clear that it is enabled by default. Regardless, the `--help` string is updated to be clear. I've checked that `--enable-tiff`/ `--enable-tiff=yes`/"not specified" enable this feature and  `--enable-tiff=no`/ `--disable-tiff` disable it, as expected. And again, I can back-out these changes to `configure.ac` to keep it as `--disable-tiff`, let me know.

Other changes are to:
* Use `TIFF_ENABLED` and `CURL_ENABLED` variables, based on option and outcome
* Reword some messages and add hints
* Move CPP's `-DTIFF_ENABLED` and `-DCURL_ENABLED` from global `add_definitions()` to `target_compile_definitions()`, which is recommended practice
* Minor spelling and style consistency around `SQLITE_VERSION` check


configure.ac
```diff
@@ -256,12 +256,10 @@ dnl Check for libtiff
 dnl ---------------------------------------------------------------------------
 
 AC_ARG_ENABLE([tiff],
-              AS_HELP_STRING([--disable-tiff],
-                    [Disable TIFF support. Strongly discouraged !]),
-              [enable_tiff=no],
-              [enable_tiff=yes])
+              AS_HELP_STRING([--enable-tiff],
+                    [Enable TIFF support; strongly encouraged to read some grids [default=yes]]))
```
QuLogic, line 260: Before was incorrect, too. The third argument to `AC_ARG_ENABLE` is  action-if-provided, and the fourth argument is action-if-not-provided. That's not what it used to do. It's possible there are more of these in `configure.ac`.

However, there's not much benefit to changing the help string if the default is enabled. `AC_ARG_ENABLE` defines _both_ `--enable-tiff` and `--disable-tiff` (and also accepts `--enable-tiff=no`, too).

kbevers:
>  I found that the CMake -DDISABLE_TIFF=YES option to be a slightly awkward way to turn-off TIFF support, and -DENABLE_TIFF=NO more natura

I believe this is awkward by design - not building with tiff support is discouraged and this is trying to communicate that. I don't have strong feelings on how this should be done so I'll leave that to @rouault

rouault (approved):
I'm OK with those changes. Yes this was more or less intended to be a bit awkward to disable TIFF support, but now you've done the work to make this more standard, let's keep it.

kbevers:
> let's keep it.

Do we have docs that need updating to reflect this change?

mwtoews:
> Do we have docs that need updating to reflect this change?

No, I don't see any in docs/. However, there is `./configure --help`
```
 ...
Optional Features:
....
  --enable-tiff           Enable TIFF support; strongly encouraged to read
                          some grids [default=yes]
```
or `cmake -LAH` (list all help)
```
...
// Enable TIFF support to read some grids
ENABLE_TIFF:BOOL=ON
...
```
(this is also shown via tooltip in the CMake GUI with the "Advanced" box checked on)

---

Should we start to document two sections in `install.rst`, one for "CMake options" and another for "Autotools options"?

kbevers:
> No, I don't see any in docs/. However, there is ./configure --help

Those are auto-generated, right?

> Should we start to document two sections in install.rst, one for "CMake options" and another for "Autotools options"?

If it is feasible, that would be awesome to have. I often find that I have no clue to turn on/off various features when I am building software. Stuff like that is rarely documented. I guess because it is the type of thing you are just supposed to know, perhaps?

mwtoews:
> Those are auto-generated, right?

Yes, via `./autogen.sh` for `./configure --help`, and [`cmake`](https://cmake.org/cmake/help/latest/manual/cmake.1.html) via `-L[A][H]` to **l**ist cached variables that are **a**dvanced with the **h**elp string.

> If it is feasible, that would be awesome to have. I often find that I have no clue to turn on/off various features when I am building software. Stuff like that is rarely documented. I guess because it is the type of thing you are just supposed to know, perhaps?

I'll see if I can cobble some docs together before RC2. Some of these options are loosely documented already. They could perhaps be organised in a section listing a bunch of:
```
.. option:: --foo

    Blah
```

configure.ac
mwtoews: @QuLogic so is this [`AC_ARG_ENABLE`](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Package-Options.html) good? The third and fourth arguments are optional. And I've stress tested all the combinations to ensure that I see (or don't) a line "checking for TIFF... yes".

The only other example using `AC_ARG_ENABLE` is for `--enable-lto`, which is done similar to this.

kbevers:
> I'll see if I can cobble some docs together before RC2

Don't worry if you can't it in time. The docs are easy to touch up after a release. The code not so much :-)

PR #1960, opened 2020-02-20
Support more projections when creating coordinate operations

nyalldawson:
Adds support for creating coordinate operations using Patterson, Natural Earth (i/ii), Compact Miller, Times and Vertical Near Side Perspective based CRSes (e.g. ESRI:53079)

(is this suitable for backport?)


kbevers:
Thanks Nyall.

> (is this suitable for backport?)

If you can make the case for this being a bug fix rather than a new feature :-)

nyalldawson:
I think there's a strong case - I certainly expected this to work at first, given that the projections themselves are advertised as being supported by proj. (Plus, it's really nice to be able to use these seamlessly in qgis!)

kbevers:
> I think there's a strong case

I agree. Thanks for taking care of this!

kbevers:
manually backported since backport-bot doesn't to multi-commit backports :|

include/proj/internal/coordinateoperation_constants.hpp
```diff
@@ -729,9 +741,15 @@ static const MethodMapping projectionMethodMappings[] = {
...
      paramsObliqueStereo},
 
+    {PROJ_WKT2_NAME_METHOD_TIMES, 0, "Times", "times", nullptr,
+     paramsLonNatOrigin},
+
     {PROJ_WKT2_NAME_METHOD_VAN_DER_GRINTEN, 0, "VanDerGrinten", "vandg", "R_A",
      paramsLonNatOrigin},
 
+    {PROJ_WKT2_NAME_METHOD_VERTICAL_NEAR_SIDE_PERSPECTIVE, 0,
+     "Vertical_Near_Side_Perspective", "nsper", nullptr, paramsLonNatOrigin},
```
rouault, line 751: paramsLonNatOrigin is not appropriate for nsper. nsper has a "h" mandatory parameter and lat_0 is also handled.

include/proj/internal/esri_projection_mappings.hpp
```diff
@@ -616,6 +616,24 @@ static const ESRIParamMapping paramsESRI_Gnomonic[] = {
...
+static const ESRIParamMapping paramsESRI_Times[] = {
+    {"False_Easting", EPSG_NAME_PARAMETER_FALSE_EASTING,
+     EPSG_CODE_PARAMETER_FALSE_EASTING, 0.0},
+    {"False_Northing", EPSG_NAME_PARAMETER_FALSE_NORTHING,
+     EPSG_CODE_PARAMETER_FALSE_NORTHING, 0.0},
+    {"Central_Meridian", EPSG_NAME_PARAMETER_LONGITUDE_OF_NATURAL_ORIGIN,
+     EPSG_CODE_PARAMETER_LONGITUDE_OF_NATURAL_ORIGIN, 0.0},
+    {nullptr, nullptr, 0, 0.0}};
+
+static const ESRIParamMapping paramsESRI_Vertical_Near_Side_Perspective[] = {
```
rouault, line 628: incomplete mapping. Missing Height and Latitude_Of_Center parameters. see https://github.com/Esri/projection-engine-db-doc/blob/master/other/proj4-esri.txt#L396

scripts/build_esri_projection_mapping.py
```diff
@@ -459,9 +459,19 @@
...
+    WKT2_name: PROJ_WKT2_NAME_METHOD_TIMES
+    Params:
+        - False_Easting: EPSG_NAME_PARAMETER_FALSE_EASTING
+        - False_Northing: EPSG_NAME_PARAMETER_FALSE_NORTHING
+        - Central_Meridian: EPSG_NAME_PARAMETER_LONGITUDE_OF_NATURAL_ORIGIN
 
-# Vertical_Near_Side_Perspective: not handled, but present in PROJ
+- Vertical_Near_Side_Perspective:
+    WKT2_name: PROJ_WKT2_NAME_METHOD_VERTICAL_NEAR_SIDE_PERSPECTIVE
+    Params:
```
rouault, line 471: missing Height and Latitude_Of_Center parameters. See above remark.

kbevers:
Sorry, it seems I was too quick on the merge button. Thanks for the corrections  @rouault

scripts/build_esri_projection_mapping.py
nyalldawson: I'll fix, thanks

nyalldawson:
@rouault 

Looking at the remaining gaps, I have some questions

- Behrmann: need to use cea, but with fixed lat_0=30 +lat_ts=0 parameters. Is this possible to specify in the mapping?

- Hammer_Aitoff: is this the same as "hammer", but with the fixed W=0. parameter mentioned in https://proj.org/operations/projections/hammer.html?

- Flat_Polar_Quartic: same as McBryde-Thomas Flat-Polar Quartic ?

- "Cube": I'm guessing this is https://pro.arcgis.com/en/pro-app/help/mapping/properties/cube.htm, and differs from https://proj.org/operations/projections/qsc.html because qsc only supports a single face at a time..?

rouault:
> I have some questions

I'm afraind I have no particular insights to provide on those.

nyalldawson:
@kbevers have you any insights on those questions?

kbevers:
> @kbevers have you any insights on those questions?

Unfortunately no. I think your best bet is the mailing list at this point

PR #1962, opened 2020-02-21
CMake: rename BUILD_LIBPROJ_SHARED to BUILD_SHARED_LIBS

### LINKED ISSUE #1858: PROJ not following the CMake convention of BUILD_SHARED_LIBS
_Directed by hobu on glitter to raise a ticket_

#### Example of problem
Using the PROJ6.3.0 source release, I am unable to build proj.dll (everything else builds fine).  According to what I have read, I should be able to add the following to the cmakelists.txt to produce a DLL:

set(BUILD_SHARED_LIBS ON)

However, this has no effect.  PROJ successfully builds the .LIB, and the executable files; but the .DLL is not created.  This is the command I am using from the VS2019 Command Prompt to start my build:

```
cmake -DCMAKE_GENERATOR_PLATFORM=x64 -DSQLITE3_INCLUDE_DIR=....\ThirdParty\sqlite\include -DSQLITE3_LIBRARY=....\ThirdParty\sqlite\Release\lib\sqlite3.lib -DEXE_SQLITE3=....\ThirdParty\sqlite\Release\bin\sqlite3.exe -DCMAKE_INSTALL_PREFIX=Release ..

cmake --build . --config RELEASE --target INSTALL
```

#### Problem description

For us, this is an issue, as we use the DLL when building a custom C# wrapper for the GDAL.  We have found a way to generate the DLL, but it involves modifying the lib_proj.cmake file:

```
if(WIN32)
  set(BUILD_LIBPROJ_SHARED_DEFAULT ON)
endif()
option(BUILD_LIBPROJ_SHARED_DEFAULT
  "Build libproj library shared." ${BUILD_LIBPROJ_SHARED_DEFAULT})
if(BUILD_LIBPROJ_SHARED_DEFAULT)
  set(PROJ_LIBRARY_TYPE SHARED)
else()
  set(PROJ_LIBRARY_TYPE STATIC)
endif()
```

However, as one would expect, this breaks the build in other ways. The DLL will now build, but the LIB and EXE's will no longer build. (I should note, I'm not a C/C++ developer).

We are hoping to have a reproducible build process if possible.

#### Expected Output
A build of the DLL, LIB, and EXEs.

#### Environment Information
 - PROJ version (`proj`)
PROJ 6.3.0 (https://github.com/OSGeo/PROJ/releases/tag/6.3.0)

Windows 10, Visual Studio 2019

#### Installation method
Downloaded the source release from (https://github.com/OSGeo/PROJ/releases/tag/6.3.0), un-archived it, and attempted to build.   See "Example of problem" above for the exact calls to cmake I'm doing to start the build after extracting the archive.

mwtoews:
CMake has a global [`BUILD_SHARED_LIBS` variable](https://cmake.org/cmake/help/v3.9/variable/BUILD_SHARED_LIBS.html), which "is often added to projects as an `option()` so that each user of a project can decide if they want to build the project using shared or static libraries". This PR essentially renames `BUILD_LIBPROJ_SHARED` to `BUILD_SHARED_LIBS`.

I'd appreciate a review from @hobu and @mloskot , if possible.

Other notes:
* Deprecate `BUILD_LIBPROJ_SHARED`, but still use it as an alias for now. The plan is to remove it by PROJ 8.0, however this can be pushed back further.
* Rename `BUILD_LIBPROJ_SHARED_DEFAULT` to `BUILD_SHARED_LIBS_DEFAULT`
* Keep previous defaults (UNIX as shared and Windows as static)
* Remove `PROJ_LIBRARY_TYPE`, since `add_library()` uses `BUILD_SHARED_LIBS`
* Closes #1858

### DIFF
--- a/src/bin_cct.cmake
+++ b/src/bin_cct.cmake
@@ -14,6 +14,6 @@ target_compile_options(cct PRIVATE ${PROJ_CXX_WARN_FLAGS})
 install(TARGETS cct
   RUNTIME DESTINATION ${BINDIR})
 
-if(MSVC AND BUILD_LIBPROJ_SHARED)
+if(MSVC AND BUILD_SHARED_LIBS)
   target_compile_definitions(cct PRIVATE PROJ_MSVC_DLL_IMPORT=1)
 endif()
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -51,7 +51,7 @@ endif()  # USE_EXTERNAL_GTEST
 # Build PROJ unit tests
 #
 
-if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC" AND BUILD_LIBPROJ_SHARED)
+if(MSVC AND BUILD_SHARED_LIBS)
   add_definitions(-DPROJ_MSVC_DLL_IMPORT=1)
 endif()
 
@@ -104,7 +104,7 @@ add_test(NAME proj_context_test COMMAND proj_context_test)
 set_property(TEST proj_context_test
         PROPERTY ENVIRONMENT "PROJ_SKIP_READ_USER_WRITABLE_DIRECTORY=YES;PROJ_LIB=${PROJECT_BINARY_DIR}/data/for_tests")
 
-if(MSVC AND BUILD_LIBPROJ_SHARED)
+if(MSVC AND BUILD_SHARED_LIBS)
   # ph_phi2_test not compatible of a .dll build
 else()
   add_executable(pj_phi2_test
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,9 +30,9 @@ build_script:
   - vcpkg install tiff:"%platform%"-windows
   - vcpkg install curl:"%platform%"-windows
   - if "%platform%" == "x64" SET VS_FULL=%VS_VERSION% Win64
-  - if "%platform%" == "x64" SET BUILD_LIBPROJ_SHARED=ON
+  - if "%platform%" == "x64" SET BUILD_SHARED_LIBS=ON
   - if "%platform%" == "x86" SET VS_FULL=%VS_VERSION%
-  - if "%platform%" == "x86" SET BUILD_LIBPROJ_SHARED=OFF
+  - if "%platform%" == "x86" SET BUILD_SHARED_LIBS=OFF
   - echo "%VS_FULL%"
 #
   - cd %APPVEYOR_BUILD_FOLDER%\data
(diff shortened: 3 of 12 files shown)


hobu (approved):

test/unit/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ endif()  # USE_EXTERNAL_GTEST
 # Build PROJ unit tests
 #
 
-if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC" AND BUILD_LIBPROJ_SHARED)
+if(MSVC AND BUILD_SHARED_LIBS)
   add_definitions(-DPROJ_MSVC_DLL_IMPORT=1)
```
mwtoews, line 55: It doesn't look like any tests use the `PROJ_MSVC_DLL_IMPORT` define, so I might remove this block.

src/bin_cct.cmake
```diff
@@ -14,6 +14,6 @@ target_compile_options(cct PRIVATE ${PROJ_CXX_WARN_FLAGS})
 install(TARGETS cct
   RUNTIME DESTINATION ${BINDIR})
 
-if(MSVC AND BUILD_LIBPROJ_SHARED)
+if(MSVC AND BUILD_SHARED_LIBS)
   target_compile_definitions(cct PRIVATE PROJ_MSVC_DLL_IMPORT=1)
 endif()
```
mloskot, line 19: A personal preference comment: I've learned to strive for declarative CMake scripting and avoid `if`-s. I'd probably have rewritten those `if`-s using the generator expressions:

Do this ones:
```
set(msvc_build_shared "$<AND:$<CXX_COMPILER_ID:MSVC>,$<BOOL:${BUILD_SHARED_LIBS}>>")
```

Then reuse
```
target_compile_definitions(cct PRIVATE $<${msvc_build_shared}:PROJ_MSVC_DLL_IMPORT>)
```

mloskot (approved):
LGTM. Thanks for your efforts

test/unit/CMakeLists.txt
mloskot: Since this PR is about renaming, the clean-up could be a separate PR.

mwtoews:
I wasn't 100% certain that `UNIX` and `WIN32` are mutually exclusive, so I've adjusted the defaults to choose shared, except static for Windows (as was the logic before). I'm more interested in the AppVeyor results, which seem to be quite slow lately...

test/unit/CMakeLists.txt
mwtoews: Hmm right, I'll keep this block for now. Restored.

PR #1968, opened 2020-02-23
Add alternative grid for JTSK - JTSK03 transform (EPSG:8364)

wonder-sk:
This is port of #1941 to master.

I was wondering what to put to `proj_grid_name` column - according to the convention it should be `sk_ugkk_jtsk03-jtsk.tif`, but that would not work as PROJ would try to use that file name, and NULL was not allowed either, so I used the original `Slovakia_JTSK03_to_JTSK.gsb` file name - hopefully it's fine like that.

I see there is also 7.0 branch - should I port it there as well?


kbevers:
> so I used the original Slovakia_JTSK03_to_JTSK.gsb file name - hopefully it's fine like that.

Since this grid is not tracked in PROJ-data I think it is best to go with the original name from the creator. If we can include it in PROJ-data at some point I think we will rename it and adjust accordingly.

@rouault can you confirm that I've got this right?

> I see there is also 7.0 branch - should I port it there as well?

No, we will take care of that for you :-)

rouault:
Looks good to me. Merging

kbevers:
@rouault You may know this already, but in case you dont. The backport bot has problems with merge commits. It will only accept "squash merge" or "rebase and merge". Multiple commits in a PR doesn't work well either.

I've backported the PR manually.

PR #1970, opened 2020-02-24
fix sql in method createCoordinateReferenceSystem  of factory.cpp

yieryi:
obviously, data view "crs_view" have "auth_name" and "code" ,the two field value are not same;


rouault:
runWithCodeParam() has logic to automatically insert the authority name value as first subsitutable parameter, so this proposed change is incorrect.

yieryi:
> runWithCodeParam() has logic to automatically insert the authority name value as first subsitutable parameter, so this proposed change is incorrect.

thanks , i mistake it and just now finded.sorry for waste time.

PR #1978, opened 2020-02-24
CompoundCRS::create(): reject combinations of components not allowed by ISO 19111

rouault:


kbevers:
What are examples of combinations of components that ISO19111 doesn't allow?

rouault:
What is allowed is described at http://docs.opengeospatial.org/as/18-005r4/18-005r4.html#34
A few tests of non-allowed combinations in https://github.com/OSGeo/PROJ/pull/1978/files#diff-bb28a64a9f1fc812fd0042cfa57982b1R3348
Basically this PR is to limit @hobu creativity ;-)

kbevers:
> Basically this PR is to limit @hobu creativity ;-)

We probably need more of those :-)

hobu:
My creativity is driven by the fact that I've been seeing these out there. Mis-specified or not.

busstoptaktik:
@rouault said:

> What is allowed is described at http://docs.opengeospatial.org/as/18-005r4/18-005r4.html#34

I have reviewed the ISO version of this document a number of times, as it found its way through the ISO system. And while I may be able to imagine a somewhat valid use case, I'm particularly disturbed by the introduction of coordinates of arbitrarily high dimension, in section *9.3.4: Spatio-parametric compound coordinate reference system*:

> More than one parametric coordinate reference system may be included if these represent independent parametric quantities

If more than one is included, (e.g. pressure *and* temperature), I would probably prefer to have one of them as an attribute, rather than coordinate.

rouault:
> I would probably prefer to have one of them as an attribute, rather than coordinate.

Yes, probably. PROJ can parse from WKT & output as WKT such non-usual CRS, but it won't make much sense of them besides that. Your remark should probably be sent to other forums than this PR :-)

Anyway the aim of this PR was mostly to avoid creating things like Geographic 3D CRS + Vertical CRS, or Projected CRS + Geographic 3D CRS, that would confuse createOperations()

busstoptaktik:
> Your remark should probably be sent to other forums than this PR :-)

Unfortunately the proper forum is closed until next standard revision cycle :-). But actually, I had **14 remarks** accepted by the ISO 19111 editing committee during the "Draft International Standard" phase. One of these 14 was about the support of dual time coordinates - i.e. a subject closely related to multiple parametric coordinates.

So my comment here were more a matter of wondering how I could have missed the multiple parametric case. Reviewing standards makes you (occasionally) blind :-)

> Anyway the aim of this PR was mostly to avoid creating things like Geographic 3D CRS + Vertical CRS

The 3D case is clearly out, according to 19111. The question is just why it is so: For the "2 time coordinates" case, one of the examples given in the (non-normative) part of 19111 explains how to add a "gregorian date" coordinate to a coordinate tuple already including a "decimal year" coordinate, to support representation of the latter in terms of the former.

But according to 19111, doing something similar for a gravity related vertical CRS is not allowed. At least it would require you to model it as a 2D geographic + a "standalone ellipsoidal height", interpreted as a parametric vertical + a gravity related height, again interpreted as a parametric vertical. i.e. a 2D geographic system supplemented by 2 different parametric vertical systems.

To me, it makes "kind of" sense, but I'm not really sure how to make it fly in practice. So reaching full 19111 coverage and compliance is probably a task better left as an exercise for a more advanced civilization in a parallel universe :-)

busstoptaktik:
> My creativity is driven by the fact that I've been seeing these out there. Mis-specified or not.

@hobu, do you maintain a list of these "creative coordinate systems" ([CCS](https://en.wikipedia.org/wiki/Collective_Consciousness_Society), perhaps...)? Since they have been observed "in the wild", this may be relevant material for input to the standard revision cycle?

hobu:
I see this one sometimes

https://henzuru.io/v1/crs/EPSG/4978+3855/WKT
https://henzuru.io/v1/crs/EPSG/4978+3855

busstoptaktik:
Wow - a 3D cartesian, combined with a (mostly) gravity related vertical.

I understand what kind of LiDAR processing can lead to that kind of system, and I see the practical use case. This is definitely food for thought!

Fun fact: Both of these systems are independent of the ellipsoid specified, so in principle this system could be transformed to a WGS84 based geographic system, but using any ellipsoid wanted.

hobu:
> Fun fact: Both of these systems are independent of the ellipsoid specified, so in principle this system could be transformed to a WGS84 based geographic system, but using any ellipsoid wanted.

yes. We have unsuccessfully argued that data should be stored as ellipsoid with an epoch instead of this frankenref.

PR #1986, opened 2020-02-26
Map the Behrmann projection to cae when converting ESRI CRSes

nyalldawson:
Adds the capability to force a fixed value for a non-specified CRS component while mapping ESRI projections, and maps the Behrman projection to cae with lat_ts=30, lon_0=0

I had to add the new "is_fixed_value" property for ESRIParamMapping in order to differentiate parameters which are specified in the ESRI WKT definition and have an expected value Vs parameters which are NOT included in the ESRI WKT definition and have to have a fixed value (such as the Standard_Parallel_1 parameter for Behrman, which is not part of the CRS definition but has to have the value of 30 so that the proj string gets the required +lat_ts=30 parameter)


rouault:
A test would be welcome. Presumably as a new entry in the esriProjDefs[] array of test/unit/test_io.cpp

nyalldawson:
@rouault I've just added a bunch of missing tests

PR #1987, opened 2020-02-26
Support conversion of Flat_Polar_Quartic projection method

nyalldawson:
I realise this conflicts with https://github.com/OSGeo/PROJ/pull/1986 , but it's an easy fix (as opposed to #1986)


rouault:
@kbevers 7.1.0 material ?

kbevers:
> @kbevers 7.1.0 material ?

I think so, yes.

PR #1998, opened 2020-02-28
createOperations(): fix wrong pipeline generation with CRS that has +nadgrids= and +pm=

rouault:
Fixes issue reported at
https://lists.osgeo.org/pipermail/gdal-dev/2020-February/051749.html

The generated pipeline assumes that the input coordinates for the grid transformation
were related to the non-Greenwich based datum, so we must compensate for that and
add logic to go back to Greenwich.


rouault:
@kbevers This bug has presumably existed since 6.0.0 but it would be good to have it fixed in 7.0.0. Would you be willing to do a RC3 ? (if so, we'd merge the pending 7.0-targetted PRs).

kbevers:
> Would you be willing to do a RC3 ?

Not really. I've already started the vote for promotion to final. I'm sure you'll find another bug in RC3 that would be nice to fix as well. This can go in 7.0.1. We can do a bug fix release April 1st to catch up if you want.

rouault:
OK fair enough. Let's queue this for 7.0.1

PR #2010, opened 2020-03-03
Document Autotools and CMake configure options

mwtoews:
This is a draft document, as I'd like to know if there are any other build configure options that are worth documenting. Drop a comment here, and I'll add it in, otherwise I'll commit this in a week or so.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -308,10 +308,6 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
 
     Build :ref:`gie`, default ON.
 
-.. option:: BUILD_GMOCK=ON
-
-    Build the googlemock subproject, default ON.
-
 .. option:: BUILD_PROJ=ON
 
     Build :ref:`proj`, default ON.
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -200,6 +200,47 @@ The test suite requires that the proj-datumgrid package is installed in
...
+
+    C++ compiler flags
+
+See ``./configure --help`` for all options, here are a few key options:
+
+.. option:: --enable-lto
+
+    Enable compiler's Link Time Optimization, default disabled.
+
+.. option:: --disable-tiff
```
kbevers, line 235: Should also include `--with-curl`  here

kbevers (approved):
Looks good to me, thanks Mike

docs/source/install.rst
mwtoews: Added, along with CMake's `CURL_INCLUDE_DIR` and `CURL_LIBRARY` options.

docs/source/install.rst
```diff
@@ -200,6 +200,51 @@ The test suite requires that the proj-datumgrid package is installed in
 :envvar:`PROJ_LIB`.
 
 
+Autotoos configure options
```
rouault:
```suggestion
Autotools configure options
```

docs/source/install.rst
```diff
@@ -240,10 +285,96 @@ Tests are run with::
...
+    Path to an ``sqlite3`` or ``sqlite3.exe`` executable.
+
+.. option:: SQLITE3_INCLUDE_DIR
+
+    Path to an include directory with the ``sqlite3.h`` header file.
+
+.. option:: SQLITE3_LIBRARY
 
+    Path to a shared or static library file, such as ``sqlite3.dll``,
+    ``libsqlite3.so``, ``sqlite3.lib`` or other name.
```
rouault, line 398: for consistency TIFF_INCLUDE_DIR and TIFF_LIBRARY should also likely be documented.

docs/source/install.rst
```diff
@@ -200,6 +200,51 @@ The test suite requires that the proj-datumgrid package is installed in
...
+
+See ``./configure --help`` for all options, here are a few key options:
+
+.. option:: --enable-lto
+
+    Enable compiler's Link Time Optimization, default disabled.
+
+.. option:: --disable-tiff
+
+    TIFF support is enabled by default to read some grids, but this can
```
rouault: Perhaps a note should be added that if PROJ is used with the proj-data package, which is the nominal use case, TIFF support must be enabled.

docs/source/install.rst
mwtoews: oddly there were `TIFF_LIBRARY_RELEASE` and `TIFF_LIBRARY_DEBUG` variables

docs/source/install.rst
mwtoews: I've reworded both places. Are we using proj-data or PROJ-data? (i.e. the package name or the GitHub repo name?)

rouault (approved):

PR #2040, opened 2020-03-09
Map ESRI Transverse_Mercator_Complex to Transverse Mercator

rouault:
According to https://gis.stackexchange.com/questions/226679/complex-utm-projection
it is highly likely that Transverse_Mercator_Complex corresponds to our
extended/enhanced/Poder-Engsager transverse mercator method (etmerc), or something
similarly precise. So we can map that to the standard Transverse Mercator
method, since etmerc is used for it.


nyalldawson:
Nice!

kbevers (approved):

PR #2042, opened 2020-03-10
igh: force es=0 (spherical case) at the beginning of the setup function

rouault:
Should solve the issue raised in
https://lists.osgeo.org/pipermail/proj/2020-January/009285.html
and
https://lists.osgeo.org/pipermail/proj/2020-March/009463.html


rouault:
Closing. This pull request doesn't change anything to the behaviour. P->es is completely ignored when instanciating the child sinu and moll projections

PR #2043, opened 2020-03-10
Speed-up inverse approximate tmerc for coordinates with same northing

rouault:
This PR is on top of #2039 for convenience, but is conceputally independent
from it. Only the second commit "Speed-up inverse approximate tmerc for coordinates with same northing" is specific to this PR

When gdalwarp is used to reproject an image to a tmerc projection, the inverse
tmerc projection is actually used, and called by batch of coordinates belonging
to the same line in the output image, that is lines of same northing. Looking
at formulas, a lot of intermediate computations, including the use of hyperbolic
functions, only depend on the northing value. So cache them to potentially speed
up following transformations.

I couldn't get a significant value for the extra cost of doing that caching in
situations where it is not needed ('random' points). It is likely below 1%
(which is largely compensated by the improvements of #2039)

By modifying my bench program, to simulate 10 times the processing of  a
2000x2000 image (so 2000 coordinates of same northing being transformed
consecutively), I get a 40% speed-up over the ones of PR #2039

Some timings with gdalwarp on a 4000x4000 Byte single-channel image in EPSG:32631:

$ time OSR_USE_APPROX_TMERC=YES gdalwarp test.tif out.tif -t_srs EPSG:32631 -et 0 -overwrite

With this PR :   3.15s (30% improvement over PR #2039, 41% improvement over master)
With PR #2039:   4.54s (15% improvement over master)
Before PR #2039: 5.39s

### DIFF
--- a/src/projections/tmerc.cpp
+++ b/src/projections/tmerc.cpp
@@ -56,13 +56,34 @@ struct pj_opaque_exact {
 /* Constant for "exact" transverse mercator */
 #define PROJ_ETMERC_ORDER 6
 
+// Determine if we should try to provide optimized versions for the Fused Multiply Addition
+// Intel instruction set. We use GCC 6 __attribute__((target_clones("fma","default")))
+// mechanism for that, where the compiler builds a default version, and one that
+// uses FMA. And at runtimes it figures out automatically which version can be used
+// by the current CPU. This allows to create general purpose binaries.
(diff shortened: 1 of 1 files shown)


kbevers:
@rouault I lost track of this one, is it still valid after all your other tmerc improvements?

rouault:
it could be valid, but I'm unsure if I want to pursue on this

PR #2044, opened 2020-03-10
data/Makefile.am: remove bashism

rouault:


kbevers:
Are `checkbashisms` and `shellcheck` quick to run against our makefiles etc? Might be a good idea to include those checks in the CI infrastructure.

rouault:
> Are `checkbashisms` and `shellcheck` quick to run

They are quick to run, but a Makefile isn't a shell script, so false positive warnings (or legit warnings that we have no influence on because this isn't in the generated part of Makefile) are probably to be expected. And are actually found, like in
```$ checkbashisms data/Makefile 
script data/Makefile does not appear to have a #! interpreter line;
you may get strange results
possible bashism in data/Makefile line 144 ([^] should be [!]):
am__nobase_list = $(am__nobase_strip_setup); \
  for p in $$list; do echo "$$p $$p"; done | \
  sed "s| $$srcdirstrip/| |;"' / .*\//!s/ .*/ ./; s,\( .*\)/[^/]*$$,\1,' | \
  $(AWK) 'BEGIN { files["."] = "" } { files[$$2] = files[$$2] " " $$1; \
    if (++n[$$2] == $(am__install_max)) \
      { print $$2, files[$$2]; n[$$2] = 0; files[$$2] = "" } } \
    END { for (dir in files) print dir, files[dir] }'
```

kbevers:
Ah, I see. Yeah, that's not going to work. I guess we will just have to be careful about this then

PR #2052, opened 2020-03-11
pj_phi2(): speed-up computation (and thus inverse ellipsoidal Mercator and LCC)

rouault:
This does not change the numeric values returned by the method,
as far as I could see on a few samplings.

The tricks used save a call to sin() and atan() at each iteration.

This directly affects speed of inverse Mercator and LCC (among others),
in their ellipsoidal formulation.

Timings on inverse Mercator show a 31% speed-up at mid-latitudes
where pj_phi2() needs 5 iterations, and 24% at latitudes close to 0 or
90deg where it needs one iteration.

### DIFF
--- a/src/phi2.cpp
+++ b/src/phi2.cpp
@@ -9,7 +9,7 @@ static const double TOL = 1.0e-10;
 static const int N_ITER = 15;
 
 /*****************************************************************************/
-double pj_phi2(projCtx ctx, double ts, double e) {
+double pj_phi2(projCtx ctx, const double ts0, const double e) {
 /******************************************************************************
 Determine latitude angle phi-2.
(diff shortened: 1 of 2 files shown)


src/phi2.cpp
```diff
@@ -9,7 +9,7 @@ static const double TOL = 1.0e-10;
 static const int N_ITER = 15;
 
 /*****************************************************************************/
-double pj_phi2(projCtx ctx, double ts, double e) {
+double pj_phi2(projCtx ctx, const double ts0, double e) {
```
mwtoews: also `const double e`? (it's an input)

rouault: done

PR #2056, opened 2020-03-11
.cirrus.yml: disable ccache as cache uploading can be super slow

rouault:


rouault:
FreeBSD build in 11 minutes vs more than 50 befores :-) !

PR #2065, opened 2020-03-12
Add proj_get_units_from_database() (fixes #2004)

### LINKED ISSUE #2004: proj_list_db_units
Currently there is `proj_list_units` for the old PROJ units. However, there is a WKT based set of units listed in: https://github.com/OSGeo/PROJ/blob/master/data/sql/unit_of_measure.sql

Thoughts on adding something like `proj_list_db_units` or `proj_list_unit_of_measure` ?

rouault:


docs/source/development/reference/functions.rst
```diff
@@ -521,6 +521,9 @@ Lists
     entry of the returned array is a NULL-entry. The array is statically
     allocated and does not need to be freed after use.
 
+    Note: starting with PROJ 7.1, this function is deprecated by
```
kbevers, line 524: How about also adding `__attribute__ ((deprecated))` to the function definition in `proj.h` ?

kbevers (commented):
Looks good to me

docs/source/development/reference/functions.rst
rouault: Added

kbevers (approved):

snowman2:
Just had the thought - what about deprecating `proj_list_angular_units` with this change as well?

rouault:
> what about deprecating `proj_list_angular_units` with this change as well?

==> PR #2079

PR #2075, opened 2020-03-15
Align install.rst with current version

### LINKED ISSUE #2072: Update docs - installation still references PROJ4 5.0.1
The source code installation instructions here: https://proj.org/install.html still reference an older version of PROJ4, 5.0.1.  This is fine and I would advocate for maintaining install instructions for v5+, v6+, and the newer v7+ versions.  That is because the older v5 don't require certain libs like sqlite, the older datum data, and with 7 now require a new set of *tiff data.  However, because of the mismatch it can be confusing to determine which steps are relevant for which version.

In addition, at least I found it a bit unclear what you are supposed to do with the proj-data-1.0.tar.gz file.  I surmised that I should unpack that in the /data folder based on the approach in v5 and v6 but that wasnt clear.

[now that I submitted I realized this may not be a FR, maybe more a bug?  Wasnt sure since you dont have a docs request]

kbevers:
Using {PROJVERSION} and {PROJDATAVERSION} substitutes the macro for the current
version numbers of PROJ and PROJ-data.

Rephrased a few sections regarding grid packages.

Closes #2072

### DIFF
--- a/HOWTO-RELEASE
+++ b/HOWTO-RELEASE
@@ -63,7 +63,7 @@ Update the following files with the new ABI version number:
 Man pages are (mostly) automatically generated from the Sphinx docs. Follow
 these steps:
 
-    - Temporarily update version and dataversion in `docs/source/conf.py`
+    - Temporarily update `version` and `data_version` in `docs/source/conf.py`
     - run `scripts/update_man.sh`
     - Revert version number in `docs/source/conf.py`
 
--- a/HOWTO-RELEASE
+++ b/HOWTO-RELEASE
@@ -63,7 +63,7 @@ Update the following files with the new ABI version number:
 Man pages are (mostly) automatically generated from the Sphinx docs. Follow
 these steps:
 
-    - Temporarily update version in `docs/source/conf.py`
+    - Temporarily update version and dataversion in `docs/source/conf.py`
     - run `scripts/update_man.sh`
     - Revert version number in `docs/source/conf.py`
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -184,7 +184,7 @@ Tests are run with::
 
     make check
 
-With a succesful install of PROJ we can now install data files using the
+With a successful install of PROJ we can now install data files using the
 :program:`projsync` utility::
 
     projsync --system-directory
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -302,7 +302,7 @@ can be modified to suit the users needs. See :ref:`projsync` for more options.
 
     The use of :program:`projsync` requires that network support is enabled (the
     default option). If the resource files are not installed using
-    :program:`projsync` PROJ will attempt to fetch them automaticall when a
+    :program:`projsync` PROJ will attempt to fetch them automatically when a
     transformation needs a specific data file. This  requires that
     :envvar:`PROJ_NETWORK` is set to ``ON``.
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -289,7 +289,7 @@ Tests are run with::
 
     ctest
 
-With a succesful install of PROJ we can now install data files using the
+With a successful install of PROJ we can now install data files using the
 :program:`projsync` utility::
 
     projsync --system-directory
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -197,7 +197,7 @@ can be modified to suit the users needs. See :ref:`projsync` for more options.
 
     The use of :program:`projsync` requires that network support is enabled (the
     default option). If the resource files are not installed using
-    :program:`projsync` PROJ will attempt to fetch them automaticall when a
+    :program:`projsync` PROJ will attempt to fetch them automatically when a
     transformation needs a specific data file. This  requires that
     :envvar:`PROJ_NETWORK` is set to ``ON``.
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -26,13 +26,14 @@ from the ``conda-forge`` channel::
 
     conda install -c conda-forge proj
 
-Using ``conda`` you can also install the PROJ grid packages. Here's how to install
-the `proj-datumgrid-europe` package::
+Using ``conda`` you can also install the PROJ data package. Here's how to install
+the `proj-data` package::
 
-    conda install -c conda-forge proj-datumgrid-europe
+    conda install -c conda-forge proj-data
 
-Available is also the ``proj-datumgrid-north-america``, ``proj-datumgrid-oceania``
-and ``proj-datumgrid-world``.
+Available is also the legacy packages ``proj-datumgrid-europe``,
+``proj-datumgrid-north-america``, ``proj-datumgrid-oceania`` and
+``proj-datumgrid-world``.
 
 .. tip::
    Read more about the various datumgrid packages available :ref:`here<datumgrid>`.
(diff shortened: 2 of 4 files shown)


HOWTO-RELEASE
```diff
@@ -63,7 +63,7 @@ Update the following files with the new ABI version number:
 Man pages are (mostly) automatically generated from the Sphinx docs. Follow
 these steps:
 
-    - Temporarily update version in `docs/source/conf.py`
+    - Temporarily update version and dataversion in `docs/source/conf.py`
```
mwtoews:
```suggestion
    - Temporarily update `version` and `data_version` in `docs/source/conf.py`
```

docs/source/install.rst
```diff
@@ -175,17 +173,25 @@ If you are building from the git repository you have to first run::
...
+Alternatively the the data files can be installed manually in the directory
+pointed to by the :envvar:`PROJ_LIB` environment variable.
+
+.. note::
 
-The installation will automatically move the grid files to the correct location.
-Alternatively the grids can be installed manually in the directory pointed to
-by the :envvar:`PROJ_LIB` environment variable. The default location is
-``/usr/local/share/proj``.
+    Installing the proj-data package is optional when PROJ is build with
```
mwtoews: "is buid" -> "is built"

docs/source/install.rst
```diff
@@ -175,17 +173,25 @@ If you are building from the git repository you have to first run::
...
-Before proceeding with the installation we need to add the datum grids. Unzip
-the contents of the *proj-datumgrid* package into ``data/``::
+Before proceeding with the installation we can add the data files.
+Unzip the contents of the *proj-data* package into ``data/``::
+
+    unzip proj-data-{PROJDATAVERSION}.zip -d proj-{PROJVERSION}/data/
 
-    unzip proj-datumgrid-1.7.zip -d proj-5.0.1/data/
+The installation will automatically move the data files to the correct location.
+Alternatively the the data files can be installed manually in the directory
```
mwtoews: "the the" -> "the"

mwtoews (changes requested):
A few remarks to address, but otherwise all good!

docs/source/install.rst
busstoptaktik: I'm certain Kristian refers to Matt Johnson and his band here :-)

kbevers: No, but I recently got a new mechanical keyboard that i **really** like to type on, so I try to sneak in a few extra words now and then :-)

*clickety-clack*

rouault: What is currently implemented in master and 7.0.0 will *not* install the TIFF grids in $prefix/share/proj if you unzip them in data. What is only supported is to unzip them directly in the $prefix/share/proj
This could be done of course (just adding a "*.tif" in the install-data-local target of data/Makefile.am), but I don't think it is necessary to correlate PROJ build and grid installation as we did before.

kbevers: Good point, Even. I was actually wondering if that was still the case. I'll update the text to reflect that. Perhaps the "official recommendation" should be to install the grids using `projsync` after a succesfull build and install?

rouault: > Perhaps the "official recommendation" should be to install the grids using `projsync` after a succesfull build and install?

I've no strong preference between both options. The closest equivalent of unzipping to $prefix/share/proj would be "sudo projsync --system-directory" for Unix systems. Another difference is that with the unzipping method there's a clear versionning of what is installed, whereas with projsync that will be the state of the CDN at the time projsync is run.

kbevers: I think I'll just add the `--system-directory` call and then refer to the `projsync` doc page for further options that suit more specialized needs.

kbevers:
I updated the install howto for both autotools and cmake build, by changing the section on installing grids to using `projsync`. I added a note going into more details on the proj-data package, which then led to expanded the resource files page. I hope this describes current state of affairs better now.

rouault (approved):

docs/source/install.rst
```diff
@@ -196,8 +184,27 @@ Tests are run with::
...
+
+which will download all resource files currently available for PROJ. If less than
+the entire collection of resource files is needed the call to :program:`projsync`
+can be modified to suit the users needs. See :ref:`projsync` for more options.
+
+.. note::
+
+    The use of :program:`projsync` requires that network support is enabled (the
+    default option). If the resource files are not installed using
+    :program:`projsync` PROJ will attempt to fetch them automaticall when a
```
mwtoews:
```suggestion
    :program:`projsync` PROJ will attempt to fetch them automatically when a
```

docs/source/install.rst
```diff
@@ -196,8 +184,27 @@ Tests are run with::
 
     make check
 
-The test suite requires that the proj-datumgrid package is installed in
-:envvar:`PROJ_LIB`.
+With a succesful install of PROJ we can now install data files using the
```
mwtoews:
```suggestion
With a successful install of PROJ we can now install data files using the
```

docs/source/install.rst
```diff
@@ -282,8 +289,28 @@ Tests are run with::
...
+
+which will download all resource files currently available for PROJ. If less than
+the entire collection of resource files is needed the call to :program:`projsync`
+can be modified to suit the users needs. See :ref:`projsync` for more options.
+
+.. note::
+
+    The use of :program:`projsync` requires that network support is enabled (the
+    default option). If the resource files are not installed using
+    :program:`projsync` PROJ will attempt to fetch them automaticall when a
```
mwtoews:
```suggestion
    :program:`projsync` PROJ will attempt to fetch them automatically when a
```

docs/source/install.rst
```diff
@@ -282,8 +289,28 @@ Tests are run with::
 
     ctest
 
-The test suite requires that the proj-datumgrid package is installed
-in :envvar:`PROJ_LIB`.
+With a succesful install of PROJ we can now install data files using the
```
mwtoews:
```suggestion
With a successful install of PROJ we can now install data files using the
```

docs/source/resource_files.rst
```diff
@@ -123,10 +123,27 @@ all formats. Using GDAL for construction of new grids is recommended.
...
 -------------------------------------------------------------------------------
 
+proj-data
++++++++++
+
+The ``proj-data`` package is a collection of all the resource files that are
+freely available for use with PROJ. The package is maintained on
+`GitHub <https://github.com/OSGeo/PROJ-data>`_ and the contents of the package
+are show-cased on the `PROJ CDN <https://cdn.proj.org/>`_. The contents of the
+package can be installed using the :program:`projsync` package or by downlading
```
mwtoews:
```suggestion
package can be installed using the :program:`projsync` package or by downloading
```

docs/source/resource_files.rst
```diff
@@ -123,10 +123,27 @@ all formats. Using GDAL for construction of new grids is recommended.
...
+
 proj-datumgrid
 ++++++++++++++
 
-For a functioning PROJ, installation of the
+.. note::
+
+    The packages described below can be used with PROJ 7 and later but are
+    primarily meant to be used with PROJ 6 and earlier versions. The proj-datumgrid
+    series of packages are not maintained any more and are only kept available for
```
mwtoews:
```suggestion
    series of packages are not maintained anymore and are only kept available for
```

mwtoews (changes requested):

Issue #2070, opened 2020-03-12
proj.pc change between 6.3.1 and 7.0.0 breaks packaging


devrimgunduz:
Hi,

In Proj 6.3,1, the proj.pc includes this line:

`datadir=${prefix}/share/proj`

However, in 7.0.0, it has:

`datadir=${datarootdir}/proj`

...and this breaks SLES-12 builds, because datarootdir is not defined in .pc file.

Any chance we can revert to 6.3.1 version, please?

Thanks!

mwtoews:
Does #2067 resolve the issue?

devrimgunduz:
In theory, yes -- but on SLES-12, when I apply the patch I get this, because automake is 1.13 (or 1.15) there

> + /usr/bin/make -j3
/pghome/pgsql/rpm12testing/BUILD/proj-7.0.0/missing: line 81: automake-1.16: command not found
WARNING: 'automake-1.16' is missing on your system.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <https://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <https://www.gnu.org/software/autoconf>
         <https://www.gnu.org/software/m4/>
         <https://www.perl.org/>

Any ideas how can I skip this?

Regards, Devrim

mwtoews:
With a pristine proj-7.0.0 source with [this patch](https://patch-diff.githubusercontent.com/raw/OSGeo/PROJ/pull/2067.diff), I get this error too on Ubuntu 16.04.

Essentially, you'd need to re-run `./autogen.sh`, which is only part of the repo and not source distribution. And to get a minimal patch, the same versions of various autotools need to be paired (e.g. I see automake 1.16.1 was used). I don't have these tool versions, so can't make a clean looking patch for the source distribution.

devrimgunduz:
Thanks @mwtoews . I'm closing this bug  (found a way to update automake to 1.6 on SLES 12..

devrimgunduz:
Hi,

What is the timeline for 7.0.1? I cannot build 7.0.0 on RHEL  7 (which majority of users run) after applying this patch (automake 1.13 is available there. Need 7.0.1 so that I can push it to the community RPM repositories.

Thanks!

Regards, Devrim

mwtoews:
[...]

Issue #2064, opened 2020-03-12
Running pkg-config --list-all with PROJ installed causes segfault


volleio:
#### Example of problem
Found this issue [debugging the install of RMagick](https://github.com/rmagick/rmagick/issues/1153#issuecomment-598203790).

Installing PROJ and then running
```c
pkg-config --debug --list-all
```
causes a segfault while `Parsing package file '/usr/local/lib/pkgconfig/proj.pc'`. 
Running
```c
pkg-config --exists --print-errors "proj"
```
results in `Variable 'datarootdir' not defined in '/usr/local/lib/pkgconfig/proj.pc'`. Removing /usr/local/lib/pkgconfig/proj.pc (renaming it to proj.pc.backup) and re-running the `pkg-config --list-all` command succeeds without segfault.

#### Problem description
PROJ shouldn't affect `pkg-config`'s ability to list all packages. 

#### Expected Output
`pkg-config --list-all` should output all of the installed packages in the `pkg-config` pc paths. 

#### Environment Information
 - PROJ version 7.0.0
 - MacOS Catalina 10.15.3 (19D76)
 - pkg-config version 0.29.2

#### Installation method
 - brew install proj

hobu:
> PROJ shouldn't affect `pkg-config`'s ability to list all packages.

Can you past the proj.pc file here too?

`pkg-config` also shouldn't segfault ever 😄, but that's `pkg-config`'s issue

volleio:
Yep, here is my proj.pc contents! 
```c
prefix=/usr/local/Cellar/proj/7.0.0
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
datadir=${datarootdir}/proj

Name: PROJ
Description: Coordinate transformation software library
Requires:
Version: 7.0.0
Libs: -L${libdir} -lproj
Libs.private: -lsqlite3 -lstdc++
Cflags: -I${includedir}
```

hobu:
hmph, that doesn't look that interesting.

mwtoews:
@volleio I'm unable to recreate the segfault under Ubuntu 18.04, but let's see if #2067 can resolve this issue. Could you try this `proj.pc` file content (the new added line is `datarootdir=${prefix}/share`):
```
prefix=/usr/local/Cellar/proj/7.0.0
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
datarootdir=${prefix}/share
[...]

PR #2077, opened 2020-03-17
Add post-install checks for CMake and Autotools/pkg-config

mwtoews:
This PR is to ensure that PROJ can be used to build applications using CMake and Autotools/pkg-config. This idea was briefly discussed in #1910, and I could expect this to be extended further.

The tests use a simple C program 'testappversion' that includes proj.h and links to libproj.

CMake's `find_package(PROJ)` and `find_package(PROJ4)` are used and the program's linked libraries and output is tested. This hopefully prevents any unexpected changes to documented behavior.

Similarly, pkg-config is used within a simple Makefile, and the program is tested. This should also hopefully prevent issues like #2064 / #2070 .

Other changes in this PR (not exactly related, but are in the same edited files) are:
* Use `export MAKEFLAGS="-j ${NPROC}"` for various `make` calls
* Replace `$TOP_DIR` with `$TRAVIS_BUILD_DIR` (same thing)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,7 +24,7 @@ build_script:
   # update vcpkg
   - cmd: |
       cd "C:\Tools\vcpkg"
-      git pull
+      git pull > nul
       .\bootstrap-vcpkg.bat
       cd %APPVEYOR_BUILD_FOLDER%
   - vcpkg install sqlite3[core,tool]:"%platform%"-windows
@@ -52,6 +52,7 @@ test_script:
   - cd %PROJ_BUILD%
   - ctest -V -C Release
   - set PATH=%PROJ_DIR%\bin;%PATH%
+  - call %APPVEYOR_BUILD_FOLDER%\test\postinstall\test_cmake.bat %PROJ_DIR%
   - proj
 
 deploy: off
--- a/test/postinstall/test_cmake.bat
+++ b/test/postinstall/test_cmake.bat
@@ -1,0 +1,40 @@
+@echo off
+:: Post-install tests with CMake
+::
+:: First required argument is the installed prefix, which
+:: is used to set CMAKE_PREFIX_PATH
+
+echo Running post-install tests with CMake
+
+set CMAKE_PREFIX_PATH=%1
+if not defined CMAKE_PREFIX_PATH (
+    echo First positional argument CMAKE_PREFIX_PATH required
+    exit /B 1
+)
+
+echo CMAKE_PREFIX_PATH=%CMAKE_PREFIX_PATH%
+
+cd %~dp0
+
+cd testappprojinfo
+del /f /q build 2> nul
+
+:: Check CMake project name PROJ
+md build
+cd build
+cmake -G "%VS_FULL%" -DCMAKE_PREFIX_PATH=%CMAKE_PREFIX_PATH% -DUSE_PROJ_NAME=PROJ ..  || exit /B 2
+cmake --build . --config Release || exit /B 3
+ctest --build-config Release -VV || exit /B 4
+cd ..
+del /f /q build
+
+:: Check legacy CMake project name PROJ4
+md build
+cd build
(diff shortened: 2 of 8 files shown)


mwtoews:
There are three different Travis CI failures that I'd appreciate any help on.

With [3115.2](https://travis-ci.com/github/OSGeo/PROJ/jobs/298775355)
> /usr/bin/ld: skipping incompatible /tmp/proj_autoconf_install_from_dist_all/lib/libproj.so when searching for -lproj
> /usr/bin/ld: skipping incompatible /tmp/proj_autoconf_install_from_dist_all/lib/libproj.a when searching for -lproj

what makes these libraries incompatible?

With [3115.3](https://travis-ci.com/github/OSGeo/PROJ/jobs/298775356)
> /tmp/proj_autoconf_install_from_dist_all/lib/libproj.so: undefined reference to '__asan_stack_free_9'
> ...
> /tmp/proj_autoconf_install_from_dist_all/lib/libproj.so: undefined reference to '__asan_report_exp_store8'
> clang-7: error: linker command failed with exit code 1 (use -v to see invocation)

something is amiss with clang linking, I don't use this compiler so have no idea.

With [3115.4](https://travis-ci.com/github/OSGeo/PROJ/jobs/298775357) on osx:
> /Users/travis/build/OSGeo/PROJ/travis/postinstall/test_pkg-config.sh: line 33: ldd: command not found

Is `otool -L` a good tool for this? I haven't seen a Mac in years.

rouault:
> what makes these libraries incompatible?

This is a 32 bit build using the -m32 flag of a native 64 bit compiler. Your test app must also use -m32

> With 3115.3

This is a ASAN (Adress Sanitizer) enabled libproj build. Your test app must also link with -fsanitize=address

> Is otool -L a good tool for this?

Yes GDAL uses it for similar purposes

mwtoews:
After several changes from the first pass, this is now ready for review.

Here is now what's tested:
* `linux`, both pkg-config and CMake are tested
* `osx` (macOS), both pkg-config and CMake. The odd thing about linking found thru `otool -L` is that the pkg-config-built version shows the app links thru `/tmp/proj_autoconf_install_from_dist_all/lib/libproj` while the CMake-built version links thru `@rpath/libproj`; not sure how to expand `@rpath`
* Windows via AppVeyor, only CMake tests with static/shared libraries. Tests using the shared libraries need `PATH` with the installed `bin` directory to load `geos_7_1.dll`, which is slightly odd. This is why the tests are run after modifying `PATH` and not before.

Not all Travis CI builds are included, as this challenge needs a bit more planning on what flags are expected to be passed and/or what flags/properties should be stored for each installed configuration. These configurations are skipped for now:
* `linux_gcc8`: not sure the best way to handle `-m32` for i386 builds using CMake or pkg-config approaches (e.g. should the user know this library was built this way and apply these CFLAGS to the app? Or should pkg-config and/or CMake handle this?)
* `linux_clang`: again, not sure the best way to handle `LDFLAGS=-fsanitize=address` for similar reasons.
* `mingw32`: libraries need to be linked to the testapp somehow, more work needed here.

rouault:
>     * should the user know this library was built this way and apply these CFLAGS to the app? Or should pkg-config and/or CMake handle this?

no, this is just an artifact of using a 64bit native compiler / OS to do 32-bit builds. I'd say that you should probably just skip your tests on such odd configuration. This is mostly an artifact of the CI.

> again, not sure the best way to handle LDFLAGS=-fsanitize=address for similar reasons.

Same. Skip your new tests

> mingw32: libraries need to be linked to the testapp somehow, more work needed here.

Same

Basically we just need to make sure that the cmake stuff works on 1 nominal Linux config, 1 Mac and 1 Windows.

rouault:
> Basically we just need to make sure that the cmake stuff works on 1 nominal Linux config, 1 Mac and 1 Windows.

OK, looking actually at your code, that's what you did. I'd say to just remove the TODO comments, and that'll be ready for merging

mwtoews:
> OK, looking actually at your code, that's what you did. I'd say to just remove the TODO comments, and that'll be ready for merging

Done, and thanks for taking a look!

PR #2085, opened 2020-03-19
proj_get_suggested_operation(): fix when there are only operations without area of use

rouault:


rouault:
This is 7.1.0 material only. proj_get_suggested_operation() is master only

kbevers:
Okay, sorry. Looked like something that could be back-ported, I figured you just forgot to set a milestone.

rouault:
No problem. It cannot hurt people double checking what I do, especially at those times

PR #2097, opened 2020-03-26
Use jtsk03-jtsk horizontal grid from CDN

wonder-sk:
Following up on https://github.com/OSGeo/PROJ-data/pull/17

Tried with projinfo and cct with PROJ_NETWORK=ON and it seems to work!

Shall I also update 6.3 branch and add the original NTv2 grid to proj-datumgrid repo or that one is already considered as abandoned?


rouault:
> or that one is already considered as abandoned?

yes, it is no longer open for new datasets.

PR #2100, opened 2020-03-27
CMake: prefer to use use PROJ_SOURCE_DIR and PROJ_BINARY_DIR

### LINKED ISSUE #2094: Use PROJ_SOURCE_DIR instead of CMAKE_SOURCE_DIR in CMake files
#### Problem description

If wrapped into another CMakeLists, CMAKE_SOURCE_DIR does not point to what PROJ expect.
For example `include_directories(${CMAKE_SOURCE_DIR}/include)` in lib_proj.cmake would be incorrect in this case.

mwtoews:
PROJ_SOURCE_DIR and PROJ_BINARY_DIR are absolute paths to the source and binary build trees for PROJ, whereas CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR refer to the top-level CMakeLists.txt, which could be a different project. Similarly, PROJECT_SOURCE_DIR and PROJECT_BINARY_DIR are relative to the most recent project() command. It's just more direct to make these consistently PROJ_SOURCE_DIR and PROJ_BINARY_DIR.

Other items covered here:
* Set properties with a list variable rather than a long line
* Correction to 'proj_test_set_properties' function, rename ENVIRONMENT:
  - PROJ_IGNORE_USER_WRITABLE_DIRECTORY     (ignored)
  - PROJ_SKIP_READ_USER_WRITABLE_DIRECTORY  (used by filemanager.cpp)

Closes #2094

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,9 +225,9 @@ endif()
 # top of the build tree rather than in hard-to-find leaf
 # directories. This simplifies manual testing and the use of the build
 # tree rather than installed PROJ libraries.
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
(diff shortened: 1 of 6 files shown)


kbevers (approved):
Looks good to me!

PR #2126, opened 2020-04-03
createFromUserInput(): allow compound CRS with the 2 parts given by names, e.g. 'WGS 84 + EGM96 height'

rouault:


kbevers:
This is great! As recently as today I was trying to do exactly this, good job on the mind reading :)

PR #2133, opened 2020-04-04
appveyor.yml: reduce to 2 configurations (still testing the same number of variables)

rouault:
Of course, the aim is to have faster builds


mwtoews (approved):
Looks good to me, as it covers the basic configurations.

Issue #2062, opened 2020-03-12
Proj 7.0.0 build error on SLES 12


devrimgunduz:
Hi,

Proj 7.0.0 fails to build on SLES-12. This is GCC 4.8.5. Can you please take a look? The error is below.

Thanks!

```
+ /usr/bin/make -j3
grids.cpp: In constructor 'osgeo::proj::Grid::Grid(const string&, int, int, const osgeo::proj::ExtentAndRes&)':
grids.cpp:100:40: warning: declaration of 'name' shadows a member of 'this' [-Wshadow]
            const ExtentAndRes &extentIn)
                                        ^
grids.cpp: In constructor 'osgeo::proj::GTiffGrid::GTiffGrid(PJ_CONTEXT*, TIFF*, osgeo::proj::BlockCache&, osgeo::proj::File*, uint32, const string&, int, int, const osgeo::proj::ExtentAndRes&, osgeo::proj::TIFFDataType, uint16, uint16, bool)':
grids.cpp:479:25: warning: declaration of 'name' shadows a member of 'this' [-Wshadow]
             std::string name;
                         ^
grids.cpp: In static member function 'static std::unique_ptr<osgeo::proj::VerticalShiftGridSet> osgeo::proj::VerticalShiftGridSet::open(PJ_CONTEXT*, const string&)':
grids.cpp:1379:16: error: cannot bind 'std::unique_ptr<osgeo::proj::GTiffVGridShiftSet>' lvalue to 'std::unique_ptr<osgeo::proj::GTiffVGridShiftSet>&&'
         return set;
                ^
In file included from /usr/include/c++/4.8/memory:81:0,
                 from grids.hpp:31,
                 from grids.cpp:34:
/usr/include/c++/4.8/bits/unique_ptr.h:169:2: error:   initializing argument 1 of 'std::unique_ptr<_Tp, _Dp>::unique_ptr(std::unique_ptr<_Up, _Ep>&&) [with _Up = osgeo::proj::GTiffVGridShiftSet; _Ep = std::default_delete<osgeo::proj::GTiffVGridShiftSet>; <template-parameter-2-3> = void; _Tp = osgeo::proj::VerticalShiftGridSet; _Dp = std::default_delete<osgeo::proj::VerticalShiftGridSet>]'
  unique_ptr(unique_ptr<_Up, _Ep>&& __u) noexcept
  ^
[...]

PR #2138, opened 2020-04-07
Test GCC 4.8 and CMake 3.9; resolve a few shadowed declaration warnings

mwtoews:
There is a gap in the Travis CI testing for GCC 4.8 and/or the current minimum required version of CMake (3.9). This PR <s>adds yet another job to the matrix</s> modifies the existing `linux_gcc` job.

Testing with `g++-4.8` will avoid issues like #2062 and compatibility with building with [manylinux1](https://github.com/pypa/manylinux), for example see [pyproj#574](https://github.com/pyproj4/pyproj/issues/574).

A few shadowed declarations were resolved to pass tests with `-Werror`, but I'd appreciate a review of my C++ edits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,13 @@ matrix:
         - docker
       env:
         - BUILD_NAME=linux_gcc
-        - DETAILS="linux, gcc"
+        - DETAILS="linux, gcc-4.8"
+        - CC=gcc-4.8
+        - CXX=g++-4.8
+      addons:
+        apt:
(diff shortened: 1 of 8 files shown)


rouault (approved):
Maybe to avoid too many CI targets we could just modify the first one linux_gcc to use gcc 4.8 ? That's a minor point though

mwtoews:
Good suggestion, I'll amend this PR to see how that goes.

PR #2145, opened 2020-04-11
Inverse cart: speed-up computation by 33%

rouault:
Saves 2 sincos() and 1 atan2() calls.

With the following bench
```
 #include "proj.h"
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>

{
    if( argc != 2 ) {
        fprintf(stderr, "Usage: bench_inv_cart fwd/inv\n");
        exit(1);
    }
    PJ* p = proj_create(0, "+proj=cart");
    const int dir = strcmp(argv[1], "inv") == 0 ? PJ_INV : PJ_FWD;
    PJ_COORD coord;
    if( dir == PJ_FWD )
    {
        coord.xyz.x = 3.14159/2;
        coord.xyz.y = 3.14159/2;
        coord.xyz.z = 100;
    }
    else
    {
        coord.xyz.x = 3e6;
        coord.xyz.y = 3e6;
        coord.xyz.z = 3e6;
    }
    for(int i = 0; i < 10* 1024 * 1024; i++ )
    {
        proj_trans(p, dir, coord);
    }

    proj_destroy(p);
    return 0;
}
```

On Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
Time before: 2.37s
Time after:  1.57s


busstoptaktik (commented):
Royally awesome this, @rouault! I had done the copysign-to-cart.z part in my local branch, but it (obviously) didn't make much speed difference - your optimizations really makes a significant difference in the speed of an extremely important geodetic operation. Awesome!!

PR #2146, opened 2020-04-11
Remove old geocent.h/.cpp code

rouault:
Last user, apart from transform.cpp, was the SCH projection. Modify it
to use cart instead.

And move content of geocent.h and .cpp into transform.cpp directly, so
that it can be later wiped up easily.


src/projections/sch.cpp
```diff
@@ -48,62 +47,49 @@ struct pj_opaque {
...
 
     /* Apply rotation */
-    pxyz[0] = Q->transMat[0] * temp[0] + Q->transMat[1] * temp[1] + Q->transMat[2] * temp[2];
-    pxyz[1] = Q->transMat[3] * temp[0] + Q->transMat[4] * temp[1] + Q->transMat[5] * temp[2];
-    pxyz[2] = Q->transMat[6] * temp[0] + Q->transMat[7] * temp[1] + Q->transMat[8] * temp[2];
+    double pxyz[] = {
+        Q->transMat[0] * temp[0] + Q->transMat[1] * temp[1] + Q->transMat[2] * temp[2],
+        Q->transMat[3] * temp[0] + Q->transMat[4] * temp[1] + Q->transMat[5] * temp[2],
+        Q->transMat[6] * temp[0] + Q->transMat[7] * temp[1] + Q->transMat[8] * temp[2]
+    };
```
kbevers: I think it would be better to simply use `xyz` here instead of `temp`. `temp` is a bit misleading as it doesn't really say what it is, except that it is temporary.  I can see why it was simplest to just keep using the temp variable from the old code but this is a good opportunity to make the code a bit easier to read. If you still want to keep the vector-feel of using indices `xyz.v[0]` (etc) can be used.

This of course also applies in the forward function below.

kbevers (approved):
Looks good to me. I have a suggestion for a slight improvement of readability that can be implemented if you agree.

src/projections/sch.cpp
rouault: yes, I was considering that too, but wanted to make the port to the cart API more obvious. I've added a second commit that simplifies the code

PR #2157, opened 2020-04-15
Adams Square II: map ESRI WKT to PROJ string, and implement iterative inverse method

rouault:
- [X] Tests added


rouault:
Inverse projection is needed for gdalwarp for example:
```
$ gdalwarp -r cubic ../autotest/gdrivers/data/small_world.tif /tmp/out.png -t_srs "+proj=adams_ws2" -overwrite
Creating output file that is 447P x 447L.
0...10...20... 30...40...50../tmp.60...70...80../out.90...100 - done.
```
out

nyalldawson:
Nice -- I was **just** building master to see if I could map these new ones! Did you look into mapping Peirce_Quincuncial yet?

src/iso19111/coordinateoperation.cpp
```diff
@@ -6183,6 +6183,38 @@ void Conversion::_exportToPROJString(
         formatter->addParam("o_lat_p", -southPoleLat);
         formatter->addParam("lon_0", southPoleLon);
         bConversionDone = true;
+    } else if (ci_equal(methodName, "Adams_Square_II")) {
+        // A bit hacky. Map directly from ESRI method and parameter names
+        // as we haven't defined a OGC-like mapping, and we have a fairly
+        // number of restrictions on the accepted values.
```
kbevers: Could you please rephrase this? It is unclear what you mean, I think there's either missing a word or a few too many

src/projections/adams.cpp
```diff
@@ -158,24 +160,122 @@ static PJ_XY forward(PJ_LP lp, PJ *P) {
...
-        double temp = xy.x;
+        const double temp = xy.x;
         xy.x = RSQRT2 * (xy.x - xy.y);
         xy.y = RSQRT2 * (temp + xy.y);
     }
 
     return xy;
 }
 
+static PJ_LP inverse_with_newton_raphson(PJ_XY xy,
```
kbevers, line 181: Just thinking out loud here. This would be a nice thing to do for all projections without an inverse. We could make this a fallback in `pj_inv()` in the case no proper inverse function is registered with the projection. I think all that it would take is a mechanism to provide the initial guess. There may be ways to automate that as well, for instance by caching previous inverse coordinates and other hacks.

test/gie/more_builtins.gie
```diff
@@ -866,5 +866,54 @@ accept      4.5        52.5        0
 expect      5.875      55.375      0
 -------------------------------------------------------------------------------
 
+-------------------------------------------------------------------------------
+Some tests from adams_ws2
```
kbevers: Why not put these in `test/gie/adams_ws2.gie`?

src/iso19111/coordinateoperation.cpp
```diff
@@ -6183,6 +6183,38 @@ void Conversion::_exportToPROJString(
...
+            "Longitude_Of_Center", common::UnitOfMeasure::DEGREE);
+        const double latitudeOfCenter = parameterValueNumeric(
+            "Latitude_Of_Center", common::UnitOfMeasure::DEGREE);
+        const double XYPlaneRotation = parameterValueNumeric(
+            "XY_Plane_Rotation", common::UnitOfMeasure::DEGREE);
+        if (scaleFactor != 1.0 || azimuth != 0.0 || latitudeOfCenter != 0.0 ||
+            XYPlaneRotation != 0.0) {
+            throw io::FormattingException("Unsupported value for one or "
+                                          "several parameters of "
+                                          "Adams_Square_II");
```
kbevers, line 6215: We would need to implement both the azimuth and XYPlaneRotation parameters to properly map the Spilhaus projection, right?

kbevers (commented):
Looks good to me. I like the generic approach to the inverse adams_ws2, good thinking!

test/gie/more_builtins.gie
rouault: Because of "This gie file was automatically generated using libproject"

src/projections/adams.cpp
rouault: yes, this has the potential to be generic but a number of error thresholds in it have been tuned to make it work with adams_ws2 (for example it somehow assumes that (0,0) maps to (0,0), etc etc) . So for now, it is well here and we can potentially try to reuse it in a more general case if we need

src/iso19111/coordinateoperation.cpp
rouault: hum, no, all words I intended are here :-) I'll try to make it clearer if I can...

src/iso19111/coordinateoperation.cpp
rouault: and Latitude_Of_Center well

src/projections/adams.cpp
kbevers: I have a bit of extra time on my hands next week. I think I will give it a first go and try to uncover some of the pitfalls. If we can make it work we can increase the amount of projections with an inverse quite significantly!

test/gie/more_builtins.gie
kbevers: Oh right, feel free to ignore that. The file can't be reproduced anyway since I generated semi-random numbers for the input coordinates. I just wanted to highlight that the test coordinates didn't come from some "official" source. Maybe I should have phrased that line differently.

rouault: ok, I'll change its header and move my tests there

rouault:
> Did you look into mapping Peirce_Quincuncial yet?

no. Not sure I'll do

busstoptaktik:
Hooray! This is awesome @rouault!

@kbevers: I see from your comments that you consider making this a generic fill-in for one-way projections, which is even more awesome. If you should succeed, consider making the core derivative computations part of `derivs.cpp/pj_derivs()`.

Also, the material covered in `factors.cpp` may give some insight in how perversely deform the projected space may be locally.

PR #2159, opened 2020-04-17
AppVeyor: use Ninja generator, cache vcpkg installed packages

mwtoews:
This is an attempt to speed up AppVeyor testing by using a [Ninja](https://ninja-build.org/) generator and by caching `C:\Tools\vcpkg\installed\` as suggested by @mloskot

Vcpkg packages don't seem to change much, so if an upgrade happens upstream, it won't be updated without deleting the cache. The build cache is deleted when appveyor.yml is modified, forcing the packages to be re-installed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,41 +7,46 @@ environment:
 
 # VS 2015
   - platform: x86
-    VS_VERSION: Visual Studio 14
     APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
     BUILD_SHARED_LIBS: OFF
 
 # VS 2017
   - platform: x64
-    VS_VERSION: Visual Studio 15
     APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
     BUILD_SHARED_LIBS: ON
 
 shallow_clone: true
 
+cache:
+  - C:\Tools\vcpkg\installed\ -> appveyor.yml
+
 build_script:
-  - echo build_script
-  # update vcpkg
-  - cmd: |
-      cd "C:\Tools\vcpkg"
-      git pull > nul
-      .\bootstrap-vcpkg.bat
+  - set VCPKG_INSTALLED=C:\Tools\vcpkg\installed\%platform%-windows
+  # If cached directory does not exist, update vcpkg and install dependencies
+  - if not exist %VCPKG_INSTALLED%\bin (
+      cd "C:\Tools\vcpkg" &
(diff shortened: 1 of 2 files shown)


kbevers:
It seems to have shaved some time already. Nice. I guess that for this first go we aren't seing the benefits of the cache yet?

mloskot:
Yes, for next builds the cache should be restored

image

mwtoews:
I'm guessing that the settings for this AppVeyor configuration has the option "Save build cache in Pull Requests" disabled, which I think is default. I suggest it should be enabled (I can't do this as I don't have any privileges to the OSGeo account there). And yeah, what @mloskot mentioned, there are helpful messages at the top/bottom that say if the cache is restored or updated.

rouault:
> I suggest it should be enabled

I've just done it

mwtoews:
I now see the last three lines to confirm the enabled option:
> Updating build cache...
> Cache 'C:\Tools\vcpkg\installed' - Updated
> Build success

rouault:
I restarted the build after enabling the option, but the new build is slower than the initial one (12min31 vs 9min22s, but that assumes that performance would be perfectly reproducible which is hard to assume on likely virtualized machines...) and from the log at https://ci.appveyor.com/project/OSGeo/proj/builds/32249817/job/9h1y5f0lmk536k7x it seems that the caching didn't work/wasn't detected since vcpkg stuff is redone

kbevers:
> it seems that the caching didn't work/wasn't detected since vcpkg stuff is redone

you probably have to restart it once more to see the effect. If I understood correctly it is only after you have changed the settings that the cache will be saved on pull requests.

mwtoews:
Caching should work on a next commit that doesn't modify `appveyor.yml`. And looking thru the [build history](https://ci.appveyor.com/project/OSGeo/proj/history), you can see the times range from 12-19 minutes per run.

mwtoews:
I did a second commit (which I'll remove before merging) that properly restored the cache, with ~ 5-8 min runs ([link to build](https://ci.appveyor.com/project/OSGeo/proj/builds/32250594)).

kbevers:
This is starting to look really good!

mwtoews:
Should this be backported to the 7.0 branch?

kbevers:
Yeah, we may as well. There's still coming in a bunch of PR's on that branch so the speed up is a big benefit. I made a merge commit here so I'll just cherry-pick the changes into 7.0. Coming right up!

PR #2162, opened 2020-04-18
Moved proj_context_get_url_endpoint & proj_context_get_user_writable_directory to proj.h

### LINKED ISSUE #2028: Download grids by bounding box or region name
I remember seeing something about adding functionality to pre-fetch grids. Potentially by region name or bounding box. Currently I see this for manually downloading files: https://github.com/OSGeo/PROJ/blob/38ec5e662a74d40e02e38dc5dca553c3ecb04356/src/proj.h#L510

Are there plans to add functionality to download grids by a bounding box or by region name?
Also, what about a download all grids method?

snowman2:
- [x] Closes #2028
 
After some thought, I decided that being consistent with what PROJ is using internally where possible will make things simpler.

### DIFF
--- a/src/apps/projinfo.cpp
+++ b/src/apps/projinfo.cpp
@@ -1074,7 +1074,7 @@ int main(int argc, char **argv) {
 #ifdef CURL_ENABLED
             if (proj_context_is_network_enabled(nullptr)) {
                 std::cout << "Status: enabled" << std::endl;
-                std::cout << "URL: " << std::string(proj_context_get_url_endpoint(nullptr))
+                std::cout << "URL: " << proj_context_get_url_endpoint(nullptr)
                           << std::endl;
             } else {
                 std::cout << "Status: disabled" << std::endl;
--- a/src/apps/projinfo.cpp
+++ b/src/apps/projinfo.cpp
@@ -1074,7 +1074,7 @@ int main(int argc, char **argv) {
 #ifdef CURL_ENABLED
             if (proj_context_is_network_enabled(nullptr)) {
                 std::cout << "Status: enabled" << std::endl;
-                std::cout << "URL: " << pj_context_get_url_endpoint(nullptr)
+                std::cout << "URL: " << std::string(proj_context_get_url_endpoint(nullptr))
                           << std::endl;
             } else {
                 std::cout << "Status: disabled" << std::endl;
--- a/src/apps/projsync.cpp
+++ b/src/apps/projsync.cpp
@@ -111,7 +111,7 @@ int main(int argc, char *argv[]) {
     auto ctx = pj_get_default_ctx();
 
     std::string targetDir;
-    std::string endpoint(pj_context_get_url_endpoint(ctx));
+    std::string endpoint(proj_context_get_url_endpoint(ctx));
     const std::string geojsonFile("files.geojson");
     std::string queriedSourceId;
     std::string queriedAreaOfUse;
@@ -224,7 +224,7 @@ int main(int argc, char *argv[]) {
     }
 
     if (targetDir.empty()) {
-        targetDir = pj_context_get_user_writable_directory(ctx, true);
+        targetDir = std::string(proj_context_get_user_writable_directory(ctx, true));
     } else {
         if (targetDir.back() == '/') {
             targetDir.resize(targetDir.size() - 1);
--- a/src/apps/projsync.cpp
+++ b/src/apps/projsync.cpp
@@ -224,7 +224,7 @@ int main(int argc, char *argv[]) {
     }
 
     if (targetDir.empty()) {
-        targetDir = std::string(proj_context_get_user_writable_directory(ctx, true));
+        targetDir = proj_context_get_user_writable_directory(ctx, true);
     } else {
         if (targetDir.back() == '/') {
             targetDir.resize(targetDir.size() - 1);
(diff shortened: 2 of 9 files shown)


src/filemanager.cpp
```diff
@@ -1182,8 +1183,16 @@ static void CreateDirectoryRecursively(PJ_CONTEXT *ctx,
...
-                                                   bool create) {
+/** Get the PROJ user writable directory for datumgrid files.
+ *
+ * @param ctx PROJ context, or NULL
+ * @param create Create the directory if it does not exist already.
+ * @return The path to the PROJ user writable directory.
+ * @since 7.1
+*/
+
+const char *proj_context_get_user_writable_directory(PJ_CONTEXT *ctx,
```
rouault, line 1189: Doxygen complains about this function and the other not being documented. This is because currently they must be within a //! @cond Doxygen_Suppress block. So either move them in a place in the file close to already exported method, or surround them with "//! @endcond" and "//! @cond Doxygen_Suppress"

src/apps/projinfo.cpp
```diff
@@ -1074,7 +1074,7 @@ int main(int argc, char **argv) {
 #ifdef CURL_ENABLED
             if (proj_context_is_network_enabled(nullptr)) {
                 std::cout << "Status: enabled" << std::endl;
-                std::cout << "URL: " << pj_context_get_url_endpoint(nullptr)
+                std::cout << "URL: " << std::string(proj_context_get_url_endpoint(nullptr))
```
rouault: this shouldn't be needed. The << operator will be happy with a const char*

src/apps/projsync.cpp
```diff
@@ -224,7 +224,7 @@ int main(int argc, char *argv[]) {
     }
 
     if (targetDir.empty()) {
-        targetDir = pj_context_get_user_writable_directory(ctx, true);
+        targetDir = std::string(proj_context_get_user_writable_directory(ctx, true));
```
rouault: I don't think the explicit cast is needed. You can assign a (non-null) const char* to a std::string implicitly

src/filemanager.cpp
```diff
@@ -1182,8 +1183,16 @@ static void CreateDirectoryRecursively(PJ_CONTEXT *ctx,
 
 // ---------------------------------------------------------------------------
 
-std::string pj_context_get_user_writable_directory(PJ_CONTEXT *ctx,
-                                                   bool create) {
+/** Get the PROJ user writable directory for datumgrid files.
+ *
+ * @param ctx PROJ context, or NULL
+ * @param create Create the directory if it does not exist already.
```
rouault:
```suggestion
 * @param create If set to TRUE, create the directory if it does not exist already.
```

src/filemanager.cpp
```diff
@@ -1478,11 +1487,11 @@ pj_open_lib_internal(projCtx ctx, const char *name, const char *mode,
 
         else if (!dontReadUserWritableDirectory() &&
                  (fid = open_file(
-                      ctx, (pj_context_get_user_writable_directory(ctx, false) +
+                      ctx, (std::string(proj_context_get_user_writable_directory(ctx, false)) +
                             DIR_CHAR + name)
                                .c_str(),
                       mode)) != nullptr) {
-            fname = pj_context_get_user_writable_directory(ctx, false);
+            fname = std::string(proj_context_get_user_writable_directory(ctx, false));
```
rouault: std::string() explicit cast probably unneeded here

src/filemanager.cpp
```diff
@@ -1554,7 +1563,7 @@ std::vector<std::string> pj_get_default_searchpaths(PJ_CONTEXT *ctx) {
         getenv("PROJ_SKIP_READ_USER_WRITABLE_DIRECTORY");
     if (ignoreUserWritableDirectory == nullptr ||
         ignoreUserWritableDirectory[0] == '\0') {
-        ret.push_back(pj_context_get_user_writable_directory(ctx, false));
+        ret.push_back(std::string(proj_context_get_user_writable_directory(ctx, false)));
```
rouault: std::string() explicit cast probably unneeded here

src/filemanager.cpp
```diff
@@ -1760,18 +1769,23 @@ int pj_find_file(projCtx ctx, const char *short_filename,
...
 /************************************************************************/
-/*                    pj_context_get_url_endpoint()                     */
+/*                    proj_context_get_url_endpoint()                   */
 /************************************************************************/
-
-std::string pj_context_get_url_endpoint(PJ_CONTEXT *ctx) {
+/** Get the URL endpoint to query for remote grids.
+*
+* @param ctx PROJ context, or NULL
+* @return Endpoint URL.
```
rouault:
```suggestion
* @return Endpoint URL. The returned pointer would be invalidated by a later call to proj_context_set_url_endpoint()
```

src/filemanager.cpp
snowman2: That is good to know - I made another attempt at what I think will fix it. :crossed_fingers:

snowman2:
@rouault thanks for the review and the tips to fix the docstrings :+1:

rouault:
Could you rebase on top of current master ? I pushed unrelated fixes to make Doxygen work on Mac due to a  likely change of version of it on homebrew

snowman2:
> Could you rebase on top of current master ? I pushed unrelated fixes to make Doxygen work on Mac due to a likely change of version of it on homebrew

Just did a rebase.

PR #2164, opened 2020-04-19
Update proj.1

jidanni:
This change is already on cs2cs man page but for some reason not on proj man page yet, even though implemented, with proof right there in the example below it!

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/man/man1/proj.1
+++ b/man/man1/proj.1
@@ -208,7 +208,7 @@ For ASCII input data the two data values must be in the first two white space
 separated fields and when both input and output are ASCII all trailing
 portions of the input line are appended to the output line.
 .sp
-Input geographic data (longitude and latitude) must be in DMS format and input
+Input geographic data (longitude and latitude) must be in DMS or decimal degrees format and input
 cartesian data must be in units consistent with the ellipsoid major axis or
 sphere radius units. Output geographic coordinates will be in DMS (if the
 \fB\-w\fP switch is not employed) and precise to 0.001” with trailing, zero\-valued


kbevers:
Thanks for the correction. The man-pages are generated automatically from the Shpinx docs. Could you please add you fix to `docs/source/apps/proj.rst` instead? That way it will also become part of the html and pdf docs. Thanks.

jidanni:
Nope nope nope...
You've got to add big warnings to the downstream doc I edited. And make it read-only.
Else the next person might spend several hours editing a downstream doc, when he should have been editing the upstream doc!

kbevers:
> You've got to add big warnings to the downstream doc I edited.

It's right there on line 1. Admittedly it is not a big flashing warning but it shouldn't be hard to see that this is not the source file

https://github.com/OSGeo/PROJ/blob/b7f5e7ea751588914eec2b64fd2ee39ac2eefc54/man/man1/proj.1#L1

> And make it read-only.

That's not as easy as it sounds... 

I can easily close this PR and make the changes myself. I wanted to offer you the chance to redo the fix so that you'll get the proper credit it for it. And honestly, it is to expected that first-time contributor need a bit of guidance so don't feel discouraged by that.

jidanni:
Well all I know is rendered man pages don't appear on GitHub,
because they are the product of post-processing.

Then the nroff shouldn't be on GitHub either,
because they are also automatically generated too...

Anyway, even if users do see
"Man page generated from reStructuredText.'
it still needs to say "DO NOT EDIT. EDIT INSTEAD
abc/efg/hij"

kbevers:
You are very welcome to fix any discrepancies you think that our documentation setup has. In the process you'll likely experience that it isn't as straight forward as you seem to think. I'll happily accept a PR that implements automatically generation of man pages when `make dist` is called that doesn't require sphinx and other supporting packages for doc creation when `make` and `make install` is called. The last time I looked into that I quickly came to the conclusion that the effort spend to do that was not worth the minor inconvenience of manually building the man pages at release time. 

As for the leading comment of the man-pages, there is no way to configure Sphinx to change that message. Sure, you can hack something together but so far you are the only one who has somehow felt offended by the lack of warnings in that comment, so for now I can't be bothered to change that. You, however, are very welcome to submit a patch that takes care of that.

PR #2166, opened 2020-04-19
Ingestion of WKT1_GDAL: correctly map 'Cylindrical_Equal_Area'

rouault:
Map it to 'Lambert Cylindrical Equal Area' / EPSG:9835 non-spherical
method, when the ellipsoid is not a sphere. And rationalize the
handling of this for other methods with spherical vs non-spherical formulations


rouault:
Manual backport to 7.0 done in PR #2167

PR #2168, opened 2020-04-19
gie: implement a strict mode with <gie-strict> </gie-strict> (fixes #2158)

### LINKED ISSUE #2158: gie accepts invalid keywords without noticing
I just accidentally noticed that I mispelled "roundtrip" as "roundrip" and gie didn't warn me... Similarly I frequently mispell "expect" as "except".

I think gie should error out when it encounters a keyword it doesn't understand. This would imply to revise existing gie files to properly add comment markers in front of some comment text that isn't marked as such currently

rouault:
In that mode:
* All non-comment/decoration lines must start with a valid tag
* Commands split on several lines should be terminated with " \\"

@busstoptaktik  @kbevers  Before I make the effort of converting all gie files in test/gie/ with those rules, I want to have your blessing


busstoptaktik:
I have no strong feelings around this, although I probably would have preferred the syntax I proposed in #2158: `<gie-strict> ... </gie-strict>`, which more clearly indicates that it is the entire *environment* that is to be considered strict.

If I understand your proposed syntax correctly, the strictness may shift half way through the input file (in which case, I believe there should also be a way to switch back to lax mode again).

Anyway - most important to me is to keep the original syntax as the default, as it really is quite useful in its own right, as demonstrated in the PROJ teaching material, I linked to over in #2158.

rouault:
> `<gie-strict> ... </gie-strict>`

ah sorry I missed that part, and was actually thinking about something similar. Just pushed this

busstoptaktik:
> ah sorry I missed that part, and was actually thinking about something similar. Just pushed this

Groovy :-)

rouault:
All .gie files converted and documentation added in gie.rst. There were a few instances of misspelled ``expect`` in builtins.gie

kbevers (approved):

rouault:
Just got feedback from @ccrook that he also got hit by expect/except misspelling, so I see a great future for `<gie-strict>` :-)

PR #2171, opened 2020-04-21
Add limited support for non-conformant WKT1 LAS COMPD_CS[]

rouault:
* Allow importing EPSG:{horizontal_code}+{geographic_code} and
  turn it into valid Geographic 3D or Projected 3D CRS internally
* Allow importing WKT1 COMPD_CS[] with above structure
* On an object imported that way, allow exporting to WKT1_GDAL,
  with this non-standard structure of a horizontal CRS + geographic CRS

CC @hobu


hobu:
This will work just fine for our needs I think. Thanks so much!

PR #2191, opened 2020-04-24
Doc: fixes for Sphinx 3.0 compat [skip appveyor]

rouault:
Those changes are also compatible with Sphinx 2.x

The use of .. cpp:enumerator::  for a C enum is questionable, but .. c:member:: no longer works in Sphinx 3 for the value of a C enum, but .. c:enumerator:: is Sphinx 3 only ( https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-c-enumerator )

Sphinx 3 is also more strict with the use of cross-reference to C types. It rejects cross-references to things like `` :c:type:`int` `` since this isn't a type we actually define.

### DIFF
--- a/docs/source/development/reference/datatypes.rst
+++ b/docs/source/development/reference/datatypes.rst
@@ -37,15 +37,15 @@ Transformation objects
             PJ_INV   = -1    /* Inverse    */
         } PJ_DIRECTION;
 
-    .. c:member:: PJ_FWD
+    .. cpp:enumerator:: PJ_FWD
 
         Perform transformation in the forward direction.
 
-    .. c:member:: PJ_IDENT
+    .. cpp:enumerator:: PJ_IDENT
 
         Identity. Do nothing.
 
-    .. c:member:: PJ_INV
+    .. cpp:enumerator:: PJ_INV
 
         Perform transformation in the inverse direction.
 
@@ -486,7 +486,7 @@ List structures
 
         Operation entry point.
 
-    .. c:member:: char * const *
+    .. c:member:: char * const * descr
 
         Description of operation.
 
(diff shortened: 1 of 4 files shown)


docs/source/development/reference/datatypes.rst
```diff
@@ -788,9 +788,9 @@ Logging
 
         typedef void (*PJ_LOG_FUNCTION)(void *, int, const char *);
 
-    where the :c:type:`void` pointer references a data structure used by the
-    calling application, the :c:type:`int` is used to set the logging level
-    and the :c:type:`const char` pointer is the string that will be logged
+    where the first agrument (void pointer) references a data structure used by the
```
kbevers:
```suggestion
    where the first argument (void pointer) references a data structure used by the
```

docs/source/development/reference/functions.rst
```diff
@@ -93,10 +96,13 @@ paragraph for more details.
...
 
-    :param PJ_CONTEXT* ctx: Threading context
-    :param int argc: Count of arguments in :c:data:`argv`
-    :param char** argv: Vector of strings with proj-string parameters, e.g. ``+proj=merc``
-    :returns: :c:type:`PJ*`
+    :param ctx: Threading context.
+    :type ctx: :c:type:`PJ_CONTEXT` *
+    :param argc: Count of arguments in :c:data:`argv`
+    :type argc: `int`
+    :param argv: Vector of strings with proj-string parameters, e.g. ``+proj=merc``
```
kbevers: Wouldn't "array" be more correct for C? Vector sounds more like C++.

docs/source/development/reference/datatypes.rst
rouault: Fixed

docs/source/development/reference/functions.rst
rouault: Fixed

kbevers (approved):

rouault:
Should we backport this to 7.0 branch ?

kbevers:
Yes, there's no need to have master and 7.0 diverge on this

PR #2195, opened 2020-04-25
Database: map more ESRI grid names

rouault:


data/sql/esri.sql
```diff
@@ -12291,3 +12289,46 @@ INSERT INTO "grid_transformation" VALUES('ESRI','108453','AGD_1984_To_GDA_1994_N
...
+-- 'spain/BALR2009': no mapping
+-- 'japan/tky2jgd': no mapping
+-- 'spain/baleares': no mapping
+-- 'spain/peninsula': no mapping
+-- 'icegrid2004': no mapping
+-- 'ICEGRID93': no mapping
+-- 'imhpgn': no mapping
+-- 'ohdhihpgn': no mapping
+-- 'switzerland/ntv2-ch03p-etrs': no mapping
+-- 'australia/QLD_0900': no mapping
```
nyalldawson, line 12334: From https://www.dnrme.qld.gov.au/__data/assets/pdf_file/0006/105765/gday-21-user-guide.pdf

    Note that the Queensland grid QLD_0900.gsb produces identical results to the National AGD84 grid for the equivalent coverage.

rouault: Thanks

PR #2206, opened 2020-05-04
 Add a +proj=defmodel transformation for multi-component time-based deformation models

### LINKED ISSUE #1001: Add support for deformation models

rouault:
Fixes #1001

The implementation of the transformation mostly resides in the defmodel.hpp header, that is completely PROJ agnostic. This offers the advantage of potentially reusing it for other implementations, and more practically for us, to be able to unit test the building blocks of the transformation that are normally not accessible through the PROJ API. The defmodel.cpp file contains the PROJ specific glue, and implements a few utility methods left unimplemented by the header to read a value in a grid, or perform cartesian<-->geographic conversions.

CC @ccrook who has tested this on the LINZ NZGD2000 deformation model along the development and contributed a few fixes and tests


src/apps/gie.cpp
```diff
@@ -723,18 +723,26 @@ Attempt to interpret args as a PJ_COORD.
         /* This could be avoided if proj_dmstor used the same proj_strtod()  */
         /* as gie, but that is not the case (yet). When we remove projects.h */
         /* from the public API we can change that.                           */
+        // Hum: not sure about the above. Coordinates are not necessarily
```
hobu: Who is commenting here?

src/grids.cpp
```diff
@@ -1584,28 +1595,27 @@ NTv1Grid *NTv1Grid::open(PJ_CONTEXT *ctx, std::unique_ptr<File> fp,
...
-    extent.resLat = to_double(header + 88) * DEG_TO_RAD;
-    if (!(fabs(extent.westLon) <= 4 * M_PI &&
-          fabs(extent.eastLon) <= 4 * M_PI &&
-          fabs(extent.northLat) <= M_PI + 1e-5 &&
-          fabs(extent.southLat) <= M_PI + 1e-5 &&
-          extent.westLon < extent.eastLon &&
-          extent.southLat < extent.northLat && extent.resLon > 1e-10 &&
-          extent.resLat > 1e-10)) {
+    extent.isGeographic = true;
+    extent.west = -to_double(header + 72) * DEG_TO_RAD;
```
hobu, line 1599: Could these magic numbers ever change?

src/grids.cpp
```diff
@@ -2995,24 +3002,24 @@ static PJ_LP pj_hgrid_apply_internal(projCtx ctx, PJ_LP in,
...
-            tb.lam -= extent->westLon;
+            tb.lam -= extent->west;
             if (tb.lam + epsilon < 0)
                 tb.lam += 2 * M_PI;
-            else if (tb.lam - epsilon > extent->eastLon - extent->westLon)
+            else if (tb.lam - epsilon > extent->east - extent->west)
                 tb.lam -= 2 * M_PI;
-            tb.phi -= extent->southLat;
+            tb.phi -= extent->south;
             dif.lam = std::numeric_limits<double>::max();
```
hobu, line 3023: NOMINMAX guards needed?

src/grids.cpp
```diff
@@ -3356,13 +3374,20 @@ bool pj_bilinear_interpolation_three_samples(const GenericShiftGrid *grid,
...
-        grid_x = (lp.lam + 2 * M_PI - extent.westLon) / extent.resLon;
-    } else if (lp.lam > extent.eastLon) {
-        grid_x = (lp.lam - 2 * M_PI - extent.westLon) / extent.resLon;
+    if (!extent.isGeographic) {
+        pj_log(ctx, PJ_LOG_ERROR,
+               "Can only handle grids referenced in a geographic CRS");
+        pj_ctx_set_errno(ctx, PJD_ERR_FAILED_TO_LOAD_GRID);
+        return false;
+    }
+
```
hobu, line 3383: A short comment about how we're computing grid locations and what we need to do it would be a helpful echo for the future.

src/transformations/defmodel.cpp
```diff
@@ -0,0 +1,445 @@
...
+PROJ_HEAD(defmodel, "Deformation model");
+
+using namespace DeformationModel;
+
+namespace {
+
+struct Grid : public GridConcept {
+    PJ_CONTEXT *ctx;
+    const NS_PROJ::GenericShiftGrid *realGrid;
+    mutable bool checkedHorizontal = false;
```
hobu, line 51: Can you do this without mutables?

src/transformations/defmodel.cpp
```diff
@@ -0,0 +1,445 @@
...
+
+    bool getEastingNorthingOffset(int ix, int iy, double &eastingOffset,
+                                  double &northingOffset) const {
+        if (!checkHorizontal(STR_METRE)) {
+            return false;
+        }
+        float eastingOffsetFloat = 0.0f;
+        float northingOffsetFloat = 0.0f;
+        const bool ret =
+            realGrid->valueAt(ix, iy, sampleX, eastingOffsetFloat) &&
```
hobu, line 181: Are the grids always floats, or could they be doubles too?

src/transformations/defmodel.cpp
```diff
@@ -0,0 +1,445 @@
...
+                                double &phi, double &height) {
+        (void)a;
+        (void)b;
+        assert(cart->a == a);
+        assert(cart->b == b);
+        PJ_XYZ xyz;
+        xyz.x = X;
+        xyz.y = Y;
+        xyz.z = Z;
+        PJ_LPZ lpz = cart->inv3d(xyz, cart);
```
hobu, line 307: Can this fail? What's its failure condition and can we recover?

src/transformations/defmodel.cpp
```diff
@@ -0,0 +1,445 @@
...
+    explicit defmodelData(PJ_CONTEXT *ctx, PJ *cart)
+        : evaluatorIface(ctx, cart) {}
+
+    defmodelData(const defmodelData &) = delete;
+    defmodelData &operator=(const defmodelData &) = delete;
+};
+
+} // namespace
+
+static PJ *destructor(PJ *P, int errlev) {
```
hobu, line 337: This seems poorly named.

src/transformations/defmodel.cpp
```diff
@@ -0,0 +1,445 @@
...
+    if (!Q->evaluator->inverse(Q->evaluatorIface, in.xyzt.x, in.xyzt.y,
+                               in.xyzt.z, in.xyzt.t, out.xyzt.x, out.xyzt.y,
+                               out.xyzt.z)) {
+        return proj_coord_error();
+    }
+
+    return out;
+}
+
+static void reassign_context(PJ *P, PJ_CONTEXT *ctx) {
```
hobu, line 381: These freestanding functions seem a little haphazard. If they were more specifically named with some comment on what they're for, I wouldn't ignore them at my peril as a reader 😄

src/transformations/defmodel.cpp
```diff
@@ -0,0 +1,445 @@
...
+    auto *Q = (struct defmodelData *)P->opaque;
+    if (Q->evaluatorIface.ctx != ctx) {
+        Q->evaluator->clearGridCache();
+        Q->evaluatorIface.ctx = ctx;
+    }
+}
+
+PJ *TRANSFORMATION(defmodel, 1) {
```
hobu, line 389: why is this named ALL CAPS? Does that denote something? No comment to tell me why.

src/transformations/defmodel.cpp
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    auto file = NS_PROJ::FileManager::open_resource_file(P->ctx, model);
+    if (nullptr == file) {
+        proj_log_error(P, "defmodel: Cannot open %s", model);
+        return destructor(P, PJD_ERR_INVALID_ARG);
+    }
+    file->seek(0, SEEK_END);
+    unsigned long long size = file->tell();
+    if (size > 10 * 1024 * 1024) {
```
hobu, line 419: magic. why is this size too large?

src/transformations/defmodel.hpp
```diff
@@ -0,0 +1,1890 @@
...
+    std::string mHorizontalOffsetUnit{};
+    std::string mVerticalOffsetUnit{};
+    std::string mHorizontalUncertaintyType{};
+    std::string mHorizontalUncertaintyUnit{};
+    std::string mVerticalUncertaintyType{};
+    std::string mVerticalUncertaintyUnit{};
+    std::string mHorizontalOffsetMethod{};
+    SpatialExtent mSpatialExtent{};
+    TimeExtent mTimeExtent{};
+    std::vector<Component> mComponents{};
```
hobu, line 463: You might want to add a ``whatever_else`` member (NL::json?), since this looks like a metadata struct, and those historically grow things as time goes on.

src/transformations/defmodel.hpp
```diff
@@ -0,0 +1,1890 @@
...
+};
+
+// ---------------------------------------------------------------------------
+
+class UnimplementedException : public std::exception {
```
hobu: Doe PROJ throw custom exceptions derived from std::exception generally? Is there a common place for their definition?

src/transformations/defmodel.hpp
```diff
@@ -0,0 +1,1890 @@
...
+    void log(const std::string & /* msg */) {
+        throw UnimplementedException("log unimplemented");
+    }
+#endif
+};
+
+// ---------------------------------------------------------------------------
+
+constexpr double DEFMODEL_PI = 3.14159265358979323846;
+constexpr double DEG_TO_RAD_CONSTANT = 3.14159265358979323846 / 180.;
```
hobu: Is there a common place for these kinds of things?

src/transformations/defmodel.hpp
```diff
@@ -0,0 +1,1890 @@
...
+    // where x0 = -resx / 2
+    inline void getBilinearGeocentric(int ix0, int iy0, double de00,
+                                      double dn00, double de01, double dn01,
+                                      double de10, double dn10, double de11,
+                                      double dn11, double m00, double m01,
+                                      double m10, double m11, double &dX,
+                                      double &dY, double &dZ) {
+        // If interpolating in the same cell as before, then we can skip
+        // the recomputation of dXij, dYij and dZij
+        if (ix0 != last_ix0 || iy0 != last_iy0) {
```
hobu: Lots of math here and not lots of comment...

src/transformations/defmodel.hpp
```diff
@@ -0,0 +1,1890 @@
...
+
+        dX = m00 * dX00 + m01 * dX01 + m10 * dX10 + m11 * dX11;
+        dY = m00 * dY00 + m01 * dY01 + m10 * dY10 + m11 * dY11;
+        dZ = m00 * dZ00 + m01 * dZ01 + m10 * dZ10 + m11 * dZ11;
+    }
+};
+
+// ---------------------------------------------------------------------------
+
+enum class DisplacementType { NONE, HORIZONTAL, VERTICAL, THREE_D };
```
hobu: I find enums in the middle of long header files difficult to find when scanning a header. Prefer them at the top. (feel free to ignore this)

src/transformations/defmodel.hpp
```diff
@@ -0,0 +1,1890 @@
...
+    void clearGridCache() {
+        gridSet.reset();
+        mapGrids.clear();
+    }
+};
+
+// ---------------------------------------------------------------------------
+
+/** Evaluator exception. */
+class EvaluatorException : public std::exception {
```
hobu: These exceptions should be pushed into another header so this one isn't so long.

src/transformations/defmodel.hpp
```diff
@@ -0,0 +1,1890 @@
...
+    ex.mMinyRad = DegToRad(ex.mMiny);
+    ex.mMaxxRad = DegToRad(ex.mMaxx);
+    ex.mMaxyRad = DegToRad(ex.mMaxy);
+
+    return ex;
+}
+
+// ---------------------------------------------------------------------------
+
+Component Component::parse(const json &j) {
```
hobu: Move this to its own utility/class/header?

hobu (approved):

src/apps/gie.cpp
rouault: fixed

src/grids.cpp
rouault: NTv1, an obsolete format, changing ? No. Those are fixed offsets in the file

src/transformations/defmodel.cpp
kbevers: Business as usual. This is a C preprocessor macro defined in the depths of `proj_internal.h`. There's also OPERATION, PROJECTION and CONVERSION. This is how we register and set up all operations.

> No comment to tell me why.

As a core committer to PROJ you are expected to know this :-)

src/grids.cpp
rouault: I don't think that's needed. Lat/long grids do not work well close to the poles...

src/transformations/defmodel.cpp
kbevers: Follows convention from many other projections and transformations

src/grids.cpp
rouault: done

src/transformations/defmodel.cpp
rouault: We can support other data types in the TIFF storage (see GTiffGrid::valueAt()) but we normalize to float. Given that the values are corrections, and not absolute values, float is accurate enough. This is as a heritage of formats like GTX or NTv2 storing floats.

src/transformations/defmodel.cpp
rouault: Lucikly, forward and inverse cart can't fail !

src/transformations/defmodel.cpp
hobu: > As a core committer to PROJ you are expected to know this :-)

I commit docs and CI. I claim no responsibility for any maths.

src/transformations/defmodel.cpp
rouault: yeah, but this is the convention used universally in PROJ for that function... I'm a bit hesitant to rename it, so that we can easily grep for that in case we'd want to do a mass change some day

src/transformations/defmodel.cpp
rouault: comment added

src/transformations/defmodel.cpp
rouault: this is a macro used for all such objects. Complain to @kbevers :-)

src/transformations/defmodel.cpp
rouault: not without removing constness of some methods. I don't see this use as an issue here. This is within the intended use of mutable in C++ (some hidden state, not accessible to the user)

src/transformations/defmodel.cpp
rouault: There's no reason for the JSON file to be arbitrarily large, so some threshold has to be established to avoid a potential risk of denial of service. Adding comment to make that explicit

src/transformations/defmodel.hpp
rouault: hum, I'm not keen on doing that :-) The objective of this class is to have traditional C++ access to metadata. At worse people can access the extra info with NL::json itself

src/transformations/defmodel.hpp
rouault: there isn't much PROJ tradition regarding this. And this file is intended to be PROJ agnostic. That said, to make the file lighter, I've just moved the exception prototype to a defmodel_exceptions.hpp header

src/transformations/defmodel.hpp
rouault: we do have PROJ constants, but this file is intended to be PROJ agnostic, and unfortunately M_PI isn't completely standard (on Windows)

src/transformations/defmodel.hpp
rouault: the maths here come from @ccrook proposal. I suspect they are "traditional" to geodesy foks. I've added a small comment regarding that

src/transformations/defmodel.hpp
rouault: this is mostly an implementation detail (not exposed in public API), so keeping it here

src/transformations/defmodel.hpp
rouault: done

src/transformations/defmodel.cpp
kbevers: > this is a macro used for all such objects. Complain to @kbevers :-)

Some of that complaint should be directed at Jerry Evenden who originally started the macro mess (for good reasons at the time, I'm sure) but I am happy to take partial responsibility for the currrent macros along with @busstoptaktik. Afterall, this is easier to read than 

```
FREEUP; if (P) free(P); }
ENTRY0(projection) P->es = 0.; P->inv = s_inverse; P->fwd = s_forward; ENDENTRY(P)
```

src/transformations/defmodel.hpp
rouault: excellent suggestion. I've moved all implementation details to a defmodel_impl.hpp file

rouault:
Thanks @hobu for the review!

kbevers:
> @kbevers Will you want to do a review on this before I merge ?

Yes, please! Sorry for not getting round to it earlier. I've been hella busy the last couple of weeks. I'll try to give it a closer look this weekend.

rouault:
> Yes, please!

No, problem. Take your time

kbevers (approved):
I think this looks very good. I haven't looked super detailed at the code since there's quite a lot of it and my memory of the initial proposal of the deformation model "master file" concept by Chris is good enough for a detailed review. The docs and the examples are rather clear to me and I have not reason to not want to include this PR as is. Good effort!

ccrook:
Thankyou Even, Kristian, and Howard for your contributions to the development of this format specification and Even for the excellent implementation.  This is a huge benefit to the New Zealand user community.  It means we will be able to move many of our tools to use the PROJ library rather than our own custom software which is great for its long term sustainability.

PR #2209, opened 2020-05-06
.travis.yml: disable notifications as they are probably more spam than useful [skip appveyor]

rouault:


kbevers:
Good idea, that's just generating spam

PR #2223, opened 2020-05-14
Avoid 2 warnings about missing database indices

### LINKED ISSUE #2221: SQLite3 warnings triggered (and not detected)
#### Example of problem
When I enable SQLite3 error logging with the call to `sqlite3_config(SQLITE_CONFIG_LOG, errorLogCallback, pData);` it detects some error/warnings. These messages may saturate the log, and they show a posible lack of performance.

Enabling sqlite3 log (see below how) this command produces this output:
```
echo  48 -110 0 | cs2cs EPSG:4326 EPSG:32001
setting SQLITE_CONFIG_LOG
(code 284) automatic index on alias_name(code)
(code 284) automatic index on alias_name(code)
(code 284) automatic index on alias_name(code)
(code 284) automatic index on alias_name(code)
(code 284) automatic index on alias_name(code)
(code 284) automatic index on alias_name(code)
(code 284) automatic index on alias_name(code)
(code 284) automatic index on alias_name(code)
1877763.91	365187.80 0.00
```

This naïve patch enables the sqlite3 log in `cs2cs`, and dumps every message to `stderr`
```
diff --git a/src/apps/cs2cs.cpp b/src/apps/cs2cs.cpp
index 4ada83b5..567cb740 100644
--- a/src/apps/cs2cs.cpp
+++ b/src/apps/cs2cs.cpp
@@ -319,7 +319,17 @@ static std::string get_geog_crs_proj_string_from_proj_crs(PJ *src,
 /*                                main()                                */
 
+#include "sqlite3.h"
+static void errorLogCallback(void*, int iErrCode, const char* zMsg)
+{
+    fprintf(stderr, "(code %d) %s\n", iErrCode, zMsg);
+}
+
 int main(int argc, char **argv) {
+
+    fprintf(stderr, "setting SQLITE_CONFIG_LOG\n"); // to test stderr
+    sqlite3_config(SQLITE_CONFIG_LOG, errorLogCallback, nullptr);
+
     char *arg;
     char **eargv = argv;
     std::string fromStr;
```

#### Problem description
When the final application uses SQLite3 for other purposes, it may want to detect errors with the log callback. However in this case, the log can be flooded by messages from `PROJ`, depending on the transformations and CRS used.
The error is suggesting that the index should be pre-created, to avoid creating it automatically every single time.

More info in 
- https://www.sqlite.org/optoverview.html#autoindex
- https://www.sqlite.org/errlog.html 
- https://www.sqlite.org/c3ref/c_config_covering_index_scan.html#sqliteconfiglog

This problem is detected also calling the method `GetProjectionRef` reading a GeoTIFF with `GDAL`. Every time I open a GeoTIFF (and read the projection ref), the message is logged.

Side note: creating a "non-Universal" Transverse Mercator projection (with scale_factor=1, false_easting=0, central_meridian=0) in `GDAL` with `SetTM` generates the log `(284) automatic index on object_view(table_name)`

#### Expected Output
No messages due to missing indices on SQLite3 log when it is enabled in a final application.

#### Environment Information
 - PROJ version (`proj`) 7.0.1
 - Operation System Information: Ubuntu 18.04

#### Installation method
 - compiled from source tag 7.0.1

rouault:
Fixes the 'automatic index on alias_name(code)' and 'automatic index on object_view(table_name)' warnings of #2221

Fixes #2221

Note: when running the whole test suite with PROJ_LOG_SQLITE3=YES, there are remaining warnings. I tried to address a few of the remaining ones, but wasn't successful

### DIFF
--- a/data/sql/proj_db_table_defs.sql
+++ b/data/sql/proj_db_table_defs.sql
@@ -1340,6 +1340,8 @@ CREATE TABLE alias_name(
     source TEXT
 );
 
+CREATE INDEX idx_alias_name_code ON alias_name(code);
+
 CREATE TRIGGER alias_name_insert_trigger
 BEFORE INSERT ON alias_name
 FOR EACH ROW BEGIN
--- a/src/iso19111/factory.cpp
+++ b/src/iso19111/factory.cpp
@@ -5434,25 +5434,6 @@ std::string AuthorityFactory::getOfficialNameFromAlias(
 
 // ---------------------------------------------------------------------------
 
-//! @cond Doxygen_Suppress
-
-static void addToListString(std::string &out, const char *in) {
-    if (!out.empty()) {
-        out += ',';
-    }
-    out += in;
-}
-
-static void addToListStringWithOR(std::string &out, const std::string &in) {
-    if (!out.empty()) {
-        out += " OR ";
-    }
-    out += in;
-}
-//! @endcond
-
-// ---------------------------------------------------------------------------
-
 /** \brief Return a list of objects by their name
  *
  * @param searchedName Searched name. Must be at least 2 character long.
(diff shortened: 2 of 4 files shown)


hobu:
I came to this PR searching for PROJ_LOG_SQLITE3, which is not in the docs and maybe should be.

Issue #1442, opened 2019-04-24
Support for the Ocean Oriented Interrupted Goode Homolosine


sridge:
![GUID-F958670B-74CA-40B0-A4A2-26C6101F6C64-web](https://user-images.githubusercontent.com/11999411/56681706-f51e2880-6697-11e9-90b7-9de8462b3c35.gif)

Something similar to the above, from ArcGIS. It's commonly used by oceanographers but doesn't exist yet for PROJ.4. Those oceanographers that use python and Cartopy (and therefore PROJ.4), would appreciate this projection being added.

Some specifications:
- Northern Hemisphere interruptions at 100˚,-100˚
- Southern Hemisphere interruptions at  -67˚, 147˚ (corresponding with [ocean borders](https://en.wikipedia.org/wiki/Borders_of_the_oceans))
- Left and right bounds at 20˚, central longitude of -160˚

kbevers:
There's a better chance to get your wish fulfilled if your provide some material that describes how to implement this projection. Ideally a paper describing the maths and some test coordinates that can be used to verify the implementation.

sridge:
Thank you, I'll add some more background information. Not sure what test coordinates are, but assume they may already exist

https://github.com/OSGeo/proj.4/blob/master/src/projections/igh.cpp

The projection is already included in PROJ (igh), but with different interruptions. I was assuming it would be as simple as rearranging the "zones" from the code. 

From this:
```
 Land Zones:
    -180            -40                       180
      +--------------+-------------------------+    Zones 1,2,9,10,11 & 12:
      |1             |2                        |      Mollweide projection
      |              |                         |
      +--------------+-------------------------+    Zones 3,4,5,6,7 & 8:
      |3             |4                        |      Sinusoidal projection
      |              |                         |
    0 +-------+------+-+-----------+-----------+
      |5      |6       |7          |8          |
      |       |        |           |           |
      +-------+--------+-----------+-----------+
      |9      |10      |11         |12         |
      |       |        |           |           |
      +-------+--------+-----------+-----------+
    -180    -100      -20         80          180
*/
```
To this:
```
 Ocean Zones:
     20       100               -100          20
      +--------+------------------+------------+    Zones 1,2,3,10,11 & 12:
      |1       |2                 |3           |      Mollweide projection
      |        |                  |            |
      +--------+------------------+------------+    Zones 4,5,6,7, 8 & 9:
      |4       |5                 |6           |      Sinusoidal projection
[...]

PR #2226, opened 2020-05-15
Implemented IGH Oceanic View

jkrasting:
- The current implementation of the Interrupted Goode Homolosine
  projection emphasizes land area. This is a compliment projection
  that emphasizes ocean area.
- A value of lon0=-160 produces a reasonable real-world map.
- Tested update using [cartopy](https://github.com/SciTools/cartopy):

imh_ocean_view


src/projections/igh_o.cpp
```diff
@@ -0,0 +1,235 @@
...
+    else if (lp.phi >= -d4044118) {
+           if (lp.lam <=  -d75) z =  7;
+      else if (lp.lam >=   d90) z =  9;
+      else z = 8;
+    }
+    else {
+           if (lp.lam <=  -d75) z =  10;
+      else if (lp.lam >=   d90) z =  12;
+      else z = 11;
+    }
```
kbevers: What are these magic numbers? Can they be represented as an enum? That would be more descriptive

src/projections/igh_o.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+C_NAMESPACE PJ *pj_sinu(PJ *), *pj_moll(PJ *);
+
+/* 40d 44' 11.8" [degrees] */
+/*
+static const double d4044118 = (40 + 44/60. + 11.8/3600.) * DEG_TO_RAD;
+has been replaced by this define, to eliminate portability issue:
+Initializer element not computable at load time
+*/
+#define d4044118 ((40 + 44/60. + 11.8/3600.) * DEG_TO_RAD)
```
kbevers: Could this get a more descriptive name? I suspect this is a longitude with a special property but it is not entirely clear from the comment here. 

It would also be better to da a `static const double ...` as below

src/projections/igh_o.cpp
```diff
@@ -0,0 +1,235 @@
...
+      else z = 8;
+    }
+    else {
+           if (xy.x <=   -d75) z =  10;
+      else if (xy.x >=    d90) z =  11;
+      else z = 12;
+    }
+
+    if (z) {
+        int ok = 0;
```
kbevers: Make this a boolean instead

src/projections/igh_o.cpp
```diff
@@ -0,0 +1,235 @@
...
+      |7           |8               |9         |
+      |            |                |          |
+      +------------+----------------+----------+
+      |10          |11              |12        |
+      |            |                |          |
+      +------------+----------------+----------+
+    -180          -75               90        180
+*/
+
+#define SETUP(n, proj, x_0, y_0, lon_0) \
```
kbevers: This looks like an ancient PROJ.4 macro... Surely this can be redefine as an actual function that reads better

kbevers (changes requested):
Thanks for adding this. Very cool to get this projection in PROJ!

There's a few things missing from this PR:

1. `igh_o.png` does not seem to be commited
2. Tests - can be added in `test\gie\`, either as a standalone file or in e.g. `builtins.gie`.
3. Less important, but still relevant: More descriptive docs. It would be nice to have a short introduction that explains what this projection is.

src/projections/igh_o.cpp
jkrasting: The new projection parallels as closely as possible the existing `igh.cpp` projection. The whole projection is a composite of 12 sub-projections.  The integers are id numbers that correspond to the ascii table at the bottom of the file

src/projections/igh_o.cpp
jkrasting: Again, maintaining consistency with `igh.cpp`

src/projections/igh_o.cpp
jkrasting: I am not really a C++ programmer - could we leave this as an issue to for someone more experienced than me to revisit both igh* projections at a later date?

jkrasting (commented):
Thanks for your feedback - just submitted a PR with a test case and image.

src/projections/igh_o.cpp
jkrasting: d4404118 is not a longitude - its the latitude of the breakpoint between the sinusoidal and Mollweide projections.

jkrasting:
FYI -- see complement pull request with [cartopy](https://github.com/SciTools/cartopy/pull/1561)

src/projections/igh_o.cpp
kbevers: I don't care much for consistency with the other projection, that's no good either. Please add comments that explain what's going on here. Or better yet, improve the code :-)

src/projections/igh_o.cpp
kbevers: Please document that in a comment, so the next guy has a chance to figure out what's going on

src/projections/igh_o.cpp
kbevers: We can, but the likelihood of fixing this is close to zero I reckon. Honestly, this is not code I am happy to maintain. Who do you suggest revisit this?

src/projections/igh_o.cpp
jkrasting: Sure - I just committed some more documentation describing the latitude constant and the integer numbering scheme.

src/projections/igh_o.cpp
jkrasting: Fair point.  Documentation added to explain the numbering scheme.

src/projections/igh_o.cpp
jkrasting: Not sure what you mean here.

src/projections/igh_o.cpp
jkrasting: This is a bit beyond my scope in C++.

src/projections/igh_o.cpp
rouault: bool ok = false;

src/projections/igh_o.cpp
rouault: > `static const double ...` as below

The C++11 canoncial way for such compile time constants is:
constexpr double = ....

src/projections/igh_o.cpp
rouault: Something along those lines:
```
static void setup_zone(struct pj_opaque *Q, int n,  PJ*(*proj_ptr)(PJ*), double x_0, double y_0, double lon_0
{
    if (!(Q->pj[n-1] = proj_ptr(nullptr))) return false;
    if (!(Q->pj[n-1] = proj_ptr(Q->pj[n-1]))) return false;
    Q->pj[n-1]->ctx = P->ctx;
    Q->pj[n-1]->x0 = x_0;
    Q->pj[n-1]->y0 = y_0;
    Q->pj[n-1]->lam0 = lon_0;
    return true;
}
```

and then replace call sites like:
```
if (!setup_zone(Q, 4, pj_sinu, -d150, 0, -d150) ||
    !setup_zone(Q, 5, pj_sinu, -d40,  0,  -d40) ||
   ... add others ... )
{
   return destructor(P, ENOMEM);
}
```

src/projections/igh_o.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+static PJ *destructor (PJ *P, int errlev) {
+    int i;
+    if (nullptr==P)
+        return nullptr;
+
+    if (nullptr==P->opaque)
+        return pj_default_destructor (P, errlev);
+
+    for (i = 0; i < 12; ++i) {
```
rouault, line 176:
```suggestion
    struct pj_opaque *Q = static_cast<struct pj_opaque*>(P->opaque);
    for (i = 0; i < 12; ++i) {
```
and replace all 3 below uses of ``static_cast<struct pj_opaque*>(P->opaque)`` with ``Q``

rouault:
> /Users/travis/build/OSGeo/PROJ/docs/source/operations/projections/igh_o.rst:document isn't included in any toctree

==> It should be referenced in docs/source/operations/projections/index.rst

src/projections/igh_o.cpp
jkrasting: Great - thanks @rouault.  Learning C++ as I go here.

src/projections/igh_o.cpp
jkrasting: Sorry - still having trouble following.  Theres this block in the middle of the SETUP() calls that adjusts the Northern Hemisphere edge zones:

```
244     /* y0 ? */
245     xy1 = Q->pj[0]->fwd(lp, Q->pj[0]); /* zone 1 */
246     xy3 = Q->pj[4]->fwd(lp, Q->pj[4]); /* zone 3 */
247     /* y0 + xy1.y = xy3.y for lt = 40d44'11.8" */
248     Q->dy0 = xy3.y - xy1.y;
249
250     Q->pj[0]->y0 = Q->dy0;
```

rouault: > Sorry - still having trouble following. Theres this block in the middle of the SETUP()

what's the issue ? You can let this block unchanged. I just suggested to use a function instead of the SETUP macro

jkrasting: Sorry - I think I see it now ...

src/projections/igh.cpp
```diff
@@ -186,14 +186,17 @@ static PJ *destructor (PJ *P, int errlev) {
     -180    -100      -20         80          180
 */
 
-#define SETUP(n, proj, x_0, y_0, lon_0) \
-    if (!(Q->pj[n-1] = pj_##proj(nullptr))) return destructor(P, ENOMEM); \
-    if (!(Q->pj[n-1] = pj_##proj(Q->pj[n-1]))) return destructor(P, ENOMEM); \
-    Q->pj[n-1]->ctx = P->ctx; \
-    Q->pj[n-1]->x0 = x_0; \
-    Q->pj[n-1]->y0 = y_0; \
+static bool setup_zone(PJ *P, struct pj_opaque *Q, int n, \
```
rouault: no need for those trailing backslash

jkrasting: I thought they added a personal touch!  😄 

(Just pushed a commit removing them.)

src/projections/igh_o.cpp
jkrasting: Thanks. Done.

rouault:
I actually see this projection was a feature request of https://github.com/OSGeo/PROJ/issues/1442 . But in https://github.com/OSGeo/PROJ/issues/1442#issuecomment-486768702 , the interruptions values suggested seem different to the one you propose, but the projection maps look similar. I'm not sure who is right or who is wrong or maybe if both proposals are identical, but wanted to raise this and if we could know if there's interoperability with the equivalent ArcGIS projection (https://desktop.arcgis.com/en/arcmap/10.3/guide-books/map-projections/goodes-homolosine.htm)

jkrasting:
Yes I see some differences. Thanks for pointing this out.  The original reference from 1925 is here:  https://www.tandfonline.com/doi/abs/10.1080/00045602509356949

Some adjustments are needed, but this version is pretty close

kbevers (approved):
Unfortunately the original paper doesn't mention how the zones are divided in the oceanic view, so there's no help there. What's implemented now is likely good enough and I'm sure we'll made aware if it isn't.

I think this is looking good now. Thanks for putting in the extra effort improving the code, very much appreaciated.

jkrasting:
No, but the map is instructive.  It looks like `lon_0` should be -160 and some minor adjustments to the true longitudes and split lines are needed.  Will update and push a new commit today.  Thats @kbevers and @rouault for pointing this out.

rouault:
Another source for inspiration / comparison is this implementation based on d3.js: https://bl.ocks.org/Fil/d90a94d7333bfef753e92a4abb7611b4 / https://observablehq.com/@d3/interrupted-homolosine-oceans. The convention for the lobes is explained at https://github.com/d3/d3-geo-projection/#interrupted-projections . Looking at the values for the standard IGH, they seem to match with PROJ igh (at list the limits of zones)

jkrasting:
The lobes are present in the original reference.  Zones 1 and 3 have them for the ocean projection.  I added them into the latest commit.

image

docs/plot/plotdefs.json
```diff
@@ -554,6 +554,17 @@
         "res": "low",
         "type": "line"
     },
+    {
+        "filename": "igh_o.png",
+        "latmax": 90,
+        "latmin": -90,
+        "lonmax": 180,
+        "lonmin": -180,
+        "name": "igh",
```
rouault:
```suggestion
        "name": "igh_o",
```

jkrasting: Good catch. Thanks.

test/gie/builtins.gie
```diff
@@ -2194,23 +2194,23 @@ operation +proj=igh_o   +a=6400000
 -------------------------------------------------------------------------------
 tolerance 0.1 mm
 accept  2 1
-expect  222687.614453779 111701.072127637
+expect  223538.245169845 111701.072127637
```
rouault, line 2239: It would be useful to test roundtripping on those points by adding `roundtrip 1` after each expect, like:

```suggestion
expect  223538.245169845 111701.072127637
roundtrip 1
```
Actually the tests should test that for each of the 12 zones rather than on those dummy standard points.

kbevers: > Actually the tests should test that for each of the 12 zones rather than on those dummy standard points.

This is a key point to making sure this works as expected. Please adjust the test accordingly. That is, add accept/expect coordinate pairs that fall within each of the twelve zones.

jkrasting: I added in sub-projection tests for the new igh_o projection along with the existing igh projection.

test/gie/builtins.gie
```diff
@@ -2183,10 +2209,35 @@ expect  -0.001790497 0.000895247
 accept  -200 -100
 expect  -0.001790496 -0.000895247
 
+accept -100.0 22.0
```
rouault: those points in the inverse direction are all close to lat=lon=0 and not super relevant. I would just remove them completely, but as suggested previously, after each expect statement above add a "roundtrip 1" line, that will check that when using the inverse path the values of expect are back projected to the ones of accept

test/gie/builtins.gie
```diff
@@ -2212,6 +2288,30 @@ expect  -0.001790494 0.000895247
 accept  -200 -100
 expect  -0.001790496 -0.000895247
 
+accept -140.0 22.0
```
rouault: same remark as above: those tests points aren't super relevant

rouault (changes requested):
tests can be simplified & improved. See 2 comments

test/gie/builtins.gie
kbevers: Great! Even's comment about adding `roundtrip 1`'s after each `expect` line is still valid. Apart from that I'd say this PR is good to go.

test/gie/builtins.gie
jkrasting: I added in the roundtrip option for all zones in my latest commit, but I am seeing failures with infs in zones 11 and 12.  I am stumped.  Do you see anything obvious with how I implemented the tests?

test/gie/builtins.gie
jkrasting: I added in the roundtrip option for all zones in my latest commit, but I am seeing failures with infs in zones 11 and 12. I am stumped. Do you see anything obvious with how I implemented the tests?

jkrasting: Choosing slightly different values eliminates the inf issue.  Could it be because I chose values that lie on the true longitude for the Mollewiede projections?

jkrasting:
```
     FAILURE in builtins.gie(2283):
     roundtrip deviation: inf mm, expected: 0.100000 mm
```

jkrasting: > Choosing slightly different values eliminates the inf issue. Could it be because I chose values that lie on the true longitude for the Mollewiede projections?

Spoke too soon on this. It doesn’t solve the problem.

jkrasting: This ended up being a typo in the forward projection.  Fixed in 4dc21af.

jkrasting:
> 0 was fine. z is a int. assigning it a boolean value is confusing (even if false will ultimately evaluate to 0 once cast to int)

I reverted the commit.

rouault:
did you regenerate the image after your latest adjustments ?

jkrasting:
> did you regenerate the image after your latest adjustments ?

It shouldn't change the image but I'll double check to confirm.

rouault:
I would expect https://github.com/OSGeo/PROJ/pull/2226/commits/bf6a8dc1efb2ef048958f3dfeded14d4b03de801 to have impacts on the forward path, no ?

jkrasting:
The plot is identical.  The change in bf6a8dc is aesthetic and was done to be consistent to match `igh.cpp`.  That block of code identifies the delta value for latitude which did not change when comparing those two zones.

PR #2230, opened 2020-05-16
Limit peirce_q to northern hemisphere, and fix images for adams_hemi, guyou and peirce_q

rouault:
I'm a bit confused by PROJ peirce_q implementation. Looking at the projection of the world, it looks like this matches the diamond formulation of the right map shown at https://desktop.arcgis.com/en/arcmap/latest/map/projections/peirce-quincuncial.htm, but limited to the inner square of this diamond (which corresponds to the northern hemisphere). We lack the 4 triangles on left, top, right and bottom for the southern hemisphere. Furthermore, this formulation of peirce_q does not seem to have the quincuncial property of the square formulation (left images of the above ESRI doc), or the one at https://en.wikipedia.org/wiki/Peirce_quincuncial_projection ...


kbevers (approved):
Well, your changes looks absolutely fine. I have no idea what to do about the possible wrong implementation of the Peirce Q... I suppose it is possible to dig out the math from old publications but I haven't looked into that.

tcwilkinson:
Note that this MIT-licenced implementation by @jonathf in `cartopy` seems to handle southern hemisphere ok (if my reading of the code is not mistaken). Possible solution contained within? 

(https://github.com/jonathf/cartopy/blob/8172cac7fc45cafc86573d408ce85b74258a9c28/lib/cartopy/peircequincuncial.py)

PR #2243, opened 2020-05-26
Implement wink2 inverse by generic inversion of forward method

### LINKED ISSUE #35512: Problem reprojecting rasters in 54019 and 54042 CRSs
In QGIS 3.12, the Winkel Tripel projection working on the fly but Winkel II is not working.
The Winkel II on the fly coordinate transformation is not implemented yet or there is a bug?

rouault:
- Move the generic method initiated from adams_ws2 to a
  pj_generic_inverse_2d() method
- Use it in adams_ws2
- Use it in wink2

Fixes https://github.com/qgis/QGIS/issues/35512


src/generic_inverse.cpp
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * Uses Newton-Raphson method, extended to 2D variables, that is using
+ * inversion of the Jacobian 2D matrix of partial derivatives. The derivatives
+ * are estimated numerically from the P->fwd method evaluated at close points.
+ *
+ * Note: thresholds used have been verified to work with adams_ws2 and wink2
+ *
+ * Starts with initial guess provided by user in lpInitial
+ */
+PJ_LP pj_generic_inverse_2D_with_newton_raphson(PJ_XY xy, PJ *P,
```
kbevers: The name is quite long, since there's only one generic inverse function around wouldn't it be fine to simply call it `pj_generic_inverse_2d`?

kbevers (approved):
Minor detail regarding the name of the generic inverse function otherwise this looks good to me. I am happy to see that it works for more than just one projection and I am hopefull that we can extend it to all projections without and inverse at some point.

src/generic_inverse.cpp
rouault: done

kbevers: Thanks. Pedantic follow-up: The name should also be changed in the commit message which now states:

> - Move the generic method initiated from adams_ws2 to a
>   pj_generic_inverse_2D_with_newton_raphson() method
> - Use it in adams_ws2
> - Use it in wink2

rouault: ah right. done

kbevers: We need the ability to suggest changes to commit messages as well :-) Just stumpled on this one by coincidence - figured it may as well be updated.

Issue #2245, opened 2020-05-29
Build failing on Windows: SHGetFolderPathW was deprecated, has finally been removed


chrodger:
#### Problem description

Building PROJ 7.* on Windows with build toolset ~1925 gives this linker error:
LNK2019	unresolved external symbol __imp_SHGetFolderPathW

SHGetFolderPathW has been deprecated for some time and was finally removed from Shell32.dll.
https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetfolderpathw

#### Proposed fix

The new API is here: SHGetKnownFolderPath doc: https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath

PR #2246, opened 2020-05-29
Update filemanager.cpp

chrodger:
CAUTION: Compiled, but not tested.

Related: https://github.com/OSGeo/PROJ/issues/2245

### DIFF
--- a/src/filemanager.cpp
+++ b/src/filemanager.cpp
@@ -1203,12 +1203,21 @@ const char *proj_context_get_user_writable_directory(PJ_CONTEXT *ctx,
     if (ctx->user_writable_directory.empty()) {
         std::string path;
 #ifdef _WIN32
-      wchar_t* wPath;
+#ifdef __MINGW32__
+        std::wstring wPath;
+        wPath.resize(MAX_PATH);
+        if (SHGetFolderPathW(nullptr, CSIDL_LOCAL_APPDATA, nullptr, 0,
+                             &wPath[0]) == S_OK) {
+            wPath.resize(wcslen(wPath.data()));
+            path = NS_PROJ::WStringToUTF8(wPath);
+#else
+        wchar_t* wPath;
         if (SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, nullptr, &wPath) == S_OK){
             std::wstring ws(wPath);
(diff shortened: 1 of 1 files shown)


chrodger:
Travis build failed. Cursory reading seems to indicate that using the new API requires specifying a Windows version?
https://stackoverflow.com/questions/28153059/shgetknownfolderpath-was-not-declared-in-this-scope
https://docs.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers?redirectedfrom=MSDN#macros_for_conditional_declarations
Might be easier to find another workaround...
Will investigate later if I get time.

src/filemanager.cpp
```diff
@@ -1203,12 +1203,12 @@ const char *proj_context_get_user_writable_directory(PJ_CONTEXT *ctx,
     if (ctx->user_writable_directory.empty()) {
         std::string path;
 #ifdef _WIN32
-        std::wstring wPath;
-        wPath.resize(MAX_PATH);
-        if (SHGetFolderPathW(nullptr, CSIDL_LOCAL_APPDATA, nullptr, 0,
-                             &wPath[0]) == S_OK) {
-            wPath.resize(wcslen(wPath.data()));
-            path = NS_PROJ::WStringToUTF8(wPath);
+      wchar_t* wPath;
```
rouault: indentation issue

src/filemanager.cpp
```diff
@@ -1203,12 +1203,12 @@ const char *proj_context_get_user_writable_directory(PJ_CONTEXT *ctx,
...
+        if (SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, nullptr, &wPath) == S_OK){
```
rouault, line 1215: to avoid the mingw32 build failure, we could perhaps keep the old detection code and use if only under #ifdef __MINGW32__ ?

src/filemanager.cpp
```diff
@@ -1203,12 +1203,12 @@ const char *proj_context_get_user_writable_directory(PJ_CONTEXT *ctx,
...
+            std::wstring ws(wPath);
+            std::string str(ws.begin(), ws.end()); // CAUTION: wstring->string conversion will mangle or drop non-ascii characters.
```
rouault: please use instead the NS_PROJ::WStringToUTF8() method as done previously

rouault (changes requested):

Issue #2173, opened 2020-04-22
Incorrect conversion from ETRS89 to OSGB36+ODN with OSGM15 geoid


hernando:
I'm trying to get accurate conversions from ERTS89 (EPSG:4937) to OSG36B+ODN (EPSG:7405) using a geoid model grid that I derived myself from publicly available data (the grid is only 1.3 MB so I could share attach it if necessary). I will give details on how I created the grid later, but basically I followed the instructions of the the grid_tools from PROJ-data, the grid doesn't seem to be the problem anyway.

I'm using the test coordinates provided in the [OS developer pack](https://www.ordnancesurvey.co.uk/documents/resources/OSTN15-OSGM15-DevelopersPack.zip)

#### Example of problem

I modified the WKT string for 7405 to include PROJ4_GRIDS in the vertical datum, then I used this string as input to `cs2cs` to test a few transformations. 

```
$ EPSG_7405=$(projinfo -q -o WKT1:GDAL EPSG:7405)
$ EPSG_7405_ext=$(echo $EPSG_7405 | sed -e 's/Newlyn",2005/Newlyn",2005, EXTENSION["PROJ4_GRIDS", "uk_os_etrs89_osgm15.tif"]/')
$ export PROJ_LIB=paths to proj.db and the grid files

$ echo 49.92226393730 -6.29977752014 100.000 | cs2cs EPSG:4937 "$EPSG_7405" -f "%.6f"
91492.146319    11318.804552 100.000000

$ echo 49.92226393730 -6.29977752014 100.000 | cs2cs EPSG:4937 "$EPSG_7405_ext" -f "%.6f"
91495.585557    11326.276036 46.517545
```
#### Problem description

The reference output for this transformation is 91492.146 11318.804 46.519.
The first transformation gets the horizontal part right, but the vertical component is not transformed. This is expected, as the file referred by `proj.db` for OSGM15 is not found (the file is  OSTN15_OSGM15_GB.txt and I haven't found it anywhere, where does it come from?).
In the second transformation the vertical transformation is almost perfect, but the horizontal transformation is ignoring the horizontal grid shift for ETRS89 to OSGB36.

[...]

PR #2250, opened 2020-06-02
Database: register geoid file for UK added in  OSGeo/PROJ-data#25

and-marsh:
Register geoid file for UK added in OSGeo/PROJ-data#25

There are already existing transformations `ETRS89 to ODN height (2)` with code `EPSG:7711`
 that use this grid `OSTN15_OSGM15_GB.txt`. So that's enough to add only `grid_alternatives` entry


hernando:
Thanks for this PR! This was in my TODO list since I opened #2173. Could you please add:
```
INSERT INTO "geoid_model" SELECT 'OSGM15', auth_name, code FROM grid_transformation WHERE auth_name = 'EPSG' AND grid_name LIKE '%OSGM15%' AND deprecated = 0;
```
to the section for geoid models in `customizations.sql` as well?

and-marsh:
@hernando, thank you!
Of course I will add your suggestion, but first i need to resolve all issues related to https://github.com/OSGeo/PROJ-data/pull/25

PR #2251, opened 2020-06-02
Docs: Use +grids and tif file for deformation examples

kbevers:
In response to https://github.com/OSGeo/PROJ-data/pull/24/


rouault:
For VS2015 issue, see https://github.com/OSGeo/PROJ/pull/2252

PR #2258, opened 2020-06-05
Doc: use Doxygen output for proj_normalize_for_visualization() to avoid duplication of info (current .rst content is outdated)

rouault:


rouault:
Manually backported to 7.0

PR #2267, opened 2020-06-19
Database: update to EPSG v9.8.12

rouault:


jmckenna:
Wow you're fast, before I could file a ticket, you must read my mind :)  thanks!

rouault:
> you must read my mind :)

probably not, but we must be subscribed to the same notification source :-)

PR #2269, opened 2020-06-23
BUG: replace assert with returning data associated with errors in c_api.cpp

### LINKED ISSUE #2268: Assertion causing core dump in c-api?
From the issue here: https://github.com/pyproj4/pyproj/issues/664, it points to this code:

https://github.com/OSGeo/PROJ/blob/af8334a5c8f6ec7d98e7f7ce0564ed4684896426/src/iso19111/c_api.cpp#L7829

Instead of an assert, thoughts on returning NULL?

snowman2:
- [x] Closes #2268

### DIFF
--- a/src/iso19111/c_api.cpp
+++ b/src/iso19111/c_api.cpp
@@ -371,7 +371,9 @@ const char *proj_context_get_database_metadata(PJ_CONTEXT *ctx,
 PJ_GUESSED_WKT_DIALECT proj_context_guess_wkt_dialect(PJ_CONTEXT *ctx,
                                                       const char *wkt) {
     (void)ctx;
-    assert(wkt);
+    if (!wkt) {
+        return PJ_GUESSED_NOT_WKT;
+    }
     switch (WKTParser().guessDialect(wkt)) {
     case WKTParser::WKTGuessedDialect::WKT2_2019:
         return PJ_GUESSED_WKT2_2019;
@@ -447,7 +449,9 @@ PJ *proj_clone(PJ_CONTEXT *ctx, const PJ *obj) {
  */
 PJ *proj_create(PJ_CONTEXT *ctx, const char *text) {
     SANITIZE_CTX(ctx);
-    assert(text);
+    if (!text) {
+        return nullptr;
+    }
 
     // Only connect to proj.db if needed
     if (strstr(text, "proj=") == nullptr || strstr(text, "init=") != nullptr) {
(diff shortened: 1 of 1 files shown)


rouault:
At least in the first occurrences (didn't check all places) where you changed assert(xxxx) by if ( !xxx), the doc of the function specifically mentions that the parameter must not be NULL. The assert() are actually useful so that people know what they did wrong. In the C library, it is common for number of functions to have undefined behaviour is some of the documented preconditions are not met.

snowman2:
In principle, I agree. However, it is causing core dumping in cython code. In that scenario, the usefulness of the assertion is lost in many cases as the error message is lost. That makes debugging the issue difficult. If I added better error log messages for each one, would you feel better about the change?

rouault:
> If I added better error log messages for each one, would you feel better about the change?

yes

src/iso19111/c_api.cpp
```diff
@@ -1125,8 +1155,7 @@ PJ_TYPE proj_get_type(const PJ *obj) {
  * @return TRUE if it is deprecated, FALSE otherwise
  */
 int proj_is_deprecated(const PJ *obj) {
-    assert(obj);
-    if (!obj->iso_obj) {
+    if (!obj || !obj->iso_obj) {
```
snowman2, line 1158: I didn't raise an error here since it didn't pass a context through. There are a couple like this one.

snowman2:
Just updated - I used `"missing required input"` as the error message in most cases for uniformity/clarity.

kbevers:
@rouault Are you satisfied with the solution provided in this PR? Apart from the failing travis target (which I assumes can be fixed without radical changes to the PR). 

I can include this in 7.1.0RC2 if this is merged within the next 48 hours or so.

src/iso19111/c_api.cpp
```diff
@@ -1169,15 +1201,14 @@ PJ_OBJ_LIST *proj_get_non_deprecated(PJ_CONTEXT *ctx, const PJ *obj) {
 static int proj_is_equivalent_to_internal(PJ_CONTEXT *ctx, const PJ *obj,
                                           const PJ *other,
                                           PJ_COMPARISON_CRITERION criterion) {
-    assert(obj);
-    assert(other);
-    if (!obj->iso_obj) {
+
+    if (!obj || !other){
+        proj_log_error(ctx, __FUNCTION__, "missing required input");
```
rouault: As raised by https://travis-ci.com/github/OSGeo/PROJ/jobs/352928854, ctx is nullptr when called from proj_is_equivalent_to()

rouault: so proj_log_error() should only be called if (ctx )

rouault:
> Are you satisfied with the solution provided in this PR?

yes, pending fixing the clang static analyzer issue with my above direction (https://github.com/OSGeo/PROJ/pull/2269#discussion_r444837282)

PR #2274, opened 2020-06-27
ENH: Added temporal, engineering, and parametric datum PJ_TYPE enumerations

### LINKED ISSUE #2273: Basic Temporal Datum Support (c-api)
Highest priority:
- Add `PJ_TYPE_TEMPORAL_DATUM` (identified with proj_get_type) - or something like it. The PJ_TYPE is currently set to unknown. Would be nice to be able to identify it. (Side note: `PJ_TYPE_ENGINEERING_DATUM` and `PJ_TYPE_PARAMETRIC_DATUM` would be nice to have as well.)

Would be nice, but currently can access with PROJ JSON, so not entirely necessary:

- `proj_temporal_datum_get_calendar()`
- `proj_temporal_datum_get_time_origin()`

```
    "TIMECRS[\"Decimal Years CE\",\n"
    "    TDATUM[\"Common Era\",\n"
    "        CALENDAR[\"proleptic Gregorian\"],\n"
    "        TIMEORIGIN[0000]],\n"
    "    CS[TemporalMeasure,1],\n"
    "        AXIS[\"decimal years (a)\",future,\n"
    "            TIMEUNIT[\"year\"]]]"
```

```python
>>> print(crs.datum.to_json(pretty=True))
{
  "$schema": "https://proj.org/schemas/v0.2/projjson.schema.json",
  "type": "TemporalDatum",
  "name": "Common Era",
  "calendar": "proleptic Gregorian",
  "time_origin": "0000"
}
```

What are the current thoughts/plans for this?

snowman2:
- [x] Closes #2273
 - [x] Tests added


snowman2:
@rouault there is the generic `DATUM` type. Is there a datum type that isn't covered by this that might pop up?

snowman2:
Any chance of adding this in 7.1?

kbevers:
> Any chance of adding this in 7.1?

Not unless there's an RC3 - this PR in itself is not enough to trigger that.

src/proj.h
```diff
@@ -710,6 +710,9 @@ typedef enum
     PJ_TYPE_VERTICAL_REFERENCE_FRAME,
     PJ_TYPE_DYNAMIC_VERTICAL_REFERENCE_FRAME,
     PJ_TYPE_DATUM_ENSEMBLE,
+    PJ_TYPE_TEMPORAL_DATUM,
```
rouault: those new values must be added at the end of the enumerations, otherwise this changes the ABI

rouault (changes requested):

snowman2:
> Not unless there's an RC3 - this PR in itself is not enough to trigger that.

From the perspective of pyproj, it is definitely worth it. 7.1 is the minimum version for the next release. It is a change that will address some current bugs.🤞

snowman2:
> It is a change that will address some current bugs.

I guess we could export each object to PROJ JSON and check the type there if this isn't added in 7.1 ... not quite as nice of a solution, but it works.

rouault:
@kbevers Is this material for 7.1.1 backport ? asking since it extends the API

kbevers:
>  asking since it extends the API

Then no. A very similar situation is adressed in this SO post: https://stackoverflow.com/questions/27300561/does-adding-enumerators-into-enum-break-abi/

PR #2278, opened 2020-07-01
improve cmake instructions

anbj:


docs/source/install.rst
```diff
@@ -266,10 +266,11 @@ Create a build directory and step into it::
     mkdir build
     cd build
 
-From the build directory you can now configure CMake and build the binaries::
+From the build directory you can now configure CMake, build and install the binaries::
 
     cmake ..
     cmake --build .
+    make install
```
mwtoews: This only works for "Unix Makefiles" generators. The generic install step is actually:

    cmake --build . --target install

anbj: Ah, I see. Any reason why the ``--target install`` step is omittted?

mwtoews: No reason, so this is a welcome contribution! Just suggesting a modification to this PR so that it is more generic.

anbj: Excellent! Thank you. PR modified accordingly.

docs/source/install.rst
```diff
@@ -266,10 +266,10 @@ Create a build directory and step into it::
     mkdir build
     cd build
 
-From the build directory you can now configure CMake and build the binaries::
+From the build directory you can now configure CMake, build and install the binaries::
 
     cmake ..
-    cmake --build .
```
mwtoews, line 272: Keep this second line in for the regular "build" step. The third new line does the "install" step (while it is confusing that it also has "build" in it too). Configure, build and install are three separate commands.

anbj: Alright, third attempt. Not sure if it's the right way still. Sorry about this.

docs/source/install.rst
```diff
@@ -266,10 +266,11 @@ Create a build directory and step into it::
     mkdir build
     cd build
 
-From the build directory you can now configure CMake and build the binaries::
+From the build directory you can now configure CMake, build and install the binaries::
 
     cmake ..
     cmake --build .
+    cmake --install .
```
mwtoews: the "install" step still needs to look like this:
```suggestion
    cmake --build . --target install
```
(however, your suggestion will be used when CMake 3.15 is the minimum, since that's the version when that switch was introduced). Nearly there!

anbj: So that's how it's done. Thank you very much @mwtoews for your help!

mwtoews (approved):

PR #2283, opened 2020-07-01
Add RFC-6 text: triangulation-based transformations

rouault:


docs/source/community/rfc/rfc-6.rst
```diff
@@ -0,0 +1,416 @@
...
+in the target CRS of the triangulation), they can provide a ``+skip_input_conv`` parameter
+(resp. ``+skip_output_conv``) to avoid the input (resp. output) conversion to be done.
+
+Example, equivalent to the previous one, using the projected coordinates:
+
+::
+
+    $ echo  3210000 6650000 0 |  cct +proj=tinshift +file=./triangulation_kkj.json +skip_input_conv +skip_output_conv
+
+    209948.5283   6647207.3168        0.0000           inf
```
kbevers: Add a time value to avoid the `inf` output, which might cause confusion for readers

docs/source/community/rfc/rfc-6.rst
```diff
@@ -0,0 +1,416 @@
...
+    Yt_i = D + Xs_i * E + Ys_i * F
+
+Pedantic note: from experiments, the interpolation using barycentric coordinates is slightly
+more numerically robust when interpolating projected coordinates of amplitude of the
+order of 1e5 / 1e6, due to computations involving differences of coordinates.
+Whereas the formulation with the A, B, C, D, E, F tends to have big values for
+the A and D constants, and values clause to 0 for C and E, and close to 1 for
+B and F. However, the difference between the two approaches is negligible for
+practical purposes (below micrometre precision)
+
```
kbevers:
```suggestion

.. note:: 

    From experiments, the interpolation using barycentric coordinates is slightly
    more numerically robust when interpolating projected coordinates of amplitude of the
    order of 1e5 / 1e6, due to computations involving differences of coordinates.
    Whereas the formulation with the A, B, C, D, E, F tends to have big values for
    the A and D constants, and values clause to 0 for C and E, and close to 1 for
    B and F. However, the difference between the two approaches is negligible for
    practical purposes (below micrometre precision)

```

kbevers (commented):
This is a very well-written RFC. I don't have anything to contribute apart from a few minor details in formatting.

hobu:
LGTM but I'm a geodetic idiot.

busstoptaktik:
I'm mostly offline these weeks, so I cannot do a proper review, as I cannot get a full view of the code, so my comment below may be based on a misunderstanding of the general working of your implementation. But it appears somewhat over-engineered to me: I seems that you do a partial reimplementation of the sqlite stuff inside the grid format in order to express transformation from geographic coordinates to geographic coordinates in a different datum, directly from the information contained in the mesh file. Then you provide a separate flag to indicate that "input and output coordinates are projected, so skip the pre and post steps".

But this kind of transformations really work *only* for projected coordinates, so the pre and post steps should, imho, be part of the EPSG transformation recipe, and hence stored in the proj sqlite database, not in the triangular mesh file, which should, in my world view contain just a list of coordinates+offsets, a list of triangle compositions, and perhaps, as metadata, information about which two *projected* systems the mesh will transform between. The operation to implement will then be just a triangular interpolation and application of the offsets determined.

Again, my impression of over-engineering may be entirely wrong, due to my limited net-connectivity, and reading code on my phone screen, in which case you should just ignore my comment.

rouault:
@busstoptaktik 

I agree this may be a bit over-engineered.
A few points:
- this kind of transformation could potentially be used with vertices expressed in geographic coordinates. Even if this is unusual. After all, most grids are referenced against a geographic CRS. So I believe we should let that possibility opened.
- I wanted to keep the possibility of having the transformation with input & output coordinates in a geographic CRS, even if internally it uses projected coordinates, because at that point I don't know how this will be registered by EPSG: between the 2 projected CRS, or the correspond geographic CRS.
- Another reason, linked to the above one, is that the transformation is between datums, and thus can apply on any projected CRS based on those datums, not just the 2 used to define the transformations, so having the transformation registered between the geographic CRS is friendly with the existing logic of createOperations().
So the solution I proposed can handle different possibilities. Code-wise, handling those geographic<-->projected CRS conversions is not a huge complication (this is handled in tinshift.cpp).

That said, I'm not opposed to removing the management of those  geographic<-->projected CRS conversions from the tinshift operation. We could in fact potentially handle them on the proj.db side either:
- by creating a concatenated operation that would do geographic -> projected + tinshift +  projected -> geographic, if EPSG registers the triangulation transformation between the projected CRS and we want to have it easily available for the geographic CRS it is based on. There should not be any particular difficulty.
- or if EPSG registers the triangulation transformation between the 2 geographic CRS, we would have to manually override the record in the database to have a +proj=pipeline that chains the forward projection, the tinshift and the inverse projection. Or add some hints "somewhere" in the DB, that the record must be translated that way. That's the most annoying situation actually, which lead me to what I currently proposed. 

We can rediscuss this when you're back to work

rouault:
@busstoptaktik Any thoughts about my above comment ?

busstoptaktik:
@rouault:

> this kind of transformation could potentially be used with vertices expressed in geographic coordinates. Even if this is unusual. After all, most grids are referenced against a geographic CRS. So I believe we should let that possibility opened

I do not think this will work: The TIN-interpolation (or rather: evaluation) works by fitting a plane to the 3 points of a triangle in euclidean space.  So the normal vector determines the slope of the plane. But for a triangle in angular coordinates, the slope is not well defined, as the longitudinal distance between latitudinal points depends on latitude.

For grid interpolation, this is not a problem, since (in principle) we interpolate first between two points at the *same latitude* (top row of the grid cell), then between another two points at the *same latitude* (the bottom row of the grid cell), to get two values at the same *longitude*, i.e. 3 interpolations with stable (but different) metrics.

Additionally, the grid can be made as dense as necessary, and it will always be easy to find the relevant grid cell for any given interpolation point, since the topology is implicit.

To get the same kind of densification of a TIN you will end up with a very large number of triangles, and hence long search times (and/or somewhat complex data structures to reduce the search time).

Of course, if someone have defined a TIN transformation using geographical coordinates as basis, that is their decision, but the geometry of the corrections will be rather strange, and probably not utilize the available data material to its full potential.

rouault:
OK, I've adjusted RFC 6 text and the candidate implementation to restrict to projected coordinates.

busstoptaktik:
> OK, I've adjusted RFC 6 text and the candidate implementation to restrict to projected coordinates

I think this is probably overly restricting: As I mentioned earlier:

> Of course, if someone have defined a TIN transformation using geographical coordinates as basis, that is their decision, but the geometry of the corrections will be rather strange, and probably not utilize the available data material to its full potential.

And I see at least one potentially useful extension, where it would make some kind of sense to use ellipsoidal triangles: The case, where one uses the TIN purely as an interpolation input data search method, while not doing the actual interpolation by evaluation on a fitted plane, but rather by using some kind of geodesic inverse distance weigthing.

rouault:
> And I see at least one potentially useful extension,

Well, I understood from your last comment that it wouldn't make much sense, hence this simplification. But if that case of geographic coordinates arise, it can be dealt as an enhancement. There would be little to change. Something to change the P->left and P->right values to indicate geographic coordinates.

rouault:
> . Something to change the P->left and P->right values to indicate geographic coordinates.

After all, actually just setting PJ_IO_UNITS_WHATEVER is sufficient to handle both cases. I've thus added back the possibility of geographic coordinates. (without the logic to do the conversion geographic<-->projected of my initial thoughts)

busstoptaktik:
Awesome  😊

ccrook:
Geographic coordinates could also be interpolated directly for more local triangulation models, eg deformation close to faulting.  In this case the difference between projected and geographic interpolation would be much less than the uncertainty of the model in the first place, so simply interpolating geographic coordinates on "TIN" may be appropriate.

PR #2303, opened 2020-07-23
update Dockerfile for osgeo/proj-docs to include spell checking

hobu:
Add spell checking capability to Sphinx and catch up to latest.

The `osgeo/proj-docs` docker image has been updated with these additions.

### DIFF
--- a/docs/docbuild/Dockerfile
+++ b/docs/docbuild/Dockerfile
@@ -4,10 +4,11 @@ RUN apt-get -y update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     python3-dev python3-pip g++ doxygen dvipng latexmk \
     cmake libjpeg8-dev zlib1g-dev texlive-latex-base \
     texlive-latex-extra git latex-cjk-all texlive-lang-all \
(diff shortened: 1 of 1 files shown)


kbevers:
Cool. How does this work? Sphinx fails if I've spelled something wrong?

hobu:
you get a list of words and line numbers. I added a ``make spell`` target to PDAL which does this and found a ton of things. https://github.com/PDAL/PDAL/pull/3186

I suppose you could make it fail, but that would involve adding a lot of special words to the dictionary.

kbevers:
> I added a make spell target to PDAL which does this and found a ton of things

Do you mind adding that here as well? As far as I can tell this PR only prepares the Docker image for this, right?

> I suppose you could make it fail, but that would involve adding a lot of special words to the dictionary.

Right, that could be a pain in the ass to maintain. `make spell` is great addition in itself, I am sure I have contributed my fair share of spelling errors!

hobu:
> Do you mind adding that here as well?

This should be merged and a new one opened adding spelling support.

PR #2323, opened 2020-08-15
ENH: Add support for custum CA Bundle path

### LINKED ISSUE #2320: Dynamically set TLS/SSL certificate path
When building wheels for `pyproj` I have discovered that including `curl` is not enough. I am having issues using PROJ_NETWORK since the path to the CA bundle can vary depending on the operating system/virtual environment. 

Here is the error I am running into:
```
Cannot open https://cdn.proj.org/ca_nrc_ntv2_0.tif: error setting certificate verify locations:

  CAfile: /etc/pki/tls/certs/ca-bundle.crt

  CApath: none
```
One idea I have to ensure that users have an up-to-date CA bundle would be to use the python `certifi` library. Doing so would require the capability to dynamically set the path to the CA bundle. I attempted to modify the `CURL_CA_BUNDLE` environment variable and it worked on OSX, but I have not got it working on Linux (yet). It would be nice to be able to scope this so you only set it for the `curl` used by `PROJ`.

Curl has this method that can be used to dynamically set the path & type:
- https://curl.haxx.se/libcurl/c/CURLOPT_SSLCERT.html
- https://curl.haxx.se/libcurl/c/CURLOPT_SSLCERTTYPE.html

This is something that will need to be updated in `networkfilemanager.cpp`. I am thinking adding a method to set this would be useful (`const int proj_curl_set_sslcert(const char*cert_path, const char* cert_type)`).

Thoughts? Suggestions?

snowman2:
- [x] Closes #2320
 - [x] Tests added


src/filemanager.cpp
```diff
@@ -1988,3 +1988,34 @@ void pj_set_searchpath(int count, const char **path) {
...
+ *
+ * @since PROJ 7.2
+ */
+void proj_context_set_ca_bundle_path(PJ_CONTEXT *ctx, const char *path) {
+    if (!ctx)
+        ctx = pj_get_default_ctx();
+    if (!ctx)
+        return;
+    try {
+      ctx->set_ca_bundle_path(path);
```
rouault: Given that the set_ca_bundle_path() takes a std::string, care should be taken if path is nullptr, since this is undefined behaviour to instanciated a std::string from NULL. So:
```suggestion
      ctx->set_ca_bundle_path(path != nullptr ? path : "");
```

src/networkfilemanager.cpp
```diff
@@ -1683,7 +1700,8 @@ PROJ_NETWORK_HANDLE *CurlFileHandle::open(PJ_CONTEXT *ctx, const char *url,
         return nullptr;
 
     auto file =
-        std::unique_ptr<CurlFileHandle>(new CurlFileHandle(url, hCurlHandle));
+        std::unique_ptr<CurlFileHandle>(new CurlFileHandle(
+            url, hCurlHandle, ctx->ca_bundle_path.c_str()));
```
rouault: The original code will always pass a non-null pointer to CurlFileHandle constructor. So this should probably be:
```suggestion
            url, hCurlHandle, ctx->ca_bundle_path.empty() ? nullptr : ctx->ca_bundle_path.c_str()));
```

test/unit/proj_context_test.cpp
```diff
@@ -179,4 +179,24 @@ TEST(proj_context, read_grid_from_user_writable_directory) {
     MyUnlink(filename);
 }
 
+// ---------------------------------------------------------------------------
+
+TEST(proj_context, proj_context_set_ca_bundle_path) {
+
+    std::string dirname;
+    auto filename = createTempDict(dirname, "temp_proj_dic2");
```
rouault: creating the directory and deleting afterwards is probably unneeded. You could pass a completely dummy string

snowman2: Good suggestion. Thanks :+1:

src/filemanager.cpp
snowman2: Thanks :+1:

src/networkfilemanager.cpp
snowman2: Thanks :+1:

src/filemanager.cpp
```diff
@@ -1988,3 +1988,34 @@ void pj_set_searchpath(int count, const char **path) {
...
+
+/** \brief Sets CA Bundle paths.
+ *
+ * Those CA Bundle path will be used by PROJ when curl and PROJ_NETWORK
+ * are enabled.
+ *
+ * If set on the default context, they will be inherited by contexts created
+ * later.
+ *
+ * The path(s) should be encoded in UTF-8.
```
rouault: why plural (s) on path ?

snowman2: Good catch.

snowman2:
Thanks for reviewing @rouault!

PR #2324, opened 2020-08-16
Clone projCppContext context when copying projCtx_t

snowman2:
 
Reasoning:

I made some changes assuming the database path would be copied from the default context. After core dumps, I determined this was not the case. I believe copying the `cpp_context` will enable this. Thoughts?

### DIFF
--- a/include/proj/internal/io_internal.hpp
+++ b/include/proj/internal/io_internal.hpp
@@ -191,6 +191,8 @@ struct projCppContext {
     explicit projCppContext(PJ_CONTEXT *ctx, const char *dbPath = nullptr,
                             const std::vector<std::string> &auxDbPaths = {});
 
+    projCppContext* clone(PJ_CONTEXT *ctx);
+
     // cppcheck-suppress functionStatic
     inline const std::string &getDbPath() const { return dbPath_; }
 
--- a/include/proj/internal/io_internal.hpp
+++ b/include/proj/internal/io_internal.hpp
@@ -191,7 +191,7 @@ struct projCppContext {
     explicit projCppContext(PJ_CONTEXT *ctx, const char *dbPath = nullptr,
                             const std::vector<std::string> &auxDbPaths = {});
 
-    projCppContext* clone(PJ_CONTEXT *ctx);
+    projCppContext* clone(PJ_CONTEXT *ctx) const;
 
     // cppcheck-suppress functionStatic
     inline const std::string &getDbPath() const { return dbPath_; }
--- a/src/iso19111/c_api.cpp
+++ b/src/iso19111/c_api.cpp
@@ -125,7 +125,7 @@ projCppContext::toVector(const char *const *auxDbPaths) {
 
 // ---------------------------------------------------------------------------
 
-projCppContext* projCppContext::clone(PJ_CONTEXT *ctx) {
+projCppContext* projCppContext::clone(PJ_CONTEXT *ctx) const {
     projCppContext* newContext = new projCppContext(
         ctx,
         getDbPath().c_str(),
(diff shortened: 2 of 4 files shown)


include/proj/internal/io_internal.hpp
```diff
@@ -191,6 +191,8 @@ struct projCppContext {
     explicit projCppContext(PJ_CONTEXT *ctx, const char *dbPath = nullptr,
                             const std::vector<std::string> &auxDbPaths = {});
 
+    projCppContext* clone(PJ_CONTEXT *ctx);
```
rouault:
```suggestion
    projCppContext* clone(PJ_CONTEXT *ctx) const;
```

src/iso19111/c_api.cpp
```diff
@@ -125,6 +125,18 @@ projCppContext::toVector(const char *const *auxDbPaths) {
 
 // ---------------------------------------------------------------------------
 
+projCppContext* projCppContext::clone(PJ_CONTEXT *ctx) {
```
rouault:
```suggestion
projCppContext* projCppContext::clone(PJ_CONTEXT *ctx) const {
```

rouault (commented):
I'm not sure if those new code paths are exercised by the existing test suite. If not, that would be welcome

snowman2:
> I'm not sure if those new code paths are exercised by the existing test suite. If not, that would be welcome

Sounds like a good idea. I will take a look at it later.

test/unit/proj_context_test.cpp
```diff
@@ -189,4 +192,20 @@ TEST(proj_context, proj_context_set_ca_bundle_path) {
     proj_context_destroy(ctx);
 }
 
+// ---------------------------------------------------------------------------
+
+TEST(proj_context, proj_context_clone) {
+    std::string dbpath("/tmp/dummmy/path/proj.db");
+    PJ_CONTEXT *default_ctx = pj_get_default_ctx();
+    projCppContext* original_cpp_context = default_ctx->cpp_context;
+    default_ctx->cpp_context = new projCppContext(default_ctx, dbpath.c_str(), {});
```
rouault: you cannot do this. This is an internal symbol of the lib, not accessible by the tests. 
You could possibly avoid calling it directly by using proj_context_set_database_path() instead

snowman2: I thought about that. I tried this route as that method looks like it attempts to connect to the database and reverts if it fails. Sounds like I will need to go that route. Thanks for the explanation for why that failed 👍

test/unit/test_c_api.cpp
```diff
@@ -4038,6 +4049,57 @@ TEST_F(Fixture_proj_context_set_autoclose_database,
...
+    }
+    fwrite(content.data(), 1, content.size(), f);
+    fclose(f);
+
+    auto c_default_path = proj_context_get_database_path(nullptr);
+    EXPECT_TRUE(proj_context_set_database_path(nullptr, tmp_filename.c_str(),
+                                               nullptr, nullptr));
+
+    PJ_CONTEXT *new_ctx = proj_context_create();
+    EXPECT_TRUE(proj_context_set_database_path(nullptr, c_default_path,
```
rouault: There's a heap-use-after-free  error in https://travis-ci.com/github/OSGeo/PROJ/jobs/375153910 that is likely due to this line. c_default_path is likely no longer valid since the call to proj_context_set_database_path() at line 4086. You should store c_default_path in a std::string before it gets invalidated.

snowman2: Good catch :+1:

snowman2:
Thanks for reviewing @rouault :+1:

PR #2329, opened 2020-08-19
Add methods to projCtx_t for safer interactions with cpp_context

snowman2:

Because the member `cpp_context` of `projCtx_t` can be null, methods to ensure safe interactions with `cpp_context` looked like they would be helpful.


rouault:
Nice cleanup!

PR #2331, opened 2020-08-20
projCtx_t: Copy ini file settings, proj4_init_rules, etc.. when initializing context from global

snowman2:
- [x] Tests added

Ran into a couple of other settings that weren't copied from the global context. I figured initializing them here might reduce the overhead of loading in of the data elsewhere when creating new contexts. Thoughts?


test/unit/test_c_api.cpp
```diff
@@ -4797,6 +4797,51 @@ TEST_F(CApi, proj_context_set_sqlite3_vfs_name) {
...
+    proj_context_use_proj4_init_rules(ctx, true);
+    ASSERT_TRUE(proj_context_get_use_proj4_init_rules(ctx, true));
+    proj_context_use_proj4_init_rules(ctx, false);
+    ASSERT_TRUE(!proj_context_get_use_proj4_init_rules(ctx, true));
+    proj_context_destroy(ctx);
+}
+
+// ---------------------------------------------------------------------------
+
+TEST_F(CApi, use_proj4_init_rules__from_global_context) {
```
rouault:
```suggestion
TEST_F(CApi, use_proj4_init_rules_from_global_context) {
```

test/unit/test_defmodel
```diff
@@ -0,0 +1,210 @@
+#! /bin/bash
```
rouault: this file should not be committed. You may need to add it to .gitignore

test/unit/test_network
```diff
@@ -0,0 +1,210 @@
+#! /bin/bash
```
rouault: this file shouldn't be committed

test/unit/test_defmodel
snowman2: :man_facepalming:, yep. Will need to add that to the `.gitignore`. Glad you caught that :+1:

src/ctx.cpp
```diff
@@ -173,12 +173,22 @@ projCtx_t::projCtx_t(const projCtx_t& other)
     logger = other.logger;
     logger_app_data = other.logger_app_data;
     fileapi_legacy = other.fileapi_legacy;
+    use_proj4_init_rules = other.use_proj4_init_rules;
     epsg_file_exists = other.epsg_file_exists;
+    env_var_proj_lib = other.env_var_proj_lib;
```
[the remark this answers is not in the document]
snowman2: Sounds reasonable. Just pushed up the change.

snowman2:
Should I also copy `defer_grid_opening`?

snowman2:
> you may add a comment about that

There is a comment `// set by pj_obj_create()`. So, I think it should be okay. Thanks for clarifying :+1:

rouault:
> // set by pj_obj_create()

Maybe "// set transiently by pj_obj_create()" ?

snowman2:
Thanks for the review @rouault. Very helpful :+1:

PR #2344, opened 2020-09-02
Add a +proj=tinshift for triangulation-based transformations

rouault:
Implements RFC-6

### DIFF
--- a/docs/source/operations/transformations/tinshift.rst
+++ b/docs/source/operations/transformations/tinshift.rst
@@ -1,0 +1,201 @@
+.. _tinshift:
+
+================================================================================
+Triangulated Irregular Network based transformation
+================================================================================
+
+.. versionadded:: 7.2.0
+
++---------------------+--------------------------------------------------------------------+
+| **Alias**           | tinshift                                                           |
++---------------------+--------------------------------------------------------------------+
+| **Input type**      | Projected or geographic coordinates (horizontal), meters (vertical)|
++---------------------+--------------------------------------------------------------------+
+| **Output type**     | Projected or geographic coordinates (horizontal), meters (vertical)|
++---------------------+--------------------------------------------------------------------+
+| **Domain**          | 2D or 3D                                                           |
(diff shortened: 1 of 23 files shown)


rouault:
@kbevers @busstoptaktik Want to review this before it is merged ? (no particular emergency, just asking)

docs/source/operations/transformations/tinshift.rst
```diff
@@ -0,0 +1,250 @@
...
+
+File format
++++++++++++
+
+The triangulation is stored in a text-based format, using JSON as a serialization.
+
+Below a minimal example, from the KKJ to ETRS89 transformation, with just a
+single triangle:
+
+.. code-block:: json
```
kbevers: This seems to be the same content as in `data/tests/tinshift_crs_implicit.json`. You could do a `literalinclude` here in the same way as is done in `docs/source/development/quickstart.rst`.

kbevers (approved):
This looks good to me. I only have a couple of minor details to comment on. One is inlined the other below.

What's the difference between `data/triangulation.schema.json` and `schemas/triangulation.schema.json`? They seem to be the same thing, do we really need both of them?

rouault:
> What's the difference between `data/triangulation.schema.json` and `schemas/triangulation.schema.json`? They seem to be the same thing, do we really need both of them?

They're the same. That's similar to projjson schema. The one in data/ is to be installed with make install in /usr/share/proj . The one under schemas/ is for the website. But I've modified docs/Makefile to copy the one from data/ to the deployed website /schemas

docs/source/operations/transformations/tinshift.rst
rouault: Done

kbevers:
> But I've modified docs/Makefile to copy the one from data/ to the deployed website /schemas

Thanks. I prefer to only have a single copy when possible.

I'm happy to see this merged once the tests have run again. Good job on this implementation of RFC6! I'm sure our Finish friends will be very happy about this.

PR #2359, opened 2020-09-23
Norwegian height transformations added to grid_transformation_custom.sql

himsve:
Dependent resources files are has a pull request at https://github.com/OSGeo/PROJ-data


data/sql/grid_transformation_custom.sql
```diff
@@ -84,3 +84,38 @@ INSERT INTO "grid_transformation" VALUES(
     NULL,
     'EPSG','8666','Geoid (height correction) model file','Icegeoid_ISN2016.gtx',
     NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);
+
+-- Norway
+
+ INSERT INTO "grid_transformation" VALUES(
+    'PROJ','EPSG_4937_TO_EPSG_5776','ETRS89 to NN1954 height',
+    NULL,NULL,
+    'EPSG','9665','Geographic3D to GravityRelatedHeight (tif)',
```
rouault: I'd prefer we stick with the official name for EPSG:9665 'Geographic3D to GravityRelatedHeight (gtx)'  even if it is not super appropriate here, but making a new name while still using EPSG:9665 could be confusing. I guess EPSG will end up attributing a proper code for TIFF in the future (especially if you submit those grids)

data/sql/grid_transformation_custom.sql
```diff
@@ -84,3 +84,38 @@ INSERT INTO "grid_transformation" VALUES(
...
+    'EPSG','5776', -- target CRS (NN1954 height)
+    'EPSG','1352', -- area of use: Norway - Onshore
+    NULL,
+     'EPSG','8666','Geoid (height correction) model file','href2008a.tif',
+    NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);
+
+ INSERT INTO "grid_transformation" VALUES(
+    'PROJ','EPSG_5776_TO_EPSG_5941','ETRS89 to NN2000 height',
+    NULL,NULL,
+    'EPSG','9665','Geographic3D to GravityRelatedHeight (tif)',
```
rouault: same as above

data/sql/grid_transformation_custom.sql
```diff
@@ -84,3 +84,38 @@ INSERT INTO "grid_transformation" VALUES(
...
+    'EPSG','5941', -- target CRS (NN2000 height)
+    'EPSG','1352', -- area of use: Norway - Onshore
+    NULL,
+    'EPSG','8666','Geoid (height correction) model file','HREF2018B_NN2000_EUREF89.tif',
+    NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);
+     
+  INSERT INTO "grid_transformation" VALUES(
+     'PROJ','EPSG_5941_TO_EPSG_5776','NN1954 to NN2000 height',
+     NULL,NULL, 
+     'EPSG','1084', 'Vertical Offset by Grid Interpolation (tif)',
```
rouault: same as above

data/sql/grid_transformation_custom.sql
```diff
@@ -84,3 +84,38 @@ INSERT INTO "grid_transformation" VALUES(
...
+    'EPSG','9665','Geographic3D to GravityRelatedHeight (tif)',
+    'EPSG','4937', -- source CRS (ETRS89 3D)
+    'EPSG','5941', -- target CRS (NN2000 height)
+    'EPSG','1352', -- area of use: Norway - Onshore
+    NULL,
+    'EPSG','8666','Geoid (height correction) model file','HREF2018B_NN2000_EUREF89.tif',
+    NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);
+     
+  INSERT INTO "grid_transformation" VALUES(
+     'PROJ','EPSG_5941_TO_EPSG_5776','NN1954 to NN2000 height',
```
rouault: The name is not consistent with the source and target. It should be 'NN2000 height to NN1954 height' if the values in NNTrans2018B.tif are the valeus to add to a NN2000 height to get a NN1954 one

rouault (commented):
- Did you submit to EPSG your grids ? We don't want to accumulate too much custom transformations in PROJ. See https://epsg.org/dataset-change-requests.html for the procedure to submit to EPSG
- you'll need also to add entries in data/sql/grid_alternatives.sql to register your grids as known and map them to the names in PROJ-data

data/sql/grid_transformation_custom.sql
himsve: According to the method EPSG 8732 (Vertical offset file) the resources file NNTrans2018B.tif has incorrect signs. I will fix the resosurce file in Proj-data.

By the way, data/sql/grid_alternatives.sql is a better solution. We will focus on implementing the norwegian height transformation that way.

rouault: > By the way, data/sql/grid_alternatives.sql is a better solution. We will focus on implementing the norwegian height transformation that way.

For proper implementation, you need both a record in grid_transformation (either coming from EPSG, or a custom entry like that), and a record in the grid_alternative table so that PROJ is aware that the resource declared in grid_transformation is also known by PROJ-data, and provide potential renaming from the original resource name to the one used by PROJ-data, URL where to download it, etc...

data/sql/customizations.sql
```diff
@@ -262,3 +262,47 @@ JOIN compound_crs c ON gt.target_crs_code = c.code AND gt.target_crs_auth_name =
 JOIN geodetic_crs gcrs ON gt.source_crs_auth_name = gcrs.auth_name AND gt.source_crs_code = gcrs.code
 JOIN vertical_crs vcrs on vcrs.auth_name = c.vertical_crs_auth_name AND vcrs.code = c.vertical_crs_code
 WHERE method_auth_name = 'EPSG' AND method_code = '9635' AND gt.deprecated = 0;
+
+ INSERT INTO "grid_transformation" VALUES(
```
rouault: why moving that record ? It was fine in grid_transformation_custom.sql . Otherwise customizations.sql becomes a bit of a kitchen sink

data/sql/customizations.sql
```diff
@@ -262,3 +262,47 @@ JOIN compound_crs c ON gt.target_crs_code = c.code AND gt.target_crs_auth_name =
...
+    NULL,NULL,
+    'EPSG','9665','Geographic3D to GravityRelatedHeight (gtx)',
+    'EPSG','4937', -- source CRS (ETRS89 geographic 3D)
+    'EPSG','5776', -- target CRS (NN1954 height)
+    'EPSG','1352', -- area of use: Norway - Onshore
+    NULL,
+     'EPSG','8666','Geoid (height correction) model file','href2008a.gtx',
+    NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);	
+
+INSERT INTO "grid_alternatives" VALUES(
```
rouault: 2 below records should be in grid_alternatives.sql

data/sql/grid_transformation_custom.sql
```diff
@@ -84,3 +84,27 @@ INSERT INTO "grid_transformation" VALUES(
     NULL,
     'EPSG','8666','Geoid (height correction) model file','Icegeoid_ISN2016.gtx',
     NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);
+
+-- Norway Iceland
```
rouault:
```suggestion
-- Norway
```

data/sql/grid_transformation_custom.sql
```diff
@@ -84,3 +84,27 @@ INSERT INTO "grid_transformation" VALUES(
...
+	
+ INSERT INTO "grid_transformation" VALUES(
+    'PROJ','EPSG_5776_TO_EPSG_5941','ETRS89 to NN2000 height',
```
rouault:
```suggestion
    'PROJ','EPSG_4937_TO_EPSG_5941','ETRS89 to NN2000 height',
```

rouault:
>     * Did you submit to EPSG your grids ?

What about this point ? I'd like to hear that the submission process has started at least, before we add those interim records.

himsve:
Hi. I have been in contact with EPSG and Roger Lott. The EPSG should be an easy implementation in the EPSG database. The committee had the request on the agenda 1st of Oct.

himsve:
The grid transformations are registered in EPSG.org with new codes:

| EPSG | Name | Source CRS | Target CRS |
| - | - | - | - |
| **9484** | ETRS89 to NN54 height | 4937 | 5776 |
| **9485** | ETRS89 to NN2000 height | 4937 | 5941 |

rouault:
> The grid transformations are registered in EPSG.org with new codes:

yes, I've seen them in 10.005, but I don't think we will integrate that version into PROJ as I've found 2 issues in it which I've submitted to EPSG, and hopefully will be addressed into a next EPSG release. Once it is integrated, only the additions in grid_alternatives.sql will be needed.
By the way https://github.com/OSGeo/PROJ-data/pull/46 is in draft mode. Anything pending on your side @himsve  ?

rouault:
Got feedback that the 2 issues I mentionnd w.r.t 10.005 should be fixed in a 10.006 release within one or two weeks

meastp:
@rouault @himsve According to the [EPSG.org website](https://epsg.org/home.html), the current version is 10.008 now. Is this ready to merge? It does look like https://github.com/OSGeo/PROJ-data/pull/46 is also ready to merge?

rouault:
> Is this ready to merge?

It should be modified now that they're are EPSG records for the transformations. The entries in grid_transformation_custom.sql are no longer necessary

rouault:
> The entries in grid_transformation_custom.sql are no longer necessary

I've removed them

meastp:
thank you, @rouault and @himsve :)

PR #2363, opened 2020-10-02
.travis.yml: re-enable s390x and add arm64

rouault:
This reverts commit 11d2e04051001a38f45a27bac7167b483ee94650.

This is now stable again


kbevers:
@rouault is this PR still relevant?

rouault:
yes if you think it is a good idea

kbevers:
I think it's fine. The only downside I can see is that CI runs might get slightly prolonged but I don't think that is a big deal.

PR #2373, opened 2020-10-12
Add Conda package GitHub Action

hobu:
This PR adds a GitHub Action that builds the @conda-forge OSX, Windows, and Linux packages for every commit. In addition to testing a build on all three platforms, it makes the packages available as artifacts for individual download.

### DIFF
--- a/scripts/ci/conda/setup.sh
+++ b/scripts/ci/conda/setup.sh
@@ -1,18 +1,11 @@
 #!/bin/bash
 
 conda update -n base -c defaults conda -y
-
-if [ "$PLATFORM" == "ubuntu-latest" ]; then
-    conda install conda-build ninja compilers automake libtool -y
-    ./autogen.sh
-else
-    conda install conda-build ninja compilers -y
-fi
-
+conda install conda-build ninja compilers -y
 
 pwd
 ls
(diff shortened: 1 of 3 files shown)


scripts/ci/conda/setup.sh
```diff
@@ -0,0 +1,22 @@
...
+cd proj.4-feedstock
+cat > recipe/recipe_clobber.yaml <<EOL
+source:
+  git_url: https://github.com/OSGeo/PROJ.git
+  git_rev: ${GITHUB_SHA}
+  url:
+  sha256:
+
+build:
+  number: 2112
```
rouault, line 19: why 2112 ?

hobu: > why 2112 ?

@warmerdam honoring 😄  More importantly, if someone is actually using these packages, we don't want them to clash with normally-installed ones from @conda-forge that would have build numbers incrementing from 0.

hobu:
OK to merge? I've followed this up with another PR that implements automated generation of the plots and other enhancements to the doc builder that will use the conda package.

rouault:
No objection from me. We'll probably need some mention in the doc to explain where/how to fetch those conda packages

PR #2377, opened 2020-10-14
Plot building Github Action

hobu:
Implement a Github Action builder that plots all of the images and then uses them in a build of the docs.


docs/Makefile
```diff
@@ -1,229 +1,21 @@
...
-clean:
-	rm -rf $(BUILDDIR)/*
-	rm -rf .doxygen_up_to_date
-
-.doxygen_up_to_date: ../Doxyfile ../src/*.cpp ../include/proj/*.hpp
-	../scripts/doxygen.sh
-	touch .doxygen_up_to_date
-
-.PHONY: html
-html: .doxygen_up_to_date
```
rouault, line 60: the html target is highly customized, and the above clean and .doxygen_up_to_date targets need to be retained as well

.github/workflows/doc_build.yml
```diff
@@ -1,21 +1,93 @@
...
-
-      - name: Build doc
+            conda update -n base -c defaults conda
+            conda install descartes shapely geopandas matplotlib proj
+      - name: Plot
+        shell: bash -l {0}
+        working-directory: ./docs/plot
         run: |
-            (cd docs; make html)
+            ./plot.sh
```
rouault, line 27: plot.sh will run with a PROJ version which is not the current one. So if we add a new projection method, this will likely fail.

docs/Makefile
hobu: ok. I ditched the older Makefile so that extension targets from Sphinx would be automatically picked up.

.github/workflows/doc_build.yml
hobu: > plot.sh will run with a PROJ version which is not the current one. 

Indeed. My plan was to get both this and the Conda builder active and then make the plots depend on and use the conda builder.

PR #2383, opened 2020-10-17
ENH: Added proj_context_clone

### LINKED ISSUE #2382: PROPOSAL: Add proj_context_clone
Reasoning:
1, To buffer a library from the potential setting changes made by other libraries on the global context (even though not recommended to change global context settings, they can still do this).
2. To set default context settings on an application global context that will be applied to any new context.

With the `proj_context_clone` method, a library can create their application level global context with `proj_context_create` and then apply any settings they would like. Then, when creating new contexts, they can use `proj_context_clone` to create new contexts.

snowman2:
- [x] Closes #2382
 - [x] Tests added


src/4D_api.cpp
```diff
@@ -1448,11 +1448,18 @@ int proj_errno_reset (const PJ *P) {
...
 PJ_CONTEXT *proj_context_create (void) {
     return pj_ctx_alloc ();
 }
 
+/* Create a new context based on a custom context */
+PJ_CONTEXT *proj_context_clone (PJ_CONTEXT *ctx) {
+    if (nullptr==ctx)
+        return pj_ctx_alloc ();
+
+    return pj_ctx_clone (ctx);
```
rouault: this function should be moved to ctx.cpp() and merged with it

src/ctx.cpp
```diff
@@ -225,6 +225,16 @@ projCtx pj_ctx_alloc()
     return new (std::nothrow) projCtx_t(*pj_get_default_ctx());
 }
 
+/************************************************************************/
+/*                            pj_ctx_clone()                            */
+/************************************************************************/
+
+projCtx pj_ctx_clone( projCtx ctx )
```
rouault: we don't need that function

src/proj_api.h
```diff
@@ -192,6 +192,7 @@ void PROJ_DLL pj_cleanup_lock(void);
 
 void PROJ_DLL pj_set_ctx( projPJ, projCtx );
 projCtx PROJ_DLL pj_ctx_alloc(void);
+projCtx PROJ_DLL pj_ctx_clone( projCtx );
```
rouault: to remove. proj_api.h is legacy and is going to be removed soon

src/proj_symbol_rename.h
```diff
@@ -29,6 +29,7 @@
 #define pj_clear_initcache internal_pj_clear_initcache
 #define pj_compare_datums internal_pj_compare_datums
 #define pj_ctx_alloc internal_pj_ctx_alloc
+#define pj_ctx_clone internal_pj_ctx_clone
```
rouault: to remove

scripts/reference_exported_symbols.txt
```diff
@@ -752,6 +752,7 @@ pj_compare_datums
 pj_context_get_grid_cache_filename(projCtx_t*)
 pj_context_set_user_writable_directory(projCtx_t*, std::string const&)
 pj_ctx_alloc
+pj_ctx_clone
```
rouault: to remove

src/4D_api.cpp
snowman2: :+1:

src/ctx.cpp
snowman2: :+1:

src/proj_api.h
snowman2: :+1:

snowman2:
Are there any remaining changes needed to get this into 7.2? The release is coming soon and it would be nice to have it added :crossed_fingers:

snowman2:
Thanks @rouault :+1:

Issue #1064, opened 2018-06-28
Document ellipsoid parameters


mwtoews:
The [source code for `pj_ell_set.c`](https://github.com/OSGeo/proj.4/blob/master/src/pj_ell_set.c#L36-L64) describes two ellipsoid size parameters ("a", "R"), five ellipsoid shape parameters ("rf", "f", "es", "e", "b"), and seven spherification parameters ("R_A", "R_V", "R_a", "R_g", "R_h", "R_lat_a", "R_lat_g").

Currently I see `+a`/`+b` documented in `usage/projections.rst`, and `+R` via `operations/options/R.rst`.

This task is to make a section and table of ellipsoid parameters in the docs (possibly somewhere under `usage/`).

kbevers:
Yeah, there's a lot that is not described in the docs at the moment with regards to ellipsoids. Last year I gave a PROJ introduction to some of my colleagues which go in some more detail of how we handle ellipsoids. I have a few slides on the ellipsoids, starting [here](https://docs.google.com/presentation/d/1EYGeiYABCWQKvpx-FzgrNS5KnHTwGq-hWpOi3DcirCI/edit?usp=sharing).

A section on determining the shape of the Earth would be nice to have. Possibly as it's own page in the "Usage" chapter. That would allow us to simply link to that section when describing the individual projection parameters, say as part of `operations/options/R.rst` and/or `operations/options/ellps.rst`.

RohitPingale:
@kbevers @mwtoews  Can I work on this issue? Also, the pj_ell_set.c file has been changed.

kbevers:
Yes, you are most welcome to work on this! The relevant code has been moved to https://github.com/OSGeo/PROJ/blob/master/src/ell_set.cpp

RohitPingale:
Sorry for the late update. please look into this [first draft documentation](https://github.com/RohitPingale/PROJ/blob/master/docs/source/usage/Ellipsoid.rst#ellipsoid--transformation)  and suggest the changes. I have added all info that is given in the [cpp file ](https://github.com/OSGeo/PROJ/blob/master/src/ell_set.cpp) and in the shared [presentation](https://docs.google.com/presentation/d/1EYGeiYABCWQKvpx-FzgrNS5KnHTwGq-hWpOi3DcirCI/edit#slide=id.g2977ac0210_0_6). 
[...]

PR #2385, opened 2020-10-20
Ellipsoid Documentation (#1064)

RohitPingale:
- According to my understanding of the first comment made by @mwtoews  on this issue, is that we needed a specific page(under `PROJ/docs/source/usage/`) which explains the parameter used for ellipsoid projection.  So I made a general page that describes a bit about the ellipsoid projection and its parameters with its use cases mentioned in the [cpp file ](https://github.com/OSGeo/PROJ/blob/master/src/ell_set.cpp) and in the shared [presentation](https://docs.google.com/presentation/d/1EYGeiYABCWQKvpx-FzgrNS5KnHTwGq-hWpOi3DcirCI/edit#slide=id.g2977ac0210_0_6) by @kbevers .  Then I tried to link those parameters that I have made in table format to the  `PROJ/docs/source/operations/options/`  but It is failed to redirect to that directory. 

- The image which I used is not taken from any web sources I have made it using external software, so we can use it.

-  I have some doubts about the things which @kbevers would like to see changed:
    1. Do we need to make each parameter documentation(Not in tabular format) inside the index.rst?  And the link to the individual parameter's .rst made in the `docs/source/operations/options/ellps.rst`
     2. If yes, then there are some unmentioned parameter in the  `docs/source/operations/options/` or some changed parameter which was used earlier with different names and now with different names. ( like `a` is a change to `h`  as a semi-major axis)
     
Please correct me if I am wrong somewhere and explain my doubts so I can make changes accordingly. Thank you


docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+   proj-string: ``+proj=merc``
+
+
+
+A complete ellipsoid definition comprises a size (primary) and a shape (secondary)
+parameter.
+
+Ellipsoid Size parameters 
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+.. _ellipsoid_parameters : PROJ/docs/source/operations/options/
```
kbevers, line 23: It's fine to label the section so that it can be linked to from elsewhere in the docs but the labels need to be unique. Also, the path after it doesn't really make any sense in this context. What did you try to achieve with it?

```suggestion
.. _ellipsoid_size_parameters:
```

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+
+	============   ==============================================================
+	Parameter      Description
+	============   ==============================================================
+	+R             Radius of a spherical planet
+	+a             Semi-major axis of an ellipsoidal planet
+	============   ==============================================================
```
kbevers, line 30:
```suggestion
.. option:: +R=<value>

    Radius of a spherical planet

.. option:: +a=<value>

    Semi-major axis of an ellipsoidal planet
```

Also, I would probably avoid using the term "planet" as someone might want to model a celestial bodies that aren't planets. "Object" is probably a better word but someone may have a better suggestion than that.

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Ellipsoid Shape parameters 
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+.. _ellipsoid_parameters : PROJ/docs/source/operations/options/
```
kbevers, line 34:
```suggestion
.. _ellipsoid_shape_parameters :
```

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+
+	============   ==============================================================
+	Parameter      Description
+	============   ==============================================================
+	+rf            Reverse flattening of the ellipsoid
+	+f             Flattening of the ellipsoid
+	+es            Eccentricity squared
+	+e             Eccentricity
+	+b             Semi-minor axis
+	============   ==============================================================
```
kbevers, line 44: Use `.. option::` directives as shown above

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+	Parameter      Description
+	============   ==============================================================
+	+R_A           A sphere with the same surface area and volume as the ellipsoid
+	+R_a           A sphere with R = (a + b)/2   (arithmetic mean)
+	+R_g           A sphere with R = sqrt(a*b)   (geometric mean)
+	+R_h           A sphere with R = 2*a*b/(a+b) (harmonic mean)
+	+R_lat_a=phi   A sphere with R being the arithmetic mean of the corresponding ellipsoid at latitude phi.
+	+R_lat_g=phi   A sphere with R being the geometric mean of the corresponding ellipsoid at latitude phi.
+	============   ==============================================================
```
kbevers, line 62: Use `.. option::` directives as shown above

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+	+e             Eccentricity
+	+b             Semi-minor axis
+	============   ==============================================================
+
+The ellipsoid definition may be augmented with a spherification flag, turning
+the ellipsoid into a sphere with features defined by the ellipsoid.
+
+Ellipsoid Spherification parameters 
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+.. _ellipsoid_parameters : PROJ/docs/source/operations/options/
```
kbevers, line 51:
```suggestion
.. _ellipsoid_spherification_parameters:
```

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+	============   ==============================================================
+	+R_A           A sphere with the same surface area and volume as the ellipsoid
+	+R_a           A sphere with R = (a + b)/2   (arithmetic mean)
+	+R_g           A sphere with R = sqrt(a*b)   (geometric mean)
+	+R_h           A sphere with R = 2*a*b/(a+b) (harmonic mean)
+	+R_lat_a=phi   A sphere with R being the arithmetic mean of the corresponding ellipsoid at latitude phi.
+	+R_lat_g=phi   A sphere with R being the geometric mean of the corresponding ellipsoid at latitude phi.
+	============   ==============================================================
+
+If R is given as size parameter, any shape and spherification parameters
```
kbevers, line 64: Now that we are using option directives we can format and link to options in a nice way:

```suggestion
If :option:`+R` is given as size parameter, any shape and spherification parameters
```

There are several other places in the text where this should be done.

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+given are ignored.
+
+The ellps=xxx parameter provides both size and shape for a number of built in
+ellipsoid definitions.
+
+Local Ellipsoid examples 
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+.. _ellipsoid_parameters : PROJ/docs/source/operations/options/
```
kbevers, line 72:
```suggestion
.. _ellipsoid_list:
```

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+	=WGS84         a=6378137.0      rf=298.257223563    WGS 84
+	=sphere        a=6370997.0      b=6370997.0         Normal Sphere (r=6370997)
+	============   =================================    ============================
+
+If size and shape are given as ellps=xxx, later shape and size parameters
+are are taken into account as modifiers for the built in ellipsoid definition.
+
+While this may seem strange, it is in accordance with historical PROJ
+behavior. It can e.g. be used to define coordinates on the ellipsoid
+scaled to unit semimajor axis by specifying "+ellps=xxx +a=1"
```
kbevers, line 94: This improves formatting

```suggestion
scaled to unit semimajor axis by specifying ``+ellps=xxx +a=1`` 
```

docs/source/usage/Ellipsoid.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Transformation example
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+Spherical earth with radius 7000km:
+
+::
+	proj=merc R=7000000
```
kbevers, line 101: I think it is better to just express these examples in terms of geodetic coordinates. This goes for the rest of the examples below as well.

```suggestion
	proj=latlon R=7000000
```

kbevers (changes requested):
Thanks for opening this as a PR. Makes things much easier! In general I think this is very good. I've made a bunch of suggestions but they mainly concern formatting, the content is self is great.

> Then I tried to link those parameters that I have made in table format to the PROJ/docs/source/operations/options/ but It is failed to redirect to that directory.

I still don't quite understand what you've tried to do here. If you add a commit with the change that doesn't work it is easier to problem solve and give feedback.

> The image which I used is not taken from any web sources I have made it using external software, so we can use it.

Great, good job. If you have a svg-version of the image that would be a good to have as well in case we want to update the figure in the future.

> Do we need to make each parameter documentation(Not in tabular format) inside the index.rst?

No. Put them in `docs/source/usage/Ellipsoid.rst`. What I am asking is simply that the parameter tables in `docs/source/usage/Ellipsoid.rst` is replaced by a bunch of `.. option::` directives. I've put an example in an inlined comment.

As mentioned in one of the inlined comments it would be nice to have the filenames use only lower case letters. It's purely cosmetical but it conforms better with current practices in the docs.

You need to add "ellipsoids" to `docs/source/usage/index.rst` for the new page to be a part of the html and pdf output of the docs.

RohitPingale:
Sorry for the late reply, Actually this is my first PR so I might have not committed properly.  I will make the suggested changes.

PR #2395, opened 2020-10-25
Add +proj=col_urban projection, implementing a EPSG projection method used by a number of projected CRS in Colombia (fixes #589)

### LINKED ISSUE #589: Add Colombia urban projection
Some time ago, several local Colombian CRS have been added to the EPSG database. They use a special "Colombia urban grid" projection, which is based on a certain height above the ellipsoid.

Technical details and sample data can be found in paragraph 1.3.19 (page 94) of

http://www.epsg.org/Portals/0/373-07-2.pdf  

See also http://osgeo-org.1560.x6.nabble.com/gdal-dev-Adding-missing-EPSG-codes-to-PROJ-4-and-GDAL-td5333395.html which depends on the correct implementation in PROJ.4

rouault:


rouault:
@kbevers The only part I'm unsure is the PROJ id name. "colombia_urban" is larger than our habits, but I don't really expect people to have to type PROJ strings using it at hand. They'd rather just use the EPSG CRS codes for the known instances of this projection.

kbevers:
> The only part I'm unsure is the PROJ id name

It is a bit long, yes. The ISO3166 abbreviation for Colombia is "COL", so perhaps "col_urban"? Not sure if that is better. Preferable I'd want a more generic name but this projection seems to be quite specific to Colombia...

rouault:
Renamed to col_urban

kbevers (approved):

nyalldawson:
@rouault does this map to the esri IGAC_Plano_Cartesiano projection? See https://pro.arcgis.com/en/pro-app/help/mapping/properties/igac-plano-cartesiano.htm

rouault:
> @rouault does this map to the esri IGAC_Plano_Cartesiano projection? See https://pro.arcgis.com/en/pro-app/help/mapping/properties/igac-plano-cartesiano.htm

absolutely from what I can see in the records of pe_list_projcs.csv where for example ESRI:102771 just matches EPSG:6247. Mapping added

nyalldawson:
Sweet... another one bites the dust!

gacarrillor:
This one was an excuse for local institutions/users to distrust FOSS4G.
Thank you very much Even! :clap:

PR #2403, opened 2020-11-01
Remove proj_api.h

### LINKED ISSUE #837: Remove proj_api.h as a public header file

kbevers:
Removes `proj_api.h` from the public API. The contents of the header file has been moved to `proj_internal.h` and any references to `proj_api.h` has been changed to `proj_internal.h`. Most functions from `proj_api.h` that are no longer needed to keep support backward compatibility has been removed.

The documentation of proj_api.h has been removed. The only exception to this is the API migration guides which still mention the old API.

I'd like to get this one merged at the start of this development cycle so downstream projects can test their software against a PROJ master that doesn't include `proj_api.h`.

In overview, this PR does the following:

- [x] Remove `src/transform.cpp`.
- [x] Remove unused function declarations from `proj_api.h` (now in `proj_internal.h`).
- [x] Remove or rewire `test/unit/pj_transform_test.cpp`.
- [x] Remove `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` macro.
- [x] Remove projPJ, projCtx, projXY, etc and adjustment function prototypes in the internal API.
- [x] Remove/replace `proj_api.h` fileapi
- [ ] ~~Remove/replace `proj_api.h` logging functions~~ 
- [x] Remove/replace `proj_api.h` context functions
- [x] `multistresstest` rewritten to use `proj.h` API

At some point the logging function `pj_log` should be replaced by the `proj_log_*` functions but that is more suitable to a stand alone PR, as is a bit more complicated than just switching from calling one function to another. In general, it would be nice to give the whole logging side of things a big overhaul (in many cases debug output is useless).

Most if not all of the `pj_init*` functions can also be removed. Most are trivial but one or two touches some central places in PROJ which I don't have the nerve to fiddle about with right now.

I suggest reviewers go through the pull request commit by commit since it would otherwise be too messy to keep track of things.

 - [x] Closes #837
 - [x] Tests removed or modified


rouault:
There would be follow-up cleanups to do (could be part of later pull requests), like removing src/transform.cpp which implements many of the proj_api.h public functions, and cleaning up function declarations from proj_api.h which are no longer implemented.

kbevers:
> There would be follow-up cleanups

Definitely. At first I just wanted to see if this old branch could be reused as a starting point. To some extend that seems to be the case. I'll add a list of things to do to the top post.

hobu:
If `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` is defined anywhere, an `#error` should be set. I don't think the PR does this yet.

kbevers:
I'm getting closer to a finished pull request but there's still a few things missing:

1. `multistresstest` needs to be configured to use the modern API. Or be removed. Not sure what is the best approach. Do we still have a use for testing the threading capabilities?
2. Remove the last remnants from `proj_api.h`, stuff like the `projPJ`, `projXY` and `projUV` datatypes and follow-up adjustments to function prototypes in the internal API.
3. Update the symbols lists (not possible from my mac, I'll do it from a Linux machine once I am happy with the rest of the changes).

rouault:
@kbevers  if you want to include this work on the OSGeo virutal code sprint, you could mention it in https://docs.google.com/presentation/d/1jKInzGnY_9ad4srD4TtfD7jKBjMrcyL51QIR3z13Ov8/edit#slide=id.ga242b60826_2_21

src/4D_api.cpp
```diff
@@ -1389,6 +1389,10 @@ int proj_errno_set (const PJ *P, int err) {
     /* For P==0 err goes to the default context */
     proj_context_errno_set (pj_get_ctx ((PJ *) P), err);
     errno = err;
+
+    // pj_errno was also set by pj_c tx_set_errno(), left out by mistake here?
+    // proj_errno_reset also sets it.
+    pj_errno = err;
```
rouault: do we still have users of pj_errno ? Couldn't we remove it completely ? Can be a follow up task

src/ctx.cpp
```diff
@@ -54,14 +54,19 @@ projCtx pj_get_ctx( projPJ pj )
...
+/** \brief Re-assign a context to a PJ* object.
+ *
+ * This may be useful if the PJ* has been created with a context that is
+ * thread-specific, and is later used in another thread. In that case,
+ * the user may want to assign another thread-specific context to the
+ * object.
+ */
+void proj_assign_context( PJ* pj, PJ_CONTEXT *ctx )
 {
+    // Note we do not deallocate the old context!
```
rouault: this comment doesn't really make sense to me. Could be removed

src/ctx.cpp
```diff
@@ -71,36 +76,20 @@ void pj_set_ctx( projPJ pj, projCtx ctx )
...
-{
-    pj_set_ctx( pj, ctx );
 }
 
 /************************************************************************/
 /*                          createDefault()                             */
 /************************************************************************/
 
-projCtx_t projCtx_t::createDefault()
+pj_ctx pj_ctx::createDefault()
```
rouault, line 87: shouldn't that be PJ_CONTEXT ? can be for a follow-up task

kbevers:
> if you want to include this work on the OSGeo virutal code sprint, you could mention it in

I haven't paid attention to the virtual sprint. What do I gain by adding this to the list? Will someone send me a participatory pizza in celebration of my work? :-) It can of course be added to the list although it's just a coincidence that I have worked on this at roughly the same time as the sprint.

src/proj_internal.h
```diff
@@ -910,7 +899,85 @@ void pj_clear_vgridshift_knowngrids_cache();
...
+PJ_CONTEXT PROJ_DLL *pj_get_ctx( PJ *);
+
+PJ_XY PROJ_DLL pj_fwd(PJ_LP, PJ *);
+PJ_LP PROJ_DLL pj_inv(PJ_XY, PJ *);
+
+PJ_XYZ PROJ_DLL pj_fwd3d(PJ_LPZ, PJ *);
+PJ_LPZ PROJ_DLL pj_inv3d(PJ_XYZ, PJ *);
+
+
+int PROJ_DLL pj_apply_gridshift( PJ_CONTEXT *, const char *, int,
```
rouault: I believe this function can now be removed from here and src/grids.cpp

rouault (approved):
Impressive cleanup ! There are a few things I've noted that would be good to address, but can potentially be left for follow-up PRs. Maybe create a ticket with such tasks.

rouault:
> What do I gain by adding this to the list? Will someone send me a participatory pizza in celebration of my work? :-) It can of course be added to the list although it's just a coincidence that I have worked on this at roughly the same time as the sprint.

ah ah I also joined the sprint in a completely informal way. I guess this could be a way to promote PROJ and make people aware of upcoming changes (which hopefully shouldn't affect people). Up to you

kbevers:
>  ah ah I also joined the sprint in a completely informal way. I guess this could be a way to promote PROJ and make people aware of upcoming changes (which hopefully shouldn't affect people). Up to you

I added it to the list last night. As you say, it might be a good idea to highlight that we are now removing the old API.

src/4D_api.cpp
kbevers: I wasn't sure, I added it here because the two errno-setters behaved differently and I suspected something might break if `pj_errno` wasn't set. It is likely that it can be removed all together, @busstoptaktik has been working towards that for quite some time at least.

I'll start by trying to remove it here and see if that makes any difference.

src/ctx.cpp
kbevers: I don't understand it either, but it was in the old function description so I figured it might have some value. I'll remove it.

src/ctx.cpp
kbevers: It could be, but I decided to follow Thomas' original plan described at the top of `proj.h`:

https://github.com/OSGeo/PROJ/blob/c560b7957664c32e2465e8425abaccc5a6b2607d/src/proj.h#L41-L53

The `PJ` typedef has the `PJconsts` struct behind so having the `PJ_CONTEXT` typedef refer to the `pj_ctx` struct follows the same recipe.

I have no preference for one or the other, but that was my reasoning.

src/4D_api.cpp
kbevers: Seems to make no difference if it's there or not. Removed.

src/proj_internal.h
kbevers: Indeed it can. Done.

src/4D_api.cpp
rouault: my point was about removing ``PROJ_DLL extern int pj_errno;`` completely

kbevers: I know, for now I just wanted to not increase the use of it further. I'll look into how difficult it is to remove altogether now. If is easy I'll do it, if not I'll add an issue about getting it done.

kbevers: Okay, it wasn't that complicated and seems to have gone okay. `pj_errno` removed in e7a66db.

kbevers:
In e7a66db `pj_errno` is removed. The docs for `pj_malloc` says this:

https://github.com/OSGeo/PROJ/blob/c560b7957664c32e2465e8425abaccc5a6b2607d/src/malloc.cpp#L56-L92

I am not sure if something extra should be done here. Remove `pj_malloc` perhaps and simply rely on `malloc`?

rouault:
> Remove `pj_malloc` perhaps and simply rely on `malloc`?

yes, looks a good plan to me.

kbevers:
> yes, looks a good plan to me.

Great. Taken care of in 046270a. I went a bit further and also removed `pj_calloc`, `pj_dalloc` and `pj_dealloc` as they don't really have a purpose any longer as well.

kbevers:
Last chance to get a review in @hobu and whoever else might be interested. I'll merge this pull request tomorrow if I don't receive more comments.

busstoptaktik:
`proj_api.h`'s gone - it's beer o'clock!

PR #2381, opened 2020-10-15
Add multi-line PROJ string export capability, and use it by default in projinfo (unless --single-line is specified) (fixes #1543)

rouault:

PR #2406, opened 2020-11-02
Correct links to github issues in release notes from 7.1.0 to 7.2.0

mast-eu:
The [release notes](https://proj.org/news.html#release-notes) for versions 7.1.0, 7.1.1 and 7.2.0 contain incorrect links to github.com.

For example, the first item of release 7.2.0: 
`Add multi-line PROJ string export capability, and use it by default in projinfo (unless --single-line is specified) (#2381)`
links to `https://github.com/OSGeo/issues/2381` (which gives a 404 error), but should be linking to `https://github.com/OSGeo/PROJ/issues/2381` instead.

Release notes for versions 7.0.1, and earlier, are correct.
For example:
`Database: update to EPSG v9.8.9 (#2141)` links to `https://github.com/OSGeo/PROJ/issues/2141`.

### DIFF
--- a/HOWTO-RELEASE
+++ b/HOWTO-RELEASE
@@ -179,7 +179,7 @@ See step 1.6 above. Do not rename the archives with RC postfixes.
 
     These vim substitutions are helpfull when converting NEWS to rst:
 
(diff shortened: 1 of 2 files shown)


kbevers:
Thanks for fixing this! I noticed it myself yesterday but did not have the time to sort it out

PR #2417, opened 2020-11-06
COPYING: update and install COPYING, NEWS and AUTHORS

### LINKED ISSUE #2415: License only covering Proj.4?
The file COPYING contains information about licenses. As stated in master right now (2020-11-05) it only applies to Proj.4.

I assume license applies to the current code base and should be updated to reflect that.

I also suggest renaming COPYING to something more obvious such as LICENSE or LICENSE.TXT, but that's up to you. ("License" feels like a broader and more correct term than "copying".)

I would also prefer if the license information was part of the installation, similar to how GDAL (and other libraries) handle it:
/usr/share/doc/gdal-libs-1.11.4/LICENSE.TXT

rouault:
Fixes #2415

- Change mention of PROJ.4 to PROJ
- Change 'I' pronoun to Frank Warmerdam


COPYING
```diff
@@ -1,18 +1,37 @@
...
-Though I have put my own name as copyright holder, I don't mean to imply
-I did the work.  Essentially all work was done by Gerald Evenden. 
+Initial PROJ 4.3 public domain code was put as Frank Warmerdam as copyright
+holder, but he didn't mean to imply he did the work. Essentially all work was
+done by Gerald Evenden.
 
  --------------
 
- Copyright (c) 2000, Frank Warmerdam
+Copyright:             2008-2020, Even Rouault <even.rouault at spatialy.com>
```
kbevers:
```suggestion
Copyright:             2008-2020, Even Rouault <even.rouault at spatialys.com>
```

rouault: good catch. can't even spell correctly my company name lol

danme74:
Ok, fine. Any idea why the COPYING file isn't installed as part of the installation? (It usually is?)

COPYING
```diff
@@ -1,18 +1,37 @@
...
+holder, but he didn't mean to imply he did the work. Essentially all work was
+done by Gerald Evenden.
 
  --------------
 
- Copyright (c) 2000, Frank Warmerdam
+Copyright:             2008-2020, Even Rouault <even.rouault at spatialys.com>
+                       2016-2020, Kristian Evers
+                       2016-2018, Thomas Knudsen
+                       2016-2018, SDFE
```
kbevers: To be exact

```suggestion
                       2016-2018, Danish Agency for Data Supply and Efficiency
```

I guess the period should be extended to 2020 as well since the majority of my contributions is done on company time.

kbevers:
Thinking about it I am not sure about the list of copyright holders. Are we going to update the copyright info with every release? Is it good enough if we say that a list of contributors can be extracted from git if someone needs it? Looking for a low maintenance solution here. `COPYING` is unchanged for the last 20 years, I'd be happy to not touch it again for another 20 years :)

rouault:
> Thinking about it I am not sure about the list of copyright holders. Are we going to update the copyright info with every release? Is it good enough if we say that a list of contributors can be extracted from git if someone needs it?

OK, I removed the copyright holder list and replaced by "Copyright information can be found in source files."

kbevers (approved):
Looks good to me!

PR #2420, opened 2020-11-10
cct: allow @filename syntax

rouault:
Similarly as for projinfo, allow "cct @filename" to mean read filename and
use its content as if it was provided inline. Useful for WKT or PROJJSON

And a tiny improvements, when the object definition contains ':', only try
proj_create_from_database() if the left part (authority name) matches a known
authority, to avoid a warning.


src/apps/cct.cpp
```diff
@@ -79,6 +79,9 @@ Thomas Knudsen, thokn@sdfe.dk, 2016-05-25/2017-10-26
 #include <stdarg.h>
 #include <iostream>
 
+#include <fstream> // std::ifstream
+#include <iostream>
```
kbevers, line 82: Also included above in line 80

src/apps/cct.cpp
```diff
@@ -293,20 +296,55 @@ int main(int argc, char **argv) {
...
+                    }
+                }
+                proj_string_list_destroy(authorityList);
+            }
         }
         if( P == nullptr ) {
             /* if we didn't get a auth:code combo we try to see if the input matches */
             /* anything else */
-            P = proj_create(nullptr, o->fargv[0]);
+            P = proj_create(nullptr, input.c_str());
```
kbevers, line 346: Good call, I should have fixed that myself!

kbevers (approved):
Great addition, I didn't even think about this

PR #2429, opened 2020-11-17
cs2cs, cct, proj and geod: fflush(stdout) after each line to emit each result as …

rouault:
…soon as it is produced

This is needed when working with pipes, when stdout is not an interactive terminal,
and thus the behaviour is to have it buffered as a regular file, whereas with
an interactive terminal, each newline character causes an implicit flush.

### DIFF
--- a/src/apps/cct.cpp
+++ b/src/apps/cct.cpp
@@ -470,6 +470,8 @@ int main(int argc, char **argv) {
                    decimals_distances, point.xyzt.z,
                    point.xyzt.t, comment_delimiter, comment
             );
+        if( fout == stdout )
+            fflush(stdout);
     }
 
     proj_destroy(P);
--- a/src/apps/cs2cs.cpp
+++ b/src/apps/cs2cs.cpp
@@ -217,6 +217,7 @@ static void process(FILE *fid)
             printf("%s", s);
         else
             printf("\n");
+        fflush(stdout);
     }
 }
 
--- a/src/apps/geod.cpp
+++ b/src/apps/geod.cpp
@@ -124,6 +124,7 @@ process(FILE *fid) {
 				(void)fputs(rtodms(pline, al21, 0, 0), stdout);
 		}
 		(void)fputs(s, stdout);
+                fflush(stdout);
 	}
 }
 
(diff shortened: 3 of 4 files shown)


kbevers:
Does this not apply to `proj` and `geod`?

rouault:
> Does this not apply to `proj` and `geod`?

good point. I've just patched them too

rouault:
manually backported

busstoptaktik:
> This is needed when working with pipes, when stdout is not an interactive terminal

@rouault: Is this really needed? I have always done roundtrip-tests like this:

```
$ type foo

12 55 0 0
12 56 0 0
12 57 0 0

$ type foo | cct proj=utm zone=32 -- | cct -I proj=utm zone=32 --

 11.9999999994   55.0000000000        0.0000        0.0000
 11.9999999998   55.9999999996        0.0000        0.0000
 12.0000000000   56.9999999998        0.0000        0.0000
```
this works without any problems, even using `cmd.exe`: The second `cct` instance in the pipe blocks until the first autoflushes - either because buffer is full, or because it reaches `eof`. And in this example, `stdout` of the first `cct` instance is not an interactive terminal.

rouault:
> Is this really needed?

the issue of the reporter was that lines were not flushed as soon as they were emitted. Whole flushing when the buffer is full or the process is finished indeed worked before my change.

busstoptaktik:
So the problem occurs in a more specific situation, not detailed in the PR description.

I have not done any timing (and do not intend to), but it would be nice to have had some more context of the problem addressed, since flushing takes time (although probably not much, compared to the awfully slow formatting of floats, which in some cases dominates the cct run time).

I'm not against this PR - I just have a feeling it targets a very narrow use case, and hence, I think the rationale should have been more clear in the description (which I find misleading - or at least which seriously puzzled me. Blame it on my stupidity, if you wish).

rouault:
Sorry for the short PR description. This is meant to address https://lists.osgeo.org/pipermail/proj/2020-November/009921.html

busstoptaktik:
Thanks - I hadn't seen that, but it was actually exactly the one use case I could imagine, when I tried to guess the rationale. I think it makes good sense to support this kind of work.

PR #2430, opened 2020-11-17
github actions: update to conda-incubator/setup-miniconda@v2 to be compatible with latest github action security fix constraints

rouault:


kbevers:
Is this why I get emails about something failing every time I push a new commit?

rouault:
> Is this why I get emails about something failing every time I push a new commit?

probably. I've done similar changes in GDAL and PDAL earlier

rouault:
Manually cherry-picked to 7.2

PR #2444, opened 2020-11-22
Add +proj=topocentric geocentric->topocentric conversion (fixes #500)

### LINKED ISSUE #500: Add transformation to ENU coordinate system

rouault:


docs/source/operations/conversions/topocentric.rst
```diff
@@ -0,0 +1,104 @@
+.. _topocentric:
+
+================================================================================
+Geocentric to topocentric conversion
```
kbevers, line 4: I am missing some maths or a reference to some math somewhere on this page. There's probably something on Wikipedia or in a EPSG guidance note.

docs/source/operations/conversions/topocentric.rst
```diff
@@ -0,0 +1,104 @@
...
+  Public domain
+..
+
+The topocentric origin is a required parameter of the conversion, and can be
+expressed either as geocentric coordinates (``X_0``, ``Y_0`` and ``Z_0``) or
+as geographic coordinates (``lat_0``, ``lon_0``, ``h_0``).
+
+When conversion between geographic and topocentric coordinates is desired, the
+topocentric conversion must be preceded by the :ref:`cart` conversion to
+perform the initial geographic to geocentric coordinates conversion.
```
kbevers, line 46: I most often see ENU-coordinates exressed in degrees (e.g. the model input for +proj=deformation which I believe to be similar to this) would that then be a matter of doing a pipeline like `+proj=pipeline +step ... +step +proj=topocentric ... +step +proj=cart +inv`? If that is the case, I think an example for that would be useful too.

docs/source/operations/conversions/topocentric.rst
```diff
@@ -0,0 +1,104 @@
...
+
+.. option:: +lat_0=<value>
+
+    Latitude of topocentric origin (in degree)
+
+.. option:: +lon_0=<value>
+
+    Longitude of topocentric origin (in degree)
+
+.. option:: +h_0=<value>
```
kbevers, line 111: Are these mutually exclusive with the parameters above?

kbevers (approved):
Seems good to me. There's few minor details in the documentation that would be nice to sort out otherwise I have no problems with merging this.

docs/source/operations/conversions/topocentric.rst
rouault: reference to EPSG guidance note 7.2 added

docs/source/operations/conversions/topocentric.rst
rouault: yes, note added regarding that

docs/source/operations/conversions/topocentric.rst
```diff
@@ -0,0 +1,117 @@
...
+    Geocentric Z value of the topocentric origin (in metre)
+
+.. note::
+
+    Those options are mutually exclusive with the ones to express the origin as geographic coordinates.
+
+Topocentric origin described as geographic coordinates
+------------------------------------------------------
+
+(mutually exclusive with the options to express the origin as geocentric coordinates)
```
kbevers:
```suggestion
.. note:: Mutually exclusive with the options to express the origin as geocentric coordinates
```

kbevers: Wait, you've already done that above and below. I think this line should just be removed.

rouault: ah I had added this twice. Fixed

docs/source/operations/conversions/topocentric.rst
rouault: Hum, from my testing, it doesn't seem that using +proj=cart +inv would give the expected result. In defmodel_impl.hpp, there's the function
```
inline void DeltaEastingNorthingToLongLat(double cosphi, double de, double dn,
                                          double a, double b, double es,
                                          double &dlam, double &dphi) {
    const double oneMinuX = es * (1 - cosphi * cosphi);
    const double X = 1 - oneMinuX;
    const double sqrtX = sqrt(X);
    dlam = de * sqrtX / (a * cosphi);
    dphi = dn * a * sqrtX * X / (b * b);
}
```
and if doing for example "echo 4.99 55.1 73 | src/cct -d 3 +proj=pipeline +step +proj=affine +step +proj=cart +ellps=WGS84 +step +proj=topocentric +ellps=WGS84 +lon_0=5 +lat_0=55 +h_0=200", and injecting the resulting de, dn into it, it gives the expected dlon ~= -0.001 and dlat ~= 0.1. Whereas using +proj=cart +inv gives completely different and wrong results.

I'd suspect DeltaEastingNorthingToLongLat() to be an approximation only valid for small deltas. Maybe @ccrook can give some hints ?

PR #2450, opened 2020-11-24
Add option to allow export of Geographic/Projected 3D CRS in WKT1_GDAL

rouault:
as CompoundCRS with a VerticalCRS being an ellipsoidal height, which is
not conformant. But needed for LAS 1.4 that only supports WKT1

### DIFF
--- a/docs/source/apps/projinfo.rst
+++ b/docs/source/apps/projinfo.rst
@@ -23,6 +23,7 @@ Synopsis
     |    [--grid-check none|discard_missing|sort|known_available]
     |    [--pivot-crs always|if_no_direct_transformation|never|{auth:code[,auth:code]*}]
     |    [--show-superseded] [--hide-ballpark]
+    |    [--allow-ellipsoidal-height-as-vertical-crs]
     |    [--boundcrs-to-wgs84]
     |    [--main-db-path path] [--aux-db-path path]*
     |    [--identify] [--3d]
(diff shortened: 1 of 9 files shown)


jjimenezshaw:
Thanks @rouault . How does it behave with projected CRSs **not** in meters? In the US it is common to use `ft` or `ftUS`. In this case, it makes sense to use the same unit for the three axes. Even with geographic coordinates, it could be needed with feet. (this is something I miss from `promoteTo3D`)

rouault:
> How does it behave with projected CRSs **not** in meters?

it will indicate whatever unit is used for the vertical axis

jjimenezshaw:
> it will indicate whatever unit is used for the vertical axis

This PR is about the export to WKT1, not how this vertical axis was generated. Sorry for the misunderstanding.

msmitherdc:
@rouault This fix worked well, can we target it for 7.2.1 vs 8.0?

kbevers:
> This fix worked well, can we target it for 7.2.1 vs 8.0?

Nope. New features does not go into bug fix releases.

msmitherdc:
I would counter that this isn't really a new feature, its a restoration of previous (if not quite conformant) behavior. We have a fair amount of that was written this way

rouault:
For what is worth, I'd be OK to backport that in 7.2.1. The new code paths don't change existing behavior (you need to specify the new option), so there's no backward incompatibility

kbevers:
A new options is added to `projinfo` and two new functions to the C++ API. It may be fixing a problem but it is also introducing new functionality. Is this really so important that it can't way two extra months?

msmitherdc:
This is a function that PDAL uses and its preventing conversion of LAS 1.4 pointclouds due to the way that WKT is stored in LAS1.4. So it does have an impact.

rouault:
> A new options is added to `projinfo` and two new functions to the C++ API.

I could possibly make the new C++ methods internal only and remove the new projinfo option in the 7.2 backport. PDAL through GDAL will only needs the new option value in the "char**" options of proj_as_wkt()

kbevers:
Sounds like a good solution to me

msmitherdc:
Good solution, thanks @rouault

PR #2466, opened 2020-11-28
cs2cs: add --area and --bbox  options to restrict candidate coordinate operations (fixes #2423)

### LINKED ISSUE #2423: Add --area option to cs2cs
In most cases `cs2cs` does a good job of selecting the correct transformation when several options are available. The selection is based on the input coordinate and the areas of use of the transformations. There are however situations where `cs2cs` get's confused:

1. When there are overlapping areas of use, a good example of that is Copenhagen which is within the bounding boxes for both Danish and Swedish areas of use
2. If the input coordinates come in an unexpected format like geocentric coordinates

The first issue can't be fixed and the second could be fixed but with added computational complexity. Both can be avoided if a specific area of use can be given to `cs2cs`. The same mechanism as used in `--area` of `projinfo` would be suitable.

rouault:
CC @kbevers 
This works exactly as with projinfo. Note that --area could potentially not give expected results, since it ends up being a bounding box. So if for example you have extent E1 and E2, and E1 is contained within E2, and you specify --area E1, then operations for E2 could still be used. But if E1 and E2 only overlaps, this should work as expected


kbevers (approved):
I can confirm that this works as I intended in the feature request. The caveat with smaller areas being fully contained in a larger area is of course unfortunate but in reality likely not a big issue. I am happy with this solution. Thanks!

PR #2469, opened 2020-11-29
Inverse tmerc spherical: fix wrong sign of latitude when lat_0 is used (fixes #2468)

### LINKED ISSUE #2468: Spherical tmerc (still) gives inverted latitude
This is a request to reopen / revisit #138 – the merged solution #440 is incomplete.

#### Example of problem

```
$ echo 3625903.500000 88781.164062 | invproj +proj=tmerc +ellps=clrk66 +x_0=10000.0 +y_0=20000.0 +lat_0=10.0s +lon_0=10.0e
41d11'50.275"E	8d1'53.248"S

$ echo 3625903.500000 88781.164062 | invproj +proj=tmerc +ellps=sphere +x_0=10000.0 +y_0=20000.0 +lat_0=10.0s +lon_0=10.0e
41d14'32.957"E	8d2'24.724"N

```
#### Problem description

The two commands above should give roughly the same latitude. Instead, the spherical solution is inverted.

The spherical tmerc inverse projection uses an arc-sine on a square root, which is always positive. A line of code is used to correct the sign of the resulting latitude. Prior to #440 the line was:

```C
 if (xy.y < 0.) lp.phi = -lp.phi;
```

However, this correction gave awry results when a latitude of origin was present. #440 changed the line to:

```C
if (xy.y < 0. && -lp.phi+P->phi0 < 0.0) lp.phi = -lp.phi;
```

As the example demonstrates, this construction still misses many cases. I believe the correct solution is:

```C
lp.phi = copysign(lp.phi, P->phi0 + xy.y / Q->esp);
```

This ensures that the sign of the result always matches the sign of the argument to the cosine (*D* in the terminology of the USGS Working Manual, page 60).

#### Expected Output

```
$ echo 3625903.500000 88781.164062 | invproj +proj=tmerc +ellps=sphere +x_0=10000.0 +y_0=20000.0 +lat_0=10.0s +lon_0=10.0e
41d14'32.957"E	8d2'24.724"S
```

#### Environment Information
 - Rel. 8.0.0, March 1st, 2021
 - macOS 11.0

#### Installation method

git master

rouault:
Corrected formula given by @evanmiller

### DIFF
--- a/src/projections/tmerc.cpp
+++ b/src/projections/tmerc.cpp
@@ -203,11 +203,13 @@ static PJ_LP approx_s_inv (PJ_XY xy, PJ *P) {
         return proj_coord_error().lp;
     }
     g = .5 * (h - 1. / h);
-    h = cos (P->phi0 + xy.y / Q->esp);
+    /* D, as in equation 8-8 of USGS "Map Projections - A Working Manual" */
+    const double D = P->phi0 + xy.y / Q->esp;
+    h = cos (D);
     lp.phi = asin(sqrt((1. - h * h) / (1. + g * g)));
 
     /* Make sure that phi is on the correct hemisphere when false northing is used */
-    if (xy.y < 0. && -lp.phi+P->phi0 < 0.0) lp.phi = -lp.phi;
+    lp.phi = copysign(lp.phi, D);
 
     lp.lam = (g != 0.0 || h != 0.0) ? atan2 (g, h) : 0.;
     return lp;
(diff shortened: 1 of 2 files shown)


evanmiller:
I would call the variable D or d so that it is more clear to anyone familiar with the equations. In the basic formulation, the argument is passed to both sine and cosine:

image

([USGS Working Manual](https://pubs.usgs.gov/pp/1395/report.pdf) p. 60). Otherwise looks good. Thanks for the swift attention to this!

rouault:
> I would call the variable D

done

PR #2471, opened 2020-11-29
Spherical tmerc forward: do not restrict to [-90,90] longitude range

rouault:
The restriction was a copy&paste from the Evenden/Snyder approximate
ellipsoidal implementation, but the spherical one is exact, so this
restriction isn't needed.

Also tune a bit the handling of lat=0, |lon| > 90

@cffk does that look good to you ?


cffk:
Yes, this is fine.  I only gave it a cursory look, though!

PR #2488, opened 2020-12-16
Improved coordinate operation filtering

rouault:
- projinfo: add a --accuracy option to define the minimum accuracy
- proj_create_crs_to_crs_from_pj(): add ACCURACY and ALLOW_BALLPARK options


kbevers:
Good addition! Could also be useful in `cs2cs` I think.

rouault:
> Good addition! Could also be useful in `cs2cs` I think.

Done

PR #2494, opened 2020-12-21
Implementing the NKG transformations in proj.db

kbevers:
This adds the NKG 2008 and 2020 transformations to proj.db. The NKG
transformations offers transformations between global reference frames
and the national realisations of ETRS89 in Denmark, Estonia, Finland,
Latvia, Lithuania, Norway and Sweden.

The 2008 transformations are already implemented in the NKG 2008 file
but will now be more accessible with the modern API.

The 2020 transformations are new to PROJ and offers and updated version
of the 2008 transformations using a new and improved deformation model
(eu_nkg_nkgrf17vel.tif). A 2020 version of the NKG transformations are
currently not available for Norway but will in all likelyhood be
included at a later point in time.

Note that the `nkg.gie` file is included in the repo but not added to the test suite (yet?) as it relies on a couple of grids that I don't want to include here and are not smart enough to reduce to a subset that will be suitable for the test-suite. 

I intend to back-port this to the 7.2 branch as well, so that the transformations will be available on system where PROJ 7 is an option (I expect some packaging systems will be late in adopting 8.0.0).

### DIFF
--- a/test/gie/nkg.gie
+++ b/test/gie/nkg.gie
@@ -1,0 +1,240 @@
+<gie-strict>
+
+# -------------------------------------------------------------------------------
+#                                      NKG
+# -------------------------------------------------------------------------------
+operation   operation urn:ogc:def:coordinateOperation:NKG::ITRF2000_TO_NKG_ETRF00
+tolerance   1 mm
+
+accept      3541657.3778 948984.2343 5201383.5231 2020.5
+expect      3541657.9311 948983.7980 5201383.2227 2020.5
+
+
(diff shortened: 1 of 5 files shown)


test/gie/nkg.gie
```diff
@@ -0,0 +1,240 @@
+<gie>
```
rouault: it would be better to use ``<gie-strict>`` mode for stronger syntax validation

kbevers: Done

rouault (approved):

PR #2496, opened 2020-12-28
C API: add proj_crs_is_derived()

rouault:


kbevers:
Background info here: https://lists.osgeo.org/pipermail/proj/2020-December/010006.html

PR #2507, opened 2021-01-21
Add documentation entry for proj_content_set_search_paths

guiguite:

### DIFF
--- a/docs/source/development/reference/functions.rst
+++ b/docs/source/development/reference/functions.rst
@@ -783,6 +783,17 @@ Setting custom I/O functions
 
 .. doxygenfunction:: proj_context_set_sqlite3_vfs_name
    :project: doxygen_api
+   
+.. c:function:: void proj_context_set_search_paths (PJ_CONTEXT *ctx , int count_paths, const char* const* paths)
+
+    Sets the resource files search paths.
+
+    :param ctx: Threading context.
+    :type ctx: :c:type:`PJ_CONTEXT` *
+    :param `count_paths`: Number of paths.
+    :type `count_paths`: `int`
+    :param `paths`: Array of paths.
+    :type `paths`: `const char* const*`
 
 
 Network related functionality


docs/source/development/reference/functions.rst
```diff
@@ -783,6 +783,17 @@ Setting custom I/O functions
...
+.. c:function:: void proj_context_set_search_paths (PJ_CONTEXT *ctx , int count_paths, const char* const* paths)
+
+    Sets the resource files search paths.
+
+    :param ctx: Threading context.
+    :type ctx: :c:type:`PJ_CONTEXT` *
+    :param `count_paths`: Number of paths.
+    :type `count_paths`: `int`
+    :param `paths`: Array of paths.
+    :type `paths`: `const char* const*`
```
kbevers, line 796:
```suggestion
    :project: doxygen_api
```

I believe that's a solution, as the function is already documented in the code https://github.com/OSGeo/PROJ/blob/5e077729274f5d28e137e1a41f7d3350146614ef/src/filemanager.cpp#L1826-L1846)

docs/source/development/reference/functions.rst
```diff
@@ -783,6 +783,10 @@ Setting custom I/O functions
 
 .. doxygenfunction:: proj_context_set_sqlite3_vfs_name
    :project: doxygen_api
+   
+.. c:function:: void proj_context_set_search_paths (PJ_CONTEXT *ctx , int count_paths, const char* const* paths)
+
+    :project: doxygen_api
```
kbevers:
```suggestion
.. doxygenfunction:: proj_context_set_search_paths
    :project: doxygen_api
```

kbevers:
Reverted to original commit. I can't for the life of me figure out why Doxygen won't include this functinon the docs. See #2540.

snowman2:
I bet this is it https://github.com/OSGeo/PROJ/pull/2162#discussion_r410681043:
```
Doxygen complains about this function and the other not being documented. This is because currently they must be within a //! @cond Doxygen_Suppress block. So either move them in a place in the file close to already exported method, or surround them with "//! @endcond" and "//! @cond Doxygen_Suppress"
```

kbevers:
I actually did try to move them above the Doxygen_Supress block without luck but there's maybe a few more things to try. Thanks for digging that out.

PR #2513, opened 2021-02-03
Better grid_alternatives entries to include 'GDA94 to AHD (Tasmania) height' using 'au_ga_AUSGeoid98.tif'

jjimenezshaw:
- [x] Added clear title that can be used to generate release notes (I am not sure)

The other transformations are named `GDA94 to AHD height (...)` for EPSG:5711. However in Tasmania, it is called `GDA94 to AHD (Tasmania) height (1)` for EPSG:5712.
The grid file for Tasmania is `SK55_DAT.htm`

I assume that `au_ga_AUSGeoid98.tif` includes Tasmania (SK55_DAT.htm), as many other S*_DAT.htm files. If it is not the case, close this PR.

### DIFF
--- a/data/sql/customizations.sql
+++ b/data/sql/customizations.sql
@@ -18,7 +18,7 @@ INSERT INTO grid_alternatives(original_grid_name,
            0,
            NULL,
            'https://cdn.proj.org/au_ga_AUSGeoid98.tif', 1, 1, NULL FROM grid_transformation WHERE
-                grid_name LIKE '%DAT.htm' AND name LIKE 'GDA94 to AHD height%';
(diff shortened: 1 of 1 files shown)


rouault:
> I assume that `au_ga_AUSGeoid98.tif` includes Tasmania

Maybe you can check this by sampling a few values in the bounding box of Tasmania ?

jjimenezshaw:
> > I assume that `au_ga_AUSGeoid98.tif` includes Tasmania
> 
> Maybe you can check this by sampling a few values in the bounding box of Tasmania ?

Oh, that I already did in cdn.proj.org. The area of Tasmania is covered by the GeoTIFF. The statement was more if the data in the GeoTIFF is the proper one for for EPSG:5712 (that it looks like).

PR #2514, opened 2021-02-03
Enable linking against static cURL on Windows

mloskot:
This is preliminary fix to allow linking the library
as well as its dependants (e.g. PROJ apps) against cURL
built as static library on Windows with support of native
Windows TLS/SSL.

On Windows, such static cURL (and its dependants) requires linking
against Winsock, CryptoAPI and other networking libaries.

------

This patch is a minimal version confirm to work with cURL 7.75.0 built as SChannel mode (old name WinSSL, see https://github.com/curl/curl/pull/3504), that is MSVC outputs binaries into:

```
curl\builds\libcurl-vc14-x86-debug-static-ipv6-sspi-schannel\
curl\builds\libcurl-vc14-x86-release-static-ipv6-sspi-schannel\ 
```

### References

- Blocked by issue #2515

### DIFF
--- a/src/lib_proj.cmake
+++ b/src/lib_proj.cmake
@@ -409,7 +409,14 @@ endif()
 if(CURL_ENABLED)
   target_compile_definitions(${PROJ_CORE_TARGET} PRIVATE -DCURL_ENABLED)
   target_include_directories(${PROJ_CORE_TARGET} PRIVATE ${CURL_INCLUDE_DIR})
-  target_link_libraries(${PROJ_CORE_TARGET} PRIVATE ${CURL_LIBRARY})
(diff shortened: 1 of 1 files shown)


kbevers:
@mloskot with https://github.com/OSGeo/PROJ/pull/2533 are you able to make this work now? It seems a merge conflict has arised. The fix seems trivial but not sure if the rest of the PR is in good shape.

mloskot:
@kbevers Thanks for reminder. I've just updated the PR resolving the conflict.

kbevers:
>  I've just updated the PR resolving the conflict.

... and it took me like five minutes to create another conflict :-) At least I know how to fix this one.

PR #2516, opened 2021-02-07
CI: specify pip < 21.0; avoid using pip3 script

mwtoews:
Specify pip version less than 21.0 to avoid syntax error with Python 3.5 or older. This fixes CI errors since January 2021.

See [this answer](https://stackoverflow.com/a/65896996/) for details.

Furthermore, use safer `python3 -m pip` rather than `pip3`.


kbevers:
What are the CI errors? It is probably more sustainable if those can be fixed instead

mwtoews:
The CI errors were attempting to update pip beyond version 21.0 for Python 3.5 or older, which is used for Ubuntu xenial. Upgrading some of these VMs to at least bionic could help.

PR #2522, opened 2021-02-10
operations_computation.rst: update 'Filtering and sorting of coordinate operations' section

rouault:


kbevers:
Pehaps update the paragraph in line 51-55
```
The algorithm involves many cases, so we will progress in the explanation from
the most simple case to more complex ones. We document here the working of this
algorithm as implemented in PROJ 6.3.0.
The results of some examples might also be quite sensitive to the content of the
PROJ database and the PROJ version used.
```

to reflect that this is now as of PROJ 8. I assume the matches the current state of affairs with this update so we may as well indicate that.

rouault:
> Pehaps update the paragraph in line 51-55

done

kbevers (approved):

PR #2528, opened 2021-02-12
Add api method to force +over on transformation operations and unit test.

### LINKED ISSUE #2512: Using +over with EPSG codes and WKT CRS Input/Output?
Is it possible to provide the +over flag when using an EPSG code or a WKT string as the CRS input when calling the utility methods like cs2cs? It doesn't seem like it's able to recognize it as-is. I've tried various combinations of attaching the +over to no avail.

I can kind of hack around it by customizing the library, but it would be nice to be able to have this out of the box. Or, if it's actually possible...?

mlptownsend:
Code for issue #2512. Since the deadline for release is looming, I didn't bother with adding any command switches for the various applications (since I didn't end up going that route anyway in the end). I added an item to the functions.rst docs page, but let me know if the wording or location needs to be changed.

 - [x] Closes #2512 
 - [x] Tests added
 - [x] Fully documented, including updating `docs/source/functions.rst` for new API

### DIFF
--- a/test/unit/gie_self_tests.cpp
+++ b/test/unit/gie_self_tests.cpp
@@ -819,4 +819,60 @@ TEST(gie, proj_trans_generic) {
     proj_destroy(P);
 }
 
+TEST(gie, proj_create_crs_to_crs_force_over) {
+
+    PJ_CONTEXT* ctx;
+
+    ctx = proj_context_create();
+    ASSERT_TRUE(ctx != nullptr);
+    proj_context_set_force_over(ctx, 1);
+    ASSERT_TRUE(ctx->forceOver);
+
+    auto P = proj_create_crs_to_crs(ctx, "EPSG:4326", "EPSG:3857", nullptr);
+    ASSERT_TRUE(P != nullptr);
+    PJ_COORD input;
+    PJ_COORD input_over;
+
+    //Test a point along the equator. 
+    //The same point, but in two different representations.
+    input.xyz.x = 0; // Lat in deg
+    input.xyz.y = 140;  // Long in deg
+    input.xyz.z = 0;
+
+    input_over.xyz.x = 0; // Lat in deg
+    input_over.xyz.y = -220;  // Long in deg
+    input_over.xyz.z = 0;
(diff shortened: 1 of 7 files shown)


rouault:
Thinking more about it, while the current approach works, I'm wondering if setting this at the PJ_CONTEXT* level is the more elegant way of doing it. I'd rather see this at the PJ* level. As proj_create_crs_to_crs() can be a set of ``alternativeCoordinateOperations``, the flag would have to be propagated to them.

mlptownsend:
Yeah, I do agree that context flag is too general a solution. I wasn't too sure how proj was building all the operations all the way down. I just knew it was too late if I was setting the flag on the PJ after proj_create_crs_to_crs_from_pj. It had to go on the items that appeared in that step list.

mlptownsend:
I'm not beholden to it making it in the 8.0.0 release if there's a smarter way to do this :)

test/unit/gie_self_tests.cpp
```diff
@@ -819,4 +819,60 @@ TEST(gie, proj_trans_generic) {
...
+TEST(gie, proj_create_crs_to_crs_force_over) {
+
+    PJ_CONTEXT* ctx;
+
+    ctx = proj_context_create();
+    ASSERT_TRUE(ctx != nullptr);
+    proj_context_set_force_over(ctx, 1);
+    ASSERT_TRUE(ctx->forceOver);
+
+    auto P = proj_create_crs_to_crs(ctx, "EPSG:4326", "EPSG:3857", nullptr);
```
rhuijben, line 831: Why not add the flag to the options (passed as nullptr) here?

mlptownsend: I've been super busy and haven't had time to revisit moving the flag around. Maybe in a few weeks I'll have some free time again.

mlptownsend:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

I haven't forgotten. Just been super super busy.

PR #2532, opened 2021-02-14
BUG: Close database connection when autoclose set to True

### LINKED ISSUE #2531: PROPOSAL: proj_context_set_autoclose_database to close database connection when set to True
This feature would enable users to turn on and off the auto-close database feature around different methods to potentially increase the speed of those methods. It would reduce the need for re-connecting to the database as often.

For example:

```c
         proj_context_set_autoclose_database(context, false)
         projobj = proj_create_crs_to_crs(
            context,
            proj_string1,
            proj_string2,
            NULL,
        )
        proj_context_set_autoclose_database(context, true)
```

Currently this is not possible as there is no way to manually close the database connection.

snowman2:
- [x] Closes #2531
 - [ ] Tests added

### DIFF
--- a/include/proj/internal/io_internal.hpp
+++ b/include/proj/internal/io_internal.hpp
@@ -201,7 +201,7 @@ struct projCppContext {
         return auxDbPaths_;
     }
 
-    void setAutoCloseDb(bool autoClose) { autoCloseDb_ = autoClose; }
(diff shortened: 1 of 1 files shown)


snowman2:
Any blockers?

kbevers:
> Any blockers?

Probably just me not having had the time to look into this.

PR #2538, opened 2021-02-18
CMake: replace '${PROJ_CORE_TARGET}' with 'proj'

mwtoews:
This simplifies the main library target from a variable (`${PROJ_CORE_TARGET}`) to simply `proj`, which it has always been named.

With most CMake projects, targets are typically a short lowercase constant label, and not a variable.

Would appreciate a quick review by @mloskot


mloskot (approved):
Yes, I think it is a good idea.

Referring to targets via variables makes sense if a name is generated in any way

PR #2539, opened 2021-02-19
Update to EPSG v10.015

nyalldawson:
And remove temporary workarounds due to bugs in older epsg db versions


kbevers:
Looks good to me. Thanks!

PR #2542, opened 2021-02-20
LOG: Default log level PJ_LOG_ERROR

### LINKED ISSUE #2535: Default PROJ debug level
With #2527, errors are no longer raised by default.

This is because `PJ_LOG_NONE` is the default:
- https://github.com/OSGeo/PROJ/blob/cd10468ece8b66199410f274a625a5b05bf36592/src/ctx.cpp#L90
- https://github.com/OSGeo/PROJ/blob/cd10468ece8b66199410f274a625a5b05bf36592/src/proj_internal.h#L650

Thoughts about making the default level `PJ_LOG_ERROR` so errors are logged by default?

snowman2:
- [x] Closes #2535
 - [ ] Tests added

If this is something you aren't interested in changing, feel free to close. Just putting it here in case it is at least a `+-0` thing.

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -87,7 +87,7 @@ void proj_assign_context( PJ* pj, PJ_CONTEXT *ctx )
 pj_ctx pj_ctx::createDefault()
 {
     pj_ctx ctx;
-    ctx.debug_level = PJ_LOG_NONE;
+    ctx.debug_level = PJ_LOG_ERROR;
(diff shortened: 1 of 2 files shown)


kbevers:
Travis fails on the arm64 target. It seems like a fluke with the system tools. Since everything else seems to perform as usual I'll merge this one. If there are actual problems I assume they show up once RC1 is out in the wild.

PR #2543, opened 2021-02-20
BUG: Static cast PJ_LOG_ERROR to int

snowman2:
I am wondering if this will fix the issue seen in #2542


kbevers:
It didn't :(

snowman2:
Seems unrelated to this change? 🤷‍♂️

kbevers:
I am quite sure that it is, yes.

PR #2536, opened 2021-02-16
International ellipsoid designation correction

busstoptaktik:
The "International 1909 (Hayford)" designation is an unfortunate mix of different aspects of the same ellipsoid.

The Hayford ellipsoid was based on investigations in 1909, published in 1910 as J.F. Hayford: *Supplementary Investigation in 1909 of the Figure of the Earth and Isostasy*, Coast and Geodetic Survey, 80pp. (freely accessible through Google Books).

In 1924 Hayford's ellipsoid was adopted by the IUGG as "the International Ellipsoid".

Hence either of 

- Hayford, 1909 (for the year the work was done),
- Hayford, 1910 (for the year of publication), or
- International, 1924

are reasonable names, in common use, but the "International, 1909" really is a peculiar mix of concepts.

Resolves #2534

 - [x] Closes #2534
 - [x] Added clear title that can be used to generate release notes
 - [x] Fully documented, including updating `docs/source/*.rst` for new API

PR #2545, opened 2021-02-22
isEquivalentName(): fix for GDAL test failure

rouault:
https://github.com/OSGeo/PROJ/pull/2536 changed the name of the
ellps=intl to "International 1924 (Hayford 1909, 1910)"
When writing a GeoTIFF file from GDAL using a SRS built from a
PROJ string, GDAL massages the datum name to
"Unknown_based_on_International_1924_Hayford_1909_1910_ellipsoid"
Before this fix, this wasn't considered as equivaleent to the non-massaged
datum name "Unknown based on International 1924 (Hayford 1909, 1910) ellipsoid"


kbevers:
So I guess this means we need a RC2?

rouault:
> So I guess this means we need a RC2?

that would be nice

Issue #2546, opened 2021-02-24
CMake files not installed when autotools is used. pkg-config file not installed when using CMake


sebastic:
#### Example of problem
When using autotools to build PROJ like the Debian package does the `.cmake` files are not installed as they are when using CMake like the Fedora package does.

Likewise the pkg-config file is not installed when using CMake, the Fedora package had to [patch](https://src.fedoraproject.org/rpms/proj/blob/rawhide/f/proj_pkgconfig.patch) PROJ to install it.

#### Problem description

The developers of QMapShack [[1](https://sourceforge.net/p/qlandkartegt/mailman/message/36634734/)] and OpenOrienteering Mapper [[2](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=983254#34)] have complained that the Debian package does not install the CMake files forcing them to maintain a `FindProj.cmake` in their source tree.

#### Expected Output

Ideally the CMake files are installed as part of the Autotools build, and the pkg-config bits as part of the CMake build.

The result of either build should be nearly identical. And have both the pkg-config and CMake bits installed for the benefit of downstream projects.

#### Environment Information
 - Rel. 7.2.1, January 1st, 2021
 - Debian GNU/Linux unstable (amd64)

#### Installation method
 - Debian package (`libproj-dev`)

dg0yt:
For more context, as the developer of OpenOrienteering Mapper:

Debian claims that our software fails to build from source with PROJ 8.0.0, due to usage of `proj_api.h`. Now we are using CMake, and we have been adopting modern PROJ and CMake early. With the Debian PROJ package being build with autotools, it doesn't provide the CMake config files. This makes our package fall back to a legacy `FindPROJ4.cmake` which looks for `proj_api.h`. This header file was removed in PROJ 8.0.0, uncovering the unbalanced feature set of the PROJ build systems. 

> The developers of QMapShack [1] and OpenOrienteering Mapper [2] have complained that the Debian package does not install the CMake files forcing them to maintain a FindProj.cmake in their source tree.

The CMake find module is not "maintained" for Debian testing / PROJ 8.0. It is kept for Debian stable, Ubuntu 18.04 and alikes, with PROJ version down to 4.9 where there are no CMake options. For this purpose, it works as intended.

My point of view is that:
- the CMake build needs to create the missing pkgconfig files
- the CMake build shall be adjusted to support Debian's multiarch builds of PROJ
- autotools should eventually be dropped.

[...]

PR #2547, opened 2021-02-24
CMake: add support for pkg-config

mwtoews:
This addresses part of #2546 to install a pkg-config file.


mwtoews:
Does anyone feel qualified to review this? Nevertheless, the CMake-generated set-up passes post-install tests with Linux and macOS.

One observation discovered here between autotools vs CMake installs on macOS is that the dynamic library via CMake has the relative reference `@rpath/libproj.22.dylib` (via `otool -L`) whereas the autotools library uses an absolute path '/tmp/proj_cmake_install/lib/libproj.22.dylib`. This shouldn't be surprising given the CMake options, but is a difference to note.

rouault:
not a cmake expert, but that looks very reasonable to me

dg0yt:
Given the syntax of the root `proj.pc.in`, did you conside creating a function (i.e. scope) that sets the variables needed to configure that file, instead of introducing a separate `cmake/proj.pc.cmake.in`? But changes should be rare, so redundancy may be acceptable.

CMakeLists.txt
```diff
@@ -282,3 +282,24 @@ endif()
...
+    set(CURL_LIBS -lcurl)
+  endif()
+  configure_file(
+    ${CMAKE_CURRENT_SOURCE_DIR}/cmake/proj.pc.cmake.in
+    ${CMAKE_CURRENT_BINARY_DIR}/proj.pc
+    @ONLY)
+
+  install(FILES
+    ${CMAKE_CURRENT_BINARY_DIR}/proj.pc
+    DESTINATION lib/pkgconfig)
```
dg0yt: I guess Debian will be unhappy about this rule. Due to multi-arch, `proj.pc` resides in `/usr/lib/x86_64-linux-gnu/pkgconfig/`. In a `GNUInstallDirs` context, the destination would probably be `${CMAKE_INSTALL_LIBDIR}/pkgconfig`.

kbevers:
Is everyone happy with this PR in its current state?

CMakeLists.txt
```diff
@@ -282,3 +282,24 @@ endif()
 set(docfiles COPYING NEWS AUTHORS)
 install(FILES ${docfiles}
         DESTINATION "${DOCDIR}")
+
+################################################################################
+# pkg-config support
+################################################################################
+if(UNIX)
```
dg0yt: I just had nothing but MinGW (MSYS2) today, and guess what... There are other pkg-config files here. I successfully tested build/install with:
~~~
if(UNIX OR MINGW)
~~~

mwtoews: hmm, possibly this one too, as some of these paths are set from `GNUInstallDirs` https://github.com/OSGeo/PROJ/blob/1a518773616fb001483c175e8dfab00c06e83c89/cmake/ProjInstallPath.cmake#L26-L27

mwtoews:
> Given the syntax of the root `proj.pc.in`, did you conside creating a function (i.e. scope) that sets the variables needed to configure that file

Sure, I'll give this a "go"...

CMakeLists.txt
dg0yt: I see. I don't see a difference in `proj.pc` when I change `ProjInstallPath.cmake`.
And I wouldn't delay PROJ 8.0.0 for MINGW pkgconfig.

Default config for MinGW, generator Ninja:
~~~
prefix=c:/OSGeo4W
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
datarootdir=${prefix}/share
datadir=${prefix}/share/proj

Name: PROJ
Description: Coordinate transformation software library
Requires:
Version: 8.0.0
Libs: -L${libdir} -lproj
Libs.private: -lsqlite3 -ltiff -lcurl -lstdc++
Cflags: -I${includedir}
~~~

Looks good to me.

dg0yt:
LGTM. 
~~~
$ diff former-mingw current-mingw
6c6
< datadir=${prefix}/share/proj
---
> datadir=${datarootdir}/proj

mwtoews:
This last update, which uses the same `proj.pc.in` file as by autotools, creates an identical `proj.pc` file for a given prefix.

PR #2551, opened 2021-03-03
Reformat code with clang-format-10 from ubuntu 20.04

rouault:
Just run scripts/reformat_cpp.sh with clang-format-10 from ubuntu 20.04

We should probably do the same in the stable branch for easier backport


kbevers:
> We should probably do the same in the stable branch for easier backport

Would of course have been better  before the 8.0.0 release but the changes here seem harmless so if the testsuite passes on the 8.0 branch I don't see any problems in doing it.

PR #2567, opened 2021-03-09
s45b.pol: remove trailing backslashes that prevent it from being used

rouault:
CC @busstoptaktik  There must have been changes in the init file parser since you committed that file. Without that change, cct +init=./examples/s45b.pol:s45b_tc32 fails. I'd appreciate if you could review that it leads to correct results


kbevers:
There are some basic gie tests in the `DK` init-file:

https://github.com/OSGeo/PROJ-data/blob/e1f3870b7da3f78e18c78dd6676c246a9dca8584/dk_sdfe/DK#L374-L387

You could try running `gie DK` and at least verify if the above tests is okay

rouault:
> You could try running `gie DK` and at least verify if the above tests is okay

yes, it passes OK. I see that dk_sdfe_DK_bornholm.pol has no trailing backslashes, so I assume my change is correct.

kbevers:
I think this is fine. Looking deeper I can't even figure out where this file would be used anymore.

rouault:
> I think this is fine. Looking deeper I can't even figure out where this file would be used anymore.

It is referenced in https://proj.org/usage/transformation.html?highlight=s45 and this was triggered by a user trying to build a pipeline from it

PR #2570, opened 2021-03-10
Make proj_lp_dist() and proj_geod() work on a PJ* CRS object

rouault:


rouault:
Someone would be welcome to improve the documentation of those functions (https://proj.org/development/reference/functions.html) to specify that the coordinates should be specified in longitude, latitude order and radian units (whatever the CRS axis order and unit are...)

rouault:
Cherry-picked in 8.0 branch

PR #2573, opened 2021-03-11
Use breathe>=4.27 to support Doxygen \page, \section, etc.

mwtoews:
This PR removes two hacks that I think are no longer needed.

The big one is to use recent breathe features, mainly https://github.com/michaeljones/breathe/pull/635 which is part of version 4.27

With the rendered output, the layout of [cpp_general](https://proj.org/development/reference/cpp/cpp_general.html) for both html and pdf is generally improved, but with more structured headers. Cross-references to node names in this section also work as expected.

The second removed hack is to not follow the bug workaround for Doxygen 1.8.17, and to just throw an error instead. This is partly because the need for a `tmp_breathe` directory (used for the breathe hack) was removed.


.github/workflows/doc_build.yml
```diff
@@ -48,6 +48,7 @@ jobs:
       run: |
           python3 --version
           sphinx-build --version
+          python3 -m pip install "breathe>=4.27"
```
mwtoews: The above line updates the package to allow this PR to work. I'm not sure how to update the osgeo/proj container with the updated version of breathe, but would appreciate any pointers. An alternative idea is to merge the "Plots" CI job that uses conda-forge. It's not clear why there is a mixture of conda and containers used to check docs.

rouault: The proj-docs source file is in docs/docbuild/Dockerfile . I've regenerated it locally and it got breathe 4.27 in it. I've checked it still runs fine with GDAL docs too since it is used by it. I'm pushing it now to Docker hub but it will take several hours

hobu: > An alternative idea is to merge the "Plots" CI job that uses conda-forge. It's not clear why there is a mixture of conda and containers used to check docs.

Because I built the containers 5+ years ago, and I did the conda build recently. It would be reasonable to reconfigure the osgeo/proj-docs container to be conda-based so everything is the same baseline, but why break what's working (for PDAL, PROJ, MapServer, and GDAL too)

hobu: We should maybe consider pushing the osgeo/proj-docs container to ghcr.io. There will be no bandwidth restrictions and we could build it with the GHA loop. I will investigate doing so.

rouault: osgeo/proj-docs has now been refreshed. you can drop that change

mwtoews:
Hmm, with a new container update, a new issue emerges from `make spelling`:
```
...
Encoding error:
'ascii' codec can't encode character '\u201c' in position 39: ordinal not in range(128)
The full traceback has been saved in /tmp/sphinx-err-cab77k47.log, if you want to report the issue to the developers.
make: *** [spelling] Error 2
Makefile:43: recipe for target 'spelling' failed
```
I'm unable to replicate this locally. Also note that `python3 -c "print('\u201c')"` returns `“`.

rouault:
> I'm unable to replicate this locally

even within the updated container ?

rouault:
Note: I had tagged the previous version of the container as osgeo/proj-docs:2020

mwtoews:
> even within the updated container ?

Ok, I am now getting the error with `osgeo/proj-docs:latest`. However, I'm unable to get the error on my Ubuntu laptop in a fresh venv or conda environment via pip. I'll need to investigate this further ....

rouault:
Restarting the Docs job with the updated Docker image

rouault (approved):

mwtoews:
Cherry-picked into 8.0, so now you can see the new layout in [this section](https://proj.org/development/reference/cpp/cpp_general.html), or page 429 of the current [PDF](https://raw.githubusercontent.com/OSGeo/proj.4/gh-pages/proj.pdf).

hobu:
Is there a way with breathe to suppress index entity creation nowadays? It makes a sloppy mess of the indexes, and I think our indexes would be better off with a few manual entries pointing to the beginning of the API section rather than the auto-generated entries which take up a bunch of space and add little.

mwtoews:
> Is there a way with breathe to suppress index entity creation nowadays?

You mean in the PDF version? Yes I agree, and will take a look to try and suppress this generation.

PR #2574, opened 2021-03-12
CI: workaround for sphinxcontrib-spelling 7.1.0 on Python 3.6.9

mwtoews:
Trade one hack for another?

From https://github.com/OSGeo/PROJ/pull/2573#issuecomment-797015308 the container osgeo/proj-docs was updated, which unfortunately broke `make spelling`. This "fix" is really a workaround, which will need to come upstream (see https://github.com/sphinx-contrib/spelling/issues/121).

### DIFF
--- a/docs/docbuild/Dockerfile
+++ b/docs/docbuild/Dockerfile
@@ -11,4 +11,5 @@ RUN python3 -m pip install Sphinx breathe \
     sphinx_rtd_theme recommonmark sphinx-markdown-tables \
     sphinxcontrib-spelling
 
-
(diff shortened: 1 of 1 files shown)


rouault:
would you mind adding this fix in docs/docbuild/Dockerfile instead ? (as an extra step, so I can reuse the existing Docker layers...) If that works OK locally for you, I'll docker push the updated image

mwtoews:
updated change, and successful local container build and run with `make spelling`

rouault:
Updated docker image pushed

PR #1012, opened 2018-05-23
Add doc section with description of behavioural changes between versions

kbevers:
As requested in #941

PR #2575, opened 2021-03-13
Set the container locale for spelling

mwtoews:
This should be the last container modification request for now. The primary issue [noted here](https://github.com/OSGeo/PROJ/pull/2573#issuecomment-797015308) was not a sphinxcontrib-spelling bug, but that the container did not have a locale set.

This PR removes the most recent hack, and sets the locale to `en_US.UTF-8` (which implies that we should be using American spelling throughout the docs?)

Dockerfile was checked locally to ensure `make spelling` succeeds.

### DIFF
--- a/docs/docbuild/Dockerfile
+++ b/docs/docbuild/Dockerfile
@@ -4,12 +4,16 @@ RUN apt-get -y update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     python3-dev python3-pip g++ doxygen dvipng latexmk \
     cmake libjpeg8-dev zlib1g-dev texlive-latex-base \
     texlive-latex-extra git latex-cjk-all texlive-lang-all \
-    graphviz python3-matplotlib wget unzip enchant
+    graphviz python3-matplotlib wget unzip enchant locales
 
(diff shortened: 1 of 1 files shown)


kbevers:
We briefly touched upon British vs American spelling [here](https://github.com/OSGeo/PROJ/pull/1012#pullrequestreview-122658327) a couple of years back. At the time we seem to lean towards British although there was no way to enforce it at the time. We can do that now.

Can we check which dialect results in less changes?

rouault:
> We briefly touched upon British vs American spelling [here](https://github.com/OSGeo/PROJ/pull/1012#pullrequestreview-122658327) a couple of years back

I suspect that scripts/fix_typos.sh uses dictionaries with mostly American spelling

mwtoews:
Sure, locale now set to `en_GB.UTF-8`. I'll fix up the other components later, and I'll take a look at scripts/fix_typos.sh when I get a moment.

New container tested locally and ready to be rebuilt and pushed.

rouault:
> New container tested locally and ready to be rebuilt and pushed.

I've just pushed the refreshed container

PR #2576, opened 2021-03-13
Update geodesic routines from GeographicLib 1.52

### LINKED ISSUE #2520: Use of uninitialized memory at src/geodesic.c:sincosdx when running under the memory sanitizer
@schwehr 

#### Example of problem
MemorySanitizer: use-of-uninitialized-value .../proj/src/geodesic.c:184:3 in sincosdx

#### Problem description
The use of *q* at https://github.com/OSGeo/PROJ/blob/master/src/geodesic.c#L184 may be uninitialized when using LLVM12 - https://reviews.llvm.org/rG8514ecb02d4
The function remquo may not initialize q in some cases.

https://github.com/OSGeo/PROJ/blob/6ce2c1174a2ee9bb529cc9260b33b6243950e0ad/src/geodesic.c#L168

#### Expected Output
*q* is initialized by the time it reaches line 184.

#### Environment Information
 - PROJ version (`proj`) - 6.0
 - Operation System Information - Debian 5.7.17-1 x86_64 GNU/Linux

#### Installation method
 - from source

cffk:
Be more aggressive in preventing negative s12 and m12 for short lines.
Initialize reference argument to remquo.

 - [x] Closes #2520
 - [x] Tests added

### DIFF
--- a/docs/source/geodesic.rst
+++ b/docs/source/geodesic.rst
@@ -47,7 +47,7 @@ Traditionally two geodesic problems are considered:
   :math:`\alpha_2`.
 
 PROJ incorporates `C library for Geodesics
-<https://geographiclib.sourceforge.io/1.49/C/>`_ from `GeographicLib
+<https://geographiclib.sourceforge.io/1.52/C/>`_ from `GeographicLib
 <https://geographiclib.sourceforge.io>`_.  This library provides
 routines to solve the direct and inverse geodesic problems.  Full double
 precision accuracy is maintained provided that
@@ -55,7 +55,7 @@ precision accuracy is maintained provided that
 to the
 
     `application programming interface
-    <https://geographiclib.sourceforge.io/1.49/C/geodesic_8h.html>`_
+    <https://geographiclib.sourceforge.io/1.52/C/geodesic_8h.html>`_
 
 for full documentation.  A brief summary of the routines is given in
 geodesic(3).
--- a/src/geodesic.c
+++ b/src/geodesic.c
@@ -18,7 +18,7 @@
  *
  * See the comments in geodesic.h for documentation.
  *
- * Copyright (c) Charles Karney (2012-2020) <charles@karney.com> and licensed
+ * Copyright (c) Charles Karney (2012-2021) <charles@karney.com> and licensed
  * under the MIT/X11 License.  For more information, see
  * https://geographiclib.sourceforge.io/
  */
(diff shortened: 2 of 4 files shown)


rouault:
Test failure on AppVeyor, which is a MSVC 2015 x86 / 32-bit release build (the github action Windows one is a MSVC 2019 x64 release build)
```
[00:04:21]       Start  1: geodesic-test
[00:04:21] 
[00:04:21] 1: Test command: C:\projects\proj\build\bin\geodtest.exe
[00:04:21] 1: Test timeout computed to be: 1500
[00:04:21] 1: checkEquals fails: -nan(ind) != 90 +/- 1e-07
[00:04:21] 1: checkEquals fails: -nan(ind) != 0.264 +/- 0.0005
[00:04:21] 1: GeodSolve92 fail: 2
[00:04:21]  1/45 Test  #1: geodesic-test ....................***Failed    0.01 sec
```

cffk:
This failure is with a new test I added to test a bug in the python version.  I'll see if I can reproduce the bug locally.

cffk:
Yup, Visual Studio 2015 win32 has the same problem as some versions of python, an inaccurate implementation of hypot; see https://bugs.python.org/issue43088.  I'll work around the problem and submit a fix tomorrow.

cffk:
Be more aggressive in preventing negative `s12` and `m12` for short lines.
Initialize reference argument to `remquo`.
Work around inaccurate implementation of `hypot` with Visual Studio 2015.

cffk:
Be more aggressive in preventing negative s12 and m12 for short lines.
Initialize reference argument to remquo.
Work around inaccurate implementation of hypot with Visual Studio 2015.

rouault:
Perhaps a narrower fix with just the q=0 initialization should be done in the 8.0 maintenance branch ?

cffk:
I don't have an opinion about this.  The `q = 0` is just to satisfy a memory sanitizer complaint #2520 (I see that I got this issue number wrong in the initial checkin for this PR).  I would be fine with waiting.  But check with @schwehr and @metzm.

schwehr:
It's not critical to fix in the branch, but it's so simple that it wouldn't hurt to do it

cffk:
Bletch, this Visual Studio bug in `hypot` is not confined to Visual Studio 2015.  32-bit compiles with VS 2017 and 2019 exhibit the same problem.  (I'm surprised this wasn't caught by PROJ's CI?)  I'll make another PR to fix.

Also, is there a clean way to apply the workaround only for 32-bit compiles?

rouault:
> I'm surprised this wasn't caught by PROJ's CI?

we "only" have 2 Windows configurations: VS2015 32bit and VS2019 64bit 

> is there a clean way to apply the workaround only for 32-bit compiles?

See https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-140
```
#if defined(_MSC_VER) && defined(_M_IX86)
```

PR #2578, opened 2021-03-14
geodesic.c: fix conditions under which hypot workaround is applied.

cffk:
Now it applies with 32-bit compiles for all versioins of Visual Studio;
versions tested on: 2015, 2017, 2019.

### DIFF
--- a/src/geodesic.c
+++ b/src/geodesic.c
@@ -127,9 +127,15 @@ static void swapx(real* x, real* y)
 { real t = *x; *x = *y; *y = t; }
 
 static void norm2(real* sinx, real* cosx) {
-#if defined(_MSC_VER) && _MSC_VER <= 1900
-  /* Visual Studio 2015 (32-bit) has inaccurate hypot, the same as in some
-   * versions of python https://bugs.python.org/issue43088 */
+#if defined(_MSC_VER) && defined(_M_IX86)
+  /* hypot for Visual Studio (A=win32) fails monotonicity, e.g., with
+   *   x  = 0.6102683302836215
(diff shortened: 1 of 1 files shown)


rhuijben:
Is there some tracking number of this bug at Microsoft? I would assume they would want some fix upstream that we eventually should use...

From the old implementation I would have guessed that it is already fixed upstream but that we somehow mix testing compiler and c library versions for when to use which. Am I wrong here?

cffk:
I submitted a bug report to Microsoft.  It's

https://developercommunity.visualstudio.com/t/hypot-is-nonmonotonic-with-win32-compile/1369259

If you have a MS account of some kind, you can "up-vote" the problem report.

The bug in the geodesic routines (triggered by Visual Studio's buggy version of `hypot`) was probably present ever since VS offered `std::hypot`. It's present with the simple implementation of hypot as

```c++
  x = abs(x); y = abs(y);
  if (x < y) std::swap(x, y); // Now x >= y >= 0
  y /= (x != 0 ? x : 1);
  return x * sqrt(1 + y * y);
```

It was only caught as a bug in the geodesic package because of testing someone did on the python implementation; python version 3.[89] had the same bug.  (This is fixed in the latest version of python 3.x.)

It's a mystery to me why this bug appears in the 32-bit compilation but not the 64-bit version.  You might think that the implementation of the cmath functions would give identical results in both cases.

The best I'm expecting from Microsoft is that they get this fixed in the next version of Visual Studio.  It's also possible that they declare that this isn't a bug, since the C++ standard doesn't place any accuracy requirements on `std::hypot`.

PR #2581, opened 2021-03-17
proj.h: add PROJ_COMPUTE_VERSION, PROJ_VERSION_NUMBER, PROJ_AT_LEAST_VERSION macros

rouault:
Makes it easier for users to test if they build against a PROJ version
later than a given x.y.z version.

To be backported in 8.0 branch

### DIFF
--- a/docs/source/development/reference/index.rst
+++ b/docs/source/development/reference/index.rst
@@ -7,6 +7,7 @@ Reference
 .. toctree::
    :maxdepth: 1
 
+   macros
    datatypes
    functions
    cpp/index.rst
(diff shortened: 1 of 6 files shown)


kbevers:
`PROJ_AT_LEAST_VERSION` should be documented somewhere I think

rouault:
> `PROJ_AT_LEAST_VERSION` should be documented somewhere I think

I've added a source/development/reference/macros.rst for that

kbevers:
> I've added a source/development/reference/macros.rst for that

Thanks! Looks good.

rouault:
Cherry-picked to 8.0.1 per 861ae257f4f640dacc71528079ab6d56c8d39e80 and f4bd832184b5ebcc5b18b0d625628cf7dd7e969a

PR #2583, opened 2021-03-18
createFromCRSCodesWithIntermediates(): improve perf when no match

rouault:
createFromCRSCodesWithIntermediates() runs a rather costly self-join.
Only run it if the source and target CRS are the source/target of a
coordinate operation. This helps for the performance of
proj_create_crs_to_crs() when run on projected CRS for example that are
extremely unlikely to be the source/target of an operation (except
currently the Finish ones). For the EPSG:26915 to EPSG:3857 case of
https://github.com/OSGeo/gdal/issues/3470, this helps decreasing the
time of proj_create_crs_to_crs() from 18 ms to 10 ms.

### DIFF
--- a/src/iso19111/factory.cpp
+++ b/src/iso19111/factory.cpp
@@ -6081,6 +6081,22 @@ AuthorityFactory::createFromCRSCodesWithIntermediates(
         return listTmp;
     }
 
+    const auto CheckIfHasOperations = [=](const std::string &auth_name,
+                                          const std::string &code) {
+        return !(d->run("SELECT 1 FROM coordinate_operation_view WHERE "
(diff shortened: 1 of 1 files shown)


rouault:
Detail revealed by instrumentation: PROJ 8 is actually faster to do UTM coordinate transformations than PROJ 5.2. 2 ms v 5 ms for that use case :-) (excluding the 10 ms to compute the initial transformation pipeline)
And if removing the -ts 1024 1024 to generate a 10907 x 12619 output, PROJ 8 spends 82 ms in coordinate transformation, vs 144 ms for PROJ 5.2. So overall, those enhancements mostly help for reprojection of small rasters.

rouault:
Backported to 8.0 per 9393831dce0ca5271834caa32dc93ab48aca0c2c

PR #2555, opened 2021-03-05
createOperations(): fix incorrect height transformation between 3D promoted RGF93 and CH1903+ (fixes #2541)

rouault:

PR #2584, opened 2021-03-18
Add transformations for CHGeo2004, Swiss geoid model

jjimenezshaw:
- added entries in `grid_alternatives.sql` for `ch_swisstopo_chgeo2004_ETRS89_LHN95.tif` and `ch_swisstopo_chgeo2004_ETRS89_LN02.tif`, included in `PROJ-data`
- added entries in `grid_transformation_custom.sql` until EPSG is updated. This gives us more time to test the behaviour.

See related PR https://github.com/OSGeo/PROJ-data/pull/61

### DIFF
--- a/data/sql/grid_alternatives.sql
+++ b/data/sql/grid_alternatives.sql
@@ -85,6 +85,8 @@ VALUES
 -- ch_swisstopo - Swisstopo Federal Office of Topography
 ('CHENyx06a.gsb','ch_swisstopo_CHENyx06a.tif','CHENyx06a.gsb','GTiff','hgridshift',0,NULL,'https://cdn.proj.org/ch_swisstopo_CHENyx06a.tif',1,1,NULL),
 ('CHENyx06_ETRS.gsb','ch_swisstopo_CHENyx06_ETRS.tif','CHENyx06_ETRS.gsb','GTiff','hgridshift',0,NULL,'https://cdn.proj.org/ch_swisstopo_CHENyx06_ETRS.tif',1,1,NULL),
+('chgeo2004_ETRS.agr','ch_swisstopo_chgeo2004_ETRS89_LHN95.tif',NULL,'GTiff','geoid_like',0,NULL,'https://cdn.proj.org/ch_swisstopo_chgeo2004_ETRS89_LHN95.tif',1,1,NULL),
+('chgeo2004_htrans_ETRS.agr','ch_swisstopo_chgeo2004_ETRS89_LN02.tif',NULL,'GTiff','geoid_like',0,NULL,'https://cdn.proj.org/ch_swisstopo_chgeo2004_ETRS89_LN02.tif',1,1,NULL),
 
 -- de_adv - Arbeitsgemeinschaft der Vermessungsverwaltungender der Länder der Bundesrepublik Deutschland (AdV)
 ('BETA2007.gsb','de_adv_BETA2007.tif','BETA2007.gsb','GTiff','hgridshift',0,NULL,'https://cdn.proj.org/de_adv_BETA2007.tif',1,1,NULL),
--- a/data/sql/grid_transformation_custom.sql
+++ b/data/sql/grid_transformation_custom.sql
@@ -169,3 +169,45 @@ INSERT INTO "usage" VALUES(
     'EPSG','3263', -- area of use: Japan - onshore mainland
     'EPSG','1024'  -- unknown
 );
+
+-- Switzerland
+
+INSERT INTO "grid_transformation" VALUES(
+    'PROJ','EPSG_4937_TO_EPSG_5729','ETRS89 to LHN95 height',
+    NULL,
+    'EPSG','9665','Geographic3D to GravityRelatedHeight (gtx)',
+    'EPSG','4937', -- source CRS (ETRS89)
+    'EPSG','5729', -- target CRS (LHN95 height)
+    NULL,
+    'EPSG','8666','Geoid (height correction) model file','chgeo2004_ETRS.agr',
+    NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);
+
+INSERT INTO "usage" VALUES(
+    'PROJ',
+    'EPSG_4937_TO_EPSG_5729_USAGE',
+    'grid_transformation',
+    'PROJ',
+    'EPSG_4937_TO_EPSG_5729',
+    'EPSG','1286', -- area of use: Europe - Liechtenstein and Switzerland
+    'EPSG','1024'  -- unknown
+);
+
+INSERT INTO "grid_transformation" VALUES(
+    'PROJ','EPSG_4937_TO_EPSG_5728','ETRS89 to LN02 height',
+    NULL,
+    'EPSG','9665','Geographic3D to GravityRelatedHeight (gtx)',
+    'EPSG','4937', -- source CRS (ETRS89)
+    'EPSG','5728', -- target CRS (LN02 height)
+    NULL,
+    'EPSG','8666','Geoid (height correction) model file','chgeo2004_htrans_ETRS.agr',
+    NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);
+
+INSERT INTO "usage" VALUES(
+    'PROJ',
+    'EPSG_4937_TO_EPSG_5728_USAGE',
+    'grid_transformation',
+    'PROJ',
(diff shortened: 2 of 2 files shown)


jjimenezshaw:
Testing the added grid file, I found an issue,
- Transforming from ETRS89 (EPSG:4937) to CH1903+ (EPSG:4150) works fine. (No changes here)
`echo 46 9 1000 | ./cs2cs EPSG:4937 "$(./projinfo EPSG:4150 --3d -q -o WKT2:2019)" -d 7`
`46.0011946	9.0011699 949.5849266`
- Transforming from ETRS89 (EPSG:4937) to CH1903+ + LHN95 (EPSG:4150+5729) works fine. (this is the new grid file)
`echo 46 9 1000 | ./cs2cs EPSG:4937 EPSG:4150+5729 -d 7`
`46.0011946	9.0011699 951.2025986`

- However, transforming from CH1903+ + LHN95 (EPSG:4150+5729) to CH1903+ ellipsoidal height (EPSG:4150) doesn't work well
`echo 46 9 1000 | ./cs2cs EPSG:4150+5729 "$(./projinfo EPSG:4150 --3d -q -o WKT2:2019)" -d 7`
`46.0000000	9.0000000 1048.7880435`
The difference should be around 2 meters, not 50.


The first operation in projinfo is this

```
Operation No. 1:

unknown id, CH1903+ to ETRS89 (1) + Inverse of ETRS89 to LHN95 height + Inverse of CH1903+ to ETRS89 (1), unknown accuracy, Liechtenstein; Switzerland.

PROJ string:
+proj=pipeline
  +step +proj=push +v_3
  +step +proj=cart +ellps=bessel
  +step +proj=helmert +x=674.374 +y=15.056 +z=405.346
  +step +inv +proj=cart +ellps=GRS80
  +step +proj=pop +v_3
  +step +proj=vgridshift +grids=ch_swisstopo_chgeo2004_ETRS89_LHN95.tif
        +multiplier=1
  +step +proj=push +v_3
  +step +proj=cart +ellps=GRS80
  +step +proj=helmert +x=-674.374 +y=-15.056 +z=-405.346
  +step +inv +proj=cart +ellps=bessel
  +step +proj=pop +v_3

```
I am not very familiar with the pipeline syntax, but if pop +v_3 is discarding the z transformation, I guess we shouldn't discard it twice. Maybe I am very wrong.

In case you want to test with the official webpage, it is in https://www.swisstopo.admin.ch/en/maps-data-online/calculation-services/reframe.html

hernando:
I'm not very familiar with this syntax either, but I think the problem is in the second `pop +v_3`. Once the grid is applied, the coordinates are in ETRS89 3D, so the full Helmert transformation should be applied to get to CH1903+, not just on xy.

rouault:
I guess the issue is related to the one that was fixed per https://github.com/OSGeo/PROJ/pull/2555 . Likely an additional fix would be needed here in the CompoundCRS to Geog3DCRS case when we use a vertical transformation whose interpolating CRS is not the one of the source/target.

rouault:
Interestingly, removing the +push / +pop before and after the vertical transformation, which would be the right thing to do all transformations in 3D, also lead to the same result as without them. Which is quite expected after more thinking ! So I think PROJ is doing the right thing here. The issue is probably to have to do the ETRS89 <--> CH1903+ datum shift. Using a geoid grid referenced against CH1903+ directly would be the ideal solution.

jjimenezshaw:
Hi @rouault . If I remove the push and pop of the second part of the pipeline, it produces the proper output,
`46.0000000003    9.0000000003      998.3689        0.0000`
Does it make sense to you?
```
echo 46 9 1000 0 | cct +proj=pipeline \
  +step +proj=axisswap +order=2,1 \
  +step +proj=unitconvert +xy_in=deg +xy_out=rad \
  +step +proj=push +v_3 \
  +step +proj=cart +ellps=bessel \
  +step +proj=helmert +x=674.374 +y=15.056 +z=405.346 \
  +step +inv +proj=cart +ellps=GRS80 \
  +step +proj=pop +v_3 \
  +step +proj=vgridshift +grids=ch_swisstopo_chgeo2004_ETRS89_LHN95.tif \
        +multiplier=1 \
  +step +proj=cart +ellps=GRS80 \
  +step +proj=helmert +x=-674.374 +y=-15.056 +z=-405.346 \
  +step +proj=unitconvert +xy_in=rad +xy_out=deg \
  ```

**Comment by hernando** [comment]
To me it makes total sense. Notice that the very first transformation in the pipeline is advertised as `CH1903+ to ETRS89 (1)` but it reality you want `CH1903+ + LHN95 to ETRS89 + LHN95 (1)`, which can be achieved by not transforming z as PROJ is doing, so this one is correct. The problem is the last transformation, from ETRS89 to CH1903+, which should be 3D, but for some reason it's not.
What does it happen if you try to transfrom from EPSG:4258+4729 (ETRS89 + LHN95) to EPSG:4150? (despite I know that geodetically coordinates in EPSG:4258+4729 may not make much sense).

**Comment by rouault** [comment]
> Does it make sense to you?

ah yes, I can see now that the push/pop in the first part is appropriate since the Z value at that point is an orthometric elevation, and it shouldn't be modified by the horizontal datum shift. But after appling the vgridshift we get an ellipsoid height, and it needs to be transformed using a full 3D Helmert. So there's definitely something that should be improved in the pipeline computation engine for such a case. The issue is generic and not specific to your addition, so for me there's nothing preventing the deployment of your additions. The issue spotted here should be reported as an issue.

**Comment by hernando** [comment]
A similar issue can be reproduced with a simpler pipeline:
```
echo 45 8 0 | cs2cs EPSG:4258+5773 EPSG:4150 -d 12
45.000000000000	8.000000000000 44.668468475342
```
The height should be closer to 0 as the Bessel ellipsoid is near EGM96 on Europe.

This is also surprising to me:
```
echo 45 8 0 | cs2cs EPSG:4937 "`projinfo --single-line -o WKT2:2019 -q --3d EPSG:4150`" -d 12
45.000000000000	8.000000000000 0.000000000000
```
but
```
echo 45 8 0 | cs2cs EPSG:4937 EPSG:4150 -d 12
45.001085134820	8.001001396607 0.000000000000
```
I don't know if all are related or they are separate issues, because in this case projinfo is display a ballpark transformation.

rouault:
> A similar issue can be reproduced with a simpler pipeline:

yes,  pretty much the same situation in ETRS89 + EGM96 height to CH1903+  as CH1903+ + LHN95 (EPSG:4150+5729) to CH1903+ 3D, except that it is in fact a bit more complicated since the source and target 2D datums are not the same (ETRS89 vs CH1903+). The simpler pipeline is just due to the fat that the ETRS89 to WGS84 initial step is a null Helmert shift.

The issue with EPSG:4937 to CH1903+ 3D is a different one however. The problem here is that you use a point outside of the area of use of the CH1903+ to ETRS89 (1) Helmert transformation whose southern most latitude is 45.85. In the 2D case, some logic only keeps the Helmert transformation and does not generate the no-op null transformation or remove it at some point. In the 3D case, the promotion to 3D removes the area of use in the definition of the CRS (it probably shouldn't), and thus both the CH1903+ to ETRS89 (1) Helmert transformation and a no-op one are generated.

hernando:
> The issue with EPSG:4937 to CH1903+ 3D is a different one however. The problem here is that you use a point outside of the area of use of the CH1903+ to ETRS89 (1) 

I thought I was inside the area of use, but I made a mistake picking the latitude. This was not intended.

rouault:
>     * However, transforming from CH1903+ + LHN95 (EPSG:4150+5729) to CH1903+ ellipsoidal height (EPSG:4150) doesn't work well

I haven't tested, but I'm confident this should be fixed per https://github.com/OSGeo/PROJ/pull/2592 (was tricky to fix while keeping existing use cases working)

rouault:
@jjimenezshaw Is the draft status of this PR still needed ?

jjimenezshaw:
@rouault sorry for the delay. I am "talking" with swisstopo to see if there is any issue. I hope it will be ready for review soon.

jjimenezshaw:
Thanks @rouault for #2592 . I rebased this PR on top of master after your merge, and it works.

rouault:
@jjimenezshaw I see we still have those custom entries. Has Swisstopo approached IOGP to register them in the EPSG dataset ?

jjimenezshaw:
@rouault I have written today to the contact person in swisstopo explaining it. Let's see when is it finally in EPSG.

PR #2585, opened 2021-03-19
SQL output: make it possible to export non-EPSG projection methods or…

rouault:
… methods and params lacking an explicit EPSG id


kbevers:
@rouault this is a continuation of #2577, right?

rouault:
> @rouault this is a continuation of #2577, right?

yes

kbevers:
Got it. Note to future self: Bundle this and #2577 as one item in the release notes.

Issue #2615, opened 2021-03-28
The extent of EPSG extent 3593 'New Zealand - offshore' appears to be backwards across the date line. (Maybe EPSG bug?)


rhuijben:
* Example of problem

If my demo application (part of SharpProj. A new wrapping around PROJ for C#/.Net users) displays EPSG 3851
![image](https://user-images.githubusercontent.com/2271549/112746654-8bf46d80-8fb0-11eb-87e0-88781f4eb42b.png)

It shows the entire world, it finds the bounds East-West of 160.6 - -171.2, while EPSG.org shows the region as

![image](https://user-images.githubusercontent.com/2271549/112746695-e7bef680-8fb0-11eb-9905-bdb7826f04fd.png)

* Problem description

As the region is wrapped around, the bounding box excludes what actually should be included.

* Expected Output

The region should be ok, so tools relying on the output should work correctly.

The information is in extend.sql, and appears auto generated. So I expect a bug somewhere in the conversion chain.
```
a9b6f3949 (** 2020-09-24 22:41:59 +0200 2572) INSERT INTO "extent" VALUES('EPSG','3593','New Zealand - offshore','New Zealand - offshore.',-55.95,-25.88,160.6,-171.2,0);
```

*  Environment Information
PROJ 8.0.0
Windows

rhuijben:
Looks like a different issue

rouault:
> The information is in extend.sql, and appears auto generated. So I expect a bug somewhere in the conversion chain.

It is expected, by convention, for extents crossing the antimerdian to have west_lon (close but a bit below to +180) and east_lon (close but a bit above to -180)

PR #2616, opened 2021-03-28
Fix bounding box calculations around the 180 degrees longitude.

rhuijben:
The bounding box calculations in 4D_api.cpp (which are used to calculate the 'best transform') are not capable of wrapping properly around at +/- 180 degrees.

This fixes the issue I originally misreported as issue #2615.

For EPSG 3851 this fixes the bounding box calculations from:
image

to
image

Or in text:
The bounding box used to be calculated as

* Unexpected
X: -4432415 - 8601774
Y: -5984083 - 12006385

* Expected
X: 5228059 - 8692628
Y: 1722484 - 4624385

I don't have an example where PROJ makes a bad decision based on this, but others might re-use the code like I did in SharpProj.

### DIFF
--- a/src/4D_api.cpp
+++ b/src/4D_api.cpp
@@ -988,17 +988,35 @@ static void reproject_bbox(PJ* pjGeogToCrs,
     if( !(west_lon == -180.0 && east_lon == 180.0 &&
           south_lat == -90.0 && north_lat == 90.0) )
     {
+        double east_step;
+
         minx = -minx;
         miny = -miny;
         maxx = -maxx;
         maxy = -maxy;
 
+        if (east_lon >= west_lon)
+        {
+            east_step = (east_lon - west_lon) / 20;
+        }
+        else
+        {
+            east_step = (east_lon - west_lon + 360.0) / 20;
+        }
+
         std::vector<double> x(21 * 4), y(21 * 4);
         for( int j = 0; j <= 20; j++ )
         {
-            x[j] = west_lon + j * (east_lon - west_lon) / 20;
+            double lon_step = west_lon + j * east_step;
+
+            if (lon_step > 180.0)
+            {
+                lon_step -= 360.0;
+            }
+
+            x[j] = lon_step;
             y[j] = south_lat;
-            x[21+j] = west_lon + j * (east_lon - west_lon) / 20;
+            x[21+j] = lon_step;
             y[21+j] = north_lat;
             x[21*2+j] = west_lon;
             y[21*2+j] = south_lat + j * (north_lat - south_lat) / 20;


rhuijben:
It looks like some of the callers tried to work around this issue in a way that I can't explain... And are potentially symptoms resolving for this case.

Might need further investigation from someone with more historic knowledge of proj.

rouault:
I'm curious how you tested this ?

snowman2:
@rhuijben, this is something I am working on adding to `pyproj` and your fix seems to work:
https://github.com/pyproj4/pyproj/pull/810

rhuijben:
I started writing my own code, then improved it by using this bit... And found this issue while the code was almost identical.

I also tried to do the right thing for projections of the entire world, which are explicitly excluded in the current proj code.

Current code below
https://github.com/AmpScm/SharpProj/blob/c963df96388859ecf9f1b83924cb09ee30c9ee8f/src/SharpProj/UsageArea.cpp#L101

This variant handles more bbox cases correctly than the code in proj, but proj doesn't need these for its current use case. But there is always room for improvement.. I doubt it is possible to get this always right for the most extreme projections, shaped like those two ESRI cases mentioned in SharpProj.

rhuijben:
That code I mentioned in the previous comment is work in progress. This link might work better if changes are committed.
https://github.com/AmpScm/SharpProj/blob/main/src/SharpProj/UsageArea.cpp#L101

I'm not sure how to handle the extreme edge cases like ESRI:102034
image

snowman2:
I looked into two of the projections you were working on solutions for and noticed this:
```
>>> from pyproj import CRS, Transformer
>>> crs = CRS("ESRI:102036")
>>> crs.area_of_use
AreaOfUse(west=-180.0, south=-90.0, east=180.0, north=0.0, name='Southern hemisphere.')
>>> transformer = Transformer.from_crs(crs.geodetic_crs, crs, always_xy=True)
>>> transformer.transform_bounds(*crs.area_of_use.bounds, errcheck=True)
...
pyproj.exceptions.ProjError: itransform error: Point outside of projection domain
>>> transformer.transform_bounds(-179.9, -89.9, 179.9, -0.01, errcheck=True)
(-36177001911.29046, -36543977094.82756, 36177001911.29046, 36544032754.558266)
>>> crs = CRS("ESRI:54026")
>>> crs.area_of_use
AreaOfUse(west=-180.0, south=-90.0, east=180.0, north=90.0, name='World.')
>>> transformer = Transformer.from_crs(crs.geodetic_crs, crs, always_xy=True)
>>> transformer.transform_bounds(*crs.area_of_use.bounds, errcheck=True)
(-1.5396069930641926e-07, -179545823.75395507, 1.5396069930641926e-07, 179545823.75395507)
>>> transformer.transform_bounds(-179.9, -89.9, 179.9, 89.9, errcheck=True)
(-14617609539.54119, -179719210.54606983, 14617609539.54119, 179719210.54606983)
```
It seems like the global bounds need to have some epsilon removed for it to work properly. I wonder how PROJ handles this?

rhuijben:
Proj explicitly ignores positive infinity values and my code explicitly ignores both positive and negative infinity, the value transformed at quite a few boundary values for this projection.

I also see a small rounding error round +180 degrees where it is handled differently than -180 degrees for this transform, which affects my min-max code.

But luckily I don't think anybody would really use this projection at the boundaries. (It is completely asymptotic). I think this projection was added as projection testcase by ESRI, or they would have used more sensible bounds. (Perhaps 30 degrees instead of 0 as boundary on the equator side)

rouault:
I don't think this pull request is needed, or at least in its current form. You're probably speaking about a more general issue (reprojection of a arbitrary bbox between 2 arbitary CRS) than what this part of the code deal with, or I need to be exhibited a concrete test case where the current code fails using cs2cs.
pj_create_prepared_operations() has already a special case at line 1250 to deal with the anti-meridian by cutting a bbox in 2 parts, so west_lon <= east_lon always holds true in reproject_bbox().
Testing with EPSG:3460 ("Fiji 1986 / Fiji Map Grid") or EPSG:4720 ("Fiji 1986" geographic) to EPSG:4326 with points west and east of the antimeridian shows that the datum shift between Fiji 1986 and WGS84 is properly applied.
I guess there could be issues if a *transformation* had a really huge extent, which failed to transform into the source / target coordinates, but should be exhibited...

rhuijben:
My use case is indeed just creating the proper bounding box, which is currently not needed by proj itself, but appears to be a common request outside proj.

But I still think doing the right thing here and then just make the caller use the result is better than making the caller split the world and calculating multiple bounding boxes (potentially doing network requests for gridshifts, etc.). 

The result of the multiple operations probably doesn't produce clean rectangles on both sides of the split, with overlap, etc. It should make it easier to prove the code is correct by removing edge cases.

I do agree that it shouldn't affect current behavior of proj as the current caller expects this behavior.


Thanks for the code improvements @rouault

rouault:
Your proposed patch is probably good for most bounding box reprojections, but not the geographic to geographic/projected situations used in the context of that file. At least https://github.com/OSGeo/PROJ/pull/2654 demonstrates a case where applying your patch would result in a regression.

snowman2:
Related: #2882

PR #2618, opened 2021-03-28
GitHub Action to build both the 'proj' and 'proj-docs' container 

hobu:
* Build both the 'proj' and the 'proj-docs' container
* Push the containers to both dockerhub and ghcr.io only if we are a commit in the OSGeo/PROJ repository
* Label both containers with the SHA or tag (if the commit is a tag)
* Label both with ``latest``. I think this might need some consideration. We use this code for PDAL, but we only ever push containers out of the ``master`` branch. PROJ could adopt that posture, or it could adjust how it labels the ``latest`` to follow only when a push to a maintenance branch happened.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,67 @@
+name: Docker
+
+on: [push, pull_request, workflow_dispatch]
+
+# adapted from https://raw.githubusercontent.com/stefanprodan/podinfo/master/.github/workflows/release.yml
+#
+jobs:
+  containers:
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


rouault:
Looks good to me

hobu:
I am likely to have to churn the CI a bit to get pushes properly working. I will do so on another branch at osgeo/proj and then merge it once complete.

hobu:
Pushes look good now. We are pushing to both ghcr.io at https://github.com/orgs/OSGeo/packages and DockerHub. When DockerHub finally stops allowing open source downloads, we should complete the switch to GitHub's registry, which shouldn't have such restrictions.

https://github.com/orgs/OSGeo/packages

PR #2619, opened 2021-03-28
Database: add grid aliases for transformations using BALR2009.gsb and PENR2009.gsb

rouault:
Fixes issue reported at https://lists.osgeo.org/pipermail/gdal-dev/2021-March/053771.html

The issue comes from the fact that EPSG has created 2 transformations
using grids BALR2009.gsb ad PENR2009.gsb that supersede the one which
uses the single grid SPED2ETV2 we have in PROJ-data


rouault:
More general fix in https://github.com/OSGeo/PROJ/pull/2622

PR #2620, opened 2021-03-28
Fix build on Solaris 11.4

rouault:


hobu:
This fix seems not worth it. It makes the code more confusing, and do we really need to care about ancient Solaris?

rouault:
Actually, it matches more the signature of the manpage of dladdr() on Linux too:
```
       #define _GNU_SOURCE
       #include <dlfcn.h>

       int dladdr(void *addr, Dl_info *info);
```

rouault:
But interestingly, on Linux:
```
$ grep dladdr /usr/include/dlfcn.h 
extern int dladdr (const void *__address, Dl_info *__info)
```
So it thinks that Linux has 'hardened' things, but POSIX has a looser definition.

PR #2622, opened 2021-03-28
createOperation(): make sure no to discard deprecated operations...

rouault:
if the replacement uses a grid unknown to us.

Fixes issue reported at https://lists.osgeo.org/pipermail/gdal-dev/2021-March/053771.html

The issue comes from the fact that EPSG has created 2 transformations
using grids BALR2009.gsb ad PENR2009.gsb that supersede the one which
uses the single grid SPED2ETV2 we have in PROJ-data.

More general fix than the hack of https://github.com/OSGeo/PROJ/pull/2619

### DIFF
--- a/src/iso19111/factory.cpp
+++ b/src/iso19111/factory.cpp
@@ -5810,12 +5810,15 @@ AuthorityFactory::createFromCoordinateReferenceSystemCodes(
 
     std::string sql;
     if (discardSuperseded) {
-        sql = "SELECT source_crs_auth_name, source_crs_code, "
-              "target_crs_auth_name, target_crs_code, "
+        sql = "SELECT cov.source_crs_auth_name, cov.source_crs_code, "
+              "cov.target_crs_auth_name, cov.target_crs_code, "
               "cov.auth_name, cov.code, cov.table_name, "
               "extent.south_lat, extent.west_lon, extent.north_lat, "
               "extent.east_lon, "
-              "ss.replacement_auth_name, ss.replacement_code FROM "
+              "ss.replacement_auth_name, ss.replacement_code, "
+              "(gt.auth_name IS NOT NULL) AS replacement_is_grid_transform, "
+              "(ga.proj_grid_name IS NOT NULL) AS replacement_is_known_grid "
+              "FROM "
               "coordinate_operation_view cov "
               "JOIN usage ON "
               "usage.object_table_name = cov.table_name AND "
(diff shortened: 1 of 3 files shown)


jjimenezshaw:
The "new" grids are in https://www.ign.es/web/ign/portal/gds-rejilla-cambio-datum (that are not new. They just updated in EPSG what was in their page for a long time)
However, I guess that the data in `es_ign_SPED2ETV2.tif` is the same, combining both files BALR2009.gsb and PENR2009.gsb in one, with two subdatasets.

The "problem" is that `PENR2009.gsb` covers the Balearic Islands, but with un-precise values. The file `BALR2009.gsb` has more accurate data for the Balearic Islands.

jjimenezshaw:
Due to this unfortunate overlap of both files in the Balearic Islands, with different data, I do not know what is going to do PROJ if we define two different gridfiles.

rouault:
> The "problem" is that `PENR2009.gsb` covers the Balearic Islands, but with un-precise values. The file `BALR2009.gsb` has more accurate data for the Balearic Islands.

yes, that was addressed per https://github.com/OSGeo/PROJ-data/pull/32
It would probably be cleaner to add the 2 separate grids in PROJ-data and register them in grid_alternatives. That way users would have the choice to select which grid is the most appropriate.

jjimenezshaw:
> That way users would have the choice to select which grid is the most appropriate.

I think user will not notice that. They will just let PROJ to select what PROJ considers better (even more from GDAL, as in the source problem). Right now, PROJ is doing what IGN says: use `BALR2009.gsb` for the islands, and `PENR2009.gsb` for the mainland (except a really tiny area in the coast). I guess that if we split, PROJ will use always `PENR2009.gsb` because it is bigger, right?

rouault:
> I guess that if we split, PROJ will use always `PENR2009.gsb` because it is bigger, right?

hum, good question. The algorithm is deterministic, but I have not in mind what would happen in that situation. Actually trying "projinfo -s ED50 -t ETRS89 --summary --spatial-test intersects --grid-check none --show-superseded" in master, I see that the transformation "ED50 to ETRS89 (16), 0.2 m, Spain - mainland and Ceuta" is presented before "ED50 to ETRS89 (17), 0.2 m, Spain - Balearic Islands". PROJ would probably need an extra logic potentially fragile to make sure the Baleares is used before the Peninsula

rouault:
Correction to my above comment. The difference I had was because I had BALR2009.gsb in my current directory... So with both master and 8.0 branch, the transformation for Peninsula is presented (and would then be used) before Baleares, likely because of its greater area

jjimenezshaw:
> PROJ would probably need an extra logic potentially fragile to make sure the Baleares is used before the Peninsula

What about what you did in #2619 (I mean, have a single tif)? I know it is a fix for this particular case. Here can be fixed under control having both grids in one file (`es_ign_SPED2ETV2.tif`) with two subdatasets.

I do not see a general way to cover this issue (two grids for the same transformation that overlap) efficiently. Imagine (they are not yet in PROJ-data), that there is a grid for Portugal for the ED50-ETRS89 transformation. Or for France. The overlapping areas will just use the bigger one (France over Spain, and Spain over Portugal).

rouault:
> What about what you did in #2619 (I mean, have a single tif)? I know it is a fix for this particular case

This PR has the same effect as #2619 while being more general

PR #2635, opened 2021-04-02
Database: update to EPSG v10.018

rouault:
Just fixes 4 wrong transformations of v10.017


jmckenna:
you're fast, 2 hours faster than I can receive their notification that the new version is available, ha.  Thanks @rouault !

rouault:
>  2 hours faster than I can receive their notification that the new version is available

No magic: Roger Lott to whom I had reported the issues I found in 10.017 emailed me about the new release being available with the fixes.

PR #2653, opened 2021-04-07
CI: build and test on MSYS2 with Windows

mwtoews:
A few adjustments were required to run the POSIX versions of CMake and pkg-config post-install tests.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -53,7 +53,6 @@ jobs:
             set VCPKG_INSTALLED=c:\vcpkg\installed\${{ env.ARCH }}-windows
             set PATH=%VCPKG_INSTALLED%\bin;%PATH%
             set PROJ_DIR=%GITHUB_WORKSPACE%\proj_dir
-            set PROJ_LIB=%PROJ_DIR%\share\proj
             set PROJ_BUILD=%GITHUB_WORKSPACE%\build
             cd %PROJ_BUILD%
             ctest -V -C Release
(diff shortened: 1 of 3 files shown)


.github/workflows/windows.yml
```diff
@@ -60,3 +60,54 @@ jobs:
...
+            cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX="${PROJ_DIR}" ..
+            make -j 2
+            make install
+            ls ${PROJ_DIR}/bin
+
+      - name: Run tests
+        run: |
+            PROJ_BUILD=${GITHUB_WORKSPACE}/build
+            PROJ_DIR=${GITHUB_WORKSPACE}/proj_dir
+            export PROJ_LIB=${PROJ_DIR}/share/proj
```
rouault: do we actually need setting PROJ_LIB ? This should be hard coded in the binary . The same also applies for the MSVC build actually

mwtoews: I had assumed there was a reason you did this for the MSVC build above, but let's drop these and see how it goes...

_Update:_ these were unnecessary, tests ran as expected without `PROJ_LIB` explicitly set.

rouault:
> Is there much need for a cross-compile CI run on Linux with mingw_w64.yml?

makes it easier to test Windows builds for folks who have urticaria flare-up when they have to fire a Windows VM :-)

PR #2655, opened 2021-04-09
Add projinfo option --list-crs 

jjimenezshaw:
- [x] Tests added

Add the option `--list-crs` to dump the CRSs in the database using the API `proj_get_crs_info_list_from_database`.
It dumps a line per filtered CRS with the syntax `auth_name:auth_code "name"`. If it is deprecated, adds ` [deprecated]`.

Some examples:
```
projinfo --list-crs
EPSG:3819 "HD1909"
EPSG:3821 "TWD67"
EPSG:3822 "TWD97"
...
IGNF:WGS84UTM32.IGN69 "WGS84 UTM NORD FUSEAU 32 et NGF-IGN 1969"
IGNF:WGS84UTM32.IGN78C "WGS84 UTM NORD FUSEAU 32 et NGF-IGN 1978"

projinfo --list-crs | wc -l
8559

projinfo --list-crs allow_deprecated
EPSG:3819 "HD1909"
EPSG:3821 "TWD67"
EPSG:3822 "TWD97"
EPSG:3823 "TWD97"
EPSG:3824 "TWD97"
EPSG:3887 "IGRS"
EPSG:3888 "IGRS"
EPSG:3889 "IGRS"
EPSG:3906 "MGI 1901"
EPSG:4000 "MOLDREF99"
EPSG:4001 "Unknown datum based upon the Airy 1830 ellipsoid" [deprecated]
EPSG:4002 "Unknown datum based upon the Airy Modified 1849 ellipsoid" [deprecated]
EPSG:4003 "Unknown datum based upon the Australian National Spheroid" [deprecated]
...

projinfo --list-crs allow_deprecated | wc -l
10263

projinfo --list-crs --authority EPSG
EPSG:3819 "HD1909"
...
EPSG:9705 "WGS 84 + MSL height"
EPSG:9707 "WGS 84 + EGM96 height"

projinfo --list-crs --authority EPSG,OGC
EPSG:3819 "HD1909"
...
EPSG:9707 "WGS 84 + EGM96 height"
OGC:CRS27 "NAD27 (CRS27)"
OGC:CRS83 "NAD83 (CRS83)"
OGC:CRS84 "WGS 84 (CRS84)"

projinfo --list-crs projected
EPSG:2000 "Anguilla 1957 / British West Indies Grid"
EPSG:20004 "Pulkovo 1995 / Gauss-Kruger zone 4"
...
EPSG:30171 "Tokyo / Japan Plane Rectangular CS XI"
EPSG:30172 "Tokyo / Japan Plane Rectangular CS XII"

projinfo --list-crs vertical,compound --authority EPSG
EPSG:3855 "EGM2008 height"
EPSG:3886 "Fao 1979 height"
...
EPSG:9705 "WGS 84 + MSL height"
EPSG:9707 "WGS 84 + EGM96 height"

projinfo --list-crs projected --bbox -100,40,-90,41 
EPSG:26714 "NAD27 / UTM zone 14N"
EPSG:26715 "NAD27 / UTM zone 15N"
EPSG:26716 "NAD27 / UTM zone 16N"
...
```

### DIFF
--- a/docs/source/apps/projinfo.rst
+++ b/docs/source/apps/projinfo.rst
@@ -294,6 +294,7 @@ The following control parameters can appear in any order:
     If no filter is provided all authority names and types of non deprecated CRSs are dumped.
     list-crs-filter is a comma separated combination of: allow_deprecated,geodetic,geocentric,
     geographic,geographic_2d,geographic_3d,vertical,projected,compound.
+    Affected by options `--authority`, `--bbox`, `--main-db-path` and `--aux-db-path`    
 
 .. option:: --3d
 
--- a/docs/source/apps/projinfo.rst
+++ b/docs/source/apps/projinfo.rst
@@ -294,7 +294,7 @@ The following control parameters can appear in any order:
     If no filter is provided all authority names and types of non deprecated CRSs are dumped.
     list-crs-filter is a comma separated combination of: allow_deprecated,geodetic,geocentric,
     geographic,geographic_2d,geographic_3d,vertical,projected,compound.
-    Affected by options `--authority`, `--bbox`, `--main-db-path` and `--aux-db-path`    
+    Affected by options :option:`--authority` and :option:`--bbox`
 
 .. option:: --3d
 
--- a/docs/source/apps/projinfo.rst
+++ b/docs/source/apps/projinfo.rst
@@ -292,9 +292,8 @@ The following control parameters can appear in any order:
 
     Outputs a list (authority name:code and CRS name) of the filtered CRSs from the database.
     If no filter is provided all authority names and types of non deprecated CRSs are dumped.
-    list-crs-filter is a comma separated combination of: auth_name=XXX,allow_deprecated,geodetic,geocentric,
-    geographic,geographic_2d,geographic_3d,vertical,protected,compound,temporal,engineering,bound,other
-    where XXX is the authority. When auth_name is provided multiple times, it is included for each one.
+    list-crs-filter is a comma separated combination of: allow_deprecated,geodetic,geocentric,
+    geographic,geographic_2d,geographic_3d,vertical,projected,compound.
 
 .. option:: --3d
 
--- a/docs/source/apps/projinfo.rst
+++ b/docs/source/apps/projinfo.rst
@@ -58,9 +58,6 @@ Synopsis
     - a PROJJSON string. The jsonschema is at https://proj.org/schemas/v0.2/projjson.schema.json (*added in 6.2*)
     - a compound CRS made from two object names separated with " + ". e.g. "WGS 84 + EGM96 height" (*added in 7.1*)
 
-    [list-crs-filter] is a comma separated combination of: auth_name=XXX,allow_deprecated,geodetic,geocentric,
-    geographic,geographic_2d,geographic_3d,vertical,protected,compound,temporal,engineering,bound,other
-
     {object_reference} is a filename preceded by the '@' character.  The
     file referenced by the {object_reference} must contain a valid
     {object_definition}.
--- a/docs/source/apps/projinfo.rst
+++ b/docs/source/apps/projinfo.rst
@@ -32,6 +32,7 @@ Synopsis
     |    [--output-id AUTH:CODE]
     |    [--c-ify] [--single-line]
     |    --searchpaths | --remote-data |
+    |    --list-crs [list-crs-filter] |
     |    --dump-db-structure [{object_definition} | {object_reference}] |
     |    {object_definition} | {object_reference} | (-s {srs_def} -t {srs_def})
     |
@@ -56,6 +57,9 @@ Synopsis
       (e.g. "urn:ogc:def:coordinateOperation,coordinateOperation:EPSG::3895,coordinateOperation:EPSG::1618")
     - a PROJJSON string. The jsonschema is at https://proj.org/schemas/v0.2/projjson.schema.json (*added in 6.2*)
     - a compound CRS made from two object names separated with " + ". e.g. "WGS 84 + EGM96 height" (*added in 7.1*)
+
+    [list-crs-filter] is a comma separated combination of: auth_name=XXX,allow_deprecated,geodetic,geocentric,
+    geographic,geographic_2d,geographic_3d,vertical,protected,compound,temporal,engineering,bound,other
 
     {object_reference} is a filename preceded by the '@' character.  The
     file referenced by the {object_reference} must contain a valid
@@ -284,6 +288,16 @@ The following control parameters can appear in any order:
     database. This option can be specified as the only switch of the utility.
     If also specifying a CRS object and the :option:`--output-id` option, the
     definition of the object as SQL statements will be appended.
+
+.. option:: --list-crs [list-crs-filter]
+
+    .. versionadded:: 8.1
+
+    Outputs a list (authority name:code and CRS name) of the filtered CRSs from the database.
+    If no filter is provided all authority names and types of non deprecated CRSs are dumped.
+    list-crs-filter is a comma separated combination of: auth_name=XXX,allow_deprecated,geodetic,geocentric,
+    geographic,geographic_2d,geographic_3d,vertical,protected,compound,temporal,engineering,bound,other
+    where XXX is the authority. When auth_name is provided multiple times, it is included for each one.
 
 .. option:: --3d
 
--- a/src/apps/projinfo.cpp
+++ b/src/apps/projinfo.cpp
@@ -130,10 +130,9 @@ static void usage() {
               << std::endl;
     std::cerr << std::endl;
     std::cerr << "list-crs-filter is a comma separated combination of: "
-                 "auth_name=XXX,allow_deprecated,geodetic,geocentric,"
+                 "allow_deprecated,geodetic,geocentric,"
               << std::endl;
-    std::cerr << "geographic,geographic_2d,geographic_3d,vertical,"
-                 "protected,compound,temporal,engineering,bound,other"
+    std::cerr << "geographic,geographic_2d,geographic_3d,vertical,projected,compound"
               << std::endl;
     std::cerr << std::endl;
     std::cerr << "{object_definition} might be a PROJ string, a WKT string, "
@@ -1267,15 +1266,8 @@ int main(int argc, char **argv) {
         auto params = proj_get_crs_list_parameters_create();
         bool all_types = false;
         std::vector<PJ_TYPE> types;
-        std::vector<std::string> auth_names;
         auto tokens = split(listCRS, ',');
         for (auto token : tokens) {
-            auto auth_tokens = split(token, '=');
-            if (auth_tokens.size() == 2 && ci_equal(auth_tokens[0], "auth_name")) {
-                auth_names.push_back(auth_tokens[1]);
-                continue;
-            }
-
             if (ci_equal(token, "allow_deprecated")) {
                 params->allow_deprecated = true;
             } else if (ci_equal(token, "all_types")) {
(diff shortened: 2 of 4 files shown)


src/apps/projinfo.cpp
```diff
@@ -128,6 +129,13 @@ static void usage() {
                  "by '-' to disable them"
               << std::endl;
     std::cerr << std::endl;
+    std::cerr << "list-crs-filter is a comma separated combination of: "
+                 "auth_name=XXX,allow_deprecated,geodetic,geocentric,"
+              << std::endl;
+    std::cerr << "geographic,geographic_2d,geographic_3d,vertical,"
```
rouault: you can remove temporal,engineering,bound,other : none of them can be stored in the DB currently. We can easily add them later. Removing them will make the doc clearer.

docs/source/apps/projinfo.rst
```diff
@@ -57,6 +58,9 @@ Synopsis
     - a PROJJSON string. The jsonschema is at https://proj.org/schemas/v0.2/projjson.schema.json (*added in 6.2*)
     - a compound CRS made from two object names separated with " + ". e.g. "WGS 84 + EGM96 height" (*added in 7.1*)
 
+    [list-crs-filter] is a comma separated combination of: auth_name=XXX,allow_deprecated,geodetic,geocentric,
```
rouault: you can remove the explanation for list-crs-filter from here. It belongs to the one the ``list-crs`` option.

src/apps/projinfo.cpp
```diff
@@ -1248,6 +1263,76 @@ int main(int argc, char **argv) {
...
+        }
+
+        if (all_types) {
+            types.clear();
+        }
+        params->typesCount = types.size();
+        params->types = types.data();
+
+        if (auth_names.empty()) {
+            auth_names.push_back("");
```
rouault:
```suggestion
            auth_names.push_back(std::string());
```

src/apps/projinfo.cpp
```diff
@@ -1248,6 +1263,76 @@ int main(int argc, char **argv) {
...
+                usage();
+            }
+        }
+
+        if (all_types) {
+            types.clear();
+        }
+        params->typesCount = types.size();
+        params->types = types.data();
+
```
rouault, line 1305: Could you take into account the --bbox and set it in the params ?
It would be good also to take into account --area and do post filtering

src/apps/projinfo.cpp
```diff
@@ -1248,6 +1263,76 @@ int main(int argc, char **argv) {
...
 
+    if (!listCRS.empty()) {
+        auto params = proj_get_crs_list_parameters_create();
+        bool all_types = false;
+        std::vector<PJ_TYPE> types;
+        std::vector<std::string> auth_names;
+        auto tokens = split(listCRS, ',');
+        for (auto token : tokens) {
+            auto auth_tokens = split(token, '=');
+            if (auth_tokens.size() == 2 && ci_equal(auth_tokens[0], "auth_name")) {
```
rouault: projinfo has also a ``--authority`` switch. It would probably better to re-use them instead of stuffing it again in crs-filter

src/apps/projinfo.cpp
```diff
@@ -1248,6 +1263,76 @@ int main(int argc, char **argv) {
...
+        params->typesCount = types.size();
+        params->types = types.data();
+
+        if (auth_names.empty()) {
+            auth_names.push_back("");
+        }
+
+        for (auto auth_name : auth_names) {
+            int result_count = 0;
+            auto list = proj_get_crs_info_list_from_database(nullptr, auth_name.c_str(), params, &result_count);
```
rouault: you should likely use proj_context_create() and proj_context_set_database_path() to take into account --main-db-path and --aux-db-path

rouault (changes requested):
Regarding testing, add new scenario(s) in test/cli/testprojinfo , run "make check" . It will error out. And then move the result of {build_tree}/test/cli/testprojinfo_out to {source_tree}/test/cli/testprojinfo_out.dist

src/apps/projinfo.cpp
jjimenezshaw: Is there any way to extract the w,e,n,s values from the `Extent bboxFilter`?

src/apps/projinfo.cpp
jjimenezshaw: I understand that there is not way to extract the `PJ_CONTEXT*` from `DatabaseContextPtr dbContext` that is already created.

src/apps/projinfo.cpp
rouault: yes, but that's a bit convoluted. Simpler would be to just store them in dedicated local variables when parsing --bbox

src/apps/projinfo.cpp
rouault: > I understand that there is not way to extract the `PJ_CONTEXT*` from `DatabaseContextPtr dbContext` that is already created.

not reallly indeed. A DatabaseContext could potentially exist without being linked to a PJ_CONTEXT* (historically the C++ part of the codebase has been developed quite independently from the C part, hence they are loosely decoupled). An alternative might be that you look at the implementation of proj_get_crs_info_list_from_database() to still from it what you need. The C function largely uses the C++ API and thus you could re-use the dbContext already constructed

src/apps/projinfo.cpp
jjimenezshaw: Ok. It is simple.
I tried `projinfo --list-crs vertical  --bbox 0,40,1,41` that included `Alicante height`. However `projinfo --list-crs vertical  --bbox 0,40,10,41` does not included it. It sounds strange to me. Is that right?

rouault: see the crs_area_of_use_contains_bbox member of PROJ_CRS_LIST_PARAMETERS. What is missing in the doc is that its default value is TRUE

jjimenezshaw:
> Regarding testing, add new scenario(s) in test/cli/testprojinfo , run "make check" . It will error out. And then move the result of {build_tree}/test/cli/testprojinfo_out to {source_tree}/test/cli/testprojinfo_out.dist

Using `cmake` and `ctest` does not leave the file `testprojinfo_out` anywhere. Is there any way to do it (apart from using plain `make`)?

Because the output of this feature is terribly verbose (thousands of lines), I will try something creative to reduce the size of the output.

rouault:
> Using `cmake` and `ctest` does not leave the file `testprojinfo_out` anywhere. Is there any way to do it (apart from using plain `make`)?

cmake/ctest should be fine, but those shell-based test scripts require Linux/Mac (I guess they might work with MSys2 on Windows, but some tweaking of CMakeList might be required)

> Because the output of this feature is terribly verbose (thousands of lines), I will try something creative to reduce the size of the output.

If you restrict to authorities lighly populated like OGC, or apply a --area / --bbox filter that should make it quite reasonable

src/apps/projinfo.cpp
jjimenezshaw: done

docs/source/apps/projinfo.rst
jjimenezshaw: done

src/apps/projinfo.cpp
jjimenezshaw: I added `--bbox`. Including `--area` would be more complicated. It needs some code reorganization. Is it really useful?
In that case it would make sense to create an `Extent` (either from --bbox or --area), and intersect it on a post filter.
BTW, I set `crs_area_of_use_contains_bbox` to false.

src/apps/projinfo.cpp
jjimenezshaw: done

src/apps/projinfo.cpp
jjimenezshaw: Using the code from C++ implementation is a bit ugly: repeating the logic to filter the types, bbox, etc.
Maybe I try later to see how it looks like, in a different branch to compare.

jjimenezshaw:
Regarding the tests, should I add a `| sort`, to avoid false positives just because the internal query is returning the values in a different order?
Travis answered it ;)

src/apps/projinfo.cpp
rouault: > Is it really useful?

It would be nice to have "--list-crs --area France" working. But that can be a later improvement. The .rst doc of --list-crs should mention which other options can be conbined with it

rouault:
> Regarding the tests, should I add a `| sort`, to avoid false positives just because the internal query is returning the values in a different order?

hum, no, the best solution would be to fix the SQLite3 query to include a deterministic sorting order

test/cli/testprojinfo
```diff
@@ -310,6 +310,38 @@ echo 'Checks that ED50 to ETRS89 (12) is in the output (superseded transformatio
 $EXE -s EPSG:23030 -t EPSG:25830 --bbox -6,40,-5,41 --grid-check known_available --hide-ballpark --summary >>${OUT} 2>&1
 echo "" >>${OUT}
 
+echo 'Testing --list-crs | grep "EPSG:432\|EPSG:3263\|ESRI:103668\|OGC" | sort' >> ${OUT}
```
rouault: Those tests are going to be fragile when the EPSG database is extended. Could you add more restrictive grepping to restrict to one or two max CRS in output ?

jjimenezshaw: No Problem

src/apps/projinfo.cpp
jjimenezshaw: >  But that can be a later improvement.

Yes.

>The .rst doc of --list-crs should mention which other options can be conbined with it

It already has as "Affected by options `--authority`, `--bbox`, `--main-db-path` and `--aux-db-path`". It that enough?

jjimenezshaw: > --area France

Do you mean to filter only the CRSs that exactly has the area name "France", or do a more complicated geometry, like area intersection?

test/cli/testprojinfo
jjimenezshaw: done

docs/source/apps/projinfo.rst
```diff
@@ -285,6 +286,16 @@ The following control parameters can appear in any order:
...
 
+.. option:: --list-crs [list-crs-filter]
+
+    .. versionadded:: 8.1
+
+    Outputs a list (authority name:code and CRS name) of the filtered CRSs from the database.
+    If no filter is provided all authority names and types of non deprecated CRSs are dumped.
+    list-crs-filter is a comma separated combination of: allow_deprecated,geodetic,geocentric,
+    geographic,geographic_2d,geographic_3d,vertical,projected,compound.
+    Affected by options `--authority`, `--bbox`, `--main-db-path` and `--aux-db-path`
```
rouault:
```suggestion
    Affected by options :option:`--authority` and :option:`--bbox`
```
It is kind of implied that --main-db-path and --aux-db-path affect the behavior of the whole utility.

src/apps/projinfo.cpp
rouault: > Do you mean to filter only the CRSs that exactly has the area name "France", or do a more complicated geometry, like area intersection?

The idea would be to look that the name of the extent of the CRS contains the word France.

kbevers:
It seems to me the CLI test could end up being more of a pain than a gain in the long run. Is it going to be stable when updating the contents of the DB?

rouault:
> Is it going to be stable when updating the contents of the DB?

Given the various grep's added, that should be hopefully be quite stable. We can revisit if realizing that it is not enough

src/apps/projinfo.cpp
jjimenezshaw: I have another branch using C++ function directly, and including `--area`. I will do another PR (so you can compare) soon.

rouault:
Superseded by https://github.com/OSGeo/PROJ/pull/2663

PR #2659, opened 2021-04-10
CRS::normalizeForVisualization(): propagate domains/extent of original CRS (fixes #2603)

### LINKED ISSUE #2603: When normalizing axis on a CRS the usage area gets 'lost'.
When you normalize the axis on a CRS (e.g. on EPSG 4326) with proj_normalize_for_visualization(), then you can't call proj_get_area_of_use() on the resulting CRS. It won't set output and returns false.

#### Problem description

proj_get_area_of_use() should work on 'normalized' CRS instances. It looks like it now only works on CRS instances that were already in normalized form.

The usage area is declared as string and in lat/lon, so it isn't affected by the normalization itself and should just work.

#### Expected Output
proj_get_area_of_use() should return true (and the same output as on the original CRS)

#### Environment Information
 - PROJ 8.0.0

rouault:
Only first commit to be backported in 8.0 branch

### DIFF
--- a/include/proj/crs.hpp
+++ b/include/proj/crs.hpp
@@ -142,7 +142,7 @@ class PROJ_GCC_DLL CRS : public common::ObjectUsage,
 
     PROJ_INTERNAL bool mustAxisOrderBeSwitchedForVisualization() const;
 
-    PROJ_INTERNAL CRSNNPtr normalizeForVisualization() const;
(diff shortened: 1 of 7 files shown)


kbevers:
>  projinfo: add a --normalize-axis-order undocument switch 

why is it undocumented?

rouault:
> why is it undocumented?

maybe too specialized. Didn't want to clutter the doc with something that was probably just useful for my own debugging :-)

PR #2662, opened 2021-04-11
Add proj_get_celestial_body_name() and …

rouault:
…make proj_get_crs_info_list_from_database() filter on and return celestial body name


snowman2:
Are there plans to add `proj_get_celestial_bodies_from_database` similar to `proj_get_authorities_from_database`?

rouault:
> Are there plans to add `proj_get_celestial_bodies_from_database` similar to `proj_get_authorities_from_database`?

that would be a good idea. You may create a ticket about that.

PR #2663, opened 2021-04-11
Add projinfo option --list-crs (supports --area)

jjimenezshaw:
- [x] Tests added

Variation of #2655. (Same behavior, different implementation).

It also uses the option `--area`. For that it creates an `Extent` from `--bbox` or `--area` (in the same way as other parts of `projinfo`. New function `makeBboxFilter` for that purpose) and intersects every CRS listed. The behavior is driven by `--spatial-test`

It does not use the C API, but directly the C++ function `getCRSInfoList()`.

### DIFF
--- a/src/apps/projinfo.cpp
+++ b/src/apps/projinfo.cpp
@@ -130,10 +130,9 @@ static void usage() {
               << std::endl;
     std::cerr << std::endl;
     std::cerr << "list-crs-filter is a comma separated combination of: "
-                 "auth_name=XXX,allow_deprecated,geodetic,geocentric,"
+                 "allow_deprecated,geodetic,geocentric,"
               << std::endl;
-    std::cerr << "geographic,geographic_2d,geographic_3d,vertical,"
-                 "protected,compound,temporal,engineering,bound,other"
+    std::cerr << "geographic,geographic_2d,geographic_3d,vertical,projected,compound"
               << std::endl;
     std::cerr << std::endl;
     std::cerr << "{object_definition} might be a PROJ string, a WKT string, "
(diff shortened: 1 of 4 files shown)


src/apps/projinfo.cpp
```diff
@@ -1248,6 +1332,77 @@ int main(int argc, char **argv) {
...
+                auto list = factory->getCRSInfoList();
+                for (const auto &info : list) {
+                    if (!allow_deprecated && info.deprecated) {
+                        continue;
+                    }
+                    if (!types.empty() && types.find(info.type) == types.end()) {
+                        continue;
+                    }
+                    if (bboxFilter) {
+                        auto crsExtent = Extent::createFromBBOX(
```
rouault, line 1392: you should check info.bbox_valid

src/apps/projinfo.cpp
```diff
@@ -1248,6 +1332,77 @@ int main(int argc, char **argv) {
...
+                            info.west_lon_degree, info.south_lat_degree,
+                            info.east_lon_degree, info.north_lat_degree);
+                        if (!bboxFilter->intersects(crsExtent)) {
```
rouault: I'm wondering if we shouldn't take the --spatial-test contains|intersects option into account . The default is SpatialCriterion::STRICT_CONTAINMENT that is we would only list CRS whose area of use is entirely contained in the specified bbox (thus bboxFilter->contains(crsExtent) . And if using --spatial-test intersects, what you do currently. Or if you prefer the current behaviour, in --list-crs mode change the default value to be intersects, but allow the override of --spatial-test contains

jjimenezshaw: It makes sense. Coming in next commit.

src/apps/projinfo.cpp
jjimenezshaw: So the logic should be like this?:
```if (bboxFilter && !info.bbox_valid) { continue; }```
(that means that CRS without extent definition are not listed if `--bbox` or `--area` are used)

rouault: sounds reasonable. That's more an hypothetical sense since a well behaved database should include extent for all CRS

PR #2669, opened 2021-04-13
Parse compound id with two authorities, like ESRI:103668+EPSG:5703

jjimenezshaw:
- [x] Tests added

Following the suggestion here, https://lists.osgeo.org/pipermail/proj/2021-April/010166.html
this PR parses an identifier that combines two authorities.

Useful to add NAVD88 to many ESRI crs not defined in EPSG.

### DIFF
--- a/src/iso19111/io.cpp
+++ b/src/iso19111/io.cpp
@@ -6438,31 +6438,15 @@ static BaseObjectNNPtr createFromUserInput(const std::string &text,
             const auto &authName2 = tokensCenter[1];
             const auto &code2 = tokens[2];
 
-            crs::CRSPtr crs1 = nullptr;
-            crs::CRSPtr crs2 = nullptr;
-
-            const auto authorities = dbContextNNPtr->getAuthorities();
-            for (const auto &authCandidate : authorities) {
-                if (ci_equal(authCandidate, authName1)) {
-                    auto factory =
-                        AuthorityFactory::create(dbContextNNPtr, authCandidate);
-                    crs1 =
-                        factory->createCoordinateReferenceSystem(code1, false);
-                }
-                if (ci_equal(authCandidate, authName2)) {
-                    auto factory =
-                        AuthorityFactory::create(dbContextNNPtr, authCandidate);
-                    crs2 =
-                        factory->createCoordinateReferenceSystem(code2, false);
-                }
-            }
-            if (crs1 && crs2) {
-                return CompoundCRS::createLax(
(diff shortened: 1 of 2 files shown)


src/iso19111/io.cpp
```diff
@@ -6424,6 +6424,46 @@ static BaseObjectNNPtr createFromUserInput(const std::string &text,
...
+
+            crs::CRSPtr crs1 = nullptr;
+            crs::CRSPtr crs2 = nullptr;
+
+            const auto authorities = dbContextNNPtr->getAuthorities();
+            for (const auto &authCandidate : authorities) {
+                if (ci_equal(authCandidate, authName1)) {
+                    auto factory =
+                        AuthorityFactory::create(dbContextNNPtr, authCandidate);
+                    crs1 =
```
rouault: some compilers require an explict .as_nullable() to cast from a non-null shared pointed to a regular one. But here the best option would probably be to just assign the real authority names to a std::string and then do like the tokens.size() == 2 case avoid the non-null test on crs1 / crs2.
Actually you could skip the ci_equal() test and just use what the user provides. The case-insensitive comparison in the tokens.size() == 2 case was done because of legacy practice of using "epsg:XXXX", but here for a new supported syntax we can be stricter. That will make the code simpler

jjimenezshaw: Thanks. Without `ci_equal` the code is much simpler.

jjimenezshaw: FYI, `projinfo esri:103668+epsg:5703` does work perfectly. However `projinfo ogc:CRS84+EPSG:5773` fails. (something that somehow I noticed with `--list-crs`)

`AuthorityFactory::create` does a `ci_equal` for `EPSG`, `ESRI` and `PROJ`.

rouault:
could you document the new syntax around https://github.com/OSGeo/PROJ/blob/master/src/iso19111/io.cpp#L6805 ?

tonianr:
Hi, I'm very new to GDAL and would like to understand how to convert my tif file from NAVD88 to the WGS84 ellipsoid height?

Using 		_**gdalwarp -overwrite Hernando_2007_FDEM_WGS84_NAVD88.tif Hernando_2007_FDEM_WGS84_WGS84.tif  -s_srs EPSG:3857+5703 -t_srs EPSG:4979**_   did not work for me; any ideas on the necessary steps and syntax?

rouault:
>  any ideas on the necessary steps and syntax?

see point B of https://lists.osgeo.org/pipermail/gdal-dev/2021-August/054531.html

PR #2674, opened 2021-04-16
ENH: Added proj_get_celestial_body_list_from_database

### LINKED ISSUE #2667: ENH: add `proj_get_celestial_bodies_from_database`
Add `proj_get_celestial_bodies_from_database` similar to `proj_get_authorities_from_database`

continuation of #2662

snowman2:
- [x] Closes #2667
 - [x] Tests added

Code or name? Which one would be better to expose?


rouault:
I think it should be the name. I don't see what you will be able to do with the code. At least with the current reality of how the concept is handled. You can't query a corresponding object from the code. One limitation of the celestial body concept is that there is no ISO 19111 equivalent of it, and no corresponding PROJ C++ class. Hopefully there is no real need for a complex modelling... It is purely a PROJ quick addition to be able to handle non-Earth CRS and try not to mix everything together. However as we haven't yet made the exercise of incorporating the IAU catalog (still hoping this will happen in coming months), it is a bit hard to assess the real impact (we just have a few ESRI non-Earth ellipsoids registered currently, themselves likely coming from some version of IAU)

src/iso19111/factory.cpp
```diff
@@ -7293,6 +7293,24 @@ AuthorityFactory::getAuthorityCodes(const ObjectType &type,
 
 // ---------------------------------------------------------------------------
 
+/** \brief Returns the set of celestial body codes of the given object type.
+ *
+ * @return the set of celestial body codes
+ * @throw FactoryException
+ */
+std::set<std::string> AuthorityFactory::getCelestialBodyCodes() const {
+    std::string sql = "SELECT code FROM celestial_body "
```
rouault: d->authority() might be empty when we want to query for all authorities. See AuthorityFactory::getUnitList() as how to handle that situation.

src/iso19111/c_api.cpp
```diff
@@ -2674,6 +2674,40 @@ PROJ_STRING_LIST proj_get_codes_from_database(PJ_CONTEXT *ctx,
...
+/** \brief Return the list of celestial body codes used in the database.
+ *
+ * The returned list is NULL terminated and must be freed with
+ * proj_string_list_destroy().
+ *
+ * @param ctx PROJ context, or NULL for default context
+ * @param auth_name Authority name (must not be NULL)
+ *
+ * @return a NULL terminated list of NULL-terminated strings that must be
+ * freed with proj_string_list_destroy(), or NULL in case of error.
```
rouault:
```suggestion
 * freed with proj_string_list_destroy(), or NULL in case of error.
 * @since 8.1
```

src/iso19111/c_api.cpp
```diff
@@ -2674,6 +2674,40 @@ PROJ_STRING_LIST proj_get_codes_from_database(PJ_CONTEXT *ctx,
 
 // ---------------------------------------------------------------------------
 
+/** \brief Return the list of celestial body codes used in the database.
+ *
+ * The returned list is NULL terminated and must be freed with
+ * proj_string_list_destroy().
+ *
+ * @param ctx PROJ context, or NULL for default context
+ * @param auth_name Authority name (must not be NULL)
```
rouault: you could potentially allow NULL to be passed in to allow querying in all authorities. See proj_create_from_name()

test/unit/test_c_api.cpp
```diff
@@ -1398,6 +1398,17 @@ TEST_F(CApi, proj_get_codes_from_database) {
...
 // ---------------------------------------------------------------------------
 
+TEST_F(CApi, proj_get_celestial_body_codes_from_database) {
+    auto list = proj_get_celestial_body_codes_from_database(m_ctxt, "PROJ");
+    ListFreer feer(list);
+    ASSERT_NE(list, nullptr);
+    ASSERT_TRUE(list[0] != nullptr);
+    EXPECT_EQ(list[0], std::string("EARTH"));
+    EXPECT_EQ(list[1], nullptr);
+}
```
rouault: adding a test for the auth_name=NULL situation would be useful

src/iso19111/factory.cpp
snowman2: Do you think that it would be useful for an object with authority, code, and name properties be returned then?

rouault: It could be done, but this is more work. It would make sense if we anticipated that our description of a celestial body goes more than just a simple name. Well, currently the database table also stores a rough semi_major_axis value so that we can use it to figure out to which body an ellipsoid belongs too,as this information isn't stored anywhere in ISO-191111, WKT, etc.).  Also I'm a bit hesitant to expose more the concept (I'm not an astro geophysisit) than what is strictly needed and while more people have thought about the concept and how it relates / is different from an ellipsoid description.

The use cases I've in mind are:
- be able to filter CRS by bodies (nominal case: I want only Earth related CRS) ==> already covered
- given a CRS object, be able to determine to which body it belongs too ==> already covered
- list the allowed bodies: purpose of that PR

snowman2: The reason I ask is if we allow filtering on NULL, then it would be nice to know the authority it belongs to. Maybe authority, name, and semi_major_axis (exclude code)?

snowman2:
Last commit was a bit rough, but putting out the idea. Will cleanup later if you like the concept.

src/iso19111/factory.cpp
rouault: > The reason I ask is if we allow filtering on NULL, then it would be nice to know the authority it belongs to. Maybe authority, name, and semi_major_axis (exclude code)?

I'm still hesitant about exposing too much, and would like to see some real use case we need to address. It is unlikely that any authority will actually deliver any code for such object, since it is absent from known models. This is more a PROJ artifact resulting from the need for a row in database to be identified by a code, and to have the ellipsoid table relates to it through a foreign key. I wouldn't really very enthusiastic to see the semi_major_axis to be exposed. The one used for Earth is some rounded value of an """"average""""" value, so that all known Earth ellipsoid match it (+/- 0.5 % I believe). Not something strongly scientificly founded.
As a compromise, if we believe that we might need more than just the name in the future, the function might return a list of
```
struct PROJ_CELESTIAL_BODY
{
     char* pszName;
     /* other fields might be added in the future */
}
```
(the working would then be like proj_get_units_from_database() )

snowman2: > I'm still hesitant about exposing too much, and would like to see some real use case we need to address.

I understand the part about exposing too much. I think authortiy and name should be sufficient in the return. It will provide clarity to the user when filtering that they got the right response and will also enable testing the filtering in the responses.

My main thought for this was to use this information to filter for CRS with the query crs info from database. This is something I think is very important if you only want CRS on `Earth` for example.

rouault: > I think authortiy and name should be sufficient in the return

ok.

snowman2:
Thanks for the review @rouault 👍

PR #2681, opened 2021-04-21
Get geoid models for a CRS code

jjimenezshaw:
- [x] Tests added

Given a CRS code, this new method returns the list of possible geoid models that apply to it.
(use case: give the user a list of available geoid models, to select the desired one).

The sql query has two parts. The second covers CRSs like EPSG:6360, that is the ftUS version of EPSG:5703 (NAVD88 height).

In a later PR, I was thinking on adding at least these geoid models to the db:
 - `AUSGeoid2020`,`AUSGeoid09`, `AUSGeoid98` for `AHD height` (Australia)
 - `RAF09`, `RAF18` for `NGF-IGN69 height` (France)
(those are the cases I found with more than one grid file with the same target VCRS)

CC @snowman2

### DIFF
--- a/test/unit/test_factory.cpp
+++ b/test/unit/test_factory.cpp
@@ -2146,7 +2146,7 @@ TEST(factory, AuthorityFactory_getAvailableGeoidmodels) {
 
     const std::string OSGM15{"OSGM15"};
 
-    auto chechNavd88 = [&navd88geoidmodels](const std::list<std::string> &res) {
+    auto checkNavd88 = [&navd88geoidmodels](const std::list<std::string> &res) {
         ASSERT_EQ(res.size(), navd88geoidmodels.size());
         auto it = navd88geoidmodels.cbegin();
         for (const auto &model : res) {
@@ -2158,34 +2158,34 @@ TEST(factory, AuthorityFactory_getAvailableGeoidmodels) {
         AuthorityFactory::create(DatabaseContext::create(), std::string());
 
     {
-        auto res = factory->getAvailableGeoidmodels("4326");
+        auto res = factory->getGeoidModels("4326");
         ASSERT_TRUE(res.empty());
     }
 
     {
-        auto res = factory->getAvailableGeoidmodels("5703");
-        chechNavd88(res);
+        auto res = factory->getGeoidModels("5703");
+        checkNavd88(res);
     }
     {
-        auto res = factory->getAvailableGeoidmodels("6360");
-        chechNavd88(res);
+        auto res = factory->getGeoidModels("6360");
+        checkNavd88(res);
     }
     {
-        auto res = factory->getAvailableGeoidmodels("8228");
-        chechNavd88(res);
+        auto res = factory->getGeoidModels("8228");
+        checkNavd88(res);
     }
     {
-        auto res = factory->getAvailableGeoidmodels("6357");
-        chechNavd88(res);
+        auto res = factory->getGeoidModels("6357");
+        checkNavd88(res);
     }
 
     {
-        auto res = factory->getAvailableGeoidmodels("5701");
+        auto res = factory->getGeoidModels("5701");
         ASSERT_EQ(res.size(), 1U);
         EXPECT_EQ(OSGM15, res.front());
     }
     {
-        auto res = factory->getAvailableGeoidmodels("5732");
+        auto res = factory->getGeoidModels("5732");
         ASSERT_EQ(res.size(), 1U);
         EXPECT_EQ(OSGM15, res.front());
     }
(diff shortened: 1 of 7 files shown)


test/unit/test_factory.cpp
```diff
@@ -2103,6 +2103,58 @@ TEST(
...
+    auto checkNavd88 = [&navd88geoidmodels](const std::list<std::string> &res) {
+        ASSERT_EQ(res.size(), navd88geoidmodels.size());
+        auto it = navd88geoidmodels.cbegin();
+        for (const auto &model : res) {
+            EXPECT_EQ(*it, model);
+            it++;
+        }
+    };
+    auto factory =
+        AuthorityFactory::create(DatabaseContext::create(), std::string());
```
rouault: In case we get objects under another authority with same codes...
```suggestion
        AuthorityFactory::create(DatabaseContext::create(), "EPSG");
```

test/unit/test_factory.cpp
```diff
@@ -2103,6 +2103,58 @@ TEST(
...
 
+TEST(factory, AuthorityFactory_getAvailableGeoidmodels) {
+    const std::list<std::string> navd88geoidmodels{
+        "GEOID03",  "GEOID06", "GEOID09", "GEOID12A",
+        "GEOID12B", "GEOID18", "GEOID99"};
+
+    const std::string OSGM15{"OSGM15"};
+
+    auto checkNavd88 = [&navd88geoidmodels](const std::list<std::string> &res) {
+        ASSERT_EQ(res.size(), navd88geoidmodels.size());
```
rouault: this test might be a bit fragile and will break next time we'll add a new US geoid. Can it just test that GEOID12B and GEOID18 are found for example ?

rouault:
this looks OK to me. I guess we need a C API for it to be usable by pyproj

jjimenezshaw:
> this looks OK to me. I guess we need a C API for it to be usable by pyproj

Do you mean in this PR, or in a later one? I can do any.

kbevers:
> Do you mean in this PR, or in a later one? I can do any.

Put it in this one, please

rouault:
https://github.com/OSGeo/PROJ/pull/2683 brings a new conversion method EPSG:1104 "Change Of Vertical Unit" that is a variant of EPSG:1069 bit without the conversion factor specified (it must be infered from the source and target CRS). Not sure how it impacts your PR but wanted to raise that as I see you explicitly select on 1069

rouault:
Can you rebase that on top of latest master that has now the EPSG v10.019 update ?

jjimenezshaw:
> Can you rebase that on top of latest master that has now the EPSG v10.019 update ?

I am working on that right now, also adding  EPSG:1104 "Change Of Vertical Unit"
Thanks for the heads-up yesterday.

jjimenezshaw:
C API added.
I do not know if the function names follow any rule. I tried to make it similar to others. Do not hesitate to suggest any other name.

jjimenezshaw:
About `clang-format`, I have it embedded in my IDE (vscode)... and `proj.h` got totally modified by it.
Does it make sense to add `// clang-format off` and `on` to `proj.h` to prevent that kind of problem? (In case we do not want `proj.h` to be automatically formatted by clang-format)

rouault:
> About `clang-format`, I have it embedded in my IDE (vscode)... and `proj.h` got totally modified by it.

We have a scripts/reformat_cpp.sh script that applies it only a restricted set of files. Will not probably help you on Windows (or maybe if you run it from WSL with Ubuntu 20.04)

jjimenezshaw:
> We have a scripts/reformat_cpp.sh script

I am developing in native Ubuntu 20.04. My vscode automatically applies clang-format when I save any file, what is very convenient (in general). If I have to edit `proj.h`, then the whole file is automatically formatted. I had to use `vim` (that is not a problem). I was asking for the clang-format guard to avoid other people to have the same issue, and maybe commit the file with all those modifications.

Anyway, it is not a big deal.

rouault:
> I was asking for the clang-format guard to avoid other people to have the same issue, and maybe commit the file with all those modifications.

There are tons of other files in the repo which aren't clang-format'ed. Maybe at some point we should apply a huge reformatting to all .h*/.c*. Or maybe there's a file that vscode could read which would indicate it which files to reformat and which not ?

rouault (approved):

javier-jimenez-shaw-pix4d:
`Maybe at some point we should apply a huge reformatting to all .h*/.c*.` I would say so.
At some point with low development activity, I would say to apply it everywhere. If there are specific files that (for any reason) we do not want to format, just add the guard comment.
Maybe also a github "action" that checks that the format is right too.
The main inconvenience could be the backport of bugs. To minimize the impact, we should do it a little bit before a release.

jjimenezshaw:
> Maybe at some point we should apply a huge reformatting to all .h*/.c*. 

I would say so.
At some point with low development activity, I would say to apply it everywhere. If there are specific files that (for any reason) we do not want to format, just add the guard comment.
Maybe also a github "action" that checks that the format is right too.
The main inconvenience could be the backport of bugs. To minimize the impact, we should do it a little bit before a release.

PR #2686, opened 2021-04-23
Update internal nlohmann/json to 3.9.1, and add a CMake option to be able to use external nlohmann/json

rouault:
Using external nlohmann/json can potentially help reducing binary size for people doing static builds with other components that already use it. And more in line with the general philosophy of packaging.


mwtoews:
Review in progress, and may take a few days ... my spare time is limited.

I'll probably add a commit here when ready, as I'm seeing potential conflicts from Ubuntu bionic with [nlohmann-json-dev](https://packages.ubuntu.com/bionic/nlohmann-json-dev), which is too old (2.1.1) and doesn't contain NLOHMANN_JSON_VERSION_* definitions.

mwtoews:
See some suggested edits at https://github.com/rouault/PROJ/pull/7

PR #2690, opened 2021-04-23
CMake: "make dist" workalike via CPack

mwtoews:
This PR adds "make dist" to the CMake set-up, to allow similar functionality that is provided by Autotools.

The process (at the moment) is:
```
git clean -dfx  # danger! this will delete any uncommitted files in a repo
./autogen.sh
./configure
make dist  # optional step, to generate autotools archive to compare
make distclean
mkdir build && cd build
cmake ..
make dist
```
this builds two sets of source distributions with *.tar.gz and *.zip formats. Comparing the two extracted archives:
```
$ diff -qr autotools/proj-8.1.0/ cmake/proj-8.1.0/
Only in cmake/proj-8.1.0/data: epsg-deprecated
Only in cmake/proj-8.1.0/data: proj_outIGNF.dist-real
Only in cmake/proj-8.1.0/data/sql: method_triggers.sql
Only in cmake/proj-8.1.0/m4: libtool.m4
Only in cmake/proj-8.1.0/m4: lt~obsolete.m4
Only in cmake/proj-8.1.0/m4: ltoptions.m4
Only in cmake/proj-8.1.0/m4: ltsugar.m4
Only in cmake/proj-8.1.0/m4: ltversion.m4
Only in autotools/proj-8.1.0/: README
Only in cmake/proj-8.1.0/src: general_doc.dox
Only in cmake/proj-8.1.0/src: runmultistresstest.sh
Only in cmake/proj-8.1.0/test: fuzzers
Only in cmake/proj-8.1.0/test/gie: nkg.gie
Only in cmake/proj-8.1.0/test/gigs: 5101.4-jhs.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5102.2.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5105.1.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5108.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5110.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5111.2.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5203.1.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5204.1.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5205.1.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5206.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5207.1.gie.failing
Only in cmake/proj-8.1.0/test/gigs: 5207.2.gie.failing
Only in cmake/proj-8.1.0/test: postinstall
```
this shows that the CMake version grabs a few extra files that are not included in the respective Makefile.am files. For some reason, CPack refuses to take README, and I'm not sure why.

---

There are a few potentially dangerous things with the approach:

1. While not part of this PR, the recommendation to use `git clean -dfx` will delete uncommitted work-in-progress, but is the only sure way to clear up extra files in a repo
2.  If there are any extra files that don't fit the `CPACK_SOURCE_IGNORE_FILES` pattern, these will be swept up in the archive
3. Many commands in the correct order (shown above) are required to generate a source archive with configure, etc., so this is easy to mess up
4. It's a bit awkward to `mkdir build && cd build && cmake .. && make dist` i.e. in a subfolder

An alternative idea is to leverage [git archive](https://git-scm.com/docs/git-archive) to build a source archive based on committed files, which could be brought into a Makefile. This approach could potentially be more robust, but it can't work with the generated files from ./autogen.sh

---

This PR is based on parts of [GEOS' CMakeLists.txt](https://github.com/libgeos/geos/blob/200ff644ea5682a4f6314efc6de1a2dc057d5acf/CMakeLists.txt#L382-L417); thanks GEOS devs for great ideas!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,3 +293,60 @@ if(UNIX OR MINGW)
     ${CMAKE_CURRENT_BINARY_DIR}/proj.pc
     DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
 endif()
+
+################################################################################
+# "make dist" workalike
+################################################################################
+
+set(CPACK_SOURCE_GENERATOR "TGZ;ZIP")
+set(CPACK_SOURCE_PACKAGE_FILE_NAME "proj-${PROJ_VERSION}")
+set(CPACK_PACKAGE_VENDOR "OSGeo")
+set(CPACK_PACKAGE_VERSION_MAJOR ${PROJ_VERSION_MAJOR})
+set(CPACK_PACKAGE_VERSION_MINOR ${PROJ_VERSION_MINOR})
+set(CPACK_PACKAGE_VERSION_PATCH ${PROJ_VERSION_PATCH})
+set(CPACK_VERBATIM_VARIABLES TRUE)
+set(CPACK_SOURCE_IGNORE_FILES
+  /\\..*  # any file/directory starting with .
+  /.*\\.yml
+  /.*\\.gz
+  /.*\\.zip
+  /.*build.*/
+  \\.deps
+  /autogen\\.sh
+  /autom4te\\.cache
+  /CODE_OF_CONDUCT.md
+  /CONTRIBUTING.md
+  /Dockerfile
+  /docs/
+  /Doxyfile
+  /examples/
+  /HOWTO-RELEASE
+  /m4/lt*
+  /m4/libtool*
+  /media/
+  /schemas/
+  /scripts/
+  /test/fuzzers/
+  /test/gigs/.*gie\\.failing
+  /test/postinstall/
+  /travis/
+  ${PROJECT_BINARY_DIR}
+)
+
+include(CPack)
+
+# Simplify README.md to README
+add_custom_target(README
+  COMMAND ${CMAKE_COMMAND}
+    -D PROJ_SOURCE_DIR=${PROJ_SOURCE_DIR}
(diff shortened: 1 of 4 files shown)


kbevers:
Good job, @mwtoews! I wonder if `./autogen.sh` is even necessary in a post-autoconf scenario?

mwtoews:
autogen.sh is only required for autotools. A source distribution based on cmake shouldn't have any generated files.

mwtoews:
The last update copies `README` generated by the automake (mystery solved), and adds some workflow to travis/install.sh used in many CI targets to run cmake's `make dist`, and compares the two versions of source archives. Besides #2691, I'm unsure if any other files should be ignored by CPack or added to the respective Makefile.am.

rouault:
@mwtoews Is this PR ready to merge ?

mwtoews:
my remaining question was if there were any additional files to ignore.

rouault:
The .m4 files should likely be ignored, as well as the fuzzers/ directory

mwtoews:
Actual last update here adds a custom README target, similar to Makefile.am, but much more complicated than running `fgrep -v "[!["`, as it should be cross-platform running only CMake commands to do the equivalent. This script could be expanded later to remove other Md markup to create an even simpler README file for distribution.

Also, a few other files and patterns are added to the ignore list. The two source distributions are now nearly identical.

PR #2698, opened 2021-04-27
TST/DOC: upgrade CSA to use ubuntu-18.04; sync docs

mwtoews:
According to the [Ubuntu release schedule](https://wiki.ubuntu.com/Releases), Ubuntu Xenial 16.04 is to reach it's end of standard support in a few days. Upgrading to [`ubuntu-18.04`](https://github.com/actions/virtual-environments) (Bionic).

Also synchronizing contribution docs to use same `clang+llvm-9.0.0-x86_64-linux-gnu-ubuntu-18.04` as used in CI.

### DIFF
--- a/.github/workflows/clang_static_analyzer.yml
+++ b/.github/workflows/clang_static_analyzer.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
 
   clang_static_analyzer:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]')"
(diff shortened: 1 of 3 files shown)


mwtoews:
Another option is to upgrade CSA to use [LLVM 12.0.0](https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.0) released a few days ago, which could be paired up with `ubuntu-20.04`

rouault:
Looks good to me. If you need to update to a later version, you'll probably need to do a similar change as https://github.com/OSGeo/GDAL/commit/e28a878507c801c29d93c643ab6393acf5042e6e in the jq filtering stage

mwtoews:
Let's only upgrade ubuntu-18.04 for now.

Issue #2704, opened 2021-05-01
Foucault vs. Foucaut: typo?


mwtoews:
Compare the two:

1. `Foucault`
2. `Foucaut`

Only 2 is used in PROJ (e.g. [`fouc`](https://proj.org/operations/projections/fouc.html)), and I'm 99.5% certain it is a typo that should be spelled as 1. I suppose this is named after [Léon Foucault](https://en.wikipedia.org/wiki/L%C3%A9on_Foucault)?

Similar to #2701, this is a decades old typo embedded in `PROJ_HEAD`, etc.

(Yes, I'm working on a general spelling PR, but am individually creating issues for prominent typos that may require cross-referencing, and yes I intend to fix these too.)

rouault:
Yeah, a treasure hunt ! I think that Foucaut, without l, is the correct spelling in that instance. There are indeed quite a few known French men with Foucault name, with l, including the one of the pendulum experiment you quoted (Léon Foucault) or more recently the philosopher Michel Foucault (https://en.wikipedia.org/wiki/Michel_Foucault). Although if you ask to the random French guy about Foucault, they will think to https://en.wikipedia.org/wiki/Jean-Pierre_Foucault :-)

But here, it seems this is "Henri Casimir Auguste De Prépetit Foucaut"

First hint I found was on https://bl.ocks.org/Fil/4db96f1250ad6118290e24d62e67187d which has a references to "Capitaine De Prépetit Foucaut, Notice sur la construction de nouvelles mappemondes et de nouveaux atlas, pp. 5-10, Arras, France, 1862."
But googling for that name doesn't lead to interesting results. But the second reference to "Coup d'oeil historique sur la projection des cartes de géographie, M. d'Avezac - 1863." leads to its digital version at Bibliothèque Nationale de France:  http://gallica.bnf.fr/ark:/12148/bpt6k6574475p and at page 11 of that document they mention:
"Le capitaine du génie Henri-Casimir-Auguste DE PRÉPETIT FOUCAUT, dans une Notice sur la construction de nouvelles mappemondes et de nouveaux atlas de géographie, lithographiée à Arras en 1862, s'est occupé spécialement de la projection de la mappemonde entière dans ce dernier système"

mwtoews:
Perfect, thanks! This all makes sense. Added to (to be committed) `spelling_wordlist.txt`.

Issue #2701, opened 2021-04-30
Stereographics vs. Stererographics: typo?


mwtoews:
Compare the two words:

1. `Stereographics`
2. `Stererographics`

Only 2 is currently used throughout PROJ for these:

1. [Modified Stererographics of Alaska](https://proj.org/operations/projections/alsk.html)
2. [Mod. Stererographics of 48 U.S.](https://proj.org/operations/projections/gs48.html)
2. [Mod. Stererographics of 50 U.S.](https://proj.org/operations/projections/gs50.html)

The singular version of 1 (stereographic) is used extensively i.e. [stereographic projection](https://en.wikipedia.org/wiki/Stereographic_projection).

Is 2 a typo? Should these be "Stereographics"? Or the singular "Stereographic"? I'm 99% certain this is a typo, but it's been there for a few decades, embedded in `PROJ_HEAD` and all. Page 18 of [proj_4_3.pdf](https://github.com/OSGeo/PROJ/blob/master/docs/old/proj_4_3.pdf) uses the singular Stereographic (1), suggesting that 2 is a typo.

rouault:
Most certainly a typo. Snyder, "Map Projections - a working manual" (https://pubs.usgs.gov/pp/1395/report.pdf) mentions at page 9:
```
44. Modified-Stereographic Conformal projection of Alaska -----------------------------------------------------------------------------------205 
45. Modified-Stereographic Conformal projection of 48 United States, bounded by a  near-rectangle of constant scale ---------------206 
```

Issue #2670, opened 2021-04-13
Kavrayskiy vs. Kavraysky vs. Kavraisky: typo or alternative transliteration?


mwtoews:
Wikipedia has an article [Vladimir V. Kavrayskiy](https://en.wikipedia.org/wiki/Vladimir_V._Kavrayskiy) for the  Soviet geodesist and cartographer named in several projections. Meanwhile, encyclopedia.com has an article [Kavraysky, Vladimir Vladimirovich](https://www.encyclopedia.com/science/dictionaries-thesauruses-pictures-and-press-releases/kavraysky-vladimir-vladimirovich), which is spelled slightly different. PROJ consistently uses a third spelling variation: Kavraisky
```
$ git grep -i Kavrayskiy
$ git grep -i Kavraysky
$ git grep -i Kavraisky
docs/old/libproj.pdf:(Eckert III, Putnins P1, Putnins P'1, Wagner VI and Kavraisky VII.)
docs/source/operations/projections/kav5.rst:Kavraisky V
docs/source/operations/projections/kav5.rst:   :alt:   Kavraisky V
docs/source/operations/projections/kav5.rst:.. note:: All parameters are optional for the Kavraisky V projection.
docs/source/operations/projections/kav7.rst:Kavraisky VII
docs/source/operations/projections/kav7.rst:   :alt:   Kavraisky VII
docs/source/operations/projections/kav7.rst:.. note:: All parameters are optional for the Kavraisky VII projection.
docs/source/operations/projections/wag1.rst:Wagner I (Kavraisky VI)
docs/source/operations/projections/wag1.rst:   :alt:   Wagner I (Kavraisky VI)
src/pj_list.h:PROJ_HEAD(kav5, "Kavraisky V")
src/pj_list.h:PROJ_HEAD(kav7, "Kavraisky VII")
src/pj_list.h:PROJ_HEAD(wag1, "Wagner I (Kavraisky VI)")
src/projections/eck3.cpp:PROJ_HEAD(kav7, "Kavraisky VII") "\n\tPCyl, Sph";
src/projections/sts.cpp:PROJ_HEAD(kav5,    "Kavraisky V")         "\n\tPCyl, Sph";
src/projections/urmfps.cpp:PROJ_HEAD(wag1, "Wagner I (Kavraisky VI)") "\n\tPCyl, Sph";
test/gie/builtins.gie:# Kavraisky V
test/gie/builtins.gie:# Kavraisky VII
[...]

PR #2705, opened 2021-05-01
DOC: configure and add spelling wordlist; fix typos, apply Sphinx syntax

mwtoews:
Changes mostly affect documentation, although a few are within the source code too.

Feel free to comment on any changes here.

This does not include potential changes for #2670 and #2701. Issue #2704 is resolved and added to the word list as correct.


rouault:
is the  docs/source/spelling_wordlist.txt  file used by the "make spelling" target in docs/ ?  And that doesn't fix all issues reported by it, right  (looking at https://github.com/OSGeo/PROJ/pull/2705/checks?check_run_id=2482452219 ) ?

mwtoews:
The `make spelling` target is passed through a catch-all target in [docs/Makefile](https://github.com/OSGeo/PROJ/blob/0f9e6bf1a21a6548501870c7c090811b199a57ff/docs/Makefile#L41-L43), so the configuration reads docs/source/spelling_wordlist.txt each time it is called. The missing target of course is .doxygen_up_to_date, so any edits to Doxygen source code files needs to be manually triggered before, i.e. `make html`.

As for the remaining 1690 misspelled words in this PR (so far), it's an improvement over the current 3626 misspelled words in the main branch. I don't expect this to get close to zero. Also the results vary considerably, depending on the system and locale, as I only have 765 misspelled words with this PR. I might add a few more to spelling_wordlist.txt based on what I see from the CI output.

src/iso19111/operation/transformation.cpp
```diff
@@ -366,7 +366,7 @@ TransformationNNPtr Transformation::create(
 
 // ---------------------------------------------------------------------------
 
-/** \brief Instantiate a transformation ands its OperationMethod.
+/** \brief Instantiate a transformation and it's OperationMethod.
```
rouault: that one should be reverted

rouault: oops no, I see that the actual fix was ands->and.  I meant that the it's should be its

rouault:
> The make spelling target is passed through a catch-all target in docs/Makefile, so the configuration reads docs/source/spelling_wordlist.txt each time it is called. 

Thanks for clearing that up, and the many fixes!

src/iso19111/operation/transformation.cpp
rouault:
```suggestion
/** \brief Instantiate a transformation and its OperationMethod.
```

rouault (approved):

src/iso19111/operation/transformation.cpp
mwtoews: thanks, amended edit

PR #2717, opened 2021-05-10
Update ESRI projection db to 12.8

nyalldawson:
See https://github.com/Esri/projection-engine-db-doc/pull/29


rouault:
The version and date at top of scripts/build_db_from_esri.py should also be updated so that they are updated in data/sql/esri.sql. Wondering if we should require them as command line argument of the script itself.

nyalldawson:
@rouault 
> The version and date at top of scripts/build_db_from_esri.py should also be updated so that they are updated in data/sql/esri.sql. Wondering if we should require them as command line argument of the script itself.

Done

PR #2724, opened 2021-05-16
GTiff grids: add support for half-float (16bit) 

rouault:
... in case we'd need it someday


kbevers:
> in case we'd need it someday

What's the likelyhood of that? I am kinda -1 on adding code with no immediate use case.

rouault:
ok let's drop this

PR #2730, opened 2021-05-24
Database: decrease DB size by using WITHOUT ROWID tables

rouault:
None of our tables are indexed by a INTEGER PRIMARY KEY, but most of
them are by a (auth_name, code) primary key. Consequently they can
benefit from being created as WITHOUT ROWID tables
(https://sqlite.org/withoutrowid.html), which avoids an index to be
created on the rowid we don't use. WITHOUT ROWID is a feature added in
SQLite 3.8.2, so as our baseline is 3.11, we can use it.

This decreases the DB size from 7,749,632 to 7,229,440 bytes, without
any measurable consequence on performance.

### DIFF
--- a/data/sql/proj_db_table_defs.sql
+++ b/data/sql/proj_db_table_defs.sql
@@ -14,7 +14,7 @@
 CREATE TABLE metadata(
     key TEXT NOT NULL PRIMARY KEY CHECK (length(key) >= 1),
     value TEXT NOT NULL
-);
+) WITHOUT ROWID;
 
 CREATE TABLE unit_of_measure(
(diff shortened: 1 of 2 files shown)


nyalldawson:
Nice... I can't find any downsides to this, looks like a safe change!

mwtoews:
Agree that WITHOUT ROWID is the correct option for the tables, as they generally have composite primary keys.

Reminds me of PostgreSQL's [WITH OIDS](https://www.postgresql.org/docs/11/sql-createtableas.html), which was [phased out](https://stackoverflow.com/q/5625585/).

PR #2738, opened 2021-06-03
Share SQLite database handle among all contexts

### LINKED ISSUE #2728: Sharing connections to proj.db
Each PROJ_CONTEXT owns a connection to proj.db. If using many threads (and thus many contexts), we can exhaust the number of file descriptors of a process. It would be good to share sqlite3* handles that point to the same proj.db (which is the nominal case)

rouault:
Fixes #2728

- The sqlite3* handle to a same file (which is the nominal case) is shared among all contexts of all threads using the SQLITE3_FULLMUTEX mode which enables a thread-safe use of such handle. Note that this is restricted to opening a single database for now, and not using auxiliary databases which will use a per-context handle (could probably be extended to that case later)
- Sharing of cached objects is not implemented yet.
- proj_context_set_autoclose_database() is now deprecated and a no-op. Implementing to work now would be a complication and would defeat the purpose of the whole work. proj_context_set_autoclose_database() was mostly a workaround for the non-shared sqlite3* handle

Seems to work well for me, but testing from others would be appreciated


snowman2:
Just tested it out with the pyproj test suite and the speed improvement was fantastic. Tests went from ~3 minutes with previous versions of PROJ to ~30 seconds with this MR. All tests pass (Note: I haven't done any serious threading tests).

Minimal speed test:
```bash
python -m timeit -s "from pyproj import CRS" "CRS('EPSG:4326')"
```
Results:

PROJ 8.0.1
```
20 loops, best of 5: 15.1 msec per loop
```
This MR
```
500 loops, best of 5: 759 usec per loop
```

PR #2749, opened 2021-06-17
S2 projection

marcus-elia:
I added the [S2](https://s2geometry.io/) family of projections to PROJ.  The structure of s2.cpp is based on the qsc.cpp file, and the math functions are from S2 ([s2coords.h](https://github.com/google/s2geometry/blob/0c4c460bdfe696da303641771f9def900b3e440f/src/s2/s2coords.h) and [vector.h](https://github.com/google/s2geometry/blob/0c4c460bdfe696da303641771f9def900b3e440f/src/s2/util/math/vector.h)).

The face can be specified similarly to the [QSC projection](https://proj.org/operations/projections/qsc.html), using +lat_0 and +lon_0.

 For S2, there are three choices for how to convert from "Cube-space" (u,v) to "Cell-space" (s,t) coordinates (see the [S2 Cells](https://s2geometry.io/devguide/s2cell_hierarchy.html) page). You can specify +UTtoST to be an element of {linear, quadratic, tangent, none} to control which function is used to convert from u,v to s,t coordinates (see the comments in the [S2 code](https://github.com/google/s2geometry/blob/0c4c460bdfe696da303641771f9def900b3e440f/src/s2/s2coords.h) for a description of them). It defaults to quadratic if +UVtoST is not given, the default mode in the S2 code.


include/proj/coordinateoperation.hpp
```diff
@@ -1313,6 +1313,12 @@ class PROJ_GCC_DLL Conversion : public SingleOperation {
         const util::PropertyMap &properties, const common::Angle &centerLat,
         const common::Angle &centerLong, const common::Length &falseEasting,
         const common::Length &falseNorthing);
+    
+    PROJ_DLL static ConversionNNPtr createS2(
```
rouault: unless you really need in the C++ API, this is not strictly needed and could be removed

src/fwd.cpp
```diff
@@ -182,7 +182,7 @@ PJ_XY pj_fwd(PJ_LP lp, PJ *P) {
 
     last_errno = proj_errno_reset(P);
 
-    if (!P->skip_fwd_prepare)
+    if (!P->skip_fwd_prepare && 0 != strcmp(P->short_name, "s2"))
```
rouault: Why is this hack needed ? Ideally, we shouldn't need that.

src/init.cpp
```diff
@@ -644,6 +645,17 @@ pj_init_ctx_with_allow_init_epsg(PJ_CONTEXT *ctx, int argc, char **argv, int all
         return pj_default_destructor (PIN, PROJ_ERR_INVALID_OP_ILLEGAL_ARG_VALUE);
     }
 
+    /* S2 projection parameter */
```
rouault: parsing this parameter should be done in the setup function of the projection, not in this generic code

src/proj.h
```diff
@@ -1,8 +1,8 @@
-/******************************************************************************
- * Project:  PROJ.4
- * Purpose:  Revised, experimental API for PROJ.4, intended as the foundation
- *           for added geodetic functionality.
- *
+    /******************************************************************************
```
rouault: this change should be reverted

src/proj_internal.h
```diff
@@ -172,6 +172,7 @@ enum pj_io_units {
 };
 enum pj_io_units pj_left (PJ *P);
 enum pj_io_units pj_right (PJ *P);
+enum S2ProjectionType {Linear, Quadratic, Tangent, NoUVtoST};
```
rouault: changes in this file should be reverted. They should be moved to s2.cpp itself

src/projections/s2.cpp
```diff
@@ -0,0 +1,341 @@
+/*
```
rouault: Please include a header with license & copyright info (you can take src/transformations/defmodel.cpp as example, and adapt)

src/projections/s2.cpp
```diff
@@ -0,0 +1,341 @@
...
+    } else if (fabs(P->lam0) <= M_HALFPI + M_FORTPI) {
+        Q->face = (P->lam0 > 0.0 ? FACE_RIGHT : FACE_LEFT);
+    } else {
+        Q->face = FACE_BACK;
+    }
+    /* Fill in useful values for the ellipsoid <-> sphere shift
+     * described in [LK12]. */
+    if (P->es != 0.0) {
+        Q->a_squared = P->a * P->a;
+        Q->b = P->a * sqrt(1.0 - P->es);
```
rouault: useless. available in P->b

rouault (changes requested):
Thanks for your contribution.
In addition to addressing the remarks I made inline, we'd also need tests in test/gie/builtins.gie and a documentation page in docs/source/operations/projections

src/fwd.cpp
marcus-elia: I don't understand what the purpose of `fwd/inv` `prepare/finalize` is, so I was unsure how to handle it. The S2 projection is intended to return points expressed in s,t-coordinates, which are in [0,1] x [0,1].  I observed that fwd_finalize was making the output several orders of magnitude bigger, so I wanted to disable it.   Is there a different way I can set the `skip` parameters specifically for S2?

rouault: Ah, for a classic projection, PROJ multiplies/demultiplies the result by P->a. You could perhaps try setting P->left = PJ_IO_UNITS_PROJECTED; and P->right = PJ_IO_UNITS_PROJECTED; in the setup function in s2.cpp

src/init.cpp
elfprince13: @rouault - I've been working with @marcus-elia on this. When you say the setup function of the projection, are you referring to the block of code beginning here?
https://github.com/OSGeo/PROJ/blob/a1b640a8f019ebebf2112d26f1c1ada39ef49194/src/projections/s2.cpp#L310-L314
If so, I'm wondering if you have another example of a projection that parses parameters outside of `src/init.cpp`, as it doesn't appear that `ctx` gets passed through to that code in order to call `pj_param` properly.

rouault: > When you say the setup function of the projection, are you referring to the block of code beginning here?

yes

> If so, I'm wondering if you have another example of a projection that parses parameters outside of src/init.cpp, as it doesn't appear that ctx gets passed through to that code in order to call pj_param properly.

It is available in P->ctx. See e.g. https://github.com/OSGeo/PROJ/blob/master/src/projections/lcc.cpp#L89

elfprince13: Perfect, thanks!

src/fwd.cpp
marcus-elia: This has introduced an error of `can't initialize operations that take non-angular input coordinates`. Is there a different solution I could try?

rouault: oops sorry, it should have been:
```
P->left = PJ_IO_UNITS_RADIANS; 
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+        case Linear:
+            return 2 * s - 1;
+            break;
+        case Quadratic:
+            if (s >= 0.5) return (1/3.) * (4*s*s - 1);
+            else return (1/3.) * (1 - 4*(1-s)*(1-s));
+            break;
+        case Tangent:
+            s = std::tan(M_PI_2 * s - M_PI_4);
+            return s + (1.0 / (long{1} << 53)) * s;
```
rouault:
```suggestion
            return s + (1.0 / static_cast<double>(static_cast<std::int64_t>(1)<< 53)) * s;
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+        case Linear:
+            return 0.5 * (u + 1);
+            break;
+        case Quadratic:
+            if (u >= 0) return 0.5 * std::sqrt(1 + 3*u);
+            else        return 1 - 0.5 * std::sqrt(1 - 3*u);
+            break;
+        case Tangent:
+        {
+            volatile double a = std::atan(u);
```
rouault, line 155: why volatile ?

docs/source/operations/projections/s2.rst
```diff
@@ -0,0 +1,109 @@
+.. _s2:
```
rouault, line 1: A reference to that file should be added in docs/source/operations/projections/index.rst

docs/source/operations/projections/s2.rst
```diff
@@ -0,0 +1,109 @@
...
+
+********************************************************************************
+S2
+********************************************************************************
+
++---------------------+----------------------------------------------------------+
+| **Classification**  | Azimuthal                                                |
```
rouault: Is it realy azimuthal ? If not Miscellaneous would be more appropriate

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+namespace { // anonymous namespace
+struct pj_opaque {
+    enum Face face;
+    double a_squared;
+    double one_minus_f;
+    double one_minus_f_squared;
+    S2ProjectionType UVtoST;
+};
+} // anonymous namespace
+PROJ_HEAD(s2, "S2") "\n\tAzi, Sph";
```
rouault: If not azimuthal:
```suggestion
PROJ_HEAD(s2, "S2") "\n\tMisc, Sph&Ell";
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+        Q->face = FACE_FRONT;
+    } else if (fabs(P->lam0) <= M_HALFPI + M_FORTPI) {
+        Q->face = (P->lam0 > 0.0 ? FACE_RIGHT : FACE_LEFT);
+    } else {
+        Q->face = FACE_BACK;
+    }
+    /* Fill in useful values for the ellipsoid <-> sphere shift
+     * described in [LK12]. */
+    if (P->es != 0.0) {
+        Q->a_squared = P->a * P->a;
```
rouault, line 388: I don't see a_squared to be used at all currently, but there's a P->a * P->a occurrence at line 330 that could use it

test/gie/builtins.gie
```diff
@@ -5273,6 +5273,165 @@ expect  -223368.098302014 111769.110486991
...
+accept -20 -70.12337013762533
+expect 0.4377574241549981 0.6710100716628344
+
+direction inverse
+accept 0.5 0.5
+expect 0 -90
+accept 0.5622425758450019 0.6710100716628344
+expect 20 -70.12337013762533
+accept 0.4377574241549981 0.6710100716628344
+expect -20 -70.12337013762533
```
rouault, line 5434: an additional test with an invalid value of +UVtoST could be useful

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+    AREA_3 = 3
+};
+} // anonymous namespace
+
+// =================================================
+//
+//                  S2 Math Util
+//
+// =================================================
+PJ_XYZ Abs(const PJ_XYZ& p) {
```
rouault:
```suggestion
static PJ_XYZ Abs(const PJ_XYZ& p) {
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+    return {abs(p.x), abs(p.y), abs(p.z)};
+}
+// return the index of the largest component (fabs)
+int LargestAbsComponent(const PJ_XYZ& p) {
```
rouault:
```suggestion
static int LargestAbsComponent(const PJ_XYZ& p) {
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+             temp.x > temp.z ? 0 : 2 :
+             temp.y > temp.z ? 1 : 2;
+}
+
+// =================================================
+// 
+//              S2 Projection Functions
+//
+// =================================================
+double STtoUV(double s, S2ProjectionType s2_projection) {
```
rouault:
```suggestion
static double STtoUV(double s, S2ProjectionType s2_projection) {
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+        case Tangent:
+            s = std::tan(M_PI_2 * s - M_PI_4);
+            return s + (1.0 / (long{1} << 53)) * s;
+            break;
+        default:
+            return s;
+    }
+}
+
+double UVtoST(double u, S2ProjectionType s2_projection) {
```
rouault:
```suggestion
static double UVtoST(double u, S2ProjectionType s2_projection) {
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,386 @@
...
+    if (nullptr==Q)
+        return pj_default_destructor (P, PROJ_ERR_OTHER /*ENOMEM*/);
+    P->opaque = Q;
+
+    /* Determine which UVtoST function is to be used */
+    PROJVALUE maybeUVtoST = pj_param(P->ctx, P->params, "sUVtoST");
+    if (nullptr != maybeUVtoST.s) {
+        try {
+            Q->UVtoST = stringToS2ProjectionType.at(maybeUVtoST.s);
+        } catch (std::out_of_range) {
```
rouault:
```suggestion
        } catch (const std::out_of_range&) {
```

rouault:
@kbevers This projection method forces the unit sphere. This is a bit at odds with other projection methods, but might be idiomatic to this method. Should we accept it as such ? Another alternative would be to make it honor the semi-major axis, but people could use "+ellps=WGS84 +a=1" to force to the unit ellipsoid.

kbevers:
>  This is a bit at odds with other projection methods, but might be idiomatic to this method. Should we accept it as such ? 

I don't know enough about this projection and it's potential use-cases to make an informed decision. If I understand correctly the output coordinates are always limited to be between 0 and 1. I wonder if that is the actual end product stored in databases etc or if they are scaled after the fact.

elfprince13:
This projection is typically used with a fractal tiling system built on
Hilbert curves, and the projections here are intentionally unitless for
compatibility with that system. I’m not aware of anyone who rescales the
projected coordinates to be physically meaningful.--
~Thomas

docs/source/operations/projections/s2.rst
marcus-elia: In the cases when "none" or "linear" is chosen for the area-normalization, it is azimuthal, but when "quadratic" or "tangent" is chosen, it is not azimuthal.   Should it be classified as miscellaneous?

rouault: > Should it be classified as miscellaneous?

yes, you could add the precision about when it is azimuthal or not later in the file

src/projections/s2.cpp
marcus-elia: The purpose of volatile here is to force the truncation to double precision to be consistent.

docs/source/operations/projections/s2.rst
```diff
@@ -0,0 +1,119 @@
...
++---------------------+----------------------------------------------------------+
+| **Alias**           | s2                                                       |
++---------------------+----------------------------------------------------------+
+| **Domain**          | 2D                                                       |
++---------------------+----------------------------------------------------------+
+| **Input type**      | Geodetic coordinates                                     |
++---------------------+----------------------------------------------------------+
+| **Output type**     | Projected coordinates                                    |
++---------------------+----------------------------------------------------------+
+
```
rouault, line 22:
```suggestion

.. versionadded:: 8.2
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,393 @@
...
+ * https://github.com/google/s2geometry/blob/0c4c460bdfe696da303641771f9def900b3e440f/src/s2/s2coords.h
+ * The S2 projection functions are adapted from the above link and the S2
+ * Math util functions are adapted from
+ * https://github.com/google/s2geometry/blob/0c4c460bdfe696da303641771f9def900b3e440f/src/s2/util/math/vector.h
+ ****************************************************************************/
+
+#define PJ_LIB__
+
+#include <errno.h>
+#include <math.h>
```
rouault:
```suggestion
#include <cmath>
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,393 @@
...
+};
+} // anonymous namespace
+
+// =================================================
+//
+//                  S2 Math Util
+//
+// =================================================
+static PJ_XYZ Abs(const PJ_XYZ& p) {
+    return {abs(p.x), abs(p.y), abs(p.z)};
```
rouault:
```suggestion
    return {std::fabs(p.x), std::fabs(p.y), std::fabs(p.z)};
```

src/projections/s2.cpp
```diff
@@ -0,0 +1,393 @@
...
+ *
+ * You must also choose which conversion from UV to ST coordinates
+ * is used (linear, quadratic, tangent). Read about them in
+ * https://github.com/google/s2geometry/blob/0c4c460bdfe696da303641771f9def900b3e440f/src/s2/s2coords.h
+ * The S2 projection functions are adapted from the above link and the S2
+ * Math util functions are adapted from
+ * https://github.com/google/s2geometry/blob/0c4c460bdfe696da303641771f9def900b3e440f/src/s2/util/math/vector.h
+ ****************************************************************************/
+
+#define PJ_LIB__
```
rouault, line 50:
```suggestion
#define PJ_LIB__
#define _USE_MATH_DEFINES  // needed for M_1_PI availability with MSVC
```

rouault:
Some of the tests fail on a Linux i386 32bit build. See https://github.com/OSGeo/PROJ/pull/2749/checks?check_run_id=2942734733 . Not sure if the accuracy of the computation (probably an excess of accuracy due to 80 bit double) can be fixed or if the tolerance on the expected output should be relaxed.

marcus-elia:
@rouault I am interesting running the unit tests on that failing 32-bit build using a sphere instead of an ellipsoid, in case the precision issues come from the ellipsoid -> sphere conversion.  Is there a way to modify and run the tests without committing/pushing?

rouault:
>  Is there a way to modify and run the tests without committing/pushing?

you can do similarly as the CI build:
start a Docker image from the root of the PROJ repo:
docker run -it -e CI -e WORK_DIR="$PWD" -v $PWD:$PWD ubuntu:20.04
inside it:
$WORK_DIR/.github/workflows/linux_gcc_32bit/start.sh
and then you should be able to run / debug

marcus-elia:
I'm getting an error on
`$WORK_DIR/.github/workflows/linux_gcc_32bit/start.sh`
, and I've attached the output:
[output.log](https://github.com/OSGeo/PROJ/files/6786805/output.log)
Can you help me figure out what is going wrong?

rouault:
> [](https://github.com/OSGeo/PROJ/files/6786805/output.log)
> Can you help me figure out what is going wrong?

You probably need to clean your git checkout as there's probably conflicts with already built objects.
I generally do such debugging sessions in a dedicated git worktree : ``git worktree add ../debug_linux_32bit ``, cd into it, and then start docker from there

src/projections/s2.cpp
rouault: This still needs to be applied to fix Windows build

docs/source/operations/projections/s2.rst
```diff
@@ -0,0 +1,120 @@
...
+The specific transformation can be chosen with the UVtoST parameter:
+
++-------------------------+-----------------------------+
+|   ``+UVtoST=linear``    | fastest, no normalization   |
++-------------------------+-----------------------------+
+| ``+UVtoST=quadratic``   | fast, good normalization    |
++-------------------------+-----------------------------+
+|  ``+UVtoST=tangent``    | slowest, best normalization |
++-------------------------+-----------------------------+
+|    ``+UVtoST=none``     | returns u,v-coordinates      |
```
rouault: One extra space to remove:
```suggestion
|    ``+UVtoST=none``     | returns u,v-coordinates     |
```

src/iso19111/operation/parammappings.cpp
```diff
@@ -829,6 +829,10 @@ static const MethodMapping projectionMethodMappings[] = {
 
     {PROJ_WKT2_NAME_METHOD_QUADRILATERALIZED_SPHERICAL_CUBE, 0,
      "Quadrilateralized_Spherical_Cube", "qsc", nullptr, paramsNatOrigin},
+    
+    {PROJ_WKT2_NAME_METHOD_S2, 0,
+     "S2", "s2", nullptr, paramsNatOrigin},
```
rouault: paramsNatOrigin doesn't include sUVtoST. I would just remove that definition for now

marcus-elia: Okay, should I remove those two lines from `parammappings.cpp` ?

rouault: > Can you elaborate slightly on how these definitions are used in parameter parsing,

They are only needed when converting between PROJ.4 string style and WKT.  But WKT has not support for string values for a projection parameter, so having a proper WKT definition for S2 isn't trivial. When there's no projection method mappings, PROJ will generate something that is conformant to WKT formalism, but only interoperable in the PROJ universe, which is probably good enough for S2.

elfprince13: So, if, for example, someone wanted to generate GeoTIFFs in S2 coordinates via GDAL, it would be beneficial to have the correct parameter mappings available to stuff into the WKT header info?

PR #2750, opened 2021-06-17
proj_trans/cs2cs: If two operations have the same accuracy, use the one that is contained within a larger one

rouault:
Relates to https://github.com/OSGeo/gdal/issues/3998

Before that change, cs2cs on a NAD83(HARN) to WGS84 transformation would
use the "NAD83(HARN) to WGS 84 (1)" transformation (a null Helmert
shift) that is valid for whole US, including non-CONUS areas, even when
used on points located on CONUS that has a "NAD83(HARN) to WGS 84 (3)"
transformation (non-null Helmert shift) with same accuracy (1m).

But if doing  EPSG:2874 "NAD83(HARN) / California zone 5 (ftUS)" to
WGS84, we would use this later "NAD83(HARN) to WGS 84 (3)"
transformation because the area of use of EPSG:2874 restricts to CONUS.
This isn't consistant.

With that change, we now have more consistent behavior, even if it can
be argued which of the 2 transformations is the best...

$ echo 34 -120 | src/cs2cs -d 8 EPSG:4326 "NAD83(HARN)" | src/cs2cs "NAD83(HARN)"  EPSG:2874
5955507.74	1828410.98 0.00

$ echo 34 -120 | src/cs2cs  EPSG:4326   EPSG:2874
5955507.74	1828410.98 0.00

### DIFF
--- a/docs/source/operations/operations_computation.rst
+++ b/docs/source/operations/operations_computation.rst
@@ -59,9 +59,9 @@ From a code point of view, the entry point of the algorithm is the C++
 
 It combines several strategies:
 
-- look up in the PROJ database for available operations
-- consider the pair (source CRS, target CRS) to synthetize operations depending
-  on the nature of the source and target CRS.
+  - look up in the PROJ database for available operations
+  - consider the pair (source CRS, target CRS) to synthetize operations depending
+    on the nature of the source and target CRS.
 
 Geographic CRS to Geographic CRS, with known identifiers
 --------------------------------------------------------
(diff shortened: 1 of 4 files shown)


kbevers:
Seems like a sensible change to me. Does this PR change the logic of what's described in https://proj.org/operations/operations_computation.html ?

rouault:
> Does this PR change the logic of what's described in https://proj.org/operations/operations_computation.html ?

no, this page explains how we compute and sort operations. Here the change is about how we use the resulted sorted list of operations, and we don't necessarily use the operations that was ranked first, since when we have coordinates we have more context to be able to pick up one that seems more relevant.

kbevers:
Okay, thanks for the clarification. Just wanted to make sure this was remembered in case the logic changed.

rouault:
I've just added a note in operations_computation.rst about that

PR #2763, opened 2021-07-03
Find nlohmann_json dependency in cmake config file when proj is confi…

### LINKED ISSUE #2762: CMake targets broken in 8.1.0 when using external nlohmann_json
When cmake configure finds an external nlohmann_json it will use that.

In that case a nlohmann_json::nlohmann_json dependency is set on the cmake targets. But no `find_dependency(nlohmann_json REQUIRED)` is added to the proj-config.cmake.

Adding the proj target as a dependency now causes a cmake error:
```
Target "yourtarget" links to target "nlohmann_json::nlohmann_json" but the target was not found
```

#### Environment Information
 - PROJ version 8.1.0

#### Installation method
 - from source

dirkvdb:
Possible fix for issue #2762

 - [ ] Closes #2762

### DIFF
--- a/cmake/project-config.cmake.in
+++ b/cmake/project-config.cmake.in
@@ -15,6 +15,11 @@ set (@PROJECT_VARIANT_NAME@_INCLUDE_DIRS "${_ROOT}/@INCLUDEDIR@")
 set (@PROJECT_VARIANT_NAME@_LIBRARY_DIRS "${_ROOT}/@LIBDIR@")
 set (@PROJECT_VARIANT_NAME@_BINARY_DIRS "${_ROOT}/@BINDIR@")
 
+if ("@NLOHMANN_JSON@" STREQUAL "external")
+  include(CMakeFindDependencyMacro)
+  find_dependency(nlohmann_json REQUIRED)
+endif ()
+
 set (@PROJECT_VARIANT_NAME@_LIBRARIES @PROJECT_VARIANT_NAME@::proj)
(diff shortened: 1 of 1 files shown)


rouault:
It doesn't feel right to have nlohmann/json requirement exposed to external users of PROJ, as it is purely an internal implementation detail

hobu:
I concur with @rouault. 

I guess I haven't been tracking so closely, but did we ever advertise that we would work with an external nlohmann? The established pattern is for projects that use it to vendor it, namespace it, and then not leak it into their external APIs. This is because many other libraries, PDAL in my case, are also using nlohmann, and the opportunity for misery increases substantially if libraries it is using start doing putting their own vendored copies in their public APIs.

rouault:
> I guess I haven't been tracking so closely, but did we ever advertise that we would work with an external nlohmann?

I'm culprit for that. This has been added in 8.1 as an option. If using external nlohmann, then the official namespace is used (and that way folks combining different libraries that use it can save some binary space, and this is more friendly of distributions w.r.t security issues). If using internal nlohmann, then the proj_nlohmann namespace is used to avoid potential version clash.
I'm not sure why cmake insists about having to require nlohmann when using the library. It probably lacks hints that this is a header only dependency and that the nlohmann headers aren't exposed in our own headers

rouault:
Superseded per https://github.com/OSGeo/PROJ/pull/2780

PR #2764, opened 2021-07-04
CMake: fix installation of executables on iOS

SpaceIm:
Since a CMake version I can't remember, CMake wants BUNDLE DESTINATION for executables on iOS, RUNTIME DESTINATION is ignored, otherwise configuration fails with:

```
CMake Error at PROJ/src/bin_cct.cmake:14 (install):
  install TARGETS given no BUNDLE DESTINATION for MACOSX_BUNDLE executable
  target "cct".
```

 Therefore it's easier to just use DESTINATION when TARGETS in install() command are executables.

### DIFF
--- a/src/bin_cct.cmake
+++ b/src/bin_cct.cmake
@@ -12,7 +12,7 @@ target_link_libraries(cct PRIVATE ${PROJ_LIBRARIES})
 target_compile_options(cct PRIVATE ${PROJ_CXX_WARN_FLAGS})
 
 install(TARGETS cct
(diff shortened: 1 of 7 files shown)


mwtoews (approved):
This seems correct, and is [`CMP0006 NEW`](https://cmake.org/cmake/help/latest/policy/CMP0006.html) policy behaviour (although I don't have access to a Mac). I wasn't aware that "RUNTIME" could be omitted from [install](https://cmake.org/cmake/help/latest/command/install.html#targets), but this seems to work fine across different platforms and versions of CMake.

PR #2789, opened 2021-07-26
cmake: remove unused configure checks

mathstuf:
These results are just not used anywhere, so there's no need to check
them (as the compiler will do it anyways).


rouault:
Looks good to me

mwtoews (approved):
Some of these headers are still in-use (e.g. `dlfcn.h`, `memory.h`), but others were not. The result variables (e.g. `HAVE_DLFCN_H`) are indeed not used, making these checks pointless. Let's leave it to the compiler to complain if any header file is not found.

PR #2791, opened 2021-07-30
Add support for Degree Sign on input

### LINKED ISSUE #2712: Be ready for U+00B0 DEGREE SIGN "°" on input
Sooner or later, please recognize "°" on input,
```
$ cs2cs +proj=latlon +to +proj=latlon <<< "22°56'52.2\"N 121°14'09.3\"E"
22dE    0dN 0.000°56'52.2"N 121°14'09.3"E
$ echo  "22°56'52.2\"N 121°14'09.3\"E" | sed s/°/d/g | cs2cs -f %.6f +proj=latlon +to +proj=latlon
22.947833       121.235917 0.000000

### PR DESCRIPTION
**Pull request by direvus**
This PR adds support in dsmtor() for a Degree Sign (U+00B0), encoded
as UTF-8 (`\xc2\xb0`), as an alternative symbol to `D`/`d` to designate
the degree unit.

No encodings other than UTF-8 are supported.

I had previously suggested in #2712 that I was willing to take a swing at this, for UTF-8 only, and nobody jumped in to dissuade me, so here goes.

I also correct a typo in a code comment and some fix some rogue whitespace on the way past.

I have added a test to test/unit/gie_self_test, but I wasn't able to run a clean red/green on the new test, due to an unrelated failure earlier in the same test block (line 411, currently failing on 'master').  My new test does pass if I comment out those failing tests.  As an aside, it would probably be a good idea to break this block up into smaller units that can run independently of each other.

As for the documentation, I have tweaked one of the examples in the `cs2cs` docs to demonstrate use of the degrees sign, but I didn't see anywhere in the docs that we give a general description of allowed DMS formats.  If I've missed it, or there is anywhere else in the docs that should be updated for this change, please let me know.  It's my first time offering a PR to this project so please do be gentle.

Cheers,
Brendan

Fixes #2712.

 - [x] Closes #2712
 - [x] Tests added

### DIFF
--- a/docs/source/apps/cs2cs.rst
+++ b/docs/source/apps/cs2cs.rst
@@ -261,7 +261,7 @@ The following script
 ::
 
     cs2cs +proj=latlong +datum=NAD83 +to +proj=utm +zone=10 +datum=NAD27 -r <<EOF
-    45d15'33.1" 111.5W
+    45°15'33.1" 111.5W
     45d15.551666667N -111d30
     +45.25919444444 111d30'000w
     EOF
--- a/src/dmstor.cpp
+++ b/src/dmstor.cpp
@@ -29,7 +29,7 @@ dmstor(const char *is, char **rs) {
 dmstor_ctx(PJ_CONTEXT *ctx, const char *is, char **rs) {
 	int n, nl, adv;
 	char *s, work[MAX_WORK];
-        const char* p;
+	const char* p;
 	double v, tv;
 
 	if (rs)
@@ -39,11 +39,18 @@ dmstor_ctx(PJ_CONTEXT *ctx, const char *is, char **rs) {
 	n = MAX_WORK;
 	s = work;
 	p = (char *)is;
+
+	/*
+	 * Copy characters into work until we hit a non-printable character or run
+	 * out of space in the buffer.  Make a special exception for the bytes 0xc2
+	 * and 0xb0, because they comprise Degree Sign U+00B0 in UTF-8.
+	 *
+	 * It is possible that a really odd input (like lots of leading zeros)
+	 * could be truncated in copying into work.  But ...
+	 */
 	while ((isgraph(*p) || *p == (char) 0xc2 || *p == (char) 0xb0) && --n)
 		*s++ = *p++;
 	*s = '\0';
-	/* it is possible that a really odd input (like lots of leading
-		zeros) could be truncated in copying into work.  But ... */
 	int sign = *(s = work);
 	if (sign == '+' || sign == '-') s++;
 	else sign = '+';
--- a/src/dmstor.cpp
+++ b/src/dmstor.cpp
@@ -67,7 +67,12 @@ dmstor_ctx(PJ_CONTEXT *ctx, const char *is, char **rs) {
 			return tv;
 		int adv = 1;
 
-		if (*s == 'D' || *s == 'd') {
+		if (*s == 'D' || *s == 'd' || *s == DEG_SIGN2) {
+			/*
+			 * Accept \xb0 as a single-byte degree symbol. This byte is the
+			 * degree symbol in various single-byte encodings: multiple ISO
+			 * 8859 parts, several Windows code pages and others.
+			 */
 			n = 0;
 		} else if (*s == '\'') {
 			n = 1;
--- a/src/dmstor.cpp
+++ b/src/dmstor.cpp
@@ -60,33 +60,32 @@ dmstor_ctx(PJ_CONTEXT *ctx, const char *is, char **rs) {
 		if ((tv = proj_strtod(s, &s)) == HUGE_VAL)
 			return tv;
 		adv = 1;
-		switch (*s) {
-		case 'D': case 'd':
-			n = 0; break;
-		case '\'':
-			n = 1; break;
-		case '"':
-			n = 2; break;
-		/* degree symbol ("\xc2\xb0" in UTF-8) */
-		case (char) 0xc2:
-			if (s[1] == (char) 0xb0) {
-				n = 0;
-				adv = 2;
-				break;
-			}
-		case 'r': case 'R':
+
+		if (*s == 'D' || *s == 'd') {
+			n = 0;
+		} else if (*s == '\'') {
+			n = 1;
+		} else if (*s == '"') {
+			n = 2;
+		} else if (s[0] == (char) 0xc2 && s[1] == (char) 0xb0) {
+			/* degree symbol ("\xc2\xb0" in UTF-8) */
+			n = 0;
+			adv = 2;
+		} else if (*s == 'r' || *s == 'R') {
 			if (nl) {
 				proj_context_errno_set( ctx, PROJ_ERR_INVALID_OP_ILLEGAL_ARG_VALUE );
 				return HUGE_VAL;
 			}
 			++s;
 			v = tv;
-			goto skip;
-		default:
+			n = 4;
+			continue;
+		} else {
 			v += tv * vm[nl];
-		skip:	n = 4;
+			n = 4;
 			continue;
 		}
+
 		if (n < nl) {
 			proj_context_errno_set( ctx, PROJ_ERR_INVALID_OP_ILLEGAL_ARG_VALUE );
 			return HUGE_VAL;
--- a/src/dmstor.cpp
+++ b/src/dmstor.cpp
@@ -20,6 +20,12 @@ vm[] = {
 	.0002908882086657216,
 	.0000048481368110953599
 };
+/* byte sequence for Degree Sign U+00B0 in UTF-8. */
+	static constexpr char
+DEG_SIGN1 = '\xc2';
+	static constexpr char
+DEG_SIGN2 = '\xb0';
+
 	double
 dmstor(const char *is, char **rs) {
 	return dmstor_ctx( pj_get_default_ctx(), is, rs );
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/dmstor.cpp [hunk 1]
```diff
@@ -27,19 +27,19 @@ dmstor(const char *is, char **rs) {
...
 	if (rs)
 		*rs = (char *)is;
-	/* copy sting into work space */
+	/* copy string into work space */
 	while (isspace(*is)) ++is;
 	n = MAX_WORK;
 	s = work;
 	p = (char *)is;
-	while (isgraph(*p) && --n)
+	while ((isgraph(*p) || *p == (char) 0xc2 || *p == (char) 0xb0) && --n)
```
kbevers: A comment explaining the symbols would be useful here

direvus: Hi @kbevers, thanks for your feedback.  I have added a comment explaining this line and pushed a new commit to my branch.

src/dmstor.cpp
```diff
@@ -27,38 +27,53 @@ dmstor(const char *is, char **rs) {
...
 			n = 1; break;
 		case '"':
 			n = 2; break;
+		/* degree symbol ("\xc2\xb0" in UTF-8) */
+		case (char) 0xc2:
+			if (s[1] == (char) 0xb0) {
+				n = 0;
+				adv = 2;
+				break;
+			}
```
rouault: there is missing code in the case s[1] is not 0xb0. I guess this should be what the default: case does

src/dmstor.cpp
rouault: We can test more accurately for the Degree Sign with:
```suggestion
	while ((isgraph(*p) || (p[0] == (char) 0xc2 && p[1] == (char) 0xb0)) && --n)
```

rouault (changes requested):

src/dmstor.cpp
QuLogic: That's not equivalent; the `p[1]` may go out bounds, but also this check will only match on the first byte, and not copy the second byte.

rouault: > `p[1]` may go out bounds

if p[0] == 0xc2, given that this is a null terminated string, p[1] cannot be out of bounds. But indeed, there would be an issue for copying the second byte.
This method should ideally be reworked to avoid the statically allocated work[] buffer and its potential overflows to use a std::string, but that's probably beyond the scope of this PR.
So yes, the current test might be acceptable

src/dmstor.cpp
direvus: HI @rouault, if s[1] is not 0xb0, then the code will not 'break', and so it should run through the other branches of the switch statement, and won't find any other matches, so it will flow through to the 'default' case.

At least, that was my intention.  Why do you believe it won't work?

rouault: >  and so it should run through the other branches of the switch statement, and won't find any other matches, so it will flow through to the 'default' case.

no, it will go on on the ``case 'r': case 'R':``. That's how switch / case works. The pedeantic compiler warnings we turn on in the continuous integration complain about that. You'll likely have to add the code of the ``default`` in the else branch

docs/source/apps/cs2cs.rst
```diff
@@ -261,7 +261,7 @@ The following script
 ::
 
     cs2cs +proj=latlong +datum=NAD83 +to +proj=utm +zone=10 +datum=NAD27 -r <<EOF
-    45d15'33.1" 111.5W
+    45°15'33.1" 111.5W
```
mwtoews, line 264: I'd prefer to keep this with `d`, as this is used to generate man pages, which are generally plain-text.

src/dmstor.cpp
mwtoews:
```
../../src/dmstor.cpp: In function 'double dmstor_ctx(PJ_CONTEXT*, const char*, char**)':
../../src/dmstor.cpp:72:4: error: this statement may fall through [-Werror=implicit-fallthrough=]
    if (s[1] == (char) 0xb0) {
    ^~
../../src/dmstor.cpp:77:3: note: here
   case 'r': case 'R':
   ^~~~
```

docs/source/apps/cs2cs.rst
QuLogic: Nothing wrong with UTF8 in man pages; in fact there are several in the non-English pages in `/usr/share/man/{anything but en}`.

src/dmstor.cpp
direvus: Yeah if we are really keen to only copy 0xc2 and 0xb0 bytes when they appear in sequence, I think we'd need to rewrite these 2 lines into something more like:

```
while (n > 0) {
    if (isgraph(*p)) {
        // copy 1 character, decrement n
    } else if (n > 1 && p[0] == 0xc2 && p[1] == 0xb0) {
        // copy 2 characters, decrement n by 2
    } else {
        // exit the loop
    }
}
```

src/dmstor.cpp
direvus: OK, thanks for picking this up.  Looking at it again with your comment in mind, I actually feel like this block might work better as a series of if .. else ifs, rather than a switch.  That way we can test for both bytes together and no need to duplicate the code in the "default" case.

src/dmstor.cpp
rouault: Visual Studio is not happy about the casts.
*reinterpret_cast<const unsigned char*>(p) == 0xc2 and similarly for 0xb0 would fix that.
Similarly at line 70

docs/source/apps/cs2cs.rst
direvus: @mwtoews I can put it back to `d` if it helps with man page generation, but then there would be nothing at all in the docs to clue users in to the fact that they can use the degree sign.  I feel like it ought to be mentioned *somewhere*, and the cs2cs page was the only place I found that touches on the DMS string format.  I'm open to other ideas about how to document the feature.

src/dmstor.cpp
direvus: What if we got rid of the casts altogether?

E.g.
```cpp
static const char DEG_SIGN1 = '\xc2';
static const char DEG_SIGN2 = '\xb0';

...

    if ( ... *p == DEG_SIGN1 ... )
```

g++ and my local linters seem to be perfectly fine with this. I don't have a copy of VS handy to check that.

rouault: > seem to be perfectly fine with this.

I believe that would be a good alternative. You could use "constexpr char DEG_SIGN ..."

direvus: Thanks, I have pushed that change to the PR branch.

src/dmstor.cpp
```diff
@@ -20,66 +20,86 @@ vm[] = {
...
+
 	double
 dmstor(const char *is, char **rs) {
 	return dmstor_ctx( pj_get_default_ctx(), is, rs );
 }
 
 	double
 dmstor_ctx(PJ_CONTEXT *ctx, const char *is, char **rs) {
-	int n, nl;
+	int n, nl, adv;
```
rouault: it would be better to move the declaration of the variable with its initialization at line 68

docs/source/apps/cs2cs.rst
mwtoews: Ideally it would be described in the online docs, but there is no perfect place for it. This feature could be reworked in [quickstart](https://proj.org/usage/quickstart.html) or in one of the following pages, but that's beyond what's expected here. So sure, let's leave it here for now.

mwtoews:
Another consideration is to also accept the single-byte encoding for the degree symbol `'\xb0'`, used in a wide range of encodings. Demo in Python:
```python
deg = "\N{DEGREE SIGN}"
print(deg)                   # °
print(deg.encode("utf-8"))   # b'\xc2\xb0'
print(deg.encode("cp1252"))  # b'\xb0'
```
This single byte sequence has the same meaning for: cp1250, cp1251, cp1252, cp1253, cp1254, cp1255, cp1256, cp1257, cp1258, iso8859_1, iso8859_2, iso8859_3, iso8859_7, iso8859_8, iso8859_9, iso8859_10, iso8859_13, iso8859_15, iso8859_16, koi8_t, kz1048, latin_1, mbcs, ptcp154, and others. There are other byte sequences for the degree symbol using other encodings, but they are less common or obsolete.

direvus:
Thanks @mwtoews I am comfortable to include it, given how many encodings we can cover with `\xb0` it seems like a useful addition.

src/dmstor.cpp
rouault: this isn't resolved

direvus: Agreed, thanks @rouault I've updated my branch today with this change.

test/unit/gie_self_tests.cpp
```diff
@@ -442,6 +442,12 @@ TEST(gie, info_functions) {
     /* we can't expect perfect numerical accuracy so testing with a tolerance */
     ASSERT_NEAR(-2.0, proj_dmstor(&buf[0], NULL), 1e-7);
 
+    /* test UTF-8 degree sign on DMS input */
+    ASSERT_NEAR(0.34512432, proj_dmstor("19°46'27\"E", NULL), 1e-7);
+
+    /* test ISO 8859-1, cp1252, et al. degree sign on DMS input */
+    ASSERT_NEAR(0.34512432, proj_dmstor("19\26046'27\"E", NULL), 1e-7);
```
rouault: MSVC complains about warning C4125: decimal digit terminates octal escape sequence
Perhaps the following would do ?
```suggestion
    ASSERT_NEAR(0.34512432, proj_dmstor("19" "\260" "46'27\"E", NULL), 1e-7);
```

PR #2793, opened 2021-08-03
Testing generation of auxiliary equinor db from test data. Will be removed when done

chflo:
- [ ] Closes #xxxx
 - [ ] Tests added


chflo:
Adding this to separate fork instead

PR #2797, opened 2021-08-11
Docs: Improve order of instructions for installing from source.

direvus:
The note about running autogen.sh was previously placed below the
instructions to run configure.

Instructions are easier to follow if they are written in the same order that they are carried out.


mwtoews (approved):
Thanks for the improvement!

PR #2825, opened 2021-08-26
proj.pc.in: add -lm -ldl in Libs.private

mwtoews:
This fixes static linking using `pkg-config proj --libs --static`

Note there currently isn't any coverage for this type of test, but I've been refactoring the postinstall suite locally, which isn't ready yet (preview at https://github.com/mwtoews/PROJ/tree/postinstall2)


mwtoews:
<s>If not too late, this could be backported to 8.1, with an update to NEWS.</s> Maybe not yet, as more work might be required.

rouault (approved):

PR #2833, opened 2021-09-01
Doc: mentions EPSG methods that corresponds to topocentric and ortho (fixes #2832)

### LINKED ISSUE #2832: EPGS codes of orthographic and topocentric implementations
Dear all, I am little bit confused about the EPGS codes and the corresponding implementations in Proj.
If I understand it correctly, 
https://proj.org/operations/conversions/topocentric.html?highlight=topocentric
then the `topocentric` conversion or maybe more precisely, `+proj=cart +ellps=WGS84 +step +proj=topocentric` , corresponds to EPSG  9836. Is my understanding correct?
I cannot find a reference for the orthographic projection `ortho`
https://proj.org/operations/projections/ortho.html?highlight=orthographic
I had a look at the implementation https://github.com/OSGeo/PROJ/blob/master/src/projections/ortho.cpp without digging too much into the details. If I understand it correctly, then orthographic projection is either "spherical" or "ellipsoidal".  Is it then EPSG 9839 or  EPSG 9840, respectively? And does the implementation follow the reference https://proj.org/zreferences.html#id17 as well (IOGP. Geomatics Guidance Note 7, Part 2)?
Thanks for the clarification.

rouault:


docs/source/operations/projections/ortho.rst
```diff
@@ -37,6 +37,8 @@ around a given latitude and longitude.
           ellipsoid must be forced to a sphere, for example by adding a ``+f=0``
           parameter.
 
+This projection method corresponds to ``ÈPSG:9840``.
```
stefangachter, line 40: Thanks for adding this! There might be a typo: È instead of E

rouault: fixed

PR #2835, opened 2021-09-02
Add proj_create_conversion_pole_rotation_netcdf_cf_convention() to address netCDF datasets using a pole rotation method

rouault:


snowman2:
Is there an upstream issue/discussion about this?

rouault:
> Is there an upstream issue/discussion about this?

This has certainly been discussed a number of times with different people, but not necessarily on public forums. Here's the objective is to support in a cleaner way the way the netCDF CF conventions encode rotated pole CRS. A similar work has been done in the past for the convention used in the GRIB format, which is another way of encoding the parameters.

rouault:
There will be follow-up work in GDAL to plug the netCDF driver to that new projection method

snowman2:
Good to know. I ask because pyproj has a [RotatedLatitudeLongitudeConversion](https://github.com/pyproj4/pyproj/blob/3812a2b693aacb62387199dc9878ad0189bad103/pyproj/crs/coordinate_operation.py#L1427-L1461) for CF conventions. However, what you have seems to be different by adding 180 to `lon_0`. I would be interested in knowing why it is implemented this way in PROJ.

rouault:
>  I would be interested in knowing why it is implemented this way in PROJ.

I'm not sure of the origin of the convention used by PROJ. The current formulas for this new netCDF CF convention rotated pole projection is based on experiments with netCDF products. There are several ways of expressing the rotated pole, GRIB and netCDF CF using different convention. For some formats for planetary maps (non Earth), there's even a third convention...  (but having different conventions for the same mathematical reality is not really specific to rotated pole. This is the case for Mercator, Polar stereographic or LCC)

snowman2:
Thanks for the explanation @rouault :+1:

PR #2851, opened 2021-09-14
Database: update to EPSG v10.035

rouault:
This seriously impacts French CRS users with the introduction of new
datums, geodetic CRS and projected CRS based on "RGF 93 v2" and "RGF 93 v2b",
and the previous single "RGF 93" being renamed as "RGF 93 v1".

To be noted too, the addition of a null transformation between NAD83(2011)
and WGS 84, which impacts a number of tests in the test suite.


rouault:
Some follow-up I've initiated with EPSG:
```
To: feedback@epsg.org
Subject: About new RGF93 datums

Hi,

I just discovered the renaming of RGF 93 as RGF 93 v1 and the addition of RGF 93 v2 and RGF 93 v2b. I anticipate that to cause quite a lot of chaos for most GIS practitioners for a sizeable number of years, but so be it.

To go back on technical things, I see that there's currently :

- no transformation between RGF 93 v1 and RGF 93 v2, and between RGF 93 v1 and RGF 93 v2b. I would suggest to add a null-shift Helmert transformation with some inaccuracy (I guess it would be a few centimeters at most) to help in scenarios where people will mix datasets in those different datums.

- no transformation between NTF and RGF 93 v2, and between NTF and RGF 93 v2b. I would suggest to add concatenated operations chaining the EPSG:15958 'RGF93 v1 to NTF (2)' and 'EPSG:9327 'NTF to RGF93 v1 (1)'  and a null-shift Helmert as suggested in my above bullet point.

Not sure what the procedure is to make that happen. Can you relay that to your IGN contacts ?

Best regards,

Even
```

PR #2858, opened 2021-09-17
DOC: fix the inclusion of the custom css file

jorisvandenbossche:
See mailing list message about broken css of proj.org. It seems that the readthedoc theme's css was no longer included in the `<head>` (only the css overrides of proj itself). Might be triggered due to changes in sphinx_rtd_theme 1.0, which was just released, but didn't verify that.

I _think_ this patch fixes that (because sphinx puts the parent theme's css in `css_files`, and thus the conf.py overrides this. And starting with sphinx_rtd_theme, it only relies on that https://github.com/readthedocs/sphinx_rtd_theme/commit/836467e4b92d7de5506dd8f996d87a7fc56d0b5d), and in any case it now includes the custom css as documented by rtd (https://docs.readthedocs.io/en/stable/guides/adding-custom-css.html)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -196,6 +196,10 @@ html_favicon = '../images/favicon.png'
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
+htm_css_files = [
+    'theme_overrides.css',  # override wide tables in RTD theme
+]
+
 html_context = {
     'display_github': True,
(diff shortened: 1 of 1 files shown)


kbevers:
Let's see if it works

mwtoews:
Artefact looks good, thanks @jorisvandenbossche !

kbevers:
Yeah, and I've backported it to 8.1 manually. Thanks for fixing this!

PR #2863, opened 2021-09-24
DOC: Refresh Development/Quickstart and remove Development/Threads

### LINKED ISSUE #2451: Outdated docs
https://proj.org/development/threads.html says it documents version 7.2.0., but it contains a lot about version 4 and functions like `pj_ctx_alloc()` which seem to be older functions. On the [Quickstart](https://proj.org/development/quickstart.html) the function `proj_context_create()` is mentioned. So there are some discrepancies here.

direvus:
The document Development/Threads was severely out of date, and was
focussed on contrasting the current API with an old API that was
deprecated several years ago, and removed from the current PROJ earlier
this year.

Update Development/Quickstart to give more information about the use of
threading context objects, as a replacement for the content that was previously
in Threads.

Also give Quickstart a general cleanup and reorganise the text to flow more
logically with the code samples.

Also give the example code examples/pj_obs_api_mini_demo.c a cleanup
pass to make it at least internally consistent with its own code style.

The header comments of the example code are still, much like the Threads
document, fixated on comparing proj.h against proj_api.h, which is an
interesting historical curio to be sure, but probably doesn't need to
persist here.  It might be worth cleaning this up further as a separate
exercise.

Fixes #2451

 - [x] Closes #2451 


docs/source/development/quickstart.rst
```diff
@@ -28,24 +28,48 @@ See the :doc:`reference for more info on data types <reference/datatypes>`.
...
+program.  This ensures that all :c:type:`PJ` objects created in the same
+context will be sharing resources such as error-numbers and loaded grids.
+
+If you are sure that ``P`` will only be used by a single program thread, you
+may pass ``NULL`` for the threading context.  This will assign the default
+thread context to ``P``.
+
+The strings for the source and target CRS may be any of:
+
+-  PROJ strings, e.g. ``+proj=longlat +datum=WGS84``,
```
kbevers: For the PROJ string to be equivalent to `EPSG:4326` we need to specify that it is a CRS:
```suggestion
-  PROJ strings, e.g. ``+proj=longlat +datum=WGS84 +type=crs``,
```

kbevers (commented):
This is a great contribution! Thanks for taking the time to improve this part of the docs.

direvus:
You're welcome, and thank you for the positive response.  I hope to be able to contribute in a more technical capacity soon, but right now it seems like the docs are where I can add the most value.

kbevers:
> You're welcome, and thank you for the positive response. I hope to be able to contribute in a more technical capacity soon, but right now it seems like the docs are where I can add the most value.

Any contribution is very welcome but you are correct in saying that the docs need a bit of love!

PR #2868, opened 2021-09-24
proj_factors(): accept P to be a projected CRS (fixes #2854)

### LINKED ISSUE #2854: proj_factors() only seems to work with a proj string
#### Example of problem
Patching the PROJ test suite as below demonstrates the issue:

```patch
index 5fc8f0f1..2ef339cc 100644
@@ -454,7 +454,7 @@ TEST(gie, info_functions) {
                 1e-7);
 
     /* test proj_derivatives_retrieve() and proj_factors_retrieve() */
-    P = proj_create(PJ_DEFAULT_CTX, "+proj=merc +ellps=WGS84");
+    P = proj_create(PJ_DEFAULT_CTX, "EPSG:3395");
     a = proj_coord(0, 0, 0, 0);
     a.lp.lam = proj_torad(12);
     a.lp.phi = proj_torad(55);
```
#### Problem description

With the patch above applied the PROJ test suite fails like so:

```
$ make check
```
[....]
```
[ RUN      ] gie.info_functions
Unrecognized vertical grid format for filename 'proj.db'
Unrecognized horizontal grid format for filename 'proj.db'
Invalid latitude or longitude
gie_self_tests.cpp:463: Failure
Value of: proj_errno(P)
  Actual: true
Expected: false
[  FAILED  ] gie.info_functions (2 ms)
```

More generally I can't seem to get `proj_factors()` to work in **any** case unless I've called `proj_create()` or similar with a PROJ string.

This is a problem because current PROJ docs say:

> The use of proj-string to describe a CRS is discouraged. It is a legacy means of conveying CRS descriptions: use of object codes (EPSG:XXXX typically) or WKT description is recommended for better expressivity.

It appears I'm currently prevented from following this advice if I want to use `proj_factors()`.

#### Expected Output

I'd expect your test suite to still pass with the above patch applied (assuming that `EPSG:3395` is actually equivalent to `+proj=merc +ellps=WGS84` - it appears to be to me at least).

#### Environment Information
 - PROJ version: git master as of 47fb85bdf0e45bf5660e0b1dbde7b76f8824a87a
 - Operation System Information: Debian unstable amd64

#### Installation method
I'm testing as above in a clone of your git repo.

(Issue originally encountered with debian proj package version 7.2.1-1 - I checked then with git master to avoid reporting if it was already fixed.)

rouault:
Updated doc:

    Starting with PROJ 8.2, the P object can be a projected CRS, for example
    instantiated from a EPSG CRS code. The factors computed will be those of the
    map projection implied by the transformation from the base geographic CRS of
    the projected CRS to the projected CRS.

    The input geodetic coordinate lp should be such that lp.lam is the longitude
    in radian, and lp.phi the latitude in radian (thus independently of the
    definition of the base CRS, if P is a projected CRS).

CC @busstoptaktik

### DIFF
--- a/src/4D_api.cpp
+++ b/src/4D_api.cpp
@@ -1896,6 +1896,48 @@ PJ_FACTORS proj_factors(PJ *P, PJ_COORD lp) {
     if (nullptr==P)
         return factors;
 
+    const auto type = proj_get_type(P);
+
+    if( type == PJ_TYPE_PROJECTED_CRS )
+    {
+        // If it is a projected CRS, then compute the factors on the conversion
+        // associated to it. We need to start from a temporary geographic CRS
+        // using the same datum as the one of the projected CRS, and with
+        // input coordinates being in longitude, latitude order in radian,
+        // to be consistent with the expectations of the lp input parameter.
+
+        auto ctx = P->ctx;
+        auto geodetic_crs = proj_get_source_crs(ctx, P);
+        assert(geodetic_crs);
+        auto datum = proj_crs_get_datum(ctx, geodetic_crs);
+        auto datum_ensemble = proj_crs_get_datum_ensemble(ctx, geodetic_crs);
+        auto cs = proj_create_ellipsoidal_2D_cs(
+            ctx, PJ_ELLPS2D_LONGITUDE_LATITUDE, "Radian", 1.0);
+        auto temp = proj_create_geographic_crs_from_datum(
+            ctx, "unnamed crs", datum ? datum : datum_ensemble,
+            cs);
+        proj_destroy(datum);
+        proj_destroy(datum_ensemble);
+        proj_destroy(cs);
+        proj_destroy(geodetic_crs);
+        auto newOp = proj_create_crs_to_crs_from_pj(ctx, temp, P, nullptr, nullptr);
+        proj_destroy(temp);
+        assert(newOp);
+        auto ret = proj_factors(newOp, lp);
+        proj_destroy(newOp);
+        return ret;
+    }
+
+    if( type != PJ_TYPE_CONVERSION &&
+        type != PJ_TYPE_TRANSFORMATION &&
+        type != PJ_TYPE_CONCATENATED_OPERATION &&
+        type != PJ_TYPE_OTHER_COORDINATE_OPERATION )
+    {
+        proj_log_error(P, _("Invalid type for P object"));
+        proj_errno_set (P, PROJ_ERR_INVALID_OP_ILLEGAL_ARG_VALUE);
+        return factors;
+    }
+
     if (pj_factors(lp.lp, P, 0.0, &f))
         return factors;
 
--- a/docs/source/development/reference/functions.rst
+++ b/docs/source/development/reference/functions.rst
@@ -756,6 +756,15 @@ Various
     distortion and meridian convergence. Depending on the underlying projection
     values will be calculated either numerically (default) or analytically.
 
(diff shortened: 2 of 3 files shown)


busstoptaktik (approved):
I have no objections to this - as I have stated in the discussion in #2854 I believe the code "as is" is not broken; I do, however recognize the use cases presented by the participants  in the discussion, and laid out especially clearly by @snowman2, so I think this patch holds merit and am happy to endorse it, and hope it fulfills the needs of the intended audience, although it is not entirely clear to me whether this will be the case: The fixed lp.lam/lp.phi interpretation of the input may be a problem - I cannot comment on that: To me at least, this looks good!

rouault:
> The fixed lp.lam/lp.phi interpretation of the input may be a problem - I cannot comment on that: To me at least, this looks good!

yeah I was unsure about the best approach for that. I'm afraid there's no clear winner. I also considered the option to use the axis unit and order of the base geographic CRS of the projected CRS, as expected with proj_trans().
That said we do have somehow a problem with the PJ_COORD structure. It was thought prior to all the PROJ 6 work, where the axis order of the authority wasn't taken into account, and if you for example do proj_trans() from EPSG:4326 to something else, you need to provide input as latitude, longitude degree. But in which member of the PJ_COORD structure to do that ? A "naive" user might be tempted to use the lp / lpz / lpzt, but he would get things wrong as the implied semantics of lp.lam and lp.phi is infact completely lost when you just get a PJ_COORD. So the most reasonable is to use the xyzt or v[4] members.

busstoptaktik:
> That said we do have somehow a problem with the PJ_COORD structure. It was thought prior to all the PROJ 6 work, where the axis order of the authority wasn't taken into account

Yes - over in https://github.com/busstoptaktik/geodesy I try a different approach, by simply calling the coordinates "first, second, third and fourth", i.e. the v[4] solution, and proclaiming that internally all operators work in "gis" order with radians for angular units, then having "`adapt`ors" handling the ormat of input and output systems in both ends of pipelines. In Rust Geodesy, a coordinate-tuple is "just a 4-tuple", its interpretation is up to the operator.

I think it's fine to merge this, but perhaps @snowman2 or @ojwb would have an opinion on this?

ojwb:
Thanks for working on this.

How the coordinates are specified here probably makes it simpler for my existing code, but I can certainly see the point that it's not entirely consistent with the modern PROJ approach.  It's clearly and explicitly documented here at least.

I'll aim to find time to try it out over the weekend and see if I have any feedback in the light of actually trying to use it.

src/4D_api.cpp
```diff
@@ -1896,6 +1896,48 @@ PJ_FACTORS proj_factors(PJ *P, PJ_COORD lp) {
...
 
+    const auto type = proj_get_type(P);
+
+    if( type == PJ_TYPE_PROJECTED_CRS )
+    {
+        // If it is a projected CRS, then compute the factors on the conversion
+        // associated to it. We need to start from a temporary geographic CRS
+        // using the same datum as the one of the projected CRS, and with
+        // input coordinates being in longitude, latitude order in radian,
+        // to be consistant with the expectations of the lp input parameter.
```
ojwb: Typo: "consistent" (e not a)

ojwb:
I've now tested and this branch does indeed solve my problem with getting convergence values from PROJ, so thanks for that.

I don't think I really have any further useful insights about the coordinate order and units.  They seems OK from my code, though that was written already.

PR #2880, opened 2021-09-29
RFC 7: Drop Autotools, maintain CMake

mwtoews:
This RFC proposes to drop Autotools for PROJ 9.0, and to maintain CMake for build automation, testing and packaging.

To see the rendered document on GitHub, click "View file" in the "..." pop-up beside rfc-7.rst.

### DIFF
--- a/docs/source/community/rfc/index.rst
+++ b/docs/source/community/rfc/index.rst
@@ -17,3 +17,4 @@ the project.
    rfc-4
    rfc-5
    rfc-6
+   rfc-7
--- a/docs/source/community/rfc/rfc-7.rst
+++ b/docs/source/community/rfc/rfc-7.rst
@@ -15,7 +15,7 @@ Summary
 
 This RFC proposes to drop Autotools for PROJ 9.0, and to maintain CMake
 for build automation, testing and packaging. This will reduce the overall
-maintenance for PROJ, and will enable the library to be integrated into other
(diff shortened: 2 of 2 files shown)


cffk:
Yes, I'd like to see Autotools fade way.

On the question of providing static and shared libraries, you might want
to check out GeographicLib's cmake support.

  https://github.com/geographiclib/geographiclib/tree/devel

Running cmake with

    -DGEOGRAPHICLIB_LIB_TYPE=BOTH

get both versions of the library built.  The end user can request a
package containing the static library

    find_package (GeographicLib COMPONENTS STATIC)

${GeographicLib_LIBRARIES} points to the shared library, if it's
available, unless

    set (GeographicLib_USE_STATIC_LIBS ON)

(mimicking FindBoost).

There's nothing particularly difficult in providing this support and I'm
not recommending the PROJ use exactly the same approach as GeographicLib
(I'm in the process of figuring out how to simplify the way
GeographicLib handles things).

A big impediment, in my view, is the lack of clear guidance from the
CMake developers on the recommended way of doing things like this.  It
would be better if there was a well established pattern to follow rather
than each package doing its own idiosyncratic implementation.

rouault:
This can probably be now submitted as a motion for approval

mwtoews:
[Motion passed](https://lists.osgeo.org/pipermail/proj/2021-October/010407.html)

PR #2882, opened 2021-09-30
ENH: Add proj_trans_bounds

### LINKED ISSUE #2779: PROPOSAL: proj_transform_bounds
Related:
https://github.com/pyproj4/pyproj/issues/809
https://lists.osgeo.org/pipermail/gdal-dev/2021-July/054450.html

Any interest in porting the logic from pyproj into PROJ?

snowman2:
- [x] Closes #2779
 - [x] Tests added

### DIFF
--- a/docs/source/development/reference/functions.rst
+++ b/docs/source/development/reference/functions.rst
@@ -391,6 +391,11 @@ Coordinate transformation
               reasons.
 
 
+
+.. doxygenfunction:: proj_trans_bounds
+   :project: doxygen_api
+
+
 Error reporting
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
--- a/src/4D_api.cpp
+++ b/src/4D_api.cpp
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
     return ret;
 }
 
+
+// ---------------------------------------------------------------------------
+static double simple_min(const double* data, const int arr_len) {
+    double min_value = data[0];
+    for( int iii = 1; iii < arr_len; iii++ ) {
+        if (data[iii] < min_value)
+            min_value = data[iii];
+    }
+    return min_value;
+}
+
+
+// ---------------------------------------------------------------------------
+static double simple_max(const double* data, const int arr_len) {
+    double max_value = data[0];
+    for( int iii = 1; iii < arr_len; iii++ ) {
+        if ((data[iii] > max_value || max_value == HUGE_VAL) && data[iii] != HUGE_VAL)
+            max_value = data[iii];
+    }
+    return max_value;
+ }
+
+
+// ---------------------------------------------------------------------------
+static int _find_previous_index(const int iii, const double* data, const int arr_len) {
+    // find index of nearest valid previous value if exists
+    int prev_iii = iii - 1;
+    if (prev_iii == -1)  // handle wraparound
+        prev_iii = arr_len - 1;
+    while (data[prev_iii] == HUGE_VAL && prev_iii != iii) {
+        prev_iii --;
+        if (prev_iii == -1)  // handle wraparound
+            prev_iii = arr_len - 1;
+    }
+    return prev_iii;
+}
+
+
+// ---------------------------------------------------------------------------
+/******************************************************************************
+Handles the case when longitude values cross the antimeridian
+when calculating the minimum.
+Note: The data array must be in a linear ring.
+Note: This requires a densified ring with at least 2 additional
+        points per edge to correctly handle global extents.
+If only 1 additional point:
+    |        |
+    |RL--x0--|RL--
+    |        |
+-180    180|-180
+If they are evenly spaced and it crosses the antimeridian:
+x0 - L = 180
+R - x0 = -180
+For example:
+Let R = -179.9, x0 = 0.1, L = -179.89
+x0 - L = 0.1 - -179.9 = 180
+R - x0 = -179.89 - 0.1 ~= -180
+This is the same in the case when it didn't cross the antimeridian.
+If you have 2 additional points:
+    |            |
+    |RL--x0--x1--|RL--
+    |            |
+-180        180|-180
+If they are evenly spaced and it crosses the antimeridian:
+x0 - L = 120
+x1 - x0 = 120
+R - x1 = -240
+For example:
+Let R = -179.9, x0 = -59.9, x1 = 60.1 L = -179.89
+x0 - L = 59.9 - -179.9 = 120
+x1 - x0 = 60.1 - 59.9 = 120
+R - x1 = -179.89 - 60.1 ~= -240
+However, if they are evenly spaced and it didn't cross the antimeridian:
+x0 - L = 120
+x1 - x0 = 120
+R - x1 = 120
+From this, we have a delta that is guaranteed to be significantly
+large enough to tell the difference reguarless of the direction
+the antimeridian was crossed.
+However, even though the spacing was even in the source projection, it isn't
+guaranteed in the target geographic projection. So, instead of 240, 200 is used
+as it significantly larger than 120 to be sure that the antimeridian was crossed
+but smalller than 240 to account for possible irregularities in distances
+when re-projecting. Also, 200 ensures latitudes are ignored for axis order handling.
+******************************************************************************/
+static double antimeridian_min(const double* data, const int arr_len) {
+    int prev_iii = 0;
+    double positive_min = HUGE_VAL;
+    double min_value = HUGE_VAL;
+    double delta = 0;
+    int crossed_meridian_count = 0;
+    bool positive_meridian = false;
+
+    for( int iii = 0; iii < arr_len; iii++ ) {
+        if (data[iii] == HUGE_VAL)
+            continue;
+        prev_iii = _find_previous_index(iii, data, arr_len);
+        // check if crossed meridian
+        delta = data[prev_iii] - data[iii];
+        // 180 -> -180
+        if (delta >= 200 && delta != HUGE_VAL) {
+            if (crossed_meridian_count == 0)
+                positive_min = min_value;
+            crossed_meridian_count ++;
+            positive_meridian = false;
+        // -180 -> 180
+        } else if (delta <= -200 && delta != HUGE_VAL) {
+            if (crossed_meridian_count == 0)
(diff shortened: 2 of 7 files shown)


src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+ * Transform boundary densifying the edges to account for nonlinear
+ * transformations along these edges and extracting the outermost bounds.
+ *
+ * If the destination CRS is geographic and right < left then the bounds
+ * crossed the antimeridian. In this scenario there are two polygons,
+ * one on each side of the antimeridian. The first polygon should be
+ * constructed with (left, bottom, 180, top) and the second with
+ * (-180, bottom, top, right).
+ *
+ * When projecting from polar projections to geographic,
```
rouault: you probably lon, lat *input* order ? (as it is a "requirement"). But even that is annoying. For consistence, we should try to respect the axis order of the input CRS implied by the P object. The contains_north_pole/contains_south_pole should rather get the target CRS of the PJ* object with proj_get_target_crs(), check if it is geographic with proj_get_type() (if it is not, then you can give up), query the coordinate system and then the axis

src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+ * lon, lat output order is required.
+ *
+ * @param P The PJ object representing the transformation.
+ * @param direction The direction of the transformation.
+ * @param left The input left bounding coordinate.
```
rouault: we should probably mention that left is according to the first axis of the source CRS of P* (or the target CRS if using the reverse direction). Same  kind of remark for the other 7 arguments

src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+        *out_left = left;
+        *out_right = right;
+        *out_bottom = bottom;
+        *out_top = top;
+        return true;
+    }
+
+    bool degree_output = proj_degree_output(P, direction) != 0;
+    bool degree_input = proj_degree_input(P, direction) != 0;
+    if (degree_output && densify_pts < 2) {
```
rouault, line 1347: wondering if densify_pts == 0 shouldn't be used as a hint to mean "please, algorithm use a sensible default value for me" ?

src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+        proj_log_error(P, _("densify_pts must be at least 2 if the output is geograpic."));
+        proj_errno_set (P, PROJ_ERR_INVALID_OP_ILLEGAL_ARG_VALUE);
+        return false;
+    }
+
+    int side_pts = densify_pts + 1;  // add one because we are densifying
```
rouault, line 1353: it would be reasonable to have if ( densify_pts < 0 || densify_pts > 10000 ) sanity check to avoid crazy memory allocations

src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+    int boundary_len = side_pts * 4;
+    std::vector<double> x_boundary_array(boundary_len);
```
rouault: it could be prudent to do in 2 steps
```
std::vector<double> x_boundary_array;
std::vector<double> y_boundary_array;
try
{
    x_boundary_array.resize(boundary_len);
    t_boundary_array.resize(boundary_len);
}
catch( const std::exception & e ) // memory allocation failure
{
     // return an error using e.what()
}
```

src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+    bool degree_output = proj_degree_output(P, direction) != 0;
+    bool degree_input = proj_degree_input(P, direction) != 0;
+    if (degree_output && densify_pts < 2) {
+        proj_log_error(P, _("densify_pts must be at least 2 if the output is geograpic."));
+        proj_errno_set (P, PROJ_ERR_INVALID_OP_ILLEGAL_ARG_VALUE);
+        return false;
+    }
+
+    int side_pts = densify_pts + 1;  // add one because we are densifying
+    int boundary_len = side_pts * 4;
```
```suggestion
    const int boundary_len = side_pts * 4;
```

src/4D_api.cpp
rouault: and the left / bottom / right / top names aren't probably appropriate if we just follow the axis order. xmin, ymin, xmax, ymax ? where xmin is a latitude in EPSG:4326 ?

src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+            P,
+            direction,
+            left,
+            bottom,
+            right,
+            top
+        );
+    }
+
+    if (degree_input && right < left) {
```
rouault: that logic would have to be adapted to the actual axis order if we enforce the one of the source CRS

src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+        // handle antimeridian
+        delta_x = (right - left + 360.0) / ( (double) side_pts );
```
rouault: implict cast from int to double will be done here
```suggestion
        delta_x = (right - left + 360.0) / side_pts;
```
applies for below similar occurences

src/proj.h
```diff
@@ -600,6 +600,19 @@ size_t PROJ_DLL proj_trans_generic (
     double *t, size_t st, size_t nt
 );
 
+int PROJ_DLL proj_trans_bounds(
```
rouault, line 603: reference to add to it in docs/source/development/reference/functions.rst using a .. doxygenfunction:: style of inclusion at the appropriate place

src/4D_api.cpp
```diff
@@ -966,6 +966,423 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+large enough to tell the difference reguarless of the direction
+the antimeridian was crossed.
+However, even though the spacing was even in the source projection, it isn't
+guaranteed in the target geographic projection. So, instead of 240, 200 is used
+as it significantly larger than 120 to be sure that the antimeridian was crossed
+but smalller than 240 to account for possible irregularities in distances
+when re-projecting. Also, 200 ensures latitudes are ignored for axis order handling.
+******************************************************************************/
+static double antimeridian_min(const double* data, const int arr_len) {
+    int prev_iii = 0;
```
rouault: all declarations that can be moved inside the loop should, with combined declaration-definition and const when possible. Same for _max function

src/4D_api.cpp
snowman2: That is a good thought, however 0 is valid when you are going to a projected CRS.

src/4D_api.cpp
snowman2: Good thinking :+1:

src/4D_api.cpp
snowman2: Thanks for the suggestion :+1:

src/4D_api.cpp
snowman2: Good to know, I will remove it.

src/4D_api.cpp
snowman2: Will do, Thanks :+1:

src/4D_api.cpp
snowman2: Sounds like a good idea, thanks :+1:

docs/source/development/reference/functions.rst
```diff
@@ -391,6 +391,11 @@ Coordinate transformation
               reasons.
 
 
+
+.. doxygenfunction:: proj_trans_bounds
```
rouault, line 395: the failure is due to 4D_api.cpp not being set in INPUT of Doxyfile (there might be issues since this will be the first time Doxygen is run onto it)

src/4D_api.cpp
snowman2: That seems to do the trick (pyproj [MR](https://github.com/pyproj4/pyproj/pull/966)). Will port over to PROJ next.

busstoptaktik: The bboxen seem to suffer from axis perplexity:  `(left, bottom, 180, top)` versus `(-180, bottom, top, right)`. I suppose `top` and `right` should be swapped in the latter case (or `180` and `top` in the former?).

snowman2: That example assumes that longitude is the first axis. It would probably be good to add an example for the scenario that latitude is the first axis.

snowman2: Oh, I see what you mean. The example needs to be fixed :-)

src/4D_api.cpp
snowman2: Based on the tests, I think this can be resolved. If I am missing something, please let me know.

docs/source/development/reference/functions.rst
snowman2: I see the issues in the build. 
```
Warning, treated as error:
PROJ/docs/source/development/reference/functions.rst:395:doxygenfunction: Cannot find function "proj_trans_bounds" in doxygen xml output for project "doxygen_api" from directory: ../build/xml/
make: *** [html] Error 2
Makefile:31: recipe for target 'html' failed
Error: Process completed with exit code 2.
```
Is this something that will fix itself in future runs?

snowman2: Think it is resolved now.

src/4D_api.cpp
```diff
@@ -966,6 +966,517 @@ std::string pj_add_type_crs_if_needed(const std::string& str)
...
+            direction,
+            xmin,
+            ymin,
+            xmax,
+            ymax,
+            output_lon_lat_order
+        );
+    }
+
+    if (degree_input && xmax < xmin) {
```
rouault, line 1407: the source CRS lon/lat order should be tested here

snowman2: Are you thinking that an error should be raised if this condition is true and the axis order is incorrect?

rouault: > Are you thinking that an error should be raised if this condition is true and the axis order is incorrect?

that would be reasonable

snowman2:
Thanks for the reviews :+1:

PR #2889, opened 2021-10-07
Add transformations for GGM10, Mexican geoid model

jjimenezshaw:
- added entry in `grid_alternatives.sql` for `mx_inegi_ggm10.tif`, included in `PROJ-data`
 - added entry in `grid_transformation_custom.sql` until EPSG is updated. This gives us more time to test the behaviour.
 - added entry in `customizations.sql` in table `geoid_model` for `GGM10`

See related PR OSGeo/PROJ-data#72

### DIFF
--- a/data/sql/grid_transformation_custom.sql
+++ b/data/sql/grid_transformation_custom.sql
@@ -230,6 +230,6 @@ INSERT INTO "usage" VALUES(
     'grid_transformation',
     'PROJ',
     'EPSG_6364_TO_EPSG_5703',
-    'EPSG','1160', -- area of use: Mexico - onshore and offshore.
+    'EPSG','3278', -- area of use: Mexico - onshore
     'EPSG','1024'  -- unknown
 );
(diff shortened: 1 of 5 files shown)


jjimenezshaw:
For Area of Use I set `'EPSG:1160 Mexico - onshore and offshore.` I do not know if `EPSG:3278 Mexico-onshore` would be better (the tif area is in between the two, if I checked correctly).

data/sql/grid_transformation_custom.sql
```diff
@@ -211,3 +211,25 @@ INSERT INTO "usage" VALUES(
...
+    'EPSG','8666','Geoid (height correction) model file','GGM10.txt',
+    NULL,NULL,NULL,NULL,NULL,NULL,NULL,0);
+
+INSERT INTO "usage" VALUES(
+    'PROJ',
+    'EPSG_6364_TO_EPSG_5703_USAGE',
+    'grid_transformation',
+    'PROJ',
+    'EPSG_6364_TO_EPSG_5703',
+    'EPSG','1160', -- area of use: Mexico - onshore and offshore.
```
rouault: we should select an area of use that is smaller or equal to the one of the grid, otherwise the grid could be selected for points outside of it

rouault (commented):
the metadata.sql file should also be updated to indicate 1.8 for proj-data

PR #2895, opened 2021-10-09
CMake: add a BUILD_APPS to be able to disable build of all applications

rouault:
This makes it easier to turn off all programs, rather than individually.
Useful for example to avoid https://github.com/OSGeo/gdal/blob/master/gdal/fuzzers/build.sh#L138

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -322,33 +322,43 @@ CMake configure options
 Options to configure a CMake are provided using ``-D<var>=<value>``.
 All cached entries can be viewed using ``cmake -LAH`` from a build directory.
 
+.. option:: BUILD_APPS=ON
+
+    Build PROJ applications. Default is ON. Control the default value for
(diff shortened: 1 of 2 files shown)


mwtoews (commented):
Good meta-option for users that only need the library. Another consideration for the name is `BUILD_APPS` which I see other projects using, and also fitting here given we have a `src/apps` directory.

Suggest to add some content to [`docs/source/install.rst`](https://github.com/OSGeo/PROJ/blob/master/docs/source/install.rst), at the top of the "CMake configure options" section.

Checking with `BUILD_PROGRAMS=OFF` showed many errors; see #2898 which I can handle next.

rouault:
> for the name is `BUILD_APPS`

done

> Suggest to add some content to docs/source/install.rst

done

mwtoews (approved):

nyalldawson (approved):

PR #2899, opened 2021-10-10
CMake: Always build gie if testing is requested

### LINKED ISSUE #2898: CMake: No target "gie"
When configuring a CMake build of PROJ with:
```
cmake -DBUILD_GIE=OFF -DBUILD_TESTING=ON ..
```
an error is repeated over 30 times:
```
CMake Error at cmake/ProjTest.cmake:31 (add_test):
  Error evaluating generator expression:

    $<TARGET_FILE_NAME:gie>

  No target "gie"
  test/CMakeLists.txt:6 (proj_add_gie_test)
```
Some options are:

1. Skip tests that require gie
2. Build gie so it can be used for tests, but don't install it

Unless anyone has an objection, I'll go for option 2.

mwtoews:
Closes #2898

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -70,7 +70,7 @@ if(BUILD_PROJINFO)
 endif()
 
 # Always build gie if testing is requested
-if(BUILD_TESTING)
+if(BUILD_GIE OR BUILD_TESTING)
   include(bin_gie.cmake)
   set(BIN_TARGETS ${BIN_TARGETS} gie)
 endif()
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -69,7 +69,8 @@ if(BUILD_PROJINFO)
   set(BIN_TARGETS ${BIN_TARGETS} binprojinfo)
 endif()
 
-if(BUILD_GIE)
+# Always build gie if testing is requested
+if(BUILD_TESTING)
```
rouault: We should also build it if requested, independently of BUILD_TESTING
```suggestion
if(BUILD_GIE OR BUILD_TESTING)
```

rouault (approved):

PR #2900, opened 2021-10-10
CMake build: add generate_wkt1_parser and generate_wkt2_parser manual target, and logic to detect when they must be run

rouault:


src/generate_wkt_parser.cmake
```diff
@@ -0,0 +1,19 @@
...
+                RESULT_VARIABLE STATUS)
+
+if(STATUS AND NOT STATUS EQUAL 0)
+  message(FATAL_ERROR "bison failed")
+endif()
+
+# Post processing of the generated file
+# All those replacements are to please MSVC
+file(READ ${OUT_FILE} CONTENTS)
+string(REPLACE "*yyssp = yystate" "$(yytype_int16)yystate" CONTENTS "${CONTENTS}")
```
nyalldawson: Is this replacement intentional? Or should it be something like `*yyssp = (yytype_int16)yystate" ?

rouault: the replacement was wrong indeed, and useless too since it was for a earlier bison version. Removed

Issue #2903, opened 2021-10-14
Default to nearest triangle in tinshift transformation


josch:
Hi,

I'm using proj and tinshift to render openstreetmap with mapnik. One problem I'm facing is, that mapnik doesn't like there to be points that cannot be transformed because they are not inside any of the triangles I specify as input to the tinshift transformation. Would it be an acceptable feature to add a flag to tinshift, that will fall back to the closest triangle for points that are not inside any triangle? I'm willing to try and write code that does exactly that but first wanted to bring this up with you to ask whether this is something you'd consider adding to proj.

Thanks!

rouault:
I'm wondering if it should be an option of the +proj=tinshift operation, or some option in the .json file itself. No strong opinion.

busstoptaktik:
@josch I think you have resolved  this yourself with #2907, so this could probably be closed?

josch:
Yes, thank you all for your the help and advice!! :+1:

PR #2907, opened 2021-10-16
add fallback strategy for tinshift transform to use closest triangle for points not in any

josch:
Hi,

this is a WIP pull request for #2903 and I'm asking for guidance to finish it up. Open questions are:

 - is "fallback_strategy" a good name? The default does not change the current behaviour.
 - should "fallback_strategy" be part of the json or the proj string?
 - currently "fallback_strategy" is a string -- should it be an enum? Maybe there are other strategies that could be useful in future implementations?
 - should the json version be bumped because of the new field?
 - is there a way to compute distance to a triangle with less computations? I can only come up with heuristics like the commented out piece of code which is not precise but only checks the vertices of the triangles.
 - the quadtree does not seem to be ready for performing a nearest-neighbor kind of search. It would need to be extended to support a distance function and a shared variable to track the currently found closest feature. For my own use-case, the current implementation that does not use the quad tree is sufficient.

TODO:

 - [x] Tests added

### DIFF
--- a/src/transformations/tinshift.hpp
+++ b/src/transformations/tinshift.hpp
@@ -53,6 +53,12 @@
 
 namespace TINSHIFT_NAMESPACE {
 
+enum FallbackStrategy {
+    FALLBACK_NONE,
+    FALLBACK_NEAREST_SIDE,
+    FALLBACK_NEAREST_CENTROID,
+};
+
 using json = nlohmann::json;
 
 // ---------------------------------------------------------------------------
@@ -92,6 +98,10 @@ class TINShiftFile {
     /** Get a text description of the model. Intended to be longer than name()
      */
     const std::string &publicationDate() const { return mPublicationDate; }
+
+    const enum FallbackStrategy &fallbackStrategy() const {
+        return mFallbackStrategy;
+    }
 
     /** Basic information on the agency responsible for the model. */
     struct Authority {
@@ -204,6 +214,7 @@ class TINShiftFile {
     std::string mLicense{};
     std::string mDescription{};
     std::string mPublicationDate{};
+    enum FallbackStrategy mFallbackStrategy {};
     Authority mAuthority{};
     std::vector<Link> mLinks{};
     std::string mInputCRS{};
--- a/src/transformations/tinshift_impl.hpp
+++ b/src/transformations/tinshift_impl.hpp
@@ -93,6 +93,24 @@ std::unique_ptr<TINShiftFile> TINShiftFile::parse(const std::string &text) {
     tinshiftFile->mDescription = getOptString(j, "description");
     tinshiftFile->mPublicationDate = getOptString(j, "publication_date");
 
+    tinshiftFile->mFallbackStrategy = FALLBACK_NONE;
+    if (j.contains("fallback_strategy")) {
+        if (tinshiftFile->mFormatVersion != "1.1") {
+            throw ParsingException(
+                "fallback_strategy needs format_version 1.1");
+        }
+        const auto fallback_strategy = getOptString(j, "fallback_strategy");
+        if (fallback_strategy == "nearest_side") {
+            tinshiftFile->mFallbackStrategy = FALLBACK_NEAREST_SIDE;
+        } else if (fallback_strategy == "nearest_centroid") {
+            tinshiftFile->mFallbackStrategy = FALLBACK_NEAREST_CENTROID;
+        } else if (fallback_strategy == "none") {
+            tinshiftFile->mFallbackStrategy = FALLBACK_NONE;
+        } else {
+            throw ParsingException("invalid fallback_strategy");
+        }
+    }
+
     if (j.contains("authority")) {
         const json jAuthority = j["authority"];
         if (!jAuthority.is_object()) {
@@ -410,6 +428,28 @@ Evaluator::Evaluator(std::unique_ptr<TINShiftFile> &&fileIn)
 
 // ---------------------------------------------------------------------------
 
+static inline double sqr(double x) { return x * x; }
+static inline double squared_distance(double x1, double y1, double x2,
+                                      double y2) {
+    return sqr(x1 - x2) + sqr(y1 - y2);
+}
+static double distance_point_segment(double x, double y, double x1, double y1,
+                                     double x2, double y2, double dist12) {
+    // squared distance of point x/y to line segment x1/y1 -- x2/y2
+    double t = ((x - x1) * (x2 - x1) + (y - y1) * (y2 - y1)) / dist12;
+    if (t <= 0) {
+        // closest to x1/y1
+        return squared_distance(x, y, x1, y1);
+    }
+    if (t >= 1) {
+        // closest to y2/y2
+        return squared_distance(x, y, x2, y2);
+    }
+
+    // closest to line segment x1/y1 -- x2/y2
+    return squared_distance(x, y, x1 + t * (x2 - x1), y1 + t * (y2 - y1));
+}
+
 static const TINShiftFile::VertexIndices *
 FindTriangle(const TINShiftFile &file,
              const NS_PROJ::QuadTree::QuadTree<unsigned> &quadtree,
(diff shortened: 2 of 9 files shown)


src/transformations/tinshift_impl.hpp
```diff
@@ -453,7 +454,119 @@ FindTriangle(const TINShiftFile &file,
...
+        if (dist < closest_dist) {
+            closest_dist = dist;
+            closest_i = i;
+            continue;
+        }
+        */
+
+        double l2, t, dist2;
+        // squared distance to line segment x1/y1 -- x2/y2
+        l2 = (x1 - x2) * (x1 - x2) + (y1 - y2) * (y1 - y2);
```
rouault: a inline function squared_distance(x1, y1, x2, y2) would be welcome, mostly to avoid potential copy&paste issues

rouault:
> * is "fallback_strategy" a good name? The default does not change the current behaviour.

Sounds good to me

> * should "fallback_strategy" be part of the json or the proj string?

I'm fine with it in the json.

> * currently "fallback_strategy" is a string -- should it be an enum? Maybe there are other strategies that could be useful in future implementations?

In the C++ code you mean I guess ?(as you already made it an enum in the json schema)  . Yes that would be fine

> * should the json version be bumped because of the new field?

Yes, adding 1.1 in the format_version enum of the json schema

> * is there a way to compute distance to a triangle with less computations? I can only come up with heuristics like the commented out piece of code which is not precise but only checks the vertices of the triangles.

I don't really have ready-made suggestions here. This is certainly a well-known problem, and must be implemented in GEOS for the Distance() function.

> * the quadtree does not seem to be ready for performing a nearest-neighbor kind of search. It would need to be extended to support a distance function and a shared variable to track the currently found closest feature. For my own use-case, the current implementation that does not use the quad tree is sufficient.

That would probably be fine to use that "naive" approach for now.

josch:
Thanks a lot for your input! I will work on these points and push a new version once I'm ready.

> > * is there a way to compute distance to a triangle with less computations? I can only come up with heuristics like the commented out piece of code which is not precise but only checks the vertices of the triangles.
> 
> I don't really have ready-made suggestions here. This is certainly a well-known problem, and must be implemented in GEOS for the Distance() function.

Can you elaborate on this point? This is my first contact with PROJ and I don't know what you mean with GEOS and the Distance() function.

Thanks!

rouault:
> Can you elaborate on this point? This is my first contact with PROJ and I don't know what you mean with GEOS and the Distance() function.

I was thinking to the GEOS library that has gazillons of geometry computation algorithms : https://github.com/libgeos/geos/blob/main/capi/geos_c.h.in#L1520  . But we don't want to depend or draw GEOS in PROJ. Maybe https://stackoverflow.com/questions/849211/shortest-distance-between-a-point-and-a-line-segment could be of help ?

josch:
> I was thinking to the GEOS library that has gazillons of geometry computation algorithms : https://github.com/libgeos/geos/blob/main/capi/geos_c.h.in#L1520 . But we don't want to depend or draw GEOS in PROJ.

Ah, now I understand. You were suggesting that maybe GEOS has a better implementation. I just checked and no it does not: https://github.com/libgeos/geos/blob/5a3cebe6e9b1b1d612638a07fce6d81ac04cd684/src/operation/distance/DistanceOp.cpp#L601 Just as my code above, GEOS just computes the distance between line segments and points. Its documentation even warns that its approach is O(n²): http://geos.refractions.net/ro/doxygen_docs/html/classgeos_1_1operation_1_1distance_1_1DistanceOp.html#_details

> Maybe https://stackoverflow.com/questions/849211/shortest-distance-between-a-point-and-a-line-segment could be of help ?

The answers there do the same thing that my code does: build a line from two points, obtain the line parameter t by projecting the point onto the line and check whether t is less than 0, larger than 1 or in the middle to compute the distance.

test/gie/tinshift.gie
```diff
@@ -47,4 +47,10 @@ accept      3210000.0000 6700000.0000   10.0
 expect      3210000.0000 6700000.0000   10.2886
 roundtrip   1
 
+# format_version is invalid with format_version before 1.1
```
rouault: this comment doesn't look very relevant for the test

src/transformations/tinshift_impl.hpp
```diff
@@ -93,6 +93,21 @@ std::unique_ptr<TINShiftFile> TINShiftFile::parse(const std::string &text) {
     tinshiftFile->mDescription = getOptString(j, "description");
     tinshiftFile->mPublicationDate = getOptString(j, "publication_date");
 
+	tinshiftFile->mFallbackStrategy = FALLBACK_NONE;
```
rouault: formatting issue. Please run ./scripts/reformat_cpp.sh

src/transformations/tinshift_impl.hpp
```diff
@@ -410,6 +425,29 @@ Evaluator::Evaluator(std::unique_ptr<TINShiftFile> &&fileIn)
...
 
+static inline double sqr(double x) { return x * x; }
+static inline double squared_distance(double x1, double y1, double x2,
+                                      double y2) {
+    return sqr(x1 - x2) + sqr(y1 - y2);
+}
+static double distance_point_segment(double x, double y, double x1, double y1,
+                                     double x2, double y2) {
+    // squared distance of point x/y to line segment x1/y1 -- x2/y2
+    double t = ((x - x1) * (x2 - x1) + (y - y1) * (y2 - y1)) /
```
rouault: in case of degenerate triangles, a division by zero could occur

src/transformations/tinshift_impl.hpp
```diff
@@ -453,7 +491,74 @@ FindTriangle(const TINShiftFile &file,
...
+    const unsigned i2 = triangle.idx2;
+    const unsigned i3 = triangle.idx3;
+    const double x1 = vertices[i1 * colCount + idxX];
+    const double y1 = vertices[i1 * colCount + idxY];
+    const double x2 = vertices[i2 * colCount + idxX];
+    const double y2 = vertices[i2 * colCount + idxY];
+    const double x3 = vertices[i3 * colCount + idxX];
+    const double y3 = vertices[i3 * colCount + idxY];
+    const double det_T = (y2 - y3) * (x1 - x3) + (x3 - x2) * (y1 - y3);
+    lambda1 = ((y2 - y3) * (x - x3) + (x3 - x2) * (y - y3)) / det_T;
```
rouault, line 587: in case of degenerate triangles, a division by zero could occur

rouault (commented):
Beyond the few details pointed, overall looks good to me !

src/transformations/tinshift_impl.hpp
josch: Ah, I indeed did not think of that case. Does it make sense for a transformation to contain degenerate triangles? Should the parser maybe throw an error if any of the input triangles in degenerate? I also see that the existing code also does not guard against division by zero.

rouault: > Should the parser maybe throw an error if any of the input triangles in degenerate?

maybe, but we want the initial parsing to be as fast as possible. Anyway, this is not a critical point. That can be left for later potentially

josch: Okay, I added code that avoids division by zero and tests for the two cases of degenerate triangles (one side is zero length or two angles are 0°). This also made me fix another bug: if the list of triangles is zero, the function would try to access the non-existant triangle at position 0. Whoops. XD I added a test for that as well.

busstoptaktik:
> currently "fallback_strategy" is a string -- should it be an enum? Maybe there are other strategies that could be useful in future implementations?

Sorry for being late to the party, but I think you are right here: Actually I think, the obvious thing to do is to select the triangle for which the external point is closest to its centroid (i.e. the intersection of the medians).

In many cases, the triangle selected will not differ from "nearest side", since geodetic TINs tend to be conscoiusly and sensibly constructed. But in the case of a long narrow triangle with the short side facing the outer edge of the TIN, its neighbouring more harmoniously shaped triangles may be more fit for extrapolation?

And if having the centroids of all triangles precomputed at instantiation, finding the nearest (in the centroid sense) is simply a matter of comparing squared point-to -point distances `(x - x_n) * (x - x_n) + (y - y_n) * (y - y_n)`

josch:
> Sorry for being late to the party, but I think you are right here: Actually I think, the obvious thing to do is to select the triangle for which the external point is closest to its centroid (i.e. the intersection of the medians).

This is a good point and it also means that "nearest" is a bad name for the format. I renamed it to "nearest_side" and added a new method called "nearest_centroid" which finds the triangle with the closest centroid to the query point.

src/transformations/tinshift.hpp
```diff
@@ -53,6 +53,12 @@
 
 namespace TINSHIFT_NAMESPACE {
 
+enum FallbackStrategy {
+    FALLBACK_NONE = 0,
+    FALLBACK_NEAREST_SIDE = 1,
+    FALLBACK_NEAREST_CENTROID = 1,
```
rouault: ouch, you set the same value to the 2 nearest methods... You should remove explict assignation of values, which aren't needed. And retest carefully your new method since it was likely untested

josch: Hah! This *does* explain a weird test result that I couldn't explain -- thanks! I'll remove the explicit assignment and fix the tests.

josch: Thanks for finding that mistake. The test is now fixed and I checked that it will fail if fallback strategy is set to "nearest_side" instead of "nearest_centroid".

src/transformations/tinshift_impl.hpp
```diff
@@ -453,7 +493,101 @@ FindTriangle(const TINShiftFile &file,
...
+    const unsigned i2 = triangle.idx2;
+    const unsigned i3 = triangle.idx3;
+    const double x1 = vertices[i1 * colCount + idxX];
+    const double y1 = vertices[i1 * colCount + idxY];
+    const double x2 = vertices[i2 * colCount + idxX];
+    const double y2 = vertices[i2 * colCount + idxY];
+    const double x3 = vertices[i3 * colCount + idxX];
+    const double y3 = vertices[i3 * colCount + idxY];
+    const double det_T = (y2 - y3) * (x1 - x3) + (x3 - x2) * (y1 - y3);
+    if (det_T < EPS) {
```
rouault: can't the determinant be negative ?  Shoudln't it be std::fabs(det_T) that is tested ?

PR #2908, opened 2021-10-19
TYPE: Update const declarations in proj_trans_bounds 

snowman2:
Continuation of #2882


rouault:
is is a BUG fix ?
I don't think it is super idiomatic in C to tag non-pointer types as const in their declaration/prototype (the .h). You could just declare them as const in the definition (the C++). See https://stackoverflow.com/questions/15297644/is-it-ok-for-function-prototypes-and-function-implementation-signatures-to-use-c
It would be more interesting to tag PJ* as `const PJ*` (in both files) however

snowman2:
> I don't think it is super idiomatic in C to tag non-pointer types as const in their declaration/prototype (the .h). You could just declare them as const in the definition (the C++)

Good to know, thanks :+1: . Just updated the PR.

PR #2911, opened 2021-10-22
CMake: align the VERSION of the library with libtool practices

rouault:
Currently (at least as of 8.1.1) we don't have the same .so.x.y.z
numbers for the libtool and cmake builds (the y and z are swapped for
the cmake build). While only the SONAME (the .x part) mostly matters,
it is probably better to stick with the libtool convention to determine
the y and z.
This patch should make it easier by setting explicitly each of the
current, revision, age and computing automatically
x = current - age
y = age
z = revision


kbevers (approved):
This will certainly prevent me from screwing up the numbers every other release. Thanks!

Issue #1821, opened 2020-01-03
Should we "import" symbols by default (Windows environment) ?


FilipGh:
In file `proj.h`, some macro definitions are set in order to import or export symbols in DLL:
<pre>
#ifndef PROJ_DLL
#ifdef PROJ_MSVC_DLL_EXPORT
#define PROJ_DLL __declspec(dllexport)
#elif defined(PROJ_MSVC_DLL_IMPORT)
#define PROJ_DLL __declspec(dllimport)
#else
#define PROJ_DLL
#endif
#endif
</pre>

When linking an executable against `proj_6_x.dll`, it is *mandatory* to add the following option in the Visual C++ compiling command line:
/D PROJ_MSVC_DLL_IMPORT

Shouldn't it be easier to "import" symbols by default, as:
<pre>
#ifndef PROJ_DLL
   #ifdef PROJ_MSVC_DLL_EXPORT
      #define PROJ_DLL __declspec(dllexport)
   #else
      #define PROJ_DLL __declspec(dllimport)
   #endif
#endif
</pre>
at least in a *Windows* environment ?

rouault:
>  is _mandatory_ to add the following option in the Visual C++ compiling command line:

That's where things are mysterious to me, and there's something odd in our current setup. __declspec(dllimport) shouldn't be mandatory according to https://blogs.msdn.microsoft.com/russellk/2005/03/20/lnk4217/

QuLogic:
> linking an executable against `proj_6_x.dll`,

Shouldn't you be linking against a `.lib`, which I think would have the list of exported things in it?

FilipGh:
In bumped into some linking error when compiling PROJ-JNI.
As @QuLogic says, I do link with *proj_6_2.lib* on the command line.
The weird thing is that I have no link errors with classes/methods/functions. The only missing symbols are :
* osgeo::proj::common::UnitOfMeasure::SCALE_UNITY (METER, RADIAN, ...)
* pj_release

The only way I could compile and link PROJ-JNI C++ part was to add the following line in CMakeList.txt :
`target_compile_options(proj-bind PUBLIC "/D PROJ_MSVC_DLL_IMPORT")`

rouault:
> when compiling PROJ-JNI.

[...]

PR #2912, opened 2021-10-23
CMake: revise how we deal with symbol export and static builds

rouault:
- Remove the explicit PROJ_MSVC_DLL_IMPORT symbol used for importing
  symbols from a MSVC .dll: by default on MSVC, we use
  now __declspec(dllimport), unless PROJ_MSVC_DLL_EXPORT is defined
  by PROJ at build time. This makes it easier for users: they
  don't have to define anything special. This simplifies in particular
  the build of our binaries
- For static builds, export -DPROJ_DLL= as public, so that users
  that import PROJ through CMake mechanism don't have to do it
  manually.

### DIFF
--- a/include/proj/util.hpp
+++ b/include/proj/util.hpp
@@ -50,43 +50,38 @@ namespace proj {}
 //! @cond Doxygen_Suppress
 
 #ifndef PROJ_DLL
-#ifdef PROJ_MSVC_DLL_EXPORT
-#define PROJ_DLL __declspec(dllexport)
-#elif defined(PROJ_MSVC_DLL_IMPORT)
-#define PROJ_DLL __declspec(dllimport)
-#elif defined(__GNUC__)
-#define PROJ_DLL __attribute__((visibility("default")))
-#else
-#define PROJ_DLL
-#endif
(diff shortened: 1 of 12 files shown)


rouault:
Solves https://github.com/OSGeo/PROJ/issues/1821

kbevers (commented):
Looks alright to me although I can't claim to understand all of it in details. Should the docs be updated to reflect this?

rouault:
> Should the docs be updated to reflect this?

I don't see any mention of that in the docs. It is more than with this change we'd need less documentations as things will be more automatic and in line with general practice

kbevers:
Was just wondering since I don't really grasp the details. Is the special Windows behaviour noted in https://proj.org/install.html#cmdoption-arg-BUILD_SHARED_LIBS still present?

rouault:
>  Is the special Windows behaviour noted in https://proj.org/install.html#cmdoption-arg-BUILD_SHARED_LIBS still present?

yes, I didn't change that. This PR is only about how we import symbols

kbevers (approved):

PR #2913, opened 2021-10-24
lib_proj.cmake: add a PROJ::proj alias and add BUILD_INTERFACE include directories...

### LINKED ISSUE #2905: PROJ is not able to be built as a subdirectory
#### Example of problem

```
#include <proj.h>
#include <iostream>
int main(){
    return 0;
}
```

```
cmake_minimum_required(VERSION 3.13)
project(test)
set(CMAKE_CXX_STANDARD 17)

add_subdirectory(external/PROJ)
add_executable(test main.cpp)
target_link_libraries(test PRIVATE PROJ::proj)
```
```
-- PROJ: Configured 'dist' target
-- Configuring done
  Target "test" links to target "PROJ::proj" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
```

with out alias target

```
cmake_minimum_required(VERSION 3.13)
project(test)
set(CMAKE_CXX_STANDARD 17)

add_subdirectory(external/PROJ)
add_executable(test main.cpp)
target_link_libraries(test PRIVATE proj) 
```
```
/home/user/projects/test/main.cpp:2:10: fatal error: proj.h: No such file or directory
 #include <proj.h>
          ^~~~~~~~
```

#### Problem description
Using PROJ as a sub directory is not possible, this creates issues with many workflows, especially if one is not using a package manager with PROJ, which is very likely now that VCPKG only supports pre 8.0 versions of proj and does not support geotiff builds. 

PROJ currently creates an alias target for `PROJ::proj`, but only during installs, and it doesn't appear as if any usable target, however this is not a big deal as long as *a* target exists.  Currently it appears as if `proj` is that target, however it doesn't appear to properly propagate it's own `target_include_directories`, so it can't be used in isolation. 

#### Expected Output

The above example should compile with out error. 

#### Environment Information
 - PROJ version 8.2.0
 Centos 7
#### Installation method
 - from source

rouault:
... so that proj can be used as a subdirectory of a larger project (fixes #2905)

### DIFF
--- a/src/lib_proj.cmake
+++ b/src/lib_proj.cmake
@@ -372,6 +372,8 @@ add_library(proj
   ${ALL_LIBPROJ_HEADERS}
   ${PROJ_RESOURCES}
 )
+add_library(PROJ::proj ALIAS proj)
+
 target_compile_options(proj
   PRIVATE $<$<COMPILE_LANGUAGE:C>:${PROJ_C_WARN_FLAGS}>
   PRIVATE $<$<COMPILE_LANGUAGE:CXX>:${PROJ_CXX_WARN_FLAGS}>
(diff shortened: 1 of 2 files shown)


kbevers:
This would be nice to include in 8.2.0, right? I'll hold of the release candidate until this is merged (within reason, obviously).

PR #2914, opened 2021-10-26
Add new option to proj_create_crs_to_crs_from_pj method to force +over on transformation operations

### LINKED ISSUE #2512: Using +over with EPSG codes and WKT CRS Input/Output?

mlptownsend:
The stale bot closed my original PR 2528. I didn't have the free time to follow up on this until now.

 - [x] Closes #2512 
 - [x] Tests added

My original strategy used a method to set a flag on the context to set the PJ->over flag. The problem is getting the value down to the pipelines. As-is, there isn't really a clear path down to send options, and I didn't really feel comfortable making that big of a breaking change to the internal APIs to hurl down additional parameters.

image
For instance in this situation, the PJ created in the pipeline method is the PJ that needs to actually receive the 1 in its over flag. But it actually gets generated through calling proj_list_get. Thus, I still had to keep the flag on the context. I just clear it when proj_create_crs_to_crs_from_pj returns. I'm not the biggest fan of the approach if anyone has a better idea there.

### DIFF
--- a/docs/source/development/reference/functions.rst
+++ b/docs/source/development/reference/functions.rst
@@ -204,6 +204,17 @@ paragraph for more details.
 .. doxygenfunction:: proj_normalize_for_visualization
    :project: doxygen_api
 
+.. c:function:: void proj_context_set_force_over(PJ_CONTEXT* ctx, int enable)
+
+   .. versionadded:: 8.0.0
+
+   Force the +over flag on transformations created with this threading-context.
+
+   :param ctx: Threading context.
+   :type ctx: :c:type:`PJ_CONTEXT` *
+   :param enable: Toggle the enabled flag.
+   :type enable: `int`
+
 .. c:function:: PJ* proj_destroy(PJ *P)
 
     Deallocate a :c:type:`PJ` transformation object.
--- a/docs/source/development/reference/functions.rst
+++ b/docs/source/development/reference/functions.rst
@@ -201,19 +201,11 @@ paragraph for more details.
     - ALLOW_BALLPARK=YES/NO: can be set to NO to disallow the use of
       :term:`Ballpark transformation` in the candidate coordinate operations.
 
+    - FORCEOVER=YES/NO: can be set to YES to force the +over flag on the transformation
+      returned by this function.
+
 .. doxygenfunction:: proj_normalize_for_visualization
    :project: doxygen_api
-
-.. c:function:: void proj_context_set_force_over(PJ_CONTEXT* ctx, int enable)
-
-   .. versionadded:: 8.0.0
-
-   Force the +over flag on transformations created with this threading-context.
-
-   :param ctx: Threading context.
-   :type ctx: :c:type:`PJ_CONTEXT` *
-   :param enable: Toggle the enabled flag.
-   :type enable: `int`
 
 .. c:function:: PJ* proj_destroy(PJ *P)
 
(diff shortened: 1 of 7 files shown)


docs/source/development/reference/functions.rst
```diff
@@ -201,6 +201,9 @@ paragraph for more details.
     - ALLOW_BALLPARK=YES/NO: can be set to NO to disallow the use of
       :term:`Ballpark transformation` in the candidate coordinate operations.
 
+    - FORCEOVER=YES/NO: can be set to YES to force the +over flag on the transformation
```
rouault: maybe with an underscore (and change accordingly in the rest of the code) ?
```suggestion
    - FORCE_OVER=YES/NO: can be set to YES to force the +over flag on the transformation
```

src/4D_api.cpp
```diff
@@ -1843,6 +1843,7 @@ PJ  *proj_create_crs_to_crs_from_pj (PJ_CONTEXT *ctx, const PJ *source_crs, cons
     const char* authority = nullptr;
     double accuracy = -1;
     bool allowBallparkTransformations = true;
+    int forceOver = 0;
```
```suggestion
    bool forceOver = false;
```
and use true/false in all other places

src/proj_internal.h
```diff
@@ -683,6 +683,7 @@ struct pj_ctx{
     void    *logger_app_data = nullptr;
     struct projCppContext* cpp_context = nullptr; /* internal context for C++ code */
     int     use_proj4_init_rules = -1; /* -1 = unknown, 0 = no, 1 = yes */
+    int     forceOver = 0; /* 0 = no, 1 = yes */
```
rouault:
```suggestion
    bool    forceOver = false;
```

src/4D_api.cpp
```diff
@@ -1930,6 +1940,7 @@ PJ  *proj_create_crs_to_crs_from_pj (PJ_CONTEXT *ctx, const PJ *source_crs, cons
     if( preparedOpList.empty() )
     {
         proj_destroy(P);
+        ctx->forceOver = 0;
```
rouault: instead of doing that here and in the below cases, you could likely do that just after pj_create_prepared_operations()

mlptownsend:
Adjusted. No rush on trying to meet any sort of release deadline on it since you're trying to get 8.2 out the door.

I created another flag that does something a lil' bit similar. When I have another bout of free time I'll try and whip up another PR for it. It might end up working itself in the create function in the same kind of way.

test/unit/gie_self_tests.cpp
```diff
@@ -890,4 +890,185 @@ TEST(gie, proj_trans_generic) {
...
 }
 
+TEST(gie, proj_create_crs_to_crs_from_pj_force_over) {
+
+    PJ_CONTEXT* ctx;
+
+    ctx = proj_context_create();
+    ASSERT_TRUE(ctx != nullptr);
+
+    auto epsg27700 = proj_create(ctx, "EPSG:27700");
```
rouault, line 900: that one and the 2 other below objects should be proj_destroy()'ed

PR #2919, opened 2021-11-03
fix MSVC 2019 build: C2491 - definitions can't be `dllimport`

mmomtchev:
Fix building with VS 2019, C2491: definitions can't be `dllimport`

 - [*] Added clear title that can be used to generate release notes


mmomtchev:
Superseded by #2920

PR #2920, opened 2021-11-04
MSVC build: fix all C2491 errors: Function definitions cannot be dllimport

mmomtchev:
MSVC build: fix all C2491 errors

Function definitions and static members cannot be dllimport


mmomtchev:
In fact, this is not needed - what is needed is a CMake option to completely remove PROJ_DLL when building a static lib

These files are not included in the application

rouault:
I'll still take your PR as it removes non-idiomatic uses of PROJ_DLL

Issue #2739, opened 2021-06-04
ETRS89 to Chart datum transformation in Norway


himsve:
## ETRS89 to Chart datum transformation in Norway

The Norwegian Mapping Authority has defined the vertical transformation between ETRS89 and chart datum with vertical correction grid (GeoTiff). The transformation involves to steps:

1) Vertical shifts from correction grid (ETRS89 ell. height>Chart datum height)
2) Height Depth Reversal of Chart datum height. (Chart datum height>Chart datum depth)

The Norwegian chart datum has already EPSG code 9672 (https://epsg.org/home.html)  

The transformation is prototyped in branches:
[CD Norway branch - Proj](https://github.com/himsve/PROJ/tree/typo_cdnorway)
[CD Norway branch - Proj-data](https://github.com/himsve/PROJ-data/tree/type_cdnorway)

### Figure vertical datums, example

![CDNORWAY](https://user-images.githubusercontent.com/56073535/120648512-02fc2580-c47c-11eb-9a1f-0f73f7e6255e.png)

### Coverage vertical correction grid

![CDNORWAY_coverage](https://user-images.githubusercontent.com/56073535/120648597-1b6c4000-c47c-11eb-80fc-1c1e0fa73b54.jpeg)

### Discussions

- So far, neither the national transformation from ETRS89 ell. to Chart datum, or intermediate vertical datum are defined in the EPSG registry. Do we need the register all datums and transformation steps?
- The vertical correction grid does not cover the entire extent of Chart datum (EPSG:4615). Do the vertical correction grid and the vertical datum need to cover the same extent?   
- What other changes do we have to do at the above branches before pull requests?

rouault:
>     * So far, neither the national transformation from ETRS89 ell. to Chart datum, or intermediate vertical datum are defined in the EPSG registry. Do we need the register all datums and transformation steps?

Yes, please. That makes their use and maintenance by PROJ much easier

> The vertical correction grid does not cover the entire extent of Chart datum (EPSG:4615). Do the vertical correction grid and the vertical datum need to cover the same extent?

Not necessarily. Each transformation/grid has its own extent, which can be a subset of the source & target CRS

> What other changes do we have to do at the above branches before pull requests?

For PROJ-data, probably none. For PROJ, once the CRS & transformations are registered in EPSG, all changes but the ones in grid_alternatives.sql are no longer needed once we've imported the new version of the EPSG database

PR #2921, opened 2021-11-04
Database: update to EPSG v10.039

rouault:
@himsve In this EPSG update, the HREF2018B_NN2000_EUREF89.gtx and href2008a.gtx grid files have been modified to have a .bin extension; I presume the content is identical and this doesn't impact the GeoTIFF files we use in PROJ that were generated from the .gtx ?

I also see that there's a new EPSG: "ETRS89 to CD Norway depth (1)" transformation that uses a grid file. As far as I know this is the first such "geoid model" for depths and we don't handle that in PROJ yet. I presume this is mostly a matter of putting a minus sign something which would require some additions in the way we map EPSG transformation methods to PROJ pipelines. I didn't implement this in in this update, so this transformation isn't usable by PROJ currently


himsve:
Yes, the original grid files of HREF2018B_NN2000_EUREF89.gtx and href2008a.gtx were made in GravSoft bin format. But, the content of gtx and bin file are identical.

Just  for your information. The CD Norway implementation is related to this issue:

https://github.com/OSGeo/PROJ/issues/2739

Issue #2540, opened 2021-02-19
Some API functions are not handled correctly by Doxygen


kbevers:
Originally discovered in https://github.com/OSGeo/PROJ/pull/2507. If the function `proj_context_set_search_paths()` is added to the Sphinx/Breathe docs using a `doxygenfunction` directive like

```
.. doxygenfunction:: proj_context_set_search_paths
   :project: doxygen_api

```
 the doc generation fails:

```
$ make html
../scripts/doxygen.sh
No Doxygen warnings found
No Doxygen warnings found
touch .doxygen_up_to_date
sphinx-build -b html -d build/doctrees   source -W build/html
Running Sphinx v3.0.3
Initializing Spelling Checker
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 239 source files that are out of date
updating environment: [new config] 239 added, 0 changed, 0 removed
checking for /Users/kevers/dev/proj/repo/docs/source/references.bib in bibtex cache... not found
parsing bibtex file /Users/kevers/dev/proj/repo/docs/source/references.bib... parsed 41 entries


Warning, treated as error:
/Users/kevers/dev/proj/repo/docs/source/development/reference/functions.rst:787:doxygenfunction: Cannot find function "proj_context_set_search_paths" in doxygen xml output for project "doxygen_api" from directory: ../build/xml/
make: *** [html] Error 2
```

Inspecting the docygen XML output reveals that the file is at least partially recognized by Doxygen

```
$ grep proj_context_set_search_paths docs/build/xml/*
docs/build/xml/filemanager_8cpp.xml:        <definition>void proj_context_set_search_paths</definition>
docs/build/xml/filemanager_8cpp.xml:        <name>proj_context_set_search_paths</name>
docs/build/xml/filemanager_8cpp.xml:<codeline lineno="1940"><highlight class="normal"></highlight><highlight class="comment">/*<sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/>proj_context_set_search_paths()<sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/><sp/>*/</highlight><highlight class="normal"></highlight></codeline>
[...]

PR #2944, opened 2021-11-19
DOCS: Add doxygen entry for proj_context_set_search_paths.

direvus:
Also remove the setting LATEXT_SOURCE_CODE from Doxyfile, because
Doxygen was warning that the setting was obsolete.

Refs #2540.


kbevers:
> Doxygen was warning that the setting was obsolete.

These are two separate fixes, do you mind splitting them into to commits? One for the latex fix and one for the doxygen entry.

direvus:
No problem Kristian, I have updated this PR to remove the Latex setting change, and will submit that as a separate PR.

PR #2948, opened 2021-11-19
CMake: add option USE_CCACHE=OFF to use ccache to compile C/C++ objs

mwtoews:
This adds a bit of convenience for CMake to use [ccache](https://ccache.dev/). To enable it:

    cmake -DUSE_CCACHE=ON ..

This was based on https://crascit.com/2016/04/09/using-ccache-with-cmake/ with a few modifications. Tested on Linux with Ninja and Unix Makefile generators, but in theory also works with Xcode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,6 +122,12 @@ set(PROJ_BUILD_VERSION "${PROJ_API_VERSION}.${PROJ_LIBTOOL_AGE}.${PROJ_LIBTOOL_R
 ################################################################################
 # Build features and variants
 ################################################################################
+
+option(USE_CCACHE "Use ccache to compile C/C++ objects" OFF)
+if(USE_CCACHE)
+  include(Ccache)
+endif()
+
 include(ProjConfig)
 include(ProjMac)
 include(policies)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,11 +123,7 @@ set(PROJ_BUILD_VERSION "${PROJ_API_VERSION}.${PROJ_LIBTOOL_AGE}.${PROJ_LIBTOOL_R
 # Build features and variants
 ################################################################################
 
-option(USE_CCACHE "Use ccache to compile C/C++ objects" OFF)
-if(USE_CCACHE)
-  include(Ccache)
-endif()
-
+include(Ccache)
 include(ProjConfig)
 include(ProjMac)
 include(policies)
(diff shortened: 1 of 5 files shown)


rouault:
It seems to me that this script is mostly related to XCode not honouring CMAKE_C_COMPILER_LAUNCHER ?

On GDAL's Windows CMake builds with Conda, I use the clcache package and can add ``-DCMAKE_C_COMPILER_LAUNCHER=clcache -DCMAKE_CXX_COMPILER_LAUNCHER=clcache``. Windering if the new script could check for clcache if ccache isn't found, or we should just document that manual invokation ?

mwtoews:
> It seems to me that this script is mostly related to XCode not honouring CMAKE_C_COMPILER_LAUNCHER ?

Correct. Perhaps the script wrappers should only be applied for only that generator. Otherwise the CMAKE_C_COMPILER_LAUNCHER/CMAKE_CXX_COMPILER_LAUNCHER variables work fine without any script.

> On GDAL's Windows CMake builds with Conda, I use the clcache package and can add `-DCMAKE_C_COMPILER_LAUNCHER=clcache -DCMAKE_CXX_COMPILER_LAUNCHER=clcache`. Windering if the new script could check for clcache if ccache isn't found, or we should just document that manual invokation ?

I've never heard of clcache before, but I see it's the same concept for MSVC's cl.exe. I'll amend this suggestion when I get a chance, using the same `USE_CCACHE` option with documentation. I'll need to also expand a few CI tests too (which yes, I'm further re-writing for [RFC 7](https://proj.org/community/rfc/rfc-7.html)).

mwtoews:
Thanks for feedback @rouault, the last commits should address it. Anything further to add?

I'd like to hold off major rewrites to CI scripts for now, as I'm going thru them separately. Other changes may eventuate in this module when I get around to Xcode and adding a MSVC conda build (I'm planning to add something similar to "buid-windows-conda" [for gdal](https://github.com/OSGeo/gdal/blob/master/.github/workflows/cmake_builds.yml#L121)).

Note that the default for `CCACHE_CPP2` is true since [ccache version 3.3 from 2016-08-27](https://github.com/ccache/ccache/blob/master/doc/NEWS.adoc#ccache-33), so this environment variable is not set, and I'll remove these when I revise the CI scripts.

mwtoews:
Note the unrelated Docs / Plots error is an upstream issue with conda-forge; see https://github.com/conda-forge/fiona-feedstock/issues/186

A "fix" is ready (https://github.com/mwtoews/PROJ/commit/c683a8b262dadec5154bcae5345417f451116a39), however, I'd rather see it fixed upstream.

rouault (approved):

PR #2953, opened 2021-11-22
Refactor parameter handling and sql generation when ingesting projcs from ESRI

nyalldawson:
The intention here is twofold:

- Make the importing process more robust and to improve the readability of the importing script
- To make it easier to extend in future with support for additional methods (without having to duplicate a ton of code)

### DIFF
--- a/scripts/build_db_from_esri.py
+++ b/scripts/build_db_from_esri.py
@@ -33,10 +33,12 @@
 import argparse
 import csv
 import os
-import sqlite3
 import re
-import sys
+import sqlite3
+from dataclasses import dataclass
+from enum import Enum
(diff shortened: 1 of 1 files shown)


nyalldawson:
@rouault The biggest impact here is that I've utilised two newer Python language features -- the Enum and dataclass classes. I figured this would be safe to do seeing as this script is only ever run in a limited set of circumstances (i.e. we don't need to worry about the script being incompatible with older distributions and development environments)

rouault:
Nice cleanup. I've added an extra commit 7ebe8202 'build_db_from_esri.py: fix pyflakes3 warning, and make it python 3.8 compatible'

PR #2954, opened 2021-11-22
Add some more direct method mapping during ESRI projection engine ingestion

nyalldawson:
Adds direct mapping for Albers, IGAC_Plano_Cartesiano, Equidistant_Cylindrical and Cassini.


scripts/build_db_from_esri.py
```diff
@@ -1411,9 +1411,216 @@ def import_projcs():
...
+
+                    assert params['False_Easting'].unit.cs_auth_name == params[
+                        'False_Northing'].unit.cs_auth_name, 'Cannot handle False_Easting CS auth {} != False_Northing CS auth {}'.format(
+                        params['False_Easting'].unit.cs_auth_name, params['False_Northing'].unit.cs_auth_name)
+                    cs_auth_name = params['False_Easting'].unit.cs_auth_name
+
+                    assert params['False_Easting'].unit.cs_code == params[
+                        'False_Northing'].unit.cs_code, 'Cannot handle False_Easting CS code {} != False_Northing CS code {}'.format(
+                        params['False_Easting'].unit.cs_code, params['False_Northing'].unit.cs_code)
+                    cs_code = params['False_Easting'].unit.cs_code
```
rouault: couldn't those checks that are common to most methods be put in a function ?

scripts/build_db_from_esri.py
```diff
@@ -1411,9 +1411,216 @@ def import_projcs():
...
+                        conv_code = map_conversion_sql_to_code[sql_extract]
+
+                    sql = """INSERT INTO "projected_crs" VALUES('ESRI','%s','%s',NULL,'%s','%s','%s','%s','%s','%s',NULL,%d);""" % (
+                        code, esri_name, cs_auth_name, cs_code, geogcs_auth_name, geogcs_code, conv_auth_name,
+                        conv_code, deprecated)
+                    all_sql.append(sql)
+
+                    sql = """INSERT INTO "usage" VALUES('ESRI', 'PCRS_%s_USAGE','projected_crs','ESRI','%s','%s','%s','%s','%s');""" % (
+                        code, code, extent_auth_name, extent_code, 'EPSG', '1024')
+                    all_sql.append(sql)
```
rouault: that part also seems to be common to most methods. would be nice to have it in a function

nyalldawson: Ok, I've just refactored out a bunch more duplicate code...

nyalldawson:
@rouault 
it's unrelated to this PR (just moved around), but there's an error somewhere here:
https://github.com/OSGeo/PROJ/pull/2954/files#diff-7c1ab90cc254afe37c4c9c5dc007a1e5295e1396d7e4aef4bfda3ecc744ab757R1156

At that line `conv_name` will ALWAYS be 'unnamed'. Do you know what the original intention was here?

(here's where that code used to live: https://github.com/OSGeo/PROJ/blob/bd251fa2b1fe929f0070dfdb80d41afb3359b974/scripts/build_db_from_esri.py#L960 and https://github.com/OSGeo/PROJ/blob/bd251fa2b1fe929f0070dfdb80d41afb3359b974/scripts/build_db_from_esri.py#L968 )

nyalldawson:
@rouault 

in the latest commit I've added a couple of new mappings, which now is done with just a few extra entries in the mapping dictionary. Looking at this, I think we could further simplify these scripts and reduce maintainance by moving the whole yaml mapping config string out of build_esri_projection_mapping.py into a standalone yaml file, and then using THAT in the build_db_from_esri.py script. Ultimately we'd end up with one definitive place in the source to setup these mappings....

Interested in your thoughts though!

rouault:
> At that line `conv_name` will ALWAYS be 'unnamed'. Do you know what the original intention was here?

Ah, the comparison of conv_name against unnamed made only sense in the first case for Transverse Mercator / Gauss Krueger, where conv_name could be set to something else at line https://github.com/OSGeo/PROJ/blob/bd251fa2b1fe929f0070dfdb80d41afb3359b974/scripts/build_db_from_esri.py#L909 . For other projection methods, this is just a left over of copy&paste

scripts/build_db_from_esri.py
```diff
@@ -1072,7 +1072,38 @@ class ConversionMapping:
...
+        epsg_name='Mercator (variant B)',
+        param_mapping={
+            8823: 'Standard_Parallel_1',
+            8802: 'Central_Meridian',
+            8806: 'False_Easting',
+            8807: 'False_Northing',
+        }
+    ),
+    'Equidistant_Conic': ConversionMapping(
+        epsg_code='9805',
```
rouault: Why mapping Equidistant_Conic to Mercator variant B ? Doesn't look appropriate according to https://github.com/Esri/projection-engine-db-doc/blob/master/other/proj4-esri.txt#L182

nyalldawson: Hmmmm, not sure how that snuck in! 😱

rouault:
> Looking at this, I think we could further simplify these scripts and reduce maintainance by moving the whole yaml mapping config string out of build_esri_projection_mapping.py into a standalone yaml file, and then using THAT in the build_db_from_esri.py script.

That would be nice to have!

nyalldawson:
@rouault 
have you got any thoughts on the failing test here?

```
Expected equality of these values:
  crs->exportToWKT( WKTFormatter::create(WKTFormatter::Convention::WKT1_ESRI, dbContext) .get())
    Which is: "PROJCS[\"WGS_1984_Web_Mercator\",GEOGCS[\"GCS_WGS_1984\",DATUM[\"D_WGS_1984\",SPHEROID[\"WGS_1984\",6378137.0,298.257223563]],PRIMEM[\"Greenwich\",0.0],UNIT[\"Degree\",0.0174532925199433]],PROJECTION[\"Mercator_Auxiliary_Sphere\"],PARAMETER[\"False_Easting\",0.0],PARAMETER[\"False_Northing\",0.0],PARAMETER[\"Central_Meridian\",0.0],PARAMETER[\"Standard_Parallel_1\",0.0],PARAMETER[\"Auxiliary_Sphere_Type\",0.0],UNIT[\"Meter\",1.0]]"
  esri_wkt
    Which is: "PROJCS[\"WGS_1984_Web_Mercator\",GEOGCS[\"GCS_WGS_1984_Major_Auxiliary_Sphere\",DATUM[\"D_WGS_1984_Major_Auxiliary_Sphere\",SPHEROID[\"WGS_1984_Major_Auxiliary_Sphere\",6378137.0,0.0]],PRIMEM[\"Greenwich\",0.0],UNIT[\"Degree\",0.0174532925199433]],PROJECTION[\"Mercator\"],PARAMETER[\"False_Easting\",0.0],PARAMETER[\"False_Northing\",0.0],PARAMETER[\"Central_Meridian\",0.0],PARAMETER[\"Standard_Parallel_1\",0.0],UNIT[\"Meter\",1.0]]"
[  FAILED  ] crs.projectedCRS_from_WKT1_ESRI_as_WKT1_ESRI (26 ms)
```

rouault:
> have you got any thoughts on the failing test here?

tricky... So the WKT1_ESRI definition of this test is the one of the deprecated EPSG:3785 / ESRI:102113 which used regular Mercator on a sphere whose radius was the semi-major axis of WGS 84. This was the inital attempt to try to model Web Mercator. Mostly worked except for datum transformations since the WGS 84 aspect was lost, and also for computations of projection factors.
When ingesting WKT, we have this particular case at https://github.com/OSGeo/PROJ/blob/master/src/iso19111/io.cpp#L4073 we we forcibly use the correct way of defining WebMercator.
So why on export did the roundtrip work ? Just because we have in the projected_crs table the original WKT definition, and ProjectedCRS::exportToWkt() will use it in priority when exporting a ESRI CRS to WKT1_ESRI.
Now that you import ESRI:102113 as standard Mercator, this original WKT definition is not there, so PROJ does its generic export logic, which maps the METHOD_POPULAR_VISUALISATION_PSEUDO_MERCATOR to Mercator_Auxiliary_Sphere, which results in a WKT that is similar to the one of the non-deprecated EPSG:3857 / ESRI:102110, but with the CRS name of EPSG:3785 / ESRI:102113
I would suggest to add an exception in the import script for ESRI:102113 to just keep the original WKT definition.
(We could argue also that the special case for "WGS_1984_Web_Mercator" in https://github.com/OSGeo/PROJ/blob/master/src/iso19111/io.cpp#L4073 is not desirable and we should just use that 'wrong' definition, but I would be nervous changing that)

nyalldawson:
@rouault This should be good to merge now -- I'll make the yaml related improvements in a follow up PR

PR #2959, opened 2021-11-25
DOC: rewrite parts of plot.py to use geojson instead of fiona

mwtoews:
This rewrite replaces fiona with geojson, which is simpler to install, and fixes a current regression from conda-forge with fiona and matplotlib. It should hopefully be a bit more future-proof, and easier to use.

The performance in using geojson instead of fiona is about the same.

Shapely version 1.7 is preferred with descartes, otherwise many warnings are generated.


rouault (approved):
LGTM. Side remark unrelated to your pull request, but as we use the proj from Conda to draw the plots, that couldn't work if we add a new projection method (I'm not sure if the static images in git in docs/source/operations/projections/images would be used). We should probably build PROJ from sources instead of relying of the one from Conda.

mwtoews:
:100: agree about building PROJ to generate plots instead of conda-forge; #2961

PR #2970, opened 2021-12-04
Refactor proj.pc to use EXTRA_LIBS; add -lpthread to Libs.private

mwtoews:
This primarily fixes a static linking issue with GCC 4.8 on a 3rd party C app:
```
$ gcc-4.8 -I/tmp/proj_autoconf_install_from_dist_all/include -Werror    -o c_app c_app-c_app.o -L/tmp/proj_autoconf_install_from_dist_all/lib -Wl,-Bstatic -lproj -Wl,-Bdynamic -lsqlite3 -ltiff -lcurl -lstdc++ -lm -ldl
/tmp/proj_autoconf_install_from_dist_all/lib/libproj.a(factory.o): In function `osgeo::proj::io::SQLiteHandleCache::getHandle(std::string const&, pj_ctx*)':
factory.cpp:(.text+0x25aa): undefined reference to `pthread_atfork'
```
this doesn't happen with other compilers, as I suppose there are some implicitly linked libraries. The `proj.pc` file for linux should normally have `Libs.private: -lstdc++ -lsqlite3 -lpthread -ltiff -lcurl -lm -ldl`

Also, this fix is separate to a post-install refactor that I'm preparing [here](https://github.com/mwtoews/PROJ/tree/postinstall2), which could be ready for a PR in the next week. Backporting to 8.2 to preserve the autotools changes somehow, as these will be dropped at some point.

### DIFF
--- a/cmake/ProjUtilities.cmake
+++ b/cmake/ProjUtilities.cmake
@@ -76,21 +76,25 @@ function(configure_proj_pc)
   set(datadir "$\{datarootdir\}")
   set(PACKAGE "proj")
   set(VERSION ${PROJ_VERSION})
-  set(SQLITE3_LIBS -lsqlite3)
+  # Build list for Libs.private
+  set(EXTRA_LIBS
+    -lstdc++
+    -lsqlite3
+    ${CMAKE_THREAD_LIBS_INIT}
+  )
   if(TIFF_ENABLED)
-    set(TIFF_LIBS -ltiff)
+    list(APPEND EXTRA_LIBS -ltiff)
   endif()
   if(CURL_ENABLED)
-    set(CURL_LIBS -lcurl)
+    list(APPEND EXTRA_LIBS -lcurl)
   endif()
-  set(EXTRA_LIBS "-lstdc++")
   if(HAVE_LIBM)
-    list(APPEND EXTRA_LIBS "-lm")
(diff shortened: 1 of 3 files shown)


dg0yt:
I just noticed this change, and I would like point out that this is insufficient to fix static linking. tiff and curl usually come with additional static dependencies. Things are even more complicated when the library name is different (debug suffix). I can submit the changes I did in a patch in vcpkg: Basically, it means to construct a `Requires.private: sqlite3 libtiff-4 libcurl` line to leverage these pkg-config modules. I think it is fair to assume the presence of these modules in a sane pkg-config context.

I would also submit changes to make the pc file usable on windows. In vcpkg, I use this data to feed proj dependencies into nmake builds...

Another change is to look for `APPLE` and `CMAKE_ANDROID_STL_TYPE` to choose between `-lc++` and `-lstdc++`.

Reference (PROJ 8.1.1): https://github.com/microsoft/vcpkg/blob/master/ports/proj4/pkgconfig.patch

mwtoews:
@dg0yt that'd be welcome to see. This PR was required before #2972 which includes static linking tests with just libproj. Static linking of dependant libraries becomes increasingly challenging to do. This is why on Linux `-Wl,-Bstatic -lproj -Wl,-Bdynamic` is used to allow dynamic linking of other components.

PR #2972, opened 2021-12-06
Refactor post-install suite to test shared and static projlib

mwtoews:
This is a re-write of the post-install tests, which now tests the following combinations:

- Library type: static, shared
- Build tools: CMake, Makefile (w/ pkg-config), Autotools (w/ pkg-config)

See README.md for further details.


kbevers:
Very impressive, Mike! Thanks for putting in the time to make this happen.

appveyor.yml
```diff
@@ -9,23 +9,26 @@ environment:
   - platform: x86
     APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
     BUILD_SHARED_LIBS: OFF
+    CMAKE_GENERATOR: Ninja
 
 # VS 2017
-#  - platform: x64
-#    APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-#    BUILD_SHARED_LIBS: ON
+  - platform: x64
```
rouault: Appveyor tends to be a contention point since we have just one single concurrent instance shared by OSGeo projects. We should limit the work done on it. Basically the only point of it compared to github actions is that it has VS2015. If we have VS2015 in appveyor and VS2019 in github actoins, we can hope for VS2017 to be OK

test/postinstall/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Each configure method is driven using a script with two arguments:
+1. Path to a PROJ install prefix (containing `bin`, `lib`, etc.)
+2. An optional build mode `shared` (default) or `static`
+
+### CMake
+
+Use `test_cmake.sh` for Unix-like or `test_cmake.bat` for Windows. This checks that `find_package(PROJ)` (or legacy `find_package(PROJ4)`) can be used with `target_link_libraries(mytarget PRIVATE PROJ::proj)`. The C-based app enables C++ when linking to static libproj. These scripts will only work with PROJ built using CMake, and will not work with PROJ built using Autotools.
+
+### Makefile / pkg-config
```
rouault, line 19: What does this configuration brings compared to autotools ?

appveyor.yml
mwtoews: Good to know, I'll limit this to one appveyor job and try to leverage more github actions.

test/postinstall/README.md
mwtoews: A few reasons: it's simpler to see what's going on, and can be run in test environments without automake/conf. The test scripts are shared between the two, so there isn't too much code overlap.

I'll check again to see if these two tests yields different results with macOS. There were are few gotchas with rpaths, which I'll explain in more detail later.

rouault: My feeling is that automake/conf should be sufficient. I'm not sure what environments do not have autotools that we'd want to support. The main thing we want to check is that the paths in pkg-config are OK, whatever hand-made makefile vs autotools consume it, don't we ?

mwtoews: Sure, these tests could probably be dropped, as they would provide identical coverage. I think all of the test environments currently have automake/conf.

mwtoews: Oh right, it was MSYS2 (in windows.yml) that only had make. Perhaps keep this test script, but don't duplicate them where autotools is available? I'll tack on a second commit to show this. The alternative is to install packages for MSYS2.

rouault (approved):

Issue #2961, opened 2021-11-26
Task: Improve CI "Plots" job to generate images from PROJ sources


mwtoews:
Idea from #2959 is to improve the "Plots" CI job to first build PROJ, and uses these fresh executables with `docs/plot/plot.py`. This would remove the `proj` dependency from `docs/plot/environment.yml`, and allow the CI to test generation of images based on source code changes to projections, rather than depending on an older version of PROJ from conda-forge.

PR #2978, opened 2021-12-09
Fix and additional options for Peirce Quincuncial projections

tcwilkinson:
This fixes the current forward implementation of Peirce Quincuncial proj to correctly flip/reflect out the southern hemisphere to four triangles, and rotate entire result to a square or diamond. (It there resolves the issues identified with pull request https://github.com/OSGeo/PROJ/pull/2230 , where southern hemisphere was wrongly projected over northern, and reverses the restriction to northern hemisphere introduced there). It also adds additional lateral projection of the hemispheres.

- This PR adds an optional parameter `+type` which allows selection of projection. The `+type=square` and `+type=diamond` types match in principle ESRI's twin implementations of square and diamond PQ projs. The **default** if not specified is `+type=diamond`.

- The previous behaviour restricted to the northern hemisphere can be reproduced using the `+type=nhemisphere`, though this is an edge case only.

- An additional `+type=horizontal` and `+type=vertical` rectangular lateral versions have been added that place each hemisphere side-by-side. This is primarily to allow creation of projections such as Greiger Triptychial, which also require the additional optional params `o_scrollx` or `o_scrolly` in order to shift parts of the projection from one side of the map to the other. 

- Additional documentation has been added to proj description, including quoting the usual meridian used in common usage of projection, and images showing the different types.

Would be grateful for review and any suggestions especially from @kbevers @rouault who are familier with these implementations.

 - [x] Tests added


src/projections/adams.cpp
```diff
@@ -33,7 +38,8 @@
 #include "proj_internal.h"
 
 PROJ_HEAD(guyou, "Guyou") "\n\tMisc Sph No inv";
-PROJ_HEAD(peirce_q, "Peirce Quincuncial") "\n\tMisc Sph No inv";
+PROJ_HEAD(peirce_q, "Peirce Quincuncial (Square)") "\n\tMisc Sph No inv";
+PROJ_HEAD(peirce_q_d, "Peirce Quincuncial (Diamond)") "\n\tMisc Sph No inv";
```
rouault: I'm wondering if we need a new projection method when the difference is only the final rotation. A possiblity would be to have a xy_rot= parameter, like in src/projections/healpix.cpp  .     peirce_q would be obtained with xy_rot=45 and peirce_q_d with xy_rot=0. The question would be what should be the default for perice_q : 45 to have backward compatibility or 0 if it makes more sense as a default ? Changing to 0 could probably be acceptable for PROJ 9, because this projection method is rather new

src/projections/adams.cpp
```diff
@@ -173,7 +178,21 @@ static PJ_XY adams_forward(PJ_LP lp, PJ *P) {
     xy.x = ell_int_5(m);
     xy.y = ell_int_5(n);
 
-    if (Q->mode == ADAMS_HEMI || Q->mode == ADAMS_WS2) { /* rotate by 45deg. */
+    if (Q->mode == PEIRCE_Q || Q->mode == PEIRCE_Q_D) {
+      /* For Quincuncial projections, spin out southern hemisphere to triangular segments of quincunx */
+      if (true && (lp.phi < 0.)) {
```
rouault:
```suggestion
      if (lp.phi < 0.) {
```

src/projections/adams.cpp
```diff
@@ -173,7 +178,21 @@ static PJ_XY adams_forward(PJ_LP lp, PJ *P) {
...
+        /* Constant complete elliptic integral of the first kind with m=0.5 https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.ellipk.html as shift distance */
+        double shd = 1.8540746773013719 * 2;
```
rouault:
```suggestion
        constexpr double shd = 1.8540746773013719 * 2;
```

docs/source/operations/projections/peirce_q_d.rst
```diff
@@ -0,0 +1,54 @@
+.. _peirce_q_d:
```
rouault: if keeping a new projection method, it needs to be referenced in docs/source/operations/projections/index.rst

docs/source/operations/projections/peirce_q_d.rst
```diff
@@ -0,0 +1,54 @@
...
++---------------------+----------------------------------------------------------+
+| **Domain**          | 2D                                                       |
++---------------------+----------------------------------------------------------+
+| **Input type**      | Geodetic coordinates                                     |
++---------------------+----------------------------------------------------------+
+| **Output type**     | Projected coordinates                                    |
++---------------------+----------------------------------------------------------+
+
+
+.. figure:: ./images/peirce_q_d.png
```
rouault: if keeping a new projection method, the corresponding file, generated with /docs/plot/plot.py should be added to ./docs/source/operations/projections/images

docs/plot/plotdefs.json
```diff
@@ -1053,11 +1053,22 @@
...
         "lonmax": 180,
         "lonmin": -180,
         "name": "peirce_q",
-        "projstring": "+proj=peirce_q",
+        "projstring": "+proj=peirce_q +lon_0=25",
+        "res": "low",
+        "type": "poly"
+    },
+    {
+        "filename": "peirce_q_d.png",
```
rouault: note: adding a new projection method makes the CI "Plots" job currently fail because of https://github.com/OSGeo/PROJ/issues/2961 . I'll try to look at solving it

rouault (commented):
Tests in ./test/gie/peirce_q.gie (they are run by ctest typically) will need to be adjusted for the southern hemisphere. Note: IMHO there are far more test points than needed. You could likely reduce to just what is needed to test each code branch.

src/projections/adams.cpp
tcwilkinson: Done. Orphaned swtich from testing code.

src/projections/adams.cpp
tcwilkinson: Done. Thanks!

docs/source/operations/projections/peirce_q_d.rst
tcwilkinson: OK, good to know. Have now pre-generated this and included.

docs/source/operations/projections/peirce_q_d.rst
tcwilkinson: done

src/projections/adams.cpp
tcwilkinson: Thank very much for the review + input.  

Can see the argument for using a xy_rot value rather than separate projection if this is better general proj practice, though I'm not exactly sure how to implement that in code.

Alternative fpr code simplicity would be to use a simple option param like ESRI (e.g. 0=square, 1=diamond) rather than angle, to avoid arbitrary angles.

Would strongly vote for the default being the quincunx square (peirce_q / xy_rot=45) or diamond (peirce_q_d / xy_rot=0). This is technically not backward compatible with the proj implementation until this fix (which only produced the inner square of the full diamond, so is xy_rot=0), but a square is surely most likely output wanted.

tcwilkinson: One additional thought on the question of `xy_rot=45/0` vs `type=0/1`. 

I was planning to build on this code to add the lateral quincuncial projection (i.e. hemispheres plotted side-by-side as rectangle). I had originally thought of making it a seperate projection with options for sliding -- to allow all continents to fit.  

But following discussion above, would this be better indicated as param in a single peirce_q proj? e.g. `type=2`?

rouault: type= would also be fine for me. I'd prefer string values like square, diamond, etc. for clarity. An example of reading a string valued parameter is at https://github.com/OSGeo/PROJ/blob/master/src/transformations/helmert.cpp#L527 . You would want to do that in the setup() function

docs/plot/plotdefs.json
rouault: Solved per https://github.com/OSGeo/PROJ/pull/2981 (you can rebase on top of latest master)

docs/plot/environment.yml
```diff
@@ -5,5 +5,4 @@ dependencies:
   - descartes
   - geojson
   - matplotlib
-  - proj
```
rouault, line 8: that change alone will not bring https://github.com/OSGeo/PROJ/pull/2981
you should rather have done:
```
git fetch origin master
git rebase origin/master
```

tcwilkinson: Thanks. Having git issues, but hopefully solved now.

tcwilkinson:
Have set the default type to `+type=nhemisphere` for backward compatibility with implementation until now, but happy to change now if you think better going forward.

rouault:
> Have set the default type to `+type=nhemisphere` for backward compatibility with implementation until now, but happy to change now if you think better going forward.

I don't think restricting to nhemisphere makes sense at all. This option shouldn't be available

tcwilkinson:
> > Have set the default type to `+type=nhemisphere` for backward compatibility with implementation until now, but happy to change now if you think better going forward.
> 
> I don't think restricting to nhemisphere makes sense at all. This option shouldn't be available

Can think of mapping applications where this and shemisphere would make more sense than convoluted alternatives (e.g. diamond with points above lat>0 will produce a square, but lat<0 will produce the four triangular spokes only).

But agree that these are edge cases and doesn't make sense as default even if backward compatible. 

Is `+type=diamond` is better as default (since it is backward compatible with coords for implementation until now, but shows full range)?

docs/source/operations/projections/peirce_q.rst
```diff
@@ -21,12 +39,33 @@ Peirce Quincuncial
...
+
+   proj-string: ``+proj=peirce_q +lon_0=25 +type=horizontal``
+
+   .. figure:: ./images/grieger_triptychial.png
+      :width: 500 px
+      :align: center
+      :alt:   Grieger Triptychial
 
-   proj-string: ``+proj=peirce_q``
+      proj-string: ``+proj=pipeline +step +proj=ob_tran +o_proj=peirce_q +o_lat_p=-45 +o_lon_p=45 +type=horizontal +o_scrollx=-0.25 +step +proj=affine +s11=-1 +s12=0 +s21=0 +s22=-1``
```
rouault: +type should be +o_type, otherwise it won't be forwarded by ob_tran
And the +o_scrollx will be translated by ob_tran as +scrollx

docs/source/operations/projections/peirce_q.rst
```diff
@@ -35,6 +74,40 @@ Parameters
...
+
+    By contrast, the ``horizontal`` and ``vertical`` forms reflect the southern hemisphere
+    laterally across the x or y axis respectively to produce a rectangular form. The origin
+    lies at the centre of the rectangle.
+
+    The other two types, ``nhemisphere`` and ``shemisphere``, discard latitudes of less
+    than 0 or more than 0, respectively, to allow single hemispheres to be selected.
+    The origin lies at the centre of the square or diamond.
+
+.. option:: +o_scrollx=<value>
```
rouault: Parameters should't start with o_  . o_ is a prefix used by +proj=ob_tran to know that the parameters must be forwarded to the projection identified by +o_proj without the o_ prefix

src/projections/adams.cpp
```diff
@@ -90,6 +114,22 @@ static double ell_int_5(double phi) {
...
+//      y = x + 90;
+//    }
+//    return y;
+//     // if ((x >= -M_HALFPI) && (x <= M_HALFPI)) return x;
+//     // x = fmod(x + M_HALFPI,M_PI);
+//     // if (x > 0)
+//     //     x -= M_PI;
+//     // return x + M_HALFPI;
+// }
+
```
rouault:
```suggestion
```

test/gie/peirce_q.gie
```diff
@@ -5,14 +5,950 @@
...
-operation  +proj=peirce_q +R=6370997
+operation  +proj=peirce_q +R=6370997 +type=diamond
 tolerance  1 mm
 ------------------------------------------------------------
 accept    -179.6126302052  -90.2440064745
 expect    failure errno coord_transfm_invalid_coord
 
 accept    -169.8486692749  -89.5742075028
-expect    failure errno coord_transfm_outside_projection_domain
+expect         -8344.698       23578003.421
```
rouault: I'm not sure we need that many test points, but we'd need tests for the other +type and new parameters

docs/source/operations/projections/peirce_q.rst
tcwilkinson: That's not how it's working for me: type is being passed directly to peirce_q. If I try o_type it fails to pass on the param. (Same now that I have changed o_scrollx to scrollx).

docs/source/operations/projections/peirce_q.rst
tcwilkinson: OK, have changed o_scrollx to scrollx in code+docs, but see note above that o_scrollx does not get passed as scrollx to peirce_q via ob_tran

docs/plot/plotdefs.json
```diff
@@ -1051,13 +1051,46 @@
...
+        "type": "line"
+    },
+    {
+        "filename": "grieger_triptychial.png",
+        "latmax": 90,
+        "latmin": -90,
+        "lonmax": 180,
+        "lonmin": -180,
+        "name": "grieger_triptychial",
+        "projstring": "+proj=pipeline +step +proj=ob_tran +o_proj=peirce_q +o_lat_p=-45 +o_lon_p=45 +type=horizontal +scrollx=-0.25 +step +proj=affine +s11=-1 +s12=0 +s21=0 +s22=-1",
```
rouault, line 1093: type and scrollx do need to be prefix with o_ here, as it is in a proj=ob_tran context

tcwilkinson: but as note above, using o_type + o_scrollx fails for me; they are not passed on.

docs/source/operations/projections/peirce_q.rst
rouault: > That's not how it's working for me: type is being passed directly to peirce_q. If I try o_type it fails to pass on the param. (Same now that I have changed o_scrollx to scrollx).

oh, I'm sorry: I was confused. So checking at how ob_tran actually works, it will forward "type", "scrollx", "scrolly" unmodified. Ignore contradicting statements I can have put regarding this.

docs/plot/plotdefs.json
rouault: ignore my above comment about prefixing with o_ : that was wrong

docs/source/operations/projections/peirce_q.rst
```diff
@@ -21,12 +38,33 @@ Peirce Quincuncial
...
+   proj-string: ``+proj=peirce_q +lon_0=25 +type=horizontal``
+
+.. figure:: ./images/grieger_triptychial.png
    :width: 500 px
    :align: center
-   :alt:   Peirce Quincuncial
+   :alt:   Grieger Triptychial
 
-   proj-string: ``+proj=peirce_q``
+   proj-string: ``+proj=pipeline +step +proj=ob_tran +o_proj=peirce_q +o_lat_p=-45 +o_lon_p=45 +o_type=horizontal +o_scrollx=-0.25 +step +proj=affine +s11=-1 +s12=0 +s21=0 +s22=-1``
```
rouault, line 67: so we do need to remove the o_ prefixes here:
```suggestion
   proj-string: ``+proj=pipeline +step +proj=ob_tran +o_proj=peirce_q +o_lat_p=-45 +o_lon_p=45 +type=horizontal +scrollx=-0.25 +step +proj=affine +s11=-1 +s12=0 +s21=0 +s22=-1``
```

tcwilkinson: I think this should remain as is: o_type and o_scrollx do not get passed by ob_tran to peirce_q, but plain type and scrollx do . Otherwise fix needed in ob_tran?

rouault: > I think this should remain as is: o_type and o_scrollx do not get passed by ob_tran to peirce_q, but plain type and scrollx do . Otherwise fix needed in ob_tran?

I believe we agree, but your sentence contradicts my suggested fix. My suggested fix consists in removing the o_ prefix of o_type and o_scrollx.

docs/source/operations/projections/peirce_q.rst
tcwilkinson: ok! no problem!

test/gie/peirce_q.gie
tcwilkinson: have reduced number of test points. probably could be rationalised further, though for now have carried over a selection from the points already included

docs/source/operations/projections/peirce_q.rst
tcwilkinson: yep, i we agree. i think all is as it should be here

tcwilkinson:
Checks currently stopped in Travis CI - Pull Request: https://github.com/OSGeo/PROJ/pull/2978/checks?check_run_id=4535751198

```
Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
The command "./travis/${BUILD_NAME}/before_install.sh" failed and exited with 1 during .
```

Presume this is temporary issue, and nothing I can do about it from my end.

src/projections/adams.cpp
tcwilkinson: implemented in later commit

docs/source/operations/projections/peirce_q.rst
rhuijben: The original proj-string `+proj=peirce_q` or more specifically `+proj=peirce_q +type=crs` now returns an error, while it didn't do this before this patch.

PR #2979, opened 2021-12-10
Fix unreachable code in projections/isea.

### LINKED ISSUE #423: Invalid return 1; in PJ_isea.c:isea_hex

direvus:
The internal helper function isea_hex() had an early unconditional
return, which made most of the code in the function totally unreachable.

Based on the comments in the unreachable section, I speculate that the
original author meant for this early return to occur for aperture values
other than 3 or 4.

In passing, change the function's return type to void, since it was
always returning 1 anyway and the return was not used (per #947), and
clean up some rogue whitespace.

Fixes #423.

 - [x] Closes #423 

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -108,9 +108,6 @@
    directories, so that proj can be used as a subdirectory of a larger
    project (#2913)
 
- o Fix unreachable internal code in Icosahedral Snyder Equal Area projection
-   (#423)
-
 
  THANKS TO
  ---------
--- a/NEWS
+++ b/NEWS
@@ -108,6 +108,9 @@
    directories, so that proj can be used as a subdirectory of a larger
    project (#2913)
 
+ o Fix unreachable internal code in Icosahedral Snyder Equal Area projection
+   (#423)
+
 
  THANKS TO
  ---------
--- a/src/projections/isea.cpp
+++ b/src/projections/isea.cpp
@@ -887,26 +887,25 @@ static long isea_disn(struct isea_dgg *g, int quadz, struct isea_pt *di) {
  * d' = d << 4 + q, d = d' >> 4, q = d' & 0xf
  */
 /* convert a q2di to global hex coord */
-static int isea_hex(struct isea_dgg *g, int tri,
+static void isea_hex(struct isea_dgg *g, int tri,
                     struct isea_pt *pt, struct isea_pt *hex) {
     struct isea_pt v;
-#ifdef FIXME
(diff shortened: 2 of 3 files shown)


NEWS
```diff
@@ -108,6 +108,9 @@
    directories, so that proj can be used as a subdirectory of a larger
    project (#2913)
 
+ o Fix unreachable internal code in Icosahedral Snyder Equal Area projection
+   (#423)
+
```
kbevers: Please leave this bit out. We only update news at release time.

direvus: Sure thing, sorry about that.  I've also been sending PRs to a different project that requires you to update NEWS yourself, so I just did it out of habit.

rouault:
it would be nice to have test/gie/builtins.gie extended to test the fixed cases

src/projections/isea.cpp
```diff
@@ -949,8 +948,7 @@ static int isea_hex(struct isea_dgg *g, int tri,
         hex->y = i + sidelength * ((g->quad-1) % 5);
     }
 
-    return 1;
-#endif
+    return;
```
rhuijben: Not needed any more.

direvus: Sorry, I don't follow.  What's not needed any more exactly?  If we put the `#endif` back in, it will not be valid because its matching `#ifdef` has been removed.

rouault: I believe @rhuijben meant that the explicit "return;" does nothing at the end method and can be safely removed

direvus: Thanks for clearing that up @rouault.  I have removed the 'return' as suggested.

rouault:
> it would be nice to have test/gie/builtins.gie extended to test the fixed cases

this point remains to be done

direvus:
> > it would be nice to have test/gie/builtins.gie extended to test the fixed cases
> 
> this point remains to be done

I have added some tests on my branch that should exercise those code paths.  I haven't independently confirmed the actual projection values are correct.

PR #2980, opened 2021-12-10
CMake: fix warning with external googletest

rouault:
Recent CMake versions throw the following warning when building
googletest 1.8.1. Bumping to 1.11.0 that has a minimum version of 2.8.12
fixes that
```
 CMake Deprecation Warning at build/googletest-src/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.
```

### DIFF
--- a/.github/workflows/linux_gcc_5_4.yml
+++ b/.github/workflows/linux_gcc_5_4.yml
@@ -1,10 +1,10 @@
-name: Linux GCC 4.8
+name: Linux GCC 5.4
 
 on: [push, pull_request]
 
 jobs:
 
-  linux_gcc_4_8:
+  linux_gcc_5_4:
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]')"
     steps:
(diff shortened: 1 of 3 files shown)


rouault:
bummer: gcc 4.8 is not able to build googletest 1.11.0

kbevers:
> bummer: gcc 4.8 is not able to build googletest 1.11.0

Maybe it's time to move on? Most recent GCC 4.8 version is from 2015

rouault:
> Maybe it's time to move on? Most recent GCC 4.8 version is from 2015

yeah, I'm changing this conf to use the default GCC 5.4 from Ubuntu 16.04

mwtoews:
If we really want to keep GCC 4.8 in the mix, another strategy is to edit test/googletest/CMakeLists.txt.in with something like:
```
if(CMAKE_VERSION VERSION_LESS 3.???)
  # GCC 4.8 compat
else()
  # more recent compat
endif()
```

rouault:
> If we really want to keep GCC 4.8 in the mix, another strategy is to edit test/googletest/CMakeLists.txt.in with something like:

I don't have any personal interest in such old versions.
I don't see the relationship between the CMake version and the GCC version. A system with gcc 4.8 has normally not CMake >= 3.9 by default, so they should use a backport version, which can be anything >= 3.9

mwtoews (approved):
Yup, lets move on from [GCC 4.8](https://gcc.gnu.org/gcc-4.8/). While interesting to see it raise different issues in the CI compared to newer compilers, it's really a bit of a distraction. [GCC 5.4 is from 2016](https://gcc.gnu.org/gcc-5/), so a bit newer. Nevertheless, I'd like to see the CMake warning go away with the upgraded googletest.

PR #2988, opened 2021-12-19
Use `Requires.Private` in proj.pc, and install it also for Windows

dg0yt:
Use sqlite3, libtiff-4 and libcurl modules via `Requires.private` instead of plain libs in `Libs.private`.
(Opt-out via `USE_PKGCONFIG_REQUIRES`.)
Install pc file for native Windows.
Determine specific c++ libs from implicit link libraries.
Adapt CI to the transitive usage requirements in pkg-config.

References: 
https://github.com/OSGeo/PROJ/pull/2970#issuecomment-995977687, 
https://github.com/microsoft/vcpkg/blob/a4f322c5672deb4ece27c88c7e85f271a286a220/ports/proj4/pkgconfig.patch


dg0yt:
Failing, but illustrating transitive dependencies:
~~~
 ccache gcc-4.8 -I/tmp/proj_autoconf_install_from_dist_all/include -I/usr/include/x86_64-linux-gnu -Werror    -o c_app c_app-c_app.o -L/tmp/proj_autoconf_install_from_dist_all/lib -L/usr/lib/x86_64-linux-gnu/mit-krb5 -Wl,-Bstatic -lproj -Wl,-Bdynamic -lstdc++ -lpthread -lm -ldl -lsqlite3 -ldl -lpthread -ltiff -llzma -ljbig -ljpeg -lz -lm -lcurl -lidn -lrtmp -lssl -lcrypto -lssl -lcrypto -Wl,-Bsymbolic-functions -Wl,-z,relro -lgssapi_krb5 -lkrb5 -lk5crypto -lcom_err -llber -lldap -lz 
/usr/bin/ld: cannot find -lidn
/usr/bin/ld: cannot find -lrtmp
/usr/bin/ld: cannot find -lssl
/usr/bin/ld: cannot find -lcrypto
/usr/bin/ld: cannot find -lssl
/usr/bin/ld: cannot find -lcrypto
/usr/bin/ld: cannot find -lgssapi_krb5
/usr/bin/ld: cannot find -lkrb5
/usr/bin/ld: cannot find -lk5crypto
/usr/bin/ld: cannot find -lcom_err
/usr/bin/ld: cannot find -llber
/usr/bin/ld: cannot find -lldap
~~~
This seems to be what the pkgconfig modules tell about the libs:
~~~
sqlite3: -lsqlite3 -ldl -lpthread 
libtiff-4: -ltiff -llzma -ljbig -ljpeg -lz -lm 
libcurl: -lcurl -lidn -lrtmp -lssl -lcrypto -lssl -lcrypto -Wl,-Bsymbolic-functions -Wl,-z,relro -lgssapi_krb5 -lkrb5 -lk5crypto -lcom_err -llber -lldap -lz
~~~
(Does PROJ use dl, or is it just a transitive dependency?)

mwtoews:
FYI, the one failure with GCC-4.8 may become irrelevant if/when #2980 is merged, which would "upgrade" the oldest GCC version to 5.4. Not sure if these will still be issues with that version.

dg0yt:
Is this test run also with other compilers?

If the change is considered too risky, it might be limited to cmake, or even hidden behind an option.

mwtoews:
Yes, post-install tests are run in a few places (e.g. [MSVC](https://github.com/OSGeo/PROJ/runs/4573931500?check_suite_focus=true#step:7:3623), [MSYS2](https://github.com/OSGeo/PROJ/runs/4573931570?check_suite_focus=true#step:6:451)) and they all passed, so this PR isn't breaking too many things. Static tests are not run on macOS, since I don't know how to control static linking (and I don't have access to a mac). Although, I just noticed that the tests are not run on `$BUILD_NAME = "linux_clang"`... so this should be revised.

Post-install tests can be run locally. It just requires PROJ to be installed somewhere (e.g.) `/tmp/proj_install`, then run:
```
./test/postinstall/test_pkg-config.sh /tmp/proj_install static
```
similar for CMake using either `.sh` or `.bat` versions.

dg0yt:
Updates for autotools:
- Prefer `libcurl.pc` over `curl-config` (for consistency with tiff and sqlite3).
  (Unchanged: The user can explicitly pass the name or path of `curl-config` if he wants to enforce it.)
- Use `Requires.private` for a module only when pkg-config was used for this module.

Updates for CMake:
- Use `Requires.private` for a module only when this pkg-config module exists (and a `pkg-config` tool is available).
  (Standard CMake behaviour: By setting `PKG_CONFIG_EXECUTABLE` to `false`, the user can effectively force the existence check to return false.)

These changes should help to deal with systems which have some libs preinstalled, but do not provide pkg-config modules. I assume this is true for sqlite3 in macOS.

dg0yt:
The pkg-config test error is due to the unfortunate mixing of static and shared libraries.
- The system installation contains development packages of curl, tiff, and sqlite3.
  https://github.com/OSGeo/PROJ/blob/11c7d0c511806e48866bb2feef7da66de995f749/.github/workflows/linux_gcc_5_4/start.sh#L15
  The development packages contain all files for shared or static linking at *build time*, such as `libtiff.a libtiff.so libtiff.so.5 libtiff.so.5.3.0`.
- The system installation contains the *run time* dependencies of curl, tiff, and sqlite3. Usually this is just a single file such as `librtmp.so.1`.
- The pkg-config test Makefiles pass the `--static` flag to `pkg-config`, enabling the private usage requirements. 
  https://github.com/OSGeo/PROJ/blob/11c7d0c511806e48866bb2feef7da66de995f749/test/postinstall/c_app/makefile.mak#L13
  With this PR, this doesn't only enable linking directly to the curl, tiff and sqlite3 libs, but also adds the privates usage requirements of the libs - which means build-time usage of additional libs (e.g. `-lrtmp`) which is not resolvable in this system installation.

For testing `libproj.a` with shared curl, tiff, sqlite3, `pkg-config` must not be used with `--static`. This flag is for using `libproj.a` with static curl, tiff, sqlite3. I will try to fix the existing test scenario (run `pkg-config` without `--static`), and add another static scenario (which needs additional packages).

dg0yt:
> I will try to fix the existing test scenario (run `pkg-config` without `--static`), and add another static scenario (which needs additional packages).

Before doing that, I would like to add that AFAICT `pkg-config` is designed to simultaneously support all-dynamic and all-static linkage *with a single pc file*, but not the mixed mode (static proj, shared objects for dependencies) which is currently under test under the "static" label. By putting the top-most dependencies in `Libs.private`, you somewhat worked around this conflict, but I don't think this is the intended use of `pkg-config --static`.

Do you really need to test the mixed mode (which needs the `sed 's/-lproj/-Wl,-Bstatic -lproj -Wl,-Bdynamic/'` hack), or should the tesed scenario be changed to all-static?

mwtoews:
Thanks for taking further investigations! Bare in mind that changes to Autoconf will not last, unless it's highly recommended these to be urgently backported to the upcoming PROJ 8.2.1 patch release to be released in a few days (now at [release candidate 1](https://lists.osgeo.org/pipermail/proj/2021-December/010482.html)). The next release of PROJ will only have a CMake build configuration ([PROJ RFC 7](https://proj.org/community/rfc/rfc-7.html)), as I think you are aware. The only use of autotools will be kept for post-install tests.

The `-Wl,-Bstatic -lproj -Wl,-Bdynamic` specification is to isolate `libproj.a` from a build that also has `libproj.so*` files, as most linkers seem to prefer the dynamic libraries if they are available. By default, autotools builds and installs both libraries. This is indeed a bit of a hack, and only works for Linux. It may not be needed after autotools are dropped (unless someone figures a nice way to somehow build/merge both static and shared into a single install prefix).

I recall looking at a full static build of PROJ with a 3rd party app using pkg-config using the `.private` options, but I was thrown off by (e.g.) `pkg-config libcurl --libs --static`. I don't know how to handle transitive dependencies. So to me, the easiest compile check that I enabled was to isolate and statically compile and link PROJ in the "mixed" mode.

Also, I should admit that I'm hardly a "pro" with the above tool sets (autoconf, pkg-config, etc). The post-install tests are designed to prevent PROJ installs from being broken by silly changes from novices like me. Improvements to the pkg-config and CMake `find_library` components are most welcome.

dg0yt:
vcpkg has working pc files for a static build of gdal and its dependencies. I started to use them now even for configuring the gdal nmake build. In any case it means (almost) all static, or all dynamic.

For this test case here, I would now just add the missing development packages for all-static.
I wouldn't mind rolling back the autoconf change. It was more illustrative, in case someone wants to transfer this to another project.

cmake/ProjUtilities.cmake
```diff
@@ -76,27 +76,38 @@ function(configure_proj_pc)
...
+    if(PROJ_PC_${MODULE}_FOUND)
+      string(APPEND EXTRA_REQUIRES " ${MODULE}")
+    else()
+      string(APPEND EXTRA_LIBS " ${LIBS}")
+    endif()
+  endmacro()
+  add_module_or_libs("sqlite3" "-lsqlite3")
   if(TIFF_ENABLED)
-    list(APPEND EXTRA_LIBS -ltiff)
+    add_module_or_libs("libtiff-4" "-ltiff")
```
rhuijben: Can we somehow obtain the '-4' suffix from the dependency instead of hardcoding it here?

configure.ac
```diff
@@ -261,9 +261,9 @@ AC_ARG_ENABLE([tiff],
 if test "x$enable_tiff" != "xno" -o "x$enable_tiff" = ""; then
     if test "x$TIFF_CFLAGS$TIFF_LIBS" = "x" ; then
         if $PKG_CONFIG libtiff; then
-            PKG_CHECK_MODULES([TIFF], [libtiff])
+            PKG_CHECK_MODULES([TIFF], [libtiff], [TIFF_PC=libtiff])
         else
-            PKG_CHECK_MODULES([TIFF], [libtiff-4])
+            PKG_CHECK_MODULES([TIFF], [libtiff-4], [TIFF_PC=libtiff-4])
```
rhuijben: More '-4' version magic.

cmake/ProjUtilities.cmake
dg0yt: The information simply is not available at this point, but add_module_or_libs checks that the module exists. 
Alternatives:
- Use an input (cache) variable.
- Attempt several names until hitting an existing/matching name.
- Use pkg-config to detect tiff, and use just that module.

configure.ac
dg0yt: Not at all: This is attempted after checking for module `libtiff`, and the matching name is recorded.

AFAIU `libtiff-4` is simply the name chosen by upstream TIFF.

mwtoews:
Hi @dg0yt, `configure.ac` is now removed. Do you think this might be ready for review anytime in the next few weeks? There'll be a release candidate sometime in February.

kbevers:
> There'll be a release candidate sometime in February.

Expect the first RC around February 15th this time

dg0yt:
Thanks for the reminder. I will try to prepare an update soon.

cmake/ProjUtilities.cmake
```diff
@@ -76,27 +76,42 @@ function(configure_proj_pc)
...
+    if(USE_PKGCONFIG_REQUIRES)
+      list(APPEND EXTRA_REQUIRES "${MODULE}")
+    else()
+      list(APPEND EXTRA_LIBS "${ARGN}")
+    endif()
+  endmacro()
+  add_module_or_libs(sqlite3 -lsqlite3)
   if(TIFF_ENABLED)
-    list(APPEND EXTRA_LIBS -ltiff)
+    add_module_or_libs(libtiff-4 -ltiff -lfoo)
```
rouault: -lfoo ?

dg0yt: Good catch :facepalm:  This was from manual testing with multiple libs.

rouault:
Looks good to me

PR #2989, opened 2021-12-19
Update website for 8.2 release

visr:
Currently the website lists 8.1.1 as the current release under downloads: https://proj.org/download.html#current-release.

This is a cherry-pick of the download.rst part of 895f5e39535c797b02e0b5fe5f4b38211d9c64d5, which is on master but not on the 8.2 branch, which this PR targets.

PR #2990, opened 2021-12-20
update "Edit on GitHub" button branch

visr:
I also removed the leading slash, since I noticed the links contained a double slash. That worked fine, but still.

(cherry picked from commit 80e32d55d076fa48f753d5014cf828d85a0d1bce)

This is the master PR companion of the second commit in #2989.


mwtoews (approved):
Thanks, the leading slash is indeed unnecessary.

PR #2997, opened 2021-12-31
Document in the tagline dat lat_0 is required for labrd

rhuijben:
Add 'lat_0=' to the labrd info lines to document that the 'lat_0' argument is needed for a minimal test on this projection.

All other projections list their required arguments on this line with only this exception. (Makes this easier to test for in the SharpProj unittests, where I now have an hardcoded exception)

### DIFF
--- a/src/projections/labrd.cpp
+++ b/src/projections/labrd.cpp
@@ -6,7 +6,7 @@
 #include "proj.h"
 #include "proj_internal.h"
 
-PROJ_HEAD(labrd, "Laborde") "\n\tCyl, Sph\n\tSpecial for Madagascar";
+PROJ_HEAD(labrd, "Laborde") "\n\tCyl, Sph\n\tlat_0= Special for Madagascar";
 #define EPS 1.e-10
 
 namespace { // anonymous namespace
--- a/src/projections/labrd.cpp
+++ b/src/projections/labrd.cpp
@@ -6,7 +6,7 @@
 #include "proj.h"
 #include "proj_internal.h"
 
-PROJ_HEAD(labrd, "Laborde") "\n\tCyl, Sph\n\tlat_0= Special for Madagascar";
+PROJ_HEAD(labrd, "Laborde") "\n\tCyl, Sph\n\tSpecial for Madagascar\n\tlat_0=";
 #define EPS 1.e-10
 
 namespace { // anonymous namespace
--- a/docs/source/operations/projections/labrd.rst
+++ b/docs/source/operations/projections/labrd.rst
@@ -34,9 +34,10 @@ Parameters
 Required
 --------------------------------------------------------------------------------
 
-.. include:: ../options/lon_0.rst
+.. option:: +lat_0=<value>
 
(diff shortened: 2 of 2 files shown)


src/projections/labrd.cpp
```diff
@@ -6,7 +6,7 @@
 #include "proj.h"
 #include "proj_internal.h"
 
-PROJ_HEAD(labrd, "Laborde") "\n\tCyl, Sph\n\tSpecial for Madagascar";
+PROJ_HEAD(labrd, "Laborde") "\n\tCyl, Sph\n\tlat_0= Special for Madagascar";
```
rouault: Perhaps ?
```suggestion
PROJ_HEAD(labrd, "Laborde") "\n\tCyl, Sph\n\tSpecial for Madagascar\n\tlat_0=";
```

rouault (commented):
see suggestion

src/projections/labrd.cpp
rhuijben: I'm not sure what the expected format is here...
All projections I found had 2 lines max where  the first line provided the projection info and the second line arguments and sometimes a bit of text appended to it. I don't know what happens to other users of the headers if we add a third line.

I'm fine with either form. But perhaps we should add a comment on the expected format while we are at it.

rouault:
I'm not sure either of what the expected format is (most of us were hardly born when this was devised), and if applications are supposed to rely on that. My quick search couldn't spot an example where there would have a line with parameters + free text, hence what I suggested

rhuijben:
> My quick search couldn't spot an example where there would have a line with parameters + free text, hence what I suggested

tmerc.cpp has
PROJ_HEAD(utm, "Universal Transverse Mercator (UTM)") "\n\tCyl, Ell\n\tzone= south approx";

rouault:
south and approx are actually boolean flags, e.g. "+proj=utm +zone=31 +south +approx"

kbevers:
I don't understand the part about Madagascar, why is that needed?

`+lat_0` is an outright requirement, as seen from an instantiation of the projection without that parameter:

```
$ proj +proj=labrd
proj_create: Error 1027 (Invalid value for an argument): labrd: Invalid value for lat_0: lat_0 should be different from 0
Rel. 8.2.0, November 1st, 2021
<proj>:
cause: Invalid value for an argument
program abnormally terminated
```

but unfortunately we state in the [docs](https://proj.org/operations/projections/labrd.html) that the default value of `+lat_0` is 0. This should be adjusted, including a sentence saying that it can't be set to 0.

rhuijben:
> I don't understand the part about Madagascar, why is that needed?

That was already there, so I assumed it should stay there... But I agree that it doesn't make much sense.

kbevers:
>  That was already there, so I assumed it should stay there... But I agree that it doesn't make much sense.

I missed that from the diff. I assume the projection originally have been made specifically for use in Madagascar. 

I've added a commit to your branch with a clarification of the parameters in the Sphinx docs.

PR #3003, opened 2022-01-03
bertin1953.rst: fix link

anbj:


mwtoews:
The original rendition of the web citation looks fine to me:
> Philippe Rivière (2017). _Bertin Projection (1953) &lt;https://visionscarto.net/bertin-projection-1953&gt;_, Visionscarto.net.

it has the author's name, year, title, print-worthy URL, and domain "publisher". The PR would only show the title, the URL would not be print-worthy (not a big deal, as no-one should be printing the PDF manual). Currently, web citations don't have a consistent format in this manual (whereas publications use a Harvard-like bibliographic engine). While it's possible to also add web citations to docs/source/references.bib, it could get busy if *every* URL is added there.

Any thoughts on web citation styles?

anbj:
>Currently, web citations don't have a consistent format in this manual (whereas publications use a Harvard-like bibliographic engine).

Ok - just close the PR.
I tried to change it to look the same as many of the other "Further reading" links.

PR #3009, opened 2022-01-05
Specify CMAKE_INSTALL_RPATH for macOS; use -rpath LDFLAGS for tests

mwtoews:
The primary change in this PR applies to macOS users, which allows the library and utilities to be relocatable, using a relative path to the library via `../lib`. There was a "TO DO: cmake 2.8.12" note in src/CMakeLists.txt that provided accurate guidance on this. Before this change, a user might see (e.g.):
```
$ /tmp/proj_cmake_install_renamed/subdir/bin/projsync
dyld: Library not loaded: @rpath/libproj.22.dylib
  Referenced from: /tmp/proj_cmake_install_renamed/subdir/bin/projsync
  Reason: image not found
```
the RPATH setting in CMake was already used for Linux so that the library and utilities are relocatable.

Other changes:
- Don't use LD_LIBRARY_PATH or DYLD_LIBRARY_PATH variables
- Quieter conda install messages, and install jsonschema required for testing

Other changes within test/postinstall:
- Rename `test_ldd` to `test_libpath` to be more generic
- For macOS, use `otool -l` to look for a matching LC_RPATH entry
- When necessary, use `export LDFLAGS="${LDFLAGS} -Wl,-rpath,$(pkg-config proj --variable=libdir)"` to "bake" the correct RPATH into objects (not needed for test_cmake or static builds)

### DIFF
--- a/.github/workflows/doc_build.yml
+++ b/.github/workflows/doc_build.yml
@@ -37,7 +37,6 @@ jobs:
         working-directory: ./docs/plot
         run: |
             export PATH=/opt/proj/bin:$PATH
-            export LD_LIBRARY_PATH=/opt/proj/lib
             export PROJ_LIB=/opt/proj/share/proj
             ./plot.py plotdefs.json images/
 
--- a/.github/workflows/mac/before_install.sh
+++ b/.github/workflows/mac/before_install.sh
@@ -3,11 +3,9 @@
 set -e
 
 
-conda update -n base -c defaults conda
-conda install compilers -y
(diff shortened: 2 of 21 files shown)


mwtoews:
I would appreciate any comment from mac users, as I don't have access to one. These tests were developed thru non-interactive runs on GitHub Actions.

Perhaps @hobu recalls changes to src/CMakeLists.txt in fab753b40728709da376680efbc503ba2643faad from 7 years ago? Do the changes look appropriate?

Issue #2688, opened 2021-04-23
cmake full-transition memo


rouault:
What is currently missing in cmake builds:

- [X]  `make dist` target - #2690
- [x] regeneration of src/wkt1_generated_parser.c/.h and src/wkt2_generated_parser.c/.h with bison : those are currently generated with "make wkt1_parser" and "make wkt2_parser" in the $builddir/src directory of a autoconf build: https://github.com/OSGeo/PROJ/pull/2900
- [x] check and migrate CI workflows, e.g. Clang Static Analyzer - #3001
- [x] migrate ossfuzz build: test/fuzzers/build.sh: tracked in https://github.com/OSGeo/PROJ/pull/2896
- [x] downstream users: GDAL CI && ossfuzz : tracked in https://github.com/OSGeo/gdal/issues/4576
- [X] downstream user: libgeotiff: https://github.com/OSGeo/libgeotiff/pull/64
- [x] Dockerfile (used by .github/workflows/docker.yml) needs to change to cmake: https://github.com/OSGeo/PROJ/pull/3018
- [x] revise `HOWTO-RELEASE` document for relevant changes - #3030
- [X] `make uninstall` target - https://github.com/OSGeo/PROJ/pull/3362
- [x] re-implement `PROJ_DB_CACHE_DIR` - #3711

mwtoews:
Just by coincidence (within minutes from each other), see #2690 for "make dist"

rouault:
>     * `make clean` target

What is missing exactly ? make clean seems to work

mwtoews:
I can't recall, as `make clean` works fine from Makefile generators from CMake 3.10.2. I'll scrub this task off.

mwtoews:
🤦 (facepalm) it was `make uninstall` that I was thinking of. Autotools supports this, although it is not tested. Any interest in bringing this to CMake too? It's not too difficult; see [how GEOS does this](https://github.com/libgeos/geos/blob/a434b1fbf6981bd6fd10dee1705fbf66b8e4fc0a/CMakeLists.txt#L394-L403).

kbevers:
Sounds like a nice thing to have

PR #3020, opened 2022-01-13
Remove autoconf build system (#2688)

rouault:
Maybe not for immediate merge, but at least to check our CI is now ready


kbevers:
This is a big milestone! I agree that it needs a bit of time before merging still. One thing that immediately comes to my mind is an overhaul to `HOWTORELEASE`. There will be other things as well, I'm sure.

mwtoews:
In a parallel world, I've been shaping a similar PR https://github.com/mwtoews/PROJ/tree/transition-ci-from-autotools-to-cmake (just the top commit). It just does the remaining CI transition. It's nearly ready. Should I create the PR for comparison?

rouault:
> Should I create the PR for comparison?

from a quick glance it looks like your approach is better, so we should likely go with it

mwtoews:
I'll push it when it's ready, perhaps tomorrow. Still typing up notes and chasing up macOS-specific failures.

mwtoews:
See #3023 for the similar PR mentioned previously.

PR #3023, opened 2022-01-16
CI: transition tests from autotools to cmake

mwtoews:
This PR rewrites most of travis/install.sh to replace autotools with cmake. There is lots going on here, so here is a brief description of some of the changes:

- Build PROJ twice, for static and dynamic builds, each in their own build and install directories; most tests are run for each group
- `cmake/ProjConfig.cmake` was adjusted to temporarily remove `-Werror` C flag while running `check_library_exists(m exp "" HAVE_LIBM)`, as this raised an error with GCC 5.4
- Update PATH on Clang Linux to fix tests with jsonschema, which fixed one issue, but created a different issue that requires a workaround for jsonschema version 3.2.0 (error was `jsonschema: error: argument -i/--instance: invalid _json_file value: 'out.json'`)
- Install jsonschema for macOS, which fixed another issue, but created several different issues (1) that `/tmp` is a symblink to `/private/tmp` only for the shared version of the library in some contexts, and (2) `sed` behaves differently than on linux, requiring a backup suffix and use of command `c\` (not `c`)
- Coverage build with `lcov` seemed to need an in-source build (i.e. `cmake .`), so this remains the last step
- No need to (e.g.) use `CC="ccache gcc"`, as this should be handled by [`USE_CCACHE=ON`](https://github.com/OSGeo/PROJ/pull/2948); move all of the `ccache -s` into the main script
- Remove `export CCACHE_CPP2=yes` as this was required for older versions of ccache
- Update test/fuzzers paths to assume a cmake build path

A PR to drop autotools support will follow after this. The cache on macOS is not working, so that'll be another follow-up task to sort-out.

### DIFF
--- a/travis/install.sh
+++ b/travis/install.sh
@@ -29,23 +29,28 @@ if test "x${CMAKE_BUILD_TYPE}" = "x"; then
     CMAKE_BUILD_TYPE=Release
 fi
 
-echo "Make dist tarball, and check consistency"
-mkdir build_dist
-cd build_dist
-cmake -D BUILD_TESTING=OFF ..
-make dist
+# For some odd reason the tar xzvf $TAR_FILENAME doesn't work on Travis-CI ...
+if test "$TRAVIS" = ""; then
+    echo "Make dist tarball, and check consistency"
+    mkdir build_dist
+    cd build_dist
+    cmake -D BUILD_TESTING=OFF ..
+    make dist
 
-TAR_FILENAME=$(ls *.tar.gz)
-TAR_DIRECTORY=$(basename $TAR_FILENAME .tar.gz)
-mkdir ../build_from_dist
-cd ../build_from_dist
-tar xvzf ../build_dist/$TAR_FILENAME
+    TAR_FILENAME=$(ls *.tar.gz)
+    TAR_DIRECTORY=$(basename $TAR_FILENAME .tar.gz)
+    mkdir ../build_from_dist
+    cd ../build_from_dist
+    tar xvzf ../build_dist/$TAR_FILENAME
+
+    # continue build from dist tarball
+    cd $TAR_DIRECTORY
+fi
 
 # There's a nasty #define CS in a Solaris system header. Avoid being caught about that again
 CXXFLAGS="-DCS=do_not_use_CS_for_solaris_compat $CXXFLAGS"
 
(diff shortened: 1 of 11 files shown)


travis/install.sh
```diff
@@ -16,113 +16,158 @@ fi
...
-cd ..
-diff -qr build_autoconf/$TAR_DIRECTORY build_cmake/$TAR_DIRECTORY || true
 
-# continue build from autoconf
-cd build_autoconf/$TAR_DIRECTORY
+TAR_FILENAME=$(ls *.tar.gz)
+TAR_DIRECTORY=$(basename $TAR_FILENAME .tar.gz)
+mkdir ../build_from_dist
+cd ../build_from_dist
+tar xvzf ../build_dist/$TAR_FILENAME
```
rouault: you seem to hit on Travis-CI the same weirdness I got in https://github.com/OSGeo/PROJ/pull/3020/ with this tar command having apparently no effect... As I couldn't figure out what was going on I just disable building from the tar on Travis-CI : https://github.com/OSGeo/PROJ/pull/3020/files#diff-c47231c4caf6347330276d1d8abd9f76d5299bd4767e395dabb4f95755b0ce8fR19 and https://github.com/OSGeo/PROJ/pull/3020/files#diff-c47231c4caf6347330276d1d8abd9f76d5299bd4767e395dabb4f95755b0ce8fR177

rouault:
There's an empty tmp.txt file that was added in the commit

travis/install.sh
mwtoews: strange! I'd need to re-arrange the source dir for the other tests.

QuLogic:
> Build PROJ twice, for static and dynamic builds, each in their own build and install directories; most tests are run for each group

How much is shared between the groups? Would it be simpler to have one script called twice with a shared/static option?

mwtoews:
There is a bit of repetition in travis/install.sh to build and run tests for shared and dynamic versions. While it could be refactored to reduce it somehow, I'd like to eventually see the CI workflows to be re-written using (e.g.) jobs and other features of GitHub Actions. For now, the main edits to travis/install.sh more-or-less cover the transition of build tools before removing the autotools files.

I'm also open to removing some of the tests and time churn (I might have gotten carried away with post-install tests, which are really only a minor component that could be checked less often).

PR #3027, opened 2022-01-23
Drop autotools; move remaining useful m4 macros

mwtoews:
Follow-up from #3023

Next step: rewrite HOWTO-RELEASE


kbevers:
This is very good, Mike! The installation instructions in the docs also needs updating. Not sure if you've forgot or another follow-up PR's is on it's way.

mwtoews:
@kbevers thanks for the reminder, I'll eventually prepare a PR to address changes to the docs. There are also a few minor tasks tracked in #2688

kbevers:
@mwtoews alright, cool. Thanks!

rouault:
I believe we can merge that now ? @mwtoews Can you send an email to the mailing list before, so that people who are late to the party and still have build scripts tracking PROJ master and using autoconf are warned that they have to take action now ?

PR #3030, opened 2022-01-30
Update HOWTO-RELEASE using only CMake (without autotools)

mwtoews:
Hopefully a bit simpler!

One packaging difference is that I recommend is building the source distribution in an isolated local directory. See what happened to libgeos [here](https://trac.osgeo.org/geos/ticket/1143) to see why. I've been tempted to rewrite the `dist` target to use git tools (like [git-archive](https://git-scm.com/docs/git-archive) or [git-ls-files](https://git-scm.com/docs/git-ls-files)) to package only git-tracked files, but not sure if it's worth the effort.

Next step: update public-facing docs.

### DIFF
--- a/HOWTO-RELEASE
+++ b/HOWTO-RELEASE
@@ -31,27 +31,22 @@ list. The PROJ PSC guidelines describes the rules for passing a motion.
 -------------------------------------------------------------------------------
 
 In case of a un-planned maintenance release, make sure that version numbers are
-propertly updated: see 3.1
+properly updated: see 3.1
 
 Determine the ABI version number. It consists of "current:revision:age". Follow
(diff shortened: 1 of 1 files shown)


rouault (commented):
Looks good to me. Regarding generation of release tarballs, we could possibly have a mkdist.sh script that automates a bit the instructions of 1.7

kbevers:
>  we could possibly have a mkdist.sh script that automates a bit the instructions of 1.7

That would be nice, yes. More or less every time I prepare a release I get annoyed by the need for first renaming the files with the RC postfixes and then doing the md5sums. Having to not do that would be great. Not necessary for this PR, but definitely a nice to have in the future.

kbevers (approved):

PR #3031, opened 2022-01-31
Allow PROJ_LIB paths wrapped with double quotes

mloskot:
This improvement is mostly useful on Windows, where users
are used to wrapping paths with whitespaces with double quotes.

For example, these should now be allowed (correctly concatenated):

```
PROJ_LIB="C:\Program Files\PROJ";C:\Users\mloskot\PROJ
```

------
By the way, as per https://github.com/OSGeo/PROJ/pull/1230#issuecomment-455826744 paths with spaces are supported:

> > Can folders with spaces in them also be added to `PROJ_LIB`?
>
> yes. The only restriction is that they don't have ':' in them on Unix or ';' on Windows as those are used as separator to delineate directories

------

### DIFF
--- a/src/filemanager.cpp
+++ b/src/filemanager.cpp
@@ -1420,6 +1420,21 @@ static void *pj_open_lib_internal(
         if (out_full_filename != nullptr && out_full_filename_size > 0)
             out_full_filename[0] = '\0';
 
+        auto open_lib_from_paths = [&ctx, open_file, &name, &fname, &sysname, &mode](const std::string & projLib) {
+            void *fid = nullptr;
+            auto paths = NS_PROJ::internal::split(projLib, dirSeparator);
+            for (const auto& path : paths) {
+                fname = NS_PROJ::internal::stripQuotes(path);;
+                fname += DIR_CHAR;
+                fname += name;
+                sysname = fname.c_str();
+                fid = open_file(ctx, sysname, mode);
+                if (fid)
+                    break;
+            }
+            return fid;
+        };
+
         /* check if ~/name */
         if (is_tilde_slash(name))
             if ((sysname = getenv("HOME")) != nullptr) {
@@ -1487,16 +1502,7 @@ static void *pj_open_lib_internal(
         else if (!gbPROJ_LIB_ENV_VAR_TRIED_LAST &&
                  !(projLib = NS_PROJ::FileManager::getProjLibEnvVar(ctx))
                       .empty()) {
-            auto paths = NS_PROJ::internal::split(projLib, dirSeparator);
-            for (const auto &path : paths) {
-                fname = path;
-                fname += DIR_CHAR;
-                fname += name;
-                sysname = fname.c_str();
-                fid = open_file(ctx, sysname, mode);
-                if (fid)
-                    break;
-            }
+            fid = open_lib_from_paths(projLib);
         }
 
         else if ((sysname = get_path_from_relative_share_proj(
@@ -1519,16 +1525,7 @@ static void *pj_open_lib_internal(
         else if (gbPROJ_LIB_ENV_VAR_TRIED_LAST &&
                  !(projLib = NS_PROJ::FileManager::getProjLibEnvVar(ctx))
                       .empty()) {
-            auto paths = NS_PROJ::internal::split(projLib, dirSeparator);
-            for (const auto &path : paths) {
-                fname = path;
-                fname += DIR_CHAR;
-                fname += name;
-                sysname = fname.c_str();
-                fid = open_file(ctx, sysname, mode);
-                if (fid)
-                    break;
-            }
+            fid = open_lib_from_paths(projLib);
         }
 
         else {
--- a/src/filemanager.cpp
+++ b/src/filemanager.cpp
@@ -1421,18 +1421,18 @@ static void *pj_open_lib_internal(
             out_full_filename[0] = '\0';
 
         auto open_lib_from_paths = [&ctx, open_file, &name, &fname, &sysname, &mode](const std::string & projLibPaths) {
-            void *fid = nullptr;
+            void *lib_fid = nullptr;
             auto paths = NS_PROJ::internal::split(projLibPaths, dirSeparator);
             for (const auto& path : paths) {
                 fname = NS_PROJ::internal::stripQuotes(path);
                 fname += DIR_CHAR;
                 fname += name;
                 sysname = fname.c_str();
-                fid = open_file(ctx, sysname, mode);
-                if (fid)
+                lib_fid = open_file(ctx, sysname, mode);
+                if (lib_fid)
                     break;
             }
-            return fid;
+            return lib_fid;
         };
 
         /* check if ~/name */
--- a/src/filemanager.cpp
+++ b/src/filemanager.cpp
@@ -1420,9 +1420,9 @@ static void *pj_open_lib_internal(
         if (out_full_filename != nullptr && out_full_filename_size > 0)
             out_full_filename[0] = '\0';
 
-        auto open_lib_from_paths = [&ctx, open_file, &name, &fname, &sysname, &mode](const std::string & projLib) {
+        auto open_lib_from_paths = [&ctx, open_file, &name, &fname, &sysname, &mode](const std::string & projLibPaths) {
             void *fid = nullptr;
-            auto paths = NS_PROJ::internal::split(projLib, dirSeparator);
+            auto paths = NS_PROJ::internal::split(projLibPaths, dirSeparator);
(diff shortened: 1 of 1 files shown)


src/filemanager.cpp
```diff
@@ -1489,7 +1489,7 @@ static void *pj_open_lib_internal(
                       .empty()) {
             auto paths = NS_PROJ::internal::split(projLib, dirSeparator);
             for (const auto &path : paths) {
-                fname = path;
+                fname = NS_PROJ::internal::stripQuotes(path);
```
rouault: should be done also on line 1524. We might use a lamba for lines 1490 et 1499 that could be use in both places

mloskot: I did not look for other uses of `projLib`. Good point. I've updated the PR.

mloskot:
I'm not sure what happened with the CI jobs after my 2nd commit, but the failures don't seem related to this change
```
CMake Error at googletest-download/googletest-prefix/src/googletest-stamp/download-googletest.cmake:170 (message):
  Each download failed!
```

rouault:
> I'm not sure what happened with the CI jobs after my 2nd commit, but the failures don't seem related to this change

indeed, see https://github.com/OSGeo/PROJ/pull/3034

rouault:
```
/home/travis/build/OSGeo/PROJ/src/filemanager.cpp: In lambda function:

/home/travis/build/OSGeo/PROJ/src/filemanager.cpp:1424:19: error: declaration of ‘fid’ shadows a previous local [-Werror=shadow]

             void *fid = nullptr;

                   ^~~

/home/travis/build/OSGeo/PROJ/src/filemanager.cpp:1413:15: note: shadowed declaration is here

         void *fid = nullptr;

               ^~~
```

jmckenna:
@mloskot thanks for your effort on this.  I am following this, because I believe I (and all MS4W/Windows users) hit this recently, with PROJ 8.2.0, where I distributed its PROJ_LIB with the following settings, that failed (users reported this to me):
```
set PROJ_LIB=\ms4w\share\proj;\ms4w\share\proj\grids
```
I had to release another beta build with the workaround (to unfortunately not use multiple directories on Windows):
```
set PROJ_LIB=\ms4w\share\proj
```

I did not use any quotes or spaces, but I will test more (now that I see your report here - sometimes it is nice seeing that I'm not alone on Windows support, ha).  

Note: I'm not looking to overtake an existing pull request, with an issue; instead I am just trying to give you a pat on the back, for tackling something that is seen as super important to Windows users.

Thanks,
-jeff

rouault:
> (to unfortunately not use multiple directories on Windows):

@jmckenna
weird, I've just tested successfully the following with builds of GDAL and PROJ master done on Windows, with proj.db in c:\dev and the grids in c:\dev\miniconda3\envs\myenv\Library\share\proj
```
(myenv) C:\dev\gdal_cmake\build_conda>set PROJ_LIB=\dev;\dev\miniconda3\envs\myenv\Library\share\proj

(myenv) C:\dev\gdal_cmake\build_conda>set PROJ_DEBUG=2

(myenv) C:\dev\gdal_cmake\build_conda>echo 2 49 0 | gdaltransform -s_srs EPSG:4979 -t_srs EPSG:4326+5773 --config CPL_DEBUG ON

PROJ: pj_open_lib(proj.db): call fopen(\dev\proj.db) - succeeded
PROJ: pj_open_lib(proj.ini): call fopen(\dev\miniconda3\envs\myenv\Library\share\proj\proj.ini) - succeeded
PROJ: Using coordinate operation WGS 84 to EGM96 height (1)
PROJ: pj_open_lib(us_nga_egm96_15.tif): call fopen(\dev\miniconda3\envs\myenv\Library\share\proj\us_nga_egm96_15.tif) - failed
PROJ: pj_open_lib(egm96_15.gtx): call fopen(\dev\miniconda3\envs\myenv\Library\share\proj\egm96_15.gtx) - succeeded
2 49 -44.6434936523438
```

Do you have local patches to PROJ ? Can you check if the line https://github.com/OSGeo/PROJ/blob/ab504340fc9cdca9044a9339acbb4e9ba0367bb6/src/filemanager.cpp#L1381 is built in your build ? You can for example add a ``#error "foo"`` line immediately after it.  The ``#ifdef WIN32`` test is not super idiomatic, it should rather be  ``#ifdef _WIN32`` , but proj_internal.h that is included by that file has https://github.com/OSGeo/PROJ/blob/ab504340fc9cdca9044a9339acbb4e9ba0367bb6/src/proj_internal.h#L149, so that makes thing work.

jmckenna:
Excellent advice @rouault I'll do exactly as you say, and report back.  (if I find that it is PROJ's "fault" I will file a new ticket here, but it sounds like it could be my build steps)  thanks for these tips, really appreciated.

dg0yt:
> For example, these should now be allowed (correctly concatenated):
> 
> ```
> PROJ_LIB="C:\Program Files\PROJ";C:\Users\mloskot\PROJ
> ```

Is this special behaviour really needed? 
Does PROJ have to deal with quotes because (Windows) users don't know how to set the environment variable in `COMMAND.COM` or powershell, but blindly copy (Unix) `bash` instructions?

~~~cmd
C:\>set PROJ_LIB=C:\Program Files\PROJ;C:\Users\mloskot\PROJ

C:\>set PROJ_LIB
PROJ_LIB=C:\Program Files\PROJ;C:\Users\mloskot\PROJ
~~~

~~~ps1
PS C:\> $env:PROJ_LIB='PROJ_LIB=C:\Program Files\PROJ;C:\Users\mloskot\PROJ'

PS C:\> $env:PROJ_LIB
PROJ_LIB=C:\Program Files\PROJ;C:\Users\mloskot\PROJ
~~~

mloskot:
@dg0yt 
> Is this special behaviour really needed?

There is nothing special in this behaviour and, mind you, there is nothing blindly copied, but native Windows Command Prompt features in use - unless one doesn't know how the TAB completion of paths work on Windows:

https://user-images.githubusercontent.com/80741/152356309-b5ceda6d-afc1-4f1a-a172-a673f9aa41f4.mp4

dg0yt:
I know tab completion on Windows. When used at the command prompt, the result is usually passed as a single string to the executable, and quotes must be interpreted by that program.
This is something different than reading an environment variable which has elements clearly separated (by `;` on Windows).

If you want some support for your PR: `PATH` seems to handle quotes as proposed here for `PROJ_LIB`. 
You don't need an elevated command prompt to check:
~~~
C:\>set PATH="C:\Windows\System32";"C:\Windows"

C:\>echo %PATH%
"C:\Windows\System32";"C:\Windows"

C:\>notepad.exe
~~~

src/filemanager.cpp
```diff
@@ -1420,6 +1420,21 @@ static void *pj_open_lib_internal(
         if (out_full_filename != nullptr && out_full_filename_size > 0)
             out_full_filename[0] = '\0';
 
+        auto open_lib_from_paths = [&ctx, open_file, &name, &fname, &sysname, &mode](const std::string & projLibPaths) {
+            void *fid = nullptr;
+            auto paths = NS_PROJ::internal::split(projLibPaths, dirSeparator);
+            for (const auto& path : paths) {
+                fname = NS_PROJ::internal::stripQuotes(path);;
```
dg0yt: I noticed a double `;` here.

mloskot:
@dg0yt 
> If you want some support for your PR

I don't really care. I just noticed the quotes handling may be a (confusing) issue to some less experienced users and I proposed a non-intrusive improvement that makes the handling of the variable a bit friendlier.
I'm not going to even try convince you or anyone else to accept this PR.

I'll leave it to @rouault to decide what to do with this PR.

src/filemanager.cpp
mloskot: Thx, that should be fixed now.

rouault:
> I'll leave it to @rouault to decide what to do with this PR.

I don't really see any harm in the PR. Once CI is happy (which implies fixing the warning about fid variable shadowing), it should be fine

mloskot:
@rouault 
> Once CI is happy (which implies fixing the warning about fid variable shadowing), it should be fine

It's all green now.

dg0yt:
Will this be applied to environment variables used by GDAL?
I'm still puzzled that legacy `COMMAND.COM` behaviour now determines environment variable interpretation on POSIX systems.

mloskot:
> Will this be applied to environment variables used by GDAL?

Yes

> legacy `COMMAND.COM` behaviour now determines environment variable interpretation on POSIX systems.

That's an overinterpretation as the change is absolutely benign and transparent to any POSIX shell.

Issue #3032, opened 2022-01-31
Question: is float GeoTIFF and GPG lossless compared to .gsb?


ValZapod:
FFmpeg that I usually use does not really have precision for float --> int stuff, so cannot verify.

If the anwser is yes, please update https://proj.org/specifications/geodetictiffgrids.html where you start talking about GDA2020 here 

>The file size is 4.8 MB using DEFLATE compression and floating-point predictor. To be compared with the 83 MB of the original .gsb file.

ValZapod:
I also find it insane that float values are not done in GeographicLib! Just nuts, they do not even support TIFF even NOT IN FLOAT (from docs here https://geographiclib.sourceforge.io/html/geoid.html)

Also can you please verify this for me, 1'x1' EGM2008 GPG file? https://www.agisoft.com/downloads/geoids/

rouault:
github issues are not for questions.

PR #3033, opened 2022-01-31
DOCS: clarify that GTG is lossless

ValZapod:
That is of course depends on #3032 that you so radically closed.

### DIFF
--- a/docs/source/specifications/geodetictiffgrids.rst
+++ b/docs/source/specifications/geodetictiffgrids.rst
@@ -636,7 +636,7 @@ It contains 5 subgrids. All essential metadata to list the subgrids and their
 georeferencing is contained within the first 3 KB of the file.
 
 The file size is 4.8 MB using DEFLATE compression and floating-point predictor.
-To be compared with the 83 MB of the original .gsb file that is lossless with it.
+It results from a lossless conversion from the 83 MB of the original .gsb file.
 
 https://github.com/OSGeo/PROJ-data/blob/master/ca_nrc/ca_nrc_ntv2_0.tif has
 been converted from https://github.com/OSGeo/proj-datumgrid/blob/master/north-america/ntv2_0.gsb
--- a/docs/source/specifications/geodetictiffgrids.rst
+++ b/docs/source/specifications/geodetictiffgrids.rst
@@ -636,7 +636,7 @@ It contains 5 subgrids. All essential metadata to list the subgrids and their
 georeferencing is contained within the first 3 KB of the file.
 
 The file size is 4.8 MB using DEFLATE compression and floating-point predictor.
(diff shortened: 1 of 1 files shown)


docs/source/specifications/geodetictiffgrids.rst
```diff
@@ -636,7 +636,7 @@ It contains 5 subgrids. All essential metadata to list the subgrids and their
 georeferencing is contained within the first 3 KB of the file.
 
 The file size is 4.8 MB using DEFLATE compression and floating-point predictor.
-To be compared with the 83 MB of the original .gsb file.
+To be compared with the 83 MB of the original .gsb file that is lossless with it.
```
rouault: I'm not sure this formulation is very clear. I'd rather change the previous line to "The file size is 4.8 MB using DEFLATE compression and floating-point predictor, which are lossless compression parameters"

ValZapod: > which are lossless compression parameters"

No, what i meant is that it is losssless with original gsb file, not that the float compressed TIFF is lossless. That is obvious.

rouault:
```suggestion
It results from a lossless conversion of the original .gsb file whose size is 83 MB.
```

ValZapod:
Sigh, will squash it. So, you are positive conversion is lossless?

ValZapod:
Can you recreate bitperfect original gsb file from it?

rouault:
> So, you are positive conversion is lossless?

yes, at least as done by the https://github.com/OSGeo/PROJ-data/blob/master/grid_tools/ntv2_to_gtiff.py script

hobu:
@ValZapod while I'm sure we are all obviously stupider than you, please try to improve your tone and interaction with us. GitHub tickets are not a support venue, and without some example code showing a specific issue on this particular topic, you are not likely to get the satisfaction you are trying to obtain through this venue.

You might try the mailing list for a wider audience, but please remember we are neither your emergency support helpline nor your bug chasers. I understand language and cultural differences can make communication challenging, but please stop acting as if your time is more valuable than ours and we have somehow wasted it with a bug you can neither illustrate or describe.

ValZapod:
> some example code showing a specific issue on this particular topic,

There is no issue. This is just docs change.

> are all obviously stupider than you

That is or opinion only.

ValZapod:
Thanks! I almost git reset'ed it to squash :) Anyway, thanks for your emergency support helpline. I was just flustrated that 16 bit quantised tiff is not lossless and thus gives errors in https://lists.osgeo.org/pipermail/proj/2019-December/009121.html (yep, I knew there was a mailing list).

PR #3034, opened 2022-02-01
Update SHA1SUM of https://github.com/google/googletest/archive/release-1.11.0.zip

rouault:
The sha1sum of the .zip has changed overnight. Comparing a previous old
cached archive and a newly downloaded one, their content is identical once
unzipped. The .zip file sizes are identical but they differ on some
bytes in the 'header' of each zipped file. Probably github changed their
version of the zip utility ?


rouault:
Dropping that one. It seems from https://github.com/OSGeo/PROJ/runs/5022725224?check_suite_focus=true that the original sha1sum has been restored...

mwtoews:
xref https://github.com/google/googletest/issues/3749

PR #3035, opened 2022-02-01
Use external gtest by default when possible

rouault:


rouault:
The clang-linux job on Ubuntu 20.04 successfully uses the libgtest-dev package:
```
-- Found GTest: /usr/lib/x86_64-linux-gnu/libgtest.a (Required is at least version "1.8.1") 
-- Using external GTest
```

PR #3042, opened 2022-02-07
CMake: make BUILD_SHARED_LIBS=ON the default even on Windows

rouault:


kbevers:
What's the reasoning behind this change?

rouault:
> What's the reasoning behind this change?

"why make a special case for Windows" ?   And more seriously, static linking is a difficult topic, especially since PROJ has now dependencies to sqlite, curl and geotiff, so it should only be used if people know they really need it.

rouault:
The current static linking for Windows come from the initial commit https://github.com/OSGeo/PROJ/commit/c90714e3b141e2ee02d110a73818f71f726a35fb . Maybe @hobu has some opinions regarding this ?

kbevers:
Thanks @rouault. I have no opinion on the topic just curious as to why it changed, mainly in order to put the right thing in the release notes.

jmckenna:
This change makes a lot of sense, very helpful.  Good to see.  Thanks.

PR #3043, opened 2022-02-08
Docker: update base image to ubuntu:20.04

mwtoews:
Upgrade the Ubuntu-based docker images to the latest Ubuntu LTS, with end of standard support in April 2025 and EOL 2030.

Specific motivation for docs/docbuild/Dockerfile is to upgrade beyond Python 3.6, which is no longer supported upstream and by many packages in use. Ubuntu 20.04 includes support for Python 3.8.

Images and docs (pdf and html) were built locally, and look fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ##
 # OSGeo/PROJ
 
-FROM ubuntu:18.04 as builder
+FROM ubuntu:20.04 as builder
 
 MAINTAINER Howard Butler <howard@hobu.co>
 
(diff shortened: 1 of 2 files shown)


hobu:
@rouault are we ready to update this to 20.04 and Python3.8?

rouault:
> are we ready to update this to 20.04 and Python3.8?

The only potential point that could have caused issues as far as I'm concerned is the update of the proj-docs image which is used by GDAL to build its doc. I've given it a local testing and things seem to work fine, so I'm go for it.

hobu:
I suppose the projects that are using this for documentation building (PDAL, PROJ, GDAL, others?) should consider pinning their container definition to a SHA rather than just the tag `osgeo/proj-docs` so that they are protected from us changing things like this.

hobu:
`osgeo/proj-docs` container has been pushed.

PR #3058, opened 2022-02-14
ITRF2008: add EURA_T, NOAM_T, etc. plate names that include translation rates (fixes #3053)

### LINKED ISSUE #3053: ITRF2008 plate motion model init strings lack translation terms
See:
- https://lists.osgeo.org/pipermail/proj/2022-February/010535.html
- https://lists.osgeo.org/pipermail/proj/2022-February/010536.html
- https://lists.osgeo.org/pipermail/proj/2022-February/010537.html

rouault:


rouault:
@kbevers  I guess we can include that in the RC. Got a semi-confirmation from Steve Riddell "I just looked at your new file. Your NOAM_T looks just like the one I’ve updated on my machine and have been running"

kbevers:
There's also this

> It turns out there was another reason, besides translation rates, that was preventing me matching HTDP results:  for the NOAM plate, the z-rotation rate should be -0.0001 as opposed to +0.0001.  The ITRF2008 init file needs to be updated.

Has that already been fixed?

rouault:
> Has that already been fixed?

yes already in master per https://github.com/OSGeo/PROJ/commit/625d4b99e0b64b3ce5ba773e4236248506ae4f99

PR #3061, opened 2022-02-15
Remove googletest source code remains

rouault:
They are no longer needed, as there were only used by autoconf builds.
For CMake, we either use external googletest, or use CMake download capabilities.


mwtoews:
Looks like a good clean-up. I was considering changing some of the googletest framework (e.g. [use FetchContent instead of ExternalProject](https://github.com/google/googletest/pull/3314)), but ultimately these will wait until the next CMake version bump, which I'll probably look into for 2023.

mwtoews (approved):

PR #3063, opened 2022-02-17
Add note on named variants of CEA

pramsey:
People wondering "does proj support Peters" will now find that it does when searching for "Peters".


rouault:
oh you beat me but I didn't notice your pull request  :-)

PR #3066, opened 2022-02-18
Simple SOVERSION

mwtoews:
See suggestion by @sebastic https://lists.osgeo.org/pipermail/proj/2022-February/010551.html

Use a much simpler SOVERSION for a CMake-only project.

This will result with the following libraries/symlinks:
```
libproj.so -> libproj.so.25
libproj.so.25 -> libproj.so.9.0.0
libproj.so.9.0.0
```
Was too much from `HOWTO-RELEASE` chopped? Rephrasing suggestions welcome.


mwtoews:
One caution is that `libproj.so.9.0.0` was a library name from proj-4.9.1 from 2015-03-04. Is this a potential issue?

sebastic:
> This will result with the following libraries/symlinks:
> 
> ```
> libproj.so -> libproj.so.25
> libproj.so.25 -> libproj.so.9.0.0
> libproj.so.9.0.0
> ```

`libproj.so` is also a symlink to `libproj.so.9.0.0`.

> Was too much from `HOWTO-RELEASE` chopped?

Yes. SOVERSION only needs to be incremented if the ABI is not backwards compatible.

> Rephrasing suggestions welcome.

```
For each release version, increment PROJ_SOVERSION in CMakeLists.txt if the 
ABI is not backwards compatible, i.e. any interfaces have been removed, or 
changed since the last release. See also: 
https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B
```

sebastic:
> One caution is that libproj.so.9.0.0 was a library name from proj-4.9.1 from 2015-03-04. Is this a potential issue?

Yes, as also raised by Even in https://lists.osgeo.org/pipermail/proj/2022-February/010553.html

It should be rare to have such old and new releases on the same system, but it cannot be ruled out.

If you want to avoid such collisions, keep the SOVERSION in VERSION, e.g.:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9bd4c1f01..9076f955c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,13 +111,8 @@ message(STATUS "Configuring PROJ:")
 ################################################################################
 include(ProjVersion)
 proj_version(MAJOR 9 MINOR 0 PATCH 0)
-# Use libtool convention to build the CMake's VERSION and SOVERSION
-# See https://github.com/pvanhoof/dir-examples#cmake-in-the-cmake-example
-set(PROJ_LIBTOOL_CURRENT    25)
-set(PROJ_LIBTOOL_REVISION   0)
-set(PROJ_LIBTOOL_AGE        0)
-math(EXPR PROJ_API_VERSION "${PROJ_LIBTOOL_CURRENT} - ${PROJ_LIBTOOL_AGE}")
-set(PROJ_BUILD_VERSION "${PROJ_API_VERSION}.${PROJ_LIBTOOL_AGE}.${PROJ_LIBTOOL_REVISION}")
+set(PROJ_SOVERSION 25)
+set(PROJ_BUILD_VERSION "${PROJ_SOVERSION}.${PROJ_VERSION}")
 
 ################################################################################
 # Build features and variants
diff --git a/src/lib_proj.cmake b/src/lib_proj.cmake
index 93b3f9ecf..1bfaae49c 100644
--- a/src/lib_proj.cmake
+++ b/src/lib_proj.cmake
@@ -416,7 +416,7 @@ else()
   set_target_properties(proj
     PROPERTIES
     VERSION "${${PROJECT_NAME}_BUILD_VERSION}"
-    SOVERSION "${${PROJECT_NAME}_API_VERSION}"
+    SOVERSION "${${PROJECT_NAME}_SOVERSION}"
     CLEAN_DIRECT_OUTPUT 1)
 endif()
 
```
Which will result in:
```
libproj.so -> libproj.so.25.9.0.0
libproj.so.25 -> libproj.so.25.9.0.0
libproj.so.25.9.0.0
```

mwtoews:
I'd prefer a simple version, but am willing to test out scenarios where an mismatched library is dynamically picked. How would this go? For instance, CentOS 7 EPEL has [proj-4.8.0](https://centos.pkgs.org/7/epel-x86_64/proj-4.8.0-4.el7.x86_64.rpm.html) with libproj.so.0.7.0. Would I need to find a binary distribution that has a specific soversion, like proj-4.9.1, like Slackware 14.1's [SlackPack](https://slackware.pkgs.org/14.1/slackpack-x86_64/proj-4.9.1-x86_64-1gds.txz.html) with `/usr/lib64/libproj.so.9.0.0`?

kbevers:
In what real world scenario can we expect a user to have both PROJ 4.9.1 *and* PROJ 9.0.0 installed at the same time? Seems to me like a theoretical problem only but I certainly don't know all the weird setups found out in the wild.

dg0yt:
> In what real world scenario can we expect a user to have both PROJ 4.9.1 _and_ PROJ 9.0.0 installed at the same time? Seems to me like a theoretical problem only but I certainly don't know all the weird setups found out in the wild.

It is 9.0.0 now but it will come back for years. There is PROJ 4.9.3 with `libproj.so.12.0.0` on Ubuntu 18.04 (LTS).
AFAIU the principle is to *never* decrement the soversion.

sebastic:
> AFAIU the principle is to _never_ decrement the soversion.

The dotted integer version in the library filename is not the SOVERSION (that is a single integer).

Refer this post for details: https://lists.osgeo.org/pipermail/proj/2022-February/010551.html

See also:
 * https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html
 * https://cmake.org/cmake/help/latest/prop_tgt/VERSION.html

kbevers:
> It is 9.0.0 now but it will come back for years. There is PROJ 4.9.3 with libproj.so.12.0.0 on Ubuntu 18.04 (LTS).
AFAIU the principle is to never decrement the soversion.

Again, what is a *real world* scenario where someone will have co-existing versions of PROJ 4.9.3 *and* 12.0.0? Not that PROJ 12.0.0 is very likely 10+ years in the future.

I'm preparing 9.0.0RC2 tonight (CET) and for now I'm leaning heavily towards the solution presented in this PR which will result in `libproj.so.9.0.0`. Now is the time to come with the good arguments against it.

sebastic:
> Again, what is a real world scenario where someone will have co-existing versions of PROJ 4.9.3 and 12.0.0?

Once can imagine academic projects where differences between versions during the PROJ lifetime are studied.

The fix to avoid this issue is simple, prepend the SOVERSION in the library filename as suggested in: https://github.com/OSGeo/PROJ/pull/3066#issuecomment-1044159912

mwtoews:
I've committed another change to prepend the SOVERSION to the filename, but I see slightly different symlinks than described in https://github.com/OSGeo/PROJ/pull/3066#issuecomment-1044159912. I get the following:
```
libproj.so -> libproj.so.25
libproj.so.25 -> libproj.so.25.9.0.0
libproj.so.25.9.0.0 (actual library, 4.9M)
```
Is this correct @sebastic ?

While it's a rare library name clash (and folks that are mixing PROJ versions that span 7+ years probably have bigger issues), I don't mind prepending SOVERSION to the library filename as suggested.

sebastic:
> Is this correct @sebastic ?

Yes, that's a CMake thing.

MapServer (built with CMake) does the same:
```
libmapserver.so -> libmapserver.so.2
libmapserver.so.2 -> libmapserver.so.7.6.4
libmapserver.so.7.6.4
```
Whereas GDAL (built with Autotools) has:
```
libgdal.so -> libgdal.so.30.0.1
libgdal.so.30 -> libgdal.so.30.0.1
libgdal.so.30.0.1
```

PR #3070, opened 2022-02-19
Fix nullptr dereference in utilities when argv[0] == NULL

rouault:
https://lwn.net/Articles/?offset=50 was an entertaining reading where we
learn that the fact that argv[0] contains the name of the binary is
purely a convention, normally taken by the shell that launches the
process, but not guaranteed by the execve() system call that does the
job.

The following test program tested against cct, cs2cs, geod, gie and proj
make them cause a null pointer dereference

```
 #include <unistd.h>
 #include <stdio.h>
extern char **environ;

int main()
{
    char* argv[] = { NULL };
    printf("%d\n", execve("/path/to/some/proj/binary", argv, environ));
    return 0;
}
```
CC @busstoptaktik @cffk  I'm sure you'll find that amusing :-)

### DIFF
--- a/src/apps/cs2cs.cpp
+++ b/src/apps/cs2cs.cpp
@@ -346,6 +346,10 @@ int main(int argc, char **argv) {
     int have_to_flag = 0, inverse = 0;
     int use_env_locale = 0;
 
+    if( argc == 0 ) {
+        exit(1);
+    }
+
     /* This is just to check that pj_init() is locale-safe */
     /* Used by nad/testvarious */
     if (getenv("PROJ_USE_ENV_LOCALE") != nullptr)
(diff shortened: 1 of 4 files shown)


cffk:
Good to get rid of obvious vulnerabilities!

cffk:
Given that we're throwing accepted conventions out, and given that `argc` is an `int`, shouldn't the `argc == 0` test be replaced with `argc < 0`?

rouault:
> shouldn't the `argc == 0` test be replaced with `argc < 0`?

that seems unnecessary. argc is not under the controlled of the user execve()'ing PROJ utilities but filled by the operating system while it iterates over argv[] until it finds a null pointer. And if the OS would fill it with a negative value, that would be a OS bug.

PR #3072, opened 2022-02-22
Fix various issues found by Coverity scan

rouault:


src/iso19111/io.cpp
```diff
@@ -4072,7 +4072,7 @@ WKTParser::Private::buildProjectedCRS(const WKTNodeNNPtr &node) {
             props.set(IdentifiedObject::NAME_KEY, "WGS 84 / UPS North (E,N)");
         } else if (projCRSName == "UPS_South") {
             props.set(IdentifiedObject::NAME_KEY, "WGS 84 / UPS South (E,N)");
-        } else {
+        } else if (cartesianCS) {
```
hobu, line 4075: Do you still need a final else clause to catch things or error?

hobu (approved):

src/iso19111/io.cpp
rouault: the lack of cartesianCS will be eventually be caught later in the method at line 4212

PR #3073, opened 2022-02-23
CMake + MacOSX + internal gtest: bump minimum version to 3.0 to avoid CMP0042 warning related to MACOSX_RPATH

rouault:


mwtoews:
I've tried this same trick too, but it didn't seem to work for me either.

rhuijben:
```
CMP0042
-------

``MACOSX_RPATH`` is enabled by default.

CMake 2.8.12 and newer has support for using ``@rpath`` in a target's install
name.  This was enabled by setting the target property
``MACOSX_RPATH``.  The ``@rpath`` in an install name is a more
flexible and powerful mechanism than ``@executable_path`` or ``@loader_path``
for locating shared libraries.

CMake 3.0 and later prefer this property to be ON by default.  Projects
wanting ``@rpath`` in a target's install name may remove any setting of
the ``INSTALL_NAME_DIR`` and ``CMAKE_INSTALL_NAME_DIR``
variables.

This policy was introduced in CMake version 3.0.  CMake version
3.20.21032501-MSVC_2 warns when the policy is not set and uses ``OLD`` behavior.  Use
the ``cmake_policy()`` command to set it to ``OLD`` or ``NEW`` explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
```

rhuijben:
This warning is now explicitly turned into an error for developers on this pull request. So the warning should be resolved by making the behavior explicit, like the documentation says.

rouault:
> This warning is now explicitly turned into an error for developers on this pull request. So the warning should be resolved by making the behavior explicit, like the documentation says.

Nope, that doesn't make any difference.

I don't think we can do anything about that warning because the root reason must be that https://github.com/google/googletest/blob/release-1.11.0/CMakeLists.txt has a "cmake_minimum_required(VERSION 2.8.12)" which overrides anything we can set before.   The master branch of googletest has bumped it to 3.5, so when we update to 1.12.0 that should be resolved.

Closing without merging

PR #3084, opened 2022-03-04
CI: use `ctest --output-on-failure` option

mwtoews:
Use [ctest](https://cmake.org/cmake/help/v3.9/manual/ctest.1.html) option `--output-on-failure`. It's usually preferred over the verbose `-V` option, as it only shows verbose output on failure.


mwtoews:
MSYS2 failures are unrelated, see #3085

PR #3099, opened 2022-03-08
CMake: fix installation of proj.pc on Windows (fixes #3078)

### LINKED ISSUE #3078: CMake: proj.pc is installed to system root dir under windows
It seems that ```CMAKE_INSTALL_LIBDIR``` can be undefined when the build system reaches the proj.pc install step.  This results in it being created as ```C:\pkgconfig\proj.pc```, and no proj.pc is installed in the target directory.

Adding a defined check to CMakeLists.txt is a workaround (see below) but perhaps there is a better way.

https://github.com/OSGeo/PROJ/blob/65c84b6e2a40ec8e6b8d1de86bcd556d01ef2aeb/CMakeLists.txt#L326-L333

Workaround (cribbed from [here](https://opensource.apple.com/source/X11libs/X11libs-60.1/libpng14/libpng-1.4.8/CMakeLists.txt.auto.html)):
```cmake
# pkg-config support
configure_proj_pc()

if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
  set(CMAKE_INSTALL_LIBDIR lib)
endif(NOT DEFINED CMAKE_INSTALL_LIBDIR)

install(FILES
  ${CMAKE_CURRENT_BINARY_DIR}/proj.pc
  DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
```

For context, the [build system](https://github.com/shawnlaffan/perl-alien-proj) I'm using is part of the stack used for the GDAL Perl bindings.  It uses [Alien::Build](https://metacpan.org/pod/Alien::Build), which runs the various build steps through an msys shell so the problem is not strictly windows.  This issue cropped up when porting from autotools to cmake.

I can work up a PR if needed.

Shawn.

rouault:


rouault:
hum, from the log of a build *before* the change (https://github.com/OSGeo/PROJ/runs/5470513722?check_suite_focus=true), proj.pc was correctly installed in D:/a/PROJ/PROJ/proj_dir/lib/pkgconfig/proj.pc

@shawnlaffan could you confirm if this pull request improves things for you

But I notice there's another issue on Windows with man pages being installed in  /man1/ ( although man/CMakeLists.txt has DESTINATION share/man/man1)

shawnlaffan:
Thanks @rouault 

This fixes the pkgconfig issue I reported in https://github.com/OSGeo/PROJ/issues/3078

shawnlaffan:
And FWIW, the man1 dir for my builds is at .../share/man/man1, not at the system root level.

dg0yt:
I would prefer if  `GNUInstallDirs` would be used consistently. It is the only common denominator for configuring packages in the CMake world. (It also facilitates contributions.)

rouault:
> I would prefer if `GNUInstallDirs` would be used consistently.

agreed, but too substantial to attack for a bug fix. I've filed that as https://github.com/OSGeo/PROJ/issues/3100

Issue #2318, opened 2020-08-10
Z-Coordinates not converted in newer versions for EPSG:31370 to EPSG:4326


aaron012:
**Description**
When converting coordinates between  EPSG:31370 and EPSG:4326 the z-coordinates are not converted anymore. In older versions of the library, this was the case. Another coordinate system that we found where this happened is EPSG:2039. 

**Minimal example**:
```bash
docker run osgeo/proj:5.2 sh -c "echo 171160.576105325 172833.962098988 -33.041749736 | cs2cs -f \"%.9f\" +init=epsg:31370 +to +init=epsg:4326"
# Output: 4.669339658	50.865500371 10.000165402

docker run osgeo/proj:7.1.0 sh -c "echo 171160.576105325 172833.962098988 -33.041749736 | cs2cs -f \"%.9f\" +init=epsg:31370 +to +init=epsg:4326"
# Output: 4.669339464	50.865500746 -33.041749736
```

As you can see in the example in version 5.2 the z-coordinate is converted from -33.041749736 to 10.000165402. In version 7.1.0 the z-coordinate stays exactly the same as in the input.

rouault:
EPSG:31370 and EPSG:4326 are 2D CRS. PROJ >= 6 will transform the Z component only if the two CRS have a Z component

Such a transformation can be used by first researching available transformations with:
```
projinfo -s EPSG:31370 -t EPSG:4979 --3d -o PROJ
```
and, for example selecting the second one listed in the above output:
```
echo "171160.576105325 172833.962098988 -33.041749736" | cct +proj=pipeline +step +inv +proj=lcc +lat_0=90 +lon_0=4.36748666666667 +lat_1=51.1666672333333 +lat_2=49.8333339 +x_0=150000.013 +y_0=5400088.438 +ellps=intl +step +proj=cart +ellps=intl +step +proj=helmert +x=-106.8686 +y=52.2978 +z=-103.7239 +rx=-0.3366 +ry=0.457 +rz=-1.8422 +s=-1.2747 +convention=coordinate_frame +step +inv +proj=cart +ellps=WGS84 +step +proj=unitconvert +xy_in=rad +z_in=m +xy_out=deg +z_out=m +step +proj=axisswap +order=2,1
```

aaron012:
Thanks for the very quick response! Then it is indeed not a bug, my apologies. 
I do not completely understand the conversion yet, since EPSG:31370 is still a 2D CRS in the projinfo command? 
[...]

Issue #1563, opened 2019-08-09
geoidgrids not found


snowman2:
[GeodPT08.gtx](https://files.gitter.im/pyproj4-pyproj/community/YSlQ/GeodPT08.gtx)

Reference:
- https://gitter.im/pyproj4-pyproj/community?at=5d4add2ebb5364334cec4556
- https://gitter.im/pyproj4-pyproj/community?at=5d4cbf00d03a7c63e6f42609

```
$ cs2cs
Rel. 5.2.0, September 15th, 2018
<cs2cs>: 
...
$ echo -19476.384 -62016.159 4226.997 | cs2cs +init=epsg:3763 +to +init=epsg:3763 +geoidgrids=./GeodPT08.gtx
-19476.38    -62016.16 4172.78
```
```
$ cs2cs
Rel. 6.1.1, July 1st, 2019
...
$ echo -19476.384 -62016.159 4226.997 | cs2cs +init=epsg:3763 +to +init=epsg:3763 +geoidgrids=./GeodPT08.gtx
-19476.38    -62016.16 4227.00
$ echo -19476.384 -62016.159 4226.997 | cs2cs +init=epsg:3763 +to +init=epsg:3763
-19476.38    -62016.16 4227.00

```

Any guidance on the issue would be appreciated. Thanks!

Issue #3075, opened 2022-02-25
Support `AUTH:nnnnn+ellipsoidal` syntax in `osgeo::proj::io::createFromUserInput`


hernando:
I would like to pitch an idea for extending the behavior of `osgeo::proj::io::createFromUserInput` that I would offer to implement if it gets accepted.
Quite often I find myself having to write things like `cs2cs "$(projinfo -q -o WKT2:2019 --single-line --3d EPSG:32632)" "$(projinfo -q -o WKT2:2019 --single-line --3d EPSG:2056)" -d 12` in order to get the vertical axis transformed between CRSs with different ellipsoids. I'm aware that this doesn't always make sense (e.g. specially with OSGB36), but in the previous example it's perfectly valid as far as I know.
Since there's already an option for 3D promotion in `projinfo`, I believe it would be handy to make the same functionality available through the CRS identifiers accepted by `createFromUserInput`, so the example above would become: `cs2cs EPSG:32632+ellipsoidal EPSG:2056+ellipsoidal`.

The open question to me is how to handle extensions to 3D of projected CRSs in units other than meter. I know that the `promoteTo3D`member function always adds the z axis in meter regardless of the unit of the other axes. I'm fine with this behaviour, however some users may find it surprising when writing something like EPSG:2234+ellipsoidal, where EPSG:2234 is a projected CRS using ftUS.

hernando:
Sorry for being pushy, but I would like to rely on this behavior for something else than just a handy shortcut in the command line, so if it's an absolute no go, I would like to know to find another solution to my problem.

kbevers:
Seems to me that implementing the `--3d` option in  `cs2cs` would solve your problem. 

With `projinfo` we can get a pipeline with almost the same syntax as would be passed to `cs2cs`:
```
$ projinfo -s EPSG:32632 -t EPSG:2056 -o PROJ --3d
Candidate operations found: 1
-------------------------------------
Operation No. 1:

unknown id, Inverse of UTM zone 32N + Inverse of CH1903+ to WGS 84 (1) + Swiss Oblique Mercator 1995, 1 m, Liechtenstein; Switzerland.

PROJ string:
+proj=pipeline
  +step +inv +proj=utm +zone=32 +ellps=WGS84
  +step +proj=cart +ellps=WGS84
  +step +proj=helmert +x=-674.374 +y=-15.056 +z=-405.346
  +step +inv +proj=cart +ellps=bessel
[...]

PR #3119, opened 2022-03-16
Transformation: no longer do vertical trasnformation when doing compound CRS to 2D CRS / add --3d to cs2cs

rouault:
Previously, when computing transformation between a compound CRS and a
geographic/projected 2D CRS, the behaviour was similar to implicitly
promoting the 2D CRS to 3D CRS in the pipeline computation logic, hence
a geoid model could be applied. But note that when doing a geographic 3D
to geographic/projected 2D CRS transformation, we *did* not do this implicit
promotion and if a Helmert transformation existed between the datums, it
was done only in 2D. So this is a bit inconsistent and triggered the
comment in https://github.com/OSGeo/PROJ/issues/2318#issuecomment-1068924513

With this commit, we no longer do any vertical transformation when doing
compound CRS to the 2D CRS, but just take the transformation of the
horizontal part of the compound CRS to the 2D CRS.
Said otherwise, NAD83+NAVD88 to NAD83 will no longer lead to the
application of the geoid model. Unless you explicitly ask for the
promotion NAD83 to 3D.

Also related, in https://github.com/OSGeo/PROJ/issues/1563 that went to 6.3.0,
I changed cs2cs to automatically promote to 3D the CRS as soon as one of
them was compound, for the sake of being consistent with the past
behaviour. But it then becomes difficult to predict PROJ behaviour
depending on which level of it you consider...
This commit undoes that and adds an explicit --3d switch to cs2cs, similarly to
projinfo, to ask for promotion.

Other bug fix found in the process, when using legacy syntax, +init=epsg:4979,
(WGS 84 3D), the resulting CRS was 2D and not 3D.

CC @kbevers @hobu @snowman2  : I'd be interested in hearing if you think this change of behaviour is a good idea. I think it is valuable as it makes PROJ behavior more predictable (and "correct"), but this has a price w.r.t backward compatibility. I believe this is material only candidate for 9.1, not 9.0.x (well, we could possibly backport the addition of the --3d switch to 9.0.x, but not the suppression of the automatic promotion, so that people can anticipate the future change of behaviour by opting in)


EvertEt:
This looks like a great improvement to the predictability of transformations, thank you!

Is there a way to "demote" the compound CRS to 2D to force this behaviour in older versions? (EPSG or WKT representation would be fine)

kbevers:
I think this change makes sense. I agree that is should only go to 9.1.x.

hobu:
> Said otherwise, NAD83+NAVD88 to NAD83 will no longer lead to the application of the geoid model. Unless you explicitly ask for the promotion NAD83 to 3D.

What are some examples of the syntax and shorthand to do the promotion? We use `EPSG:NAD83+NAVD88` quite a bit.

rouault:
> EPSG:NAD83+NAVD88

this is invalid, but "NAD83 + NAVD88 height" or "EPSG:4269+5703" are strings recognized by proj_create().

> What are some examples of the syntax and shorthand to do the promotion? 

There's no shorthand name to do the promotion (was discussed recently in https://github.com/OSGeo/PROJ/issues/3075). Promotion only makes sense for geographic 2D CRS (compound CRS are already 3D by definition). You need to use the proj_crs_promote_to_3D() C API, or the --3d switch of projinfo and (just added in this PR) cs2cs, or the WKT resulting from "projinfo --3d NAD83 -o WKT2:2019 -q"

hobu:
This change means `EPSG:4326+5703` or `EPSG:26915+5703` will not magically promote anymore, correct?

rouault:
> This change means `EPSG:4326+5703` or `EPSG:26915+5703` will not magically promote anymore, correct?

Those CRSs are compoundCRS. They are already 3D. Promoting them is a no-op. Promotion only makes sense for a 2D geographic or projected CRS, to add an ellipsoidal height to it.
The change affects trying to transform such a compoundCRS to a target 2D geographic or projected CRS. Before the change, that target CRS was implicitly promoted to 3D. With the change, it is no longer ==> no vertical transformation is done.

snowman2:
This makes sense to me 👍

PR #3125, opened 2022-03-19
Fix datum names when importing from PROJ4 crs strings (affects some transformations using geoidgrids)

rouault:
When importing from a PROJ4 string with +towgs84, +nadgrids or
+geoidgrids terms, the datum name was just set to 'unknown'. So for
example the datums of '+proj=longlat +ellps=GRS80 +towgs84=1,2,3' and
'+proj=longlat +ellps=GRS80 +towgs84=4,5,6' were considered identical,
because they had the same name 'unknown' and used the same ellipsoid.
This affected the transformation of such CRS when they had an
additional +geoidgrids term, which resulted in an erroneous +proj=push
+v_1 +v_2 step to be added to preserve the horizontal coordinates.

### DIFF
--- a/src/iso19111/io.cpp
+++ b/src/iso19111/io.cpp
@@ -9200,13 +9200,22 @@ PROJStringParser::Private::buildDatum(Step &step, const std::string &title) {
     PrimeMeridianNNPtr pm(buildPrimeMeridian(step));
     PropertyMap grfMap;
 
+    const auto &nadgrids = getParamValue(step, "nadgrids");
+    const auto &towgs84 = getParamValue(step, "towgs84");
+    std::string datumNameSuffix;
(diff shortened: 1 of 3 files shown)


kbevers:
> When importing from a PROJ4 string with +towgs84, +nadgrids or
+geoidgrids terms, the datum name was just set to 'unknown'.

What are the datum names set to after this change?

rouault:
> What are the datum names set to after this change?

"unknown using nadgrids=...." , "unknown using towgs84=...."  , "unknown using geoidgrid=...."

kbevers:
>  "unknown using nadgrids=...." , "unknown using towgs84=...." , "unknown using geoidgrid=...."

Sounds reasonable to me

PR #3132, opened 2022-03-23
Steps with omit_inv are not considered when checking if inverse pipeline operation is possible

oskaritimperi:
Hello!

While testing if we could move from our internal coordinate transformation library to using pure PROJ, I came upon one case where it would be beneficial to be able to make PROJ proceed if affine transformation is not invertible but the inverse is not going to be used.

I'm was a bit suprised that this didn't work. You would think that it's okay that an affine transformation cannot be run in reverse if it's explicitly specified that the step should be skipped when doing inverse operation.

I'm not sure if this is a good idea or not. I would like some feedback. If this is a good idea, would this need documenting?

 - [x] Tests added

### DIFF
--- a/src/pipeline.cpp
+++ b/src/pipeline.cpp
@@ -550,7 +550,7 @@ PJ *OPERATION(pipeline,0) {
     /* determine if an inverse operation is possible */
     for( auto& step: pipeline->steps) {
         PJ *Q = step.pj;
-        if ( pj_has_inverse(Q) ) {
+        if ( step.omit_inv || pj_has_inverse(Q) ) {
             continue;
         } else {
             P->inv   = nullptr;
--- a/test/gie/more_builtins.gie
+++ b/test/gie/more_builtins.gie
@@ -879,4 +879,30 @@ expect      5.875      55.375      0
 -------------------------------------------------------------------------------
 
 
+
+-------------------------------------------------------------------------------
+# Test inverse pipeline with noninvertible affine transformation. We define an
+# affine transformation that's not invertible and omit the inverse path. We
+# also define a step that handles the inverse and omits the forward path.
+#
+# Because the first affine matrix is not invertible, we specify omit_inv to
+# disable executing in reverse.
+#
+# Now we need another affine to actually make the inverse operation.
+#
+# The second affine has omit_fwd and inv. We don't execute this when running
+# forwards. When doing inverse, we call the inverse of the inverted affine which
+# means that we actually do the forward of the specified affine. As it happens,
+# the forward of this affine actually undoes what the first affine did in forward
+# path.
+-------------------------------------------------------------------------------
+
+-------------------------------------------------------------------------------
+operation   proj=pipeline step proj=affine omit_inv s11=2 s22=0 s33=0 step proj=affine omit_fwd inv s11=0.5 s22=0 s33=0
+-------------------------------------------------------------------------------
+
+accept      10 0 0
+expect      20 0 0
+roundtrip
+
 </gie-strict>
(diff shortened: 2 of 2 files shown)


src/pipeline.cpp
```diff
@@ -550,7 +550,7 @@ PJ *OPERATION(pipeline,0) {
     /* determine if an inverse operation is possible */
     for( auto& step: pipeline->steps) {
         PJ *Q = step.pj;
-        if ( pj_has_inverse(Q) ) {
+        if ( step.omit_inv || pj_has_inverse(Q) ) {
```
rouault, line 553: I'm wondering if that shouldn't be:
```suggestion
        if ( (!Q->inverted && step.omit_inv) || (Q->inverted && step.omit_fwd) || pj_has_inverse(Q) ) {
```
but I'm not 100% positive as I don't recall whet "+proj=something +inv +omit_inv" is actually supposed to do...

kbevers: > "+proj=something +inv +omit_inv" is actually supposed to do...

See https://proj.org/operations/pipeline.html#cmdoption-arg-omit_fwd. Basically, in certain circumstances it can be useful to skip a step in a pipeline, for instance if you need to treat the forward and inverse differenctly as explained in the docs

rouault: > See https://proj.org/operations/pipeline.html#cmdoption-arg-omit_fwd

yes I know. My unknown was on the double negation aspect +inv +omit_inv, which isn't covered in the example, and that in the forward and inverse paths...

So experimenting with:
$ echo 0 0 0 | bin/cct  +proj=pipeline +step +inv +proj=geogoffset +dlon=3600 +omit_inv
 -1.0000000000    0.0000000000        0.0000           inf

it means that in the forward direction, the inverse step is done (and thus +omit_inv without effect).

and when adding -I (triple negation):

$ echo 0 0 0 | bin/cct -I +proj=pipeline +step +inv +proj=geogoffset +dlon=3600 +omit_inv
  0.0000000000    0.0000000000        0.0000           inf

so +omit_inv specific effect isn't affected by the presence or absence of the +inv flag of the step.

which *probably* means that my above suggested change is not correct (my head is still hurting)

oskaritimperi: Here's a table of the combinations

| flags | invertible step forward path | invertible step inverse path | non-invertible step forward path | non-invertible step inverse path |
| --- | --- | --- | --- | --- |
| +omit_fwd | omit | inv | omit | undefined |
| +omit_fwd +inv | omit | fwd | omit | fwd |
| +omit_inv | fwd | omit | fwd | **!** (should be omit) |
| +omit_inv +inv | inv | omit | undefined | omit |
| | fwd | inv | fwd | undefined |
| +inv | inv | fwd | undefined | fwd |

The one bold exclamation mark there is the one that's not currently working and what this patch is trying to solve. The step doesn't have a valid inverse, but the inverse path is omitted anyway so it should be OK. Currently the pipeline code marks the whole pipeline as not invertible because of this.

The undefined cells are operations that would try to run the inverse operation when the inverse isn't defined.

I think I can use this table to write gie tests for each of the table cells.

kbevers: > I think I can use this table to write gie tests for each of the table cells.

That would certianly be helpful. Especially if you add comments that explain what's going on. All these negations can be very difficult to wrap your head around.

oskaritimperi: I've added more tests based on the table and I tried to add comments describing what's being tested. I agree that it's very hard to think about multiple negations.

rouault:
Thanks for those very complete tests added!

PR #3133, opened 2022-03-23
Add invertible Horner polynomials

georgeouzou:
This enables the user to create horner polynomial transformations
in the real number space without the need to supply "inv" parameters:
"+inv_origin", "+inv_u" and "+inv_v". We can get the inverse
transformation by only using the "fwd" parameters and solving
a system of equations iteratively. The system is usually solved in a
very small number of iterations.

A new optional parameter "+inv_tolerance" is added that can make the
iterative process stop earlier if a certain coordinate accuracy is
met. For example "+inv_tolerance=0.1" will make the iterative
process stop when the calculated coordinates differ less than
0.1 meters, if the units of the transformation destination is meters.

 - [x] Tests added

### DIFF
--- a/docs/source/operations/transformations/horner.rst
+++ b/docs/source/operations/transformations/horner.rst
@@ -5,7 +5,7 @@ Horner polynomial evaluation
 ================================================================================
 
 .. versionadded:: 5.0.0
-.. versionchanged:: 9.1.0 Iterative real polynormal inversion
+.. versionchanged:: 9.1.0 Iterative polynormal inversion
 
 +-----------------+-------------------------------------------------------------------+
 | **Alias**       | horner                                                            |
@@ -126,7 +126,7 @@ Evaluation of the complex polynomials are defined by the following equations:
 
 Where :math:`n` is the degree of the polynomial. :math:`U` and :math:`V` are
 defined as in :eq:`UV` and the resulting coordinates are again determined
-by :eq:`xy_out`.
+by :eq:`xy_out`. Complex polynomials can be solved iteratively similar to real polynomials.
 
 Examples
 ################################################################################
@@ -185,10 +185,6 @@ describing real and complex polynomials can't be mixed.
 
     Coordinate of origin for the forward mapping
 
-.. option:: +inv_origin=<northing,easting>
-
-    Coordinate of origin for the inverse mapping. Not required for iterative inversion.
-
 Real polynomials
 ..............................................................................
 
@@ -228,13 +224,16 @@ of the polynomial:
     Coefficients for the complex forward transformation
     as described in :eq:`complex_poly`.
 
-.. option:: +inv_c=<c_1,c_2,...,c_N>
-
-    Coefficients for the complex inverse transformation
-    as described in :eq:`complex_poly`.
-
 Optional
 -------------------------------------------------------------------------------
+
+.. option:: +inv_origin=<northing,easting>
+
+    .. versionchanged:: 9.1.0
+
+    Coordinate of origin for the inverse mapping.
+    Without this option iterative polynomial evaluation is used for
+    the inverse tranformation.
 
 .. option:: +inv_u=<u_11,u_12,...,u_ij,..,u_mn>
 
@@ -242,7 +241,7 @@ Optional
 
     Coefficients for the inverse transformation i.e. latitude to northing
     as described in :eq:`real_poly`. Only applies for real polynomials.
-    Without this option iterative real polynomial evaluation is used for
+    Without this option iterative polynomial evaluation is used for
     the inverse tranformation.
 
 .. option:: +inv_v=<v_11,v_12,...,v_ij,..,v_mn>
@@ -251,7 +250,16 @@ Optional
 
     Coefficients for the inverse transformation i.e. longitude to easting
     as described in :eq:`real_poly`. Only applies for real polynomials.
-    Without this option iterative real polynomial evaluation is used for
+    Without this option iterative polynomial evaluation is used for
+    the inverse tranformation.
+
+.. option:: +inv_c=<c_1,c_2,...,c_N>
+
+    .. versionchanged:: 9.1.0
+
+    Coefficients for the complex inverse transformation
+    as described in :eq:`complex_poly`. Only applies for complex polynomials.
+    Without this option iterative polynomial evaluation is used for
     the inverse tranformation.
 
 .. option:: +range=<value>
@@ -270,7 +278,7 @@ Optional
 
     .. versionadded:: 9.1.0
 
-    Only applies to real polynomials and iterative inversion.
+    Only applies to cases of iterative inversion.
     The procedure converges to the correct results with each step.
     Iteration stops when the result differs from the previous calculated
     result by less than <value>.
--- a/docs/source/operations/transformations/horner.rst
+++ b/docs/source/operations/transformations/horner.rst
@@ -5,7 +5,7 @@ Horner polynomial evaluation
 ================================================================================
 
 .. versionadded:: 5.0.0
-.. versionadded:: 9.1.0 Iterative real polynormal inversion
+.. versionchanged:: 9.1.0 Iterative real polynormal inversion
 
 +-----------------+-------------------------------------------------------------------+
 | **Alias**       | horner                                                            |
@@ -66,9 +66,6 @@ coefficients. If only the forward set of coefficients and origin is known the in
 be done by iteratively solving a system of equations. By writing :eq:`real_poly` as:
 
 .. math::
-    :label: real_poly_iterative_inversion
-
-    \begin{align}
         \begin{bmatrix}
             \Delta X \\
             \Delta Y \\
@@ -84,6 +81,12 @@ be done by iteratively solving a system of equations. By writing :eq:`real_poly`
         \begin{bmatrix}
             U \\
             V \\
+        \end{bmatrix} \\
+
+.. math::
+        \begin{bmatrix}
+            \Delta X \\
+            \Delta Y \\
         \end{bmatrix} =
         \begin{bmatrix}
             u_{0,0} \\
@@ -96,10 +99,9 @@ be done by iteratively solving a system of equations. By writing :eq:`real_poly`
         \begin{bmatrix}
             U \\
             V \\
-        \end{bmatrix}
-    \end{align}
+        \end{bmatrix} \\
 
-    \begin{align}
+.. math::
         \begin{bmatrix}
             U \\
             V \\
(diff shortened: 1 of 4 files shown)


georgeouzou:
I will also try to create docs for this.

georgeouzou:
I also believe there are errors in the [horner documentation](https://proj.org/operations/transformations/horner.html)

- Input and Output types can be Geodetic _or_ projected coordinates
- Equation (3) seems wrong as the final coordinates are not calculated by adding the input X,Y.
- Equation (1) seems also a little off as the  Σ u<sub>i,j</sub> U<sup>i</sup> V<sup>j</sup> needs to be  Σ u<sub>i,j</sub> U<sup>j</sup> V<sup>i</sup>
- u and v in +fwd_u, +fwd_v, +inv_u, +inv_v, +origin_u, +origin_v can express northing/easting, easting/northing, latitude/longitude,  longitude/latitude etc... Parameter documentation implies that  u, v expresses northing/easting.

docs/source/operations/transformations/horner.rst
```diff
@@ -206,6 +259,14 @@ Optional
 
     Express longitude as westing. Only applies for complex polynomials.
 
+.. option:: +inv_tolerance=<value>
+
```
kbevers, line 281:
```suggestion
.. versionadded:: 9.1.0

```

docs/source/operations/transformations/horner.rst
```diff
@@ -5,6 +5,7 @@ Horner polynomial evaluation
 ================================================================================
 
 .. versionadded:: 5.0.0
+.. versionadded:: 9.1.0 Iterative real polynormal inversion
```
kbevers:
```suggestion
.. versionchanged:: 9.1.0 Iterative real polynormal inversion
```

docs/source/operations/transformations/horner.rst
```diff
@@ -133,7 +186,7 @@ describing real and complex polynomials can't be mixed.
 
 .. option:: +inv_origin=<northing,easting>
 
-    Coordinate of origin for the inverse mapping
+    Coordinate of origin for the inverse mapping. Not required for iterative inversion.
```
kbevers: Perhaps these are better placed under optional parameters with this change in effect? As far as I understand they are only necessary if you are explicitly defining an inverse operation. Same comment applies for `+inv_u` and `+inv_v`

kbevers (changes requested):
I can't really comment on the maths but here's a few suggestions for improving the docs. I hope @busstoptaktik can find the time to give this a proper review at some point. He has the expertise in this part of the code base.

docs/source/operations/transformations/horner.rst
georgeouzou: @kbevers  thanks for the review! For +inv_u, +inv_v, yes they are  optional and i will move them to the optional parameters. As for +inv_origin we still need it in the case of complex polynomials but it is not required in the case of real polynomials. How do you think we can handle this ?

kbevers: Your current wording for`+inv_origin`is probably okay then.

kbevers (commented):
Doc changes looks good to me!

georgeouzou:
Just added complex inversion too :)

src/transformations/horner.cpp
```diff
@@ -99,6 +100,8 @@ struct horner {
     int    order;    /* maximum degree of polynomium */
     int    coefs;    /* number of coefficients for each polynomium  */
     double range;    /* radius of the region of validity */
+    int    has_only_fwd; /* inv parameters are not specified, inverse is done by gauss-newton iteration */
```
busstoptaktik: In almost all cases, when you use `has_only_fwd`, you use its inverse form, i.e. `!has_only_fwd`, so it will probably lead to more straightforward code if you switch the meaning and name of the element to `has_inv`. Also: It probably should be a `bool`, rather than `int`

src/transformations/horner.cpp
```diff
@@ -244,29 +246,93 @@ summing the tiny high order elements first.
...
+                    } else if (i == 0) {
+                        Ma += (*tcx) * pow(x0, j-1);
+                        Md += (*tcy) * pow(y0, j-1);
+                    } else {
+                        Mb += (*tcx) * pow(y0, i-1) * pow(x0, j);
+                        Mc += (*tcy) * pow(x0, i-1) * pow(y0, j);
+                    }
+                    tcx++;
+                    tcy++;
+                }
```
busstoptaktik: This seems a bit odd: The overall name of the procedure is "Horner" for a reason - because we use Horner's scheme for the polynomium evaluations. We do that for two other reasons:

1. Because of the stability of Horner's scheme due to its starting the summation with the smallest (i.e. high order) terms,
2. Because of its speed. On many platforms, the `pow(...)` call is **extremely** slow.

You probably should refactor the block labelled `The melody of this block is straight out of the great Engsager/Poder songbook` below into a separate function, and call that where needed.

src/transformations/horner.cpp
```diff
@@ -343,40 +408,89 @@ polynomial evaluation engine.
...
+            double E = *--c;
+            double N = *--c;
+            while (c > cb) {
+                double w = w0.real()*E + w0.imag()*N + *--c;
+                       N = w0.real()*N - w0.imag()*E + *--c;
+                       E = w;
+            }
+            std::complex<double> det(N, E);
+            std::complex<double> w1 = dZ / det;
+            converged = (fabs(w1.real()-w0.real()) < tol) && (fabs(w1.imag()-w0.imag()) < tol);
```
busstoptaktik, line 479: You probably could use `hypot()` or `(w1 - w0).abs()` here? (note: **could**, not necessarily **should** - you may have your reasons for treating each dimension separately)

src/transformations/horner.cpp
```diff
@@ -343,40 +408,89 @@ polynomial evaluation engine.
...
     } else {                                              /* inverse */
-        cb =  transformation->inv_c;
-        c  =  cb + sz;
-        e  =  position.u - transformation->inv_origin->u;
-        n  =  position.v - transformation->inv_origin->v;
-        if (transformation->uneg)
-            e  =  -e;
-        if (transformation->vneg)
-            n  =  -n;
+        if (!iterative_inverse) {
```
busstoptaktik, line 446: There's a number of these conditional blocks in the code. Originally this was due to the *Poder dual autochecking* approach, where forward and inverse cases were handled by the same code. Now, having 3 different cases singing in different voices in the same code, it begins to look a bit hairy.

Now, while you still have a good grasp of this code, you should consider following up with another pull request (once this one is merged), untangling the three cases into separate code paths/functions.

busstoptaktik (changes requested):
Certainly a useful addition to the Horner code. See, however, my comments regarding the probable un-soundness of evaluating the polynomium by brute force, rather than Horner's scheme.

Also, *IOGP Publication 373-7-2 – Geomatics Guidance Note number 7, part 2 – September 2019* has a very clear section on the criteria for polynomium reversibility. It would probably be a good thing to quote and refer extensively from/to that in the documentation updates. For example, the Geomatics Guidance Note says:

> **Polynomial reversibility**
>
>  Approximation polynomials are in a strict mathematical sense not reversible, i.e. the same polynomial coefficients cannot be used to execute the reverse transformation
>
> However, under certain conditions, described below, a satisfactory solution for the reverse transformation may be obtained using the forward coefficient values in a single step, rather than multiple step iteration. If such a solution is possible, in the EPSG Dataset the polynomial coordinate transformation method is classified as a reversible polynomial of degree n.
> 
> A (general) polynomial transformation is reversible only when the following conditions are met.
> 1. The coordinates of source and target evaluation point are (numerically) the same.
> 2. The unit of measure of the coordinate differences in source and target coordinate reference system are the same.
> 3. The scaling factors applied to source and target coordinate differences are the same.
> 4. The spatial variation of the differences between the coordinate reference systems around any given location is sufficiently small.

Your code solves a more general problem, and hence extends on the EPSG sense of a reversible polynomium - it probably would be a good idea to explicitly state this in the documentation.

src/transformations/horner.cpp
georgeouzou: @busstoptaktik thank you for the detailed review! I was waiting for this comment, i will of course make the optimization. I did this already with the complex case, but i left this in this form to be a little more understandable for the time being.

src/transformations/horner.cpp
georgeouzou: Using hypot or abs may result in a more strict convergence criterion.
For example if tolerance = 1.0 and dw=w1-w0 = 0.9 + 0.9i then hypot(dw) = 1.27  but both real and imag parts are below the tolerance.

georgeouzou: But yes, you are right. So when hypot <= tol, both real and imag will be equal or  less than tol.

busstoptaktik: I agree that it makes strict mathematical sense, and that is probably what matters most for communication purposes. Geodetically, however, the differences in scaling factor absorbed by the polynomals, is related to the historical problem of measuring distances accurately, so the convergence is basically two-dimensional. But as indicated above, I agree that for the iterative convergence the "twice linear" criterion is also reasonable.

src/transformations/horner.cpp
busstoptaktik: Awesome!

georgeouzou:
@busstoptaktik i will make a separate commit for the docs. 
As i understand we implemented the general case of the second bullet of Polynomial reversibility section (page 126)

> 2. By applying the polynomial transformation with the same coefficients but with their signs reversed
and then iterate to an acceptable solution, the number of iteration steps being dependent on the
desired accuracy. (Note that only the signs of the polynomial coefficients should be reversed and not
the coordinates of the evaluation points or the scaling factors!) The iteration procedure is usually
described by the information source of the polynomial transformation.

So, i will update the docs, so as to make it clear this is an iterative inverse transformation and not a "reversible polynomial" as presented in the publication.

Another thing i realized when i read the section of the polynomial transformations in the publication, is that:
- we do not support using scale factors m<sub>S</sub> and m<sub>T</sub>
- we use evaluation points in a different way than the usage presented in the publication. Specifically we do: 
X<sub>T</sub> = ΔX,
Y<sub>T</sub> = ΔY
instead of:
X<sub>T</sub> = X<sub>S</sub> – X<sub>SO</sub> + X<sub>TO</sub> + ΔX,
Y<sub>T</sub> = Y<sub>S</sub> – Y<sub>SO</sub> + Y<sub>TO</sub> + ΔY
where ΔΧ, ΔΥ the result of the polynomium evaluation.

georgeouzou:
@busstoptaktik i added the doc reference to the publication. Also i refactored the code as requested in the review.

busstoptaktik (approved):

kbevers (approved):

PR #3136, opened 2022-03-25
Use Mambaforge variant instead of Miniconda

mwtoews:
The motivation is to speed-up use of conda tools in CI using [mamba](https://github.com/mamba-org/mamba), which has experimental support in [setup-miniconda GitHub Actions](https://github.com/marketplace/actions/setup-miniconda). Mambaforge is a variant of miniforge, which only has a conda-forge channel.

While testing, docs/plot/plot.py crashed due to one invalid geometry in land.geojson, now modified to be valid.

Changes to conda.yml were not implemented, as this is not a CI bottleneck.


kbevers (approved):
I've been using mambaforge for a while now, both on desktop and in GitHub Actions, and have generally been very pleased with the experience

PR #3137, opened 2022-03-25
Fix build issue for ARM64 when using MSVC

anthony-linaro:
When compiling with MSVC for an ARM64 platform using the same configuration options as used in the github workflow, an inaccurate warning is treated as an error due to the "/Wx" option.

This disables the warning for the ARM64 target in the CMake files.

Tested with the CMake options:
* `MAKE_BUILD_TYPE="RELEASE"`
* `BUILD_SHARED_LIBS="ON"`

Using the VS2022 x64 -> ARM64 cross-compilation environment via `vcvarsamd64_arm64`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,11 @@ elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
     /wd4706  # Suppress warning about assignment within conditional expression
     /wd4996  # Suppress warning about sprintf, etc., being unsafe
   )
+  if("$ENV{VSCMD_ARG_TGT_ARCH}" STREQUAL "arm64")
+    # Suppress an inaccurate warning when compiling for an MSVC/ARM64 platform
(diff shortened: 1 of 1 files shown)


hobu:
div0 is an important warning for PROJ. It doesn't seem prudent to be globally shutting it off. I think this shut-off should be conditioned on the fact that the platform is ARM64.

anthony-linaro:
> div0 is an important warning for PROJ. It doesn't seem prudent to be globally shutting it off. I think this shut-off should be conditioned on the fact that the platform is ARM64.

That is what L76, `if("$ENV{VSCMD_ARG_TGT_ARCH}" STREQUAL "arm64")` does

hobu:
> That is what L76, if("$ENV{VSCMD_ARG_TGT_ARCH}" STREQUAL "arm64") does

yep, I missed it on first read and deleted my comment. Sorry for the chatter.

PR #3139, opened 2022-03-25
Add Interrupted Mollweide (land and ocean view) projections

erykoff:
- [x] Closes #3122
 - [x] Tests added
 - [x] Added clear title that can be used to generate release notes
 - [x] Fully documented, including updating `docs/source/*.rst` for new API

This PR is in response to #3122.

I've recently been trying to learn about various map projections and how PROJ works, and this seemed like an interesting exercise.  In this case I went back to read the original Goode (1919) paper which introduced the Interrupted Homolographic Projection (his name for the Interrupted Mollweide Projection) and the Goode (1925) paper which introduced the Interrupted Goode Homolosine Projection (combining Mollweide at high latitude with Sinusoidal at low latitude).  I agree with the oceanographers (such as @lukegre) who think that Goode's original trade-off of equatorial distortion for a discontiguous derivative at +/-40°44'11.8" is not the best aesthetic choice, particularly for the Oceanic View.

One thing I should note is that the built-in plotting code for the `proj` documentation does not do a very good job of rendering interrupted projections.  Attached please find examples of the new Interrupted Mollweide (`imoll`) and Interrupted Mollweide Oceanic View (`imoll_o`) projections as rendered by `proj`+`pyproj`+`cartopy`.
imoll
imoll_o

The code here is adapted from the `igh.cpp` and `igh_o.cpp` source files, keeping to the same style.

PR #3144, opened 2022-03-28
Improve documentation plotting of interrupted projections.

erykoff:
- [ ] Tests added

This is a spin-off of #3139 that improves the documentation plotting code by adding checks for interrupted lines and graticules.  This does not replace the continental outlines with filled continents because that is a harder job.

I also fixed a typo on the `ob_tran` plot so that it now renders as intended according to the description paragraph as an oblique Mollweide rather than oblique Miller Cylindrical, which is neither a common nor attractive projection.


kbevers (approved):
This is great, very nice to get cleaner versions of the problematic figures!!

kbevers:
@erykoff do you mind if I squash this into one commit? All the commits are working towards the same thing so I think it makes more sense to group them in one.

erykoff:
No problem.  I left them all as separate commits in case you had some comments on particular ones.  Otherwise, it's fine to bundle it up.

PR #3158, opened 2022-04-06
createBoundCRSToWGS84IfPossible(): improve selection logic to generate +towgs84= taking into account extent

rouault:
Helps for North Macedonia CRS EPSG:6204 and EPSG:9945

### DIFF
--- a/src/iso19111/crs.cpp
+++ b/src/iso19111/crs.cpp
@@ -574,6 +574,8 @@ CRSNNPtr CRS::createBoundCRSToWGS84IfPossible(
                 operation::CoordinateOperationFactory::create()
                     ->createOperations(NN_NO_CHECK(geodCRS), hubCRS, ctxt);
             CRSPtr candidateBoundCRS;
+            int candidateCount = 0;
+            bool candidateHasExactlyMachingExtent = false;
             for (const auto &op : list) {
                 auto transf =
                     util::nn_dynamic_pointer_cast<operation::Transformation>(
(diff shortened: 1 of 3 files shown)


bidrizi74:
Latest situation: 
In QGIS 3.26.0 version, in list of Coordinate Reference Systems (CRS), CRS with code EPSG 6316 doesn't cover area of North Macedonia. With this, North Macedonian CRS based on EPSG 6316 parameters cannot be used in QGIS.
This probably is because of two zones that cover EPSG 6316 (https://epsg.org/crs_6316/MGI-1901-Balkans-zone-7.html?sessionkey=z3r4rmfri7), one for other exYugoslavia countries (EPSG 1711) and second one for North Macedonia (EPSG 1148). This can be practicaly recogized if you add data in other CRS than WGS84 and if you use "Ask for datum transformation if several are available" option in "Settings/Options/Coordinate transformation and operations". In that case if you use EPSG 6316, all datum transformations will be available except datum transformation for North Macedonian territory which should be done by using EPSG 6206 as default datum transformation.
This issue can be solved only in a case to give other-separate EPSG code for North Macedonia with the second extend (EPSG 1148, extend of national area) of EPSG 6316, or to define sub-code of EPSG 6316 for second extend 1148, that should be linked to 6206 as default datum transformation for this area.
   
Example of projstring for North Macedonian area:
+proj=tmerc +lat_0=0 +lon_0=21 +k=0.9999 +x_0=7500000 +y_0=0 +ellps=bessel +towgs84=521.748,229.489,590.921,4.029,4.488,-15.521,-9.78 +units=m +no_defs

In a case of any missunderstanding, please get in touch me via email [bashkim.idrizi@yahoo.com](mailto:bashkim.idrizi@yahoo.com) any time.

   
https://epsg.org/crs_6316/MGI-1901-Balkans-zone-7.html?sessionkey=z3r4rmfri7

PR #3169, opened 2022-04-19
Fix cmake CURL dependency

fferflo:
FindCURL from cmake defines the variables `${CURL_LIBRARIES}` and `${CURL_INCLUDE_DIRS},` but `${CURL_LIBRARY}` and `${CURL_INCLUDE_DIR}` (empty variables) are used here instead. See [https://cmake.org/cmake/help/v3.23/module/FindCURL.html](https://cmake.org/cmake/help/v3.23/module/FindCURL.html). This has been the case since cmake 3.0, so this fix is backwards compatible.

(The reason why this hasn't been a problem so far is probably due to the fact that CURL libraries are often installed in system directories that are linked to anyway. I noticed this bug when using a custom CURL installation directory.)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,11 +26,9 @@ build_script:
   - set VCPKG_ROOT=C:\Tools\vcpkg
   - set VCPKG_INSTALLED=%VCPKG_ROOT%\installed\%platform%-windows
   # If cached directory does not exist, update vcpkg and install dependencies
-  # The checkout of a precise sha1 for VS2015 is a workaround for https://github.com/microsoft/vcpkg/issues/11666
   - if not exist %VCPKG_INSTALLED%\bin (
       cd "%VCPKG_ROOT%" &
       git pull > nul &
-      (if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" (git checkout a64dc07690bc8806e717e190f62eb58e198b599c)) &
       .\bootstrap-vcpkg.bat -disableMetrics &
       vcpkg install sqlite3[core,tool]:"%platform%"-windows &
       vcpkg install tiff:"%platform%"-windows &
--- a/src/lib_proj.cmake
+++ b/src/lib_proj.cmake
@@ -463,10 +463,10 @@ endif()
 
 if(CURL_ENABLED)
   target_compile_definitions(proj PRIVATE -DCURL_ENABLED)
-  target_include_directories(proj PRIVATE ${CURL_INCLUDE_DIR})
+  target_include_directories(proj PRIVATE ${CURL_INCLUDE_DIRS})
   target_link_libraries(proj
     PRIVATE
-      ${CURL_LIBRARY}
+      ${CURL_LIBRARIES}
       $<$<CXX_COMPILER_ID:MSVC>:ws2_32>
       $<$<CXX_COMPILER_ID:MSVC>:wldap32>
       $<$<CXX_COMPILER_ID:MSVC>:advapi32>


dg0yt:
WRT the appvey error: Which curl does it use? Recent versions of `FindCURL.cmake` will use CURL config if available, and I assume this config is where the `ZLIB::ZLIB` comes from - and where a `find_dependency(ZLIB)` is missing.
A workaround might be using `CURL_NO_CURL_CMAKE` set to on for the test.

rouault:
> WRT the appvey error: Which curl does it use?

probably an older one. 
@fferflo Could you try removing https://github.com/OSGeo/PROJ/blob/master/appveyor.yml#L29 and https://github.com/OSGeo/PROJ/blob/master/appveyor.yml#L33 which pinned to an older version of vcpkg due to a bug in the past and are probably no longer needed

mwtoews:
The working appveyor test in master shows this CMake configure:
```
...
-- Sqlite3 version: 3.31.1
-- Found TIFF: optimized;C:/Tools/vcpkg/installed/x86-windows/lib/tiff.lib;debug;C:/Tools/vcpkg/installed/x86-windows/debug/lib/tiffd.lib (found version "4.0.10") 
-- Found JPEG: optimized;C:/Tools/vcpkg/installed/x86-windows/lib/jpeg.lib;debug;C:/Tools/vcpkg/installed/x86-windows/debug/lib/jpegd.lib (found version "62") 
-- Found ZLIB: optimized;C:/Tools/vcpkg/installed/x86-windows/lib/zlib.lib;debug;C:/Tools/vcpkg/installed/x86-windows/debug/lib/zlibd.lib (found version "1.2.11") 
-- Found ZLIB: optimized;C:/Tools/vcpkg/installed/x86-windows/lib/zlib.lib;debug;C:/Tools/vcpkg/installed/x86-windows/debug/lib/zlibd.lib (found suitable version "1.2.11", minimum required is "1") 
-- Looking for pthread.h
...
```
while the appveyor in this PR (using updated vcpkg version) shows:
```
...
-- Sqlite3 version: 3.37.1
-- Found TIFF: optimized;C:/Tools/vcpkg/installed/x86-windows/lib/tiff.lib;debug;C:/Tools/vcpkg/installed/x86-windows/debug/lib/tiffd.lib (found version "4.3.0") 
-- Found ZLIB: optimized;C:/Tools/vcpkg/installed/x86-windows/lib/zlib.lib;debug;C:/Tools/vcpkg/installed/x86-windows/debug/lib/zlibd.lib (found suitable version "1.2.12", minimum required is "1") 
-- Looking for pthread.h
...
```
The differences are:
- Sqlite3 version 3.31.1 vs 3.37.1
- TIFF version 4.0.10 vs 4.3.0
- ZLIB version 1.2.11 vs 1.2.12

Unfortunately there are no handy messages to show the versions of CURL. Also the post-install test shows different `PROJ::proj.INTERFACE_LINK_LIBRARIES`. With this PR it is missing `lzma.lib`, `jpeg.lib` and `zlib.lib` paths (and their debug equivalents). I'm not sure what the cause is.

**Update**: looking at the appveyor error from the first commit with the precise vcpkg version, `PROJ::proj.INTERFACE_LINK_LIBRARIES` contains a path to `zlib.lib`, but the target "ZLIB::ZLIB" was not found, so that's still a bit strange.

dg0yt:
> **Update**: looking at the appveyor error from the first commit with the precise vcpkg version, `PROJ::proj.INTERFACE_LINK_LIBRARIES` contains a path to `zlib.lib`, but the target "ZLIB::ZLIB" was not found, so that's still a bit strange.

That is just coming from different CMake packages.

IIUC this is the Visual Studio 2015 appveyor image, and I found now that it comes with CMake 3.16.2. This implies an old `FindCURL.cmake`. I still think the bug is in the vcpkg curl port. This configuration is not tested, but I think a fix would be accepted quickly.

dg0yt:
Hm, I don't know enough about appveyor, but: 
`VCPKG_ROOT` is set in the `build_script` step, but not in the `test_script` step. Maybe that step inherits an even more outdated `VCPKG_ROOT` from the appveyor environment? The `Visual Studio 2015` image comes with `vcpkg 2020.02.04-nohash`.

dg0yt:
This is the pitfall of using targets in `<Pkg>_LIBRARIES` variables. Here, it is upstream implementation in CURL:
https://github.com/curl/curl/blob/37492ebbfa24ba4e700e6655b3dbc2bdd65c894a/CMakeLists.txt#L708-L710
The exported curl config includes the required `find_dependency(ZLIB)` so direct consumers don't have a problem.
But PROJ re-exports CMake config. The exported config assumes `CURL_LIBRARIES` to be fully resolved, and so it doesn't come with a `find_dependency` for CURL or ZLIB.

Resolving the CMake targets is hard - at least with support for multi-config generators. (I added code for that to GDAL for the purpose of pkg-config file generation. It might be useful for PROJ, too.)

For now, the easiest mitigation is adding a `find_dependency(CURL)` to the exported config. Apart from ZLIB, some CURL configurations might add more targets (e.g. libssh2).
I will give that a try in #3172.

dg0yt:
Appveyor for https://github.com/OSGeo/PROJ/pull/3172 is green now. My conclusions:
- It is entirely right to use `CURL_LIBRARIES` instead of `CURL_LIBRARY`. (This PR.)
- `find_package(CURL)` may add targets to `CURL_LIBRARIES`, in particular when using CURL's CMake config.
  To ensure that these targets are available to consumers of PROJ's CMake config, the PROJ's config should do a `find_dependency(CURL)`. (Currently included in https://github.com/OSGeo/PROJ/pull/3172.)
- With the extra config change added, this PR might work with the original vcpkg configuration, so that commit can be reverted here.
- I'm offering these changes as PR against fferflo's fork: https://github.com/fferflo/PROJ/pull/1.
  Please indicate to @fferflo if he should merge that PR. (I would suggest to use the "Rebase" variant.)
- After that merge, I could rework https://github.com/OSGeo/PROJ/pull/3172 to provide only the final changes which would update vcpkg for appveyor.

dg0yt:
Ping @rouault @mwtoews

rouault:
@dg0yt  I'd say we should just drop this PR and go for #3172 to make things simpler

dg0yt:
> @dg0yt I'd say we should just drop this PR and go for #3172 to make things simpler

Okay, then I will rework #3172 a little bit, for a clean history.

dg0yt:
#3172 is updated now.

rouault:
superseded per #3172

PR #3171, opened 2022-04-20
CRS PROJ string parser: accept +ellps/+a/+rf/... compatible of +datum…

### LINKED ISSUE #3170: User isn't informed that information from proj4 string has been ignored in making CRS
#### Example of problem

```python
from pyproj.crs import CRS

crss = [
    CRS.from_user_input("+proj=merc +datum=WGS84 +ellps=WGS84 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +units=m +no_defs +type=crs"),
    CRS.from_user_input("+proj=merc +a=6378137.0 +rf=298.257223563 +datum=WGS84 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +units=m +no_defs +type=crs"),
]

for crs in crss:
    print(crs)
    print(crs.datum)
    print()
```

```
+proj=merc +datum=WGS84 +ellps=WGS84 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +units=m +no_defs +type=crs
World Geodetic System 1984

+proj=merc +a=6378137.0 +rf=298.257223563 +datum=WGS84 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +units=m +no_defs +type=crs
unknown
```
#### Problem description

Currently, information (the datum) is dropped in the second example. I'm informed (https://github.com/pyproj4/pyproj/issues/1061) that this is because the string parsing in `proj` has identified that the ellipse information conflicts with the datum and thus chosen to drop the datum.

The lack of warning that this has been done (along with `pyproj`'s choice to show the string used to create the crs even though it's inaccurate) makes it harder to decode what's gone wrong. I think that `proj` should warn or error on provision of strings that make it drop data like this.

#### Environment Information
 - PROJ version (`proj`)
Rel. 8.2.0, November 1st, 2021
System:
    python: 3.8.12 | packaged by conda-forge | (default, Oct 12 2021, 21:59:51)  [GCC 9.4.0]
executable: .../.conda/envs/iris-dev/bin/python
   machine: Linux-3.10.0-1160.62.1.el7.x86_64-x86_64-with-glibc2.10

#### Installation method
 - conda

rouault:
…, but error out if there are not (fixes #3170)

Probably not appropriate for backport

### DIFF
--- a/src/iso19111/io.cpp
+++ b/src/iso19111/io.cpp
@@ -9345,9 +9345,6 @@ PROJStringParser::Private::buildDatum(Step &step, const std::string &title) {
     const auto &fStr = getParamValue(step, "f");
     const auto &esStr = getParamValue(step, "es");
     const auto &eStr = getParamValue(step, "e");
-    double a = -1.0;
-    double b = -1.0;
-    double rf = -1.0;
     const util::optional<std::string> optionalEmptyString{};
     const bool numericParamPresent =
         !RStr.empty() || !aStr.empty() || !bStr.empty() || !rfStr.empty() ||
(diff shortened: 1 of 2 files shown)


rhuijben:
If we can't get this in 9.x, can we at least get some of this in the logging or warning output from parsing, like from wkt.

I'd rather see a dependable behavior then silent ignores. So you would have my +1 for merging, but I also see the backwards compatibility story...

rouault:
@kbevers your take on the backport story of this in a stable branch ? I'm rather not confident. I wouldn't be surprised at all that there are odd cases of altering an ellipsoid +ellps with some +a/+rf/etc that haven't been identified. During development, I had to tweak the code to make a couple odd tests pass

busstoptaktik:
I'm  `-0` on this, and somewhat concerned. Not with respect to the problem at hand (#3170), but more in the general case of fear of painting ourselves deeper into some of the most unfortunate corners of current and past versions of the OGC/ISO geospatial model which, at least historically, have been at odds with the geodetic realities, focusing as they do (or at least did) on (non-existing) *definitions*  of geodetic datums, rather than (stochastically derivable) *relations between them*.

 `+datum=...` provides information about the *actual datum* in use, not just about its *associated ellipsoid*. So while equivalencing `+datum`=... and `+ellps=...` may not make any numerical difference for the problem at hand, it makes a huge difference with respect to user education, and limits the gamut of potential forward evolutionary paths for PROJ.

As the requirement for coordinate accuracy grows, so does the necessity of actually following the stochastic geodetic realities, rather than the more deterministic everyday-experience-compatible pseudo-geodetic mental models. When these are at odds, having a *specified* datum on *both* the source and the target side of the scene, at least makes it possible to check that a transformation explicitly selected by the user, *actually* transforms between the two systems at hand.

So blurring the difference between datums and ellipsoids is an efficient way of digging holes in the road toward the support of inevitable future use cases.

I can hardly claim to be against merging this PR, since it solves a concrete problem, but #3170 really sheds light on a more general problem that will bite us in the future: For higher accuracy, the API focus need to shift from *definitions* to *transformations*. Not just trying to guess what the user needs, but making it easier to ask for a specific transformation, then actually doing what the user asks for (i.e.,at the command line level, somehow making `cs2cs`and `cct` confluence, and ditto at the API level).

This is one of the 4 reasons I started work on [Rust Geodesy](https://github.com/busstoptaktik/geodesy) - not as a replacement for PROJ (not even remotely), but as an experimental platform for simplified data flows, and for developing concepts and potential solutions for predictable future problems. If these concepts and solutions make their way into PROJ in a few years, they may actually become useful for geodetic everyday work - but painting PROJ into too many corners in the meantime may impossibilize necessary innovations - from [Rust Geodesy](https://github.com/busstoptaktik/geodesy) or elsewhere.

While sounding the alarm here, I do not actually think we're dealing with a huge problem, but I find it quite necessary to raise awareness that trying to provide deterministic solutions to stochastic problems will run us all out of steam sooner or later.

kbevers:
> your take on the backport story of this in a stable branch ?

I don't think this should go into 9.0.1. It's a change in behaviour but not interface so I think it would be okay to put it in 9.1.0. I understand your worries though, we would be changing behaviour that someone might rely on at the moment. Maybe it would be smarter to just properly document the current behaviour? On the other hand we are in legacy PROJ.4-string territory that should be avoided and users can't expect this to function perfectly.

rhuijben:
The 9.1 ship sailed...

kbevers:
> The 9.1 ship sailed...

Was this important to you? Two developers are questioning the approach here and the community hasn't been particularly vocal about needing this change. I'm okay with it not landing in 9.1.0 but happy to reconsider if this proves to be more needed than I realize.

rhuijben:
No it was not important, but we should somehow make a decision for the future on how we handle those 'error' cases. We can target it for 9.1 or postpone to 10... or decide not to do it at all..

Personally I would like to see which arguments are ignored via some kind of warning notification, assuming we can't make this an error. (Which I don't think we can. Adding errors to well defined transforms that always worked... Sounds like a breaking error)

kbevers:
> Personally I would like to see which arguments are ignored via some kind of warning notification, assuming we can't make this an error

I would welcome a RFC on the topic. Especially if it's backed up by code. Handling of errors and warnings has always been difficult and it would be good to have a better defined strategy. Error-handling was refined a bit some time ago but warnings (and maybe deprecations?) could be definitely be handled better.

PR #3172, opened 2022-04-20
Fix cmake CURL dependency, update appveyor - vcpkg interaction

dg0yt:
Includes #3169.

 - [X] Ensure consistent and proper vcpkg usage

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,7 +38,7 @@ build_script:
       cd %APPVEYOR_BUILD_FOLDER%
     )
   - dir %VCPKG_INSTALLED%\bin
-  - set PATH=%VCPKG_INSTALLED%\bin;%PATH%
+  - set PATH=%VCPKG_INSTALLED%\tools;%PATH%
   # See https://www.appveyor.com/docs/lang/cpp/
   - if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" if %platform%==x86
       (call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86)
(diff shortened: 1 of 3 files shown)


dotlambda:
Because of this change the following ends up in `/lib/cmake/proj4/proj4-config.cmake`:
```cmake
include(CMakeFindDependencyMacro)
if("TRUE" STREQUAL "TRUE")
  # Chainload CURL usage requirements
  find_dependency(CURL)
endif()
```
This means that anything using `find_package(PROJ4 REQUIRED)` now also depends on cURL. Is that intentional?

dg0yt:
> This means that anything using `find_package(PROJ4 REQUIRED)` now also depends on cURL. Is that intentional?

Anything which depends on a build of PROJ which is configured with CURL. Note that the source is:
https://github.com/OSGeo/PROJ/blob/65458dfb8123a96ccd45a46c363e81e1307c1582/cmake/project-config.cmake.in#L15-L18

dotlambda:
> Anything which depends on a build of PROJ which is configured with CURL.

You're right. But I don't quite see why a project depending on PROJ (configured with cURL) should depend on cURL. Does the latter need to be available at link time?

dg0yt:
> But I don't quite see why a project depending on PROJ (configured with cURL) should depend on cURL. 

It does so anyways, at least transitively.

> Does the latter need to be available at link time?

At least it is needed at CMake configuration time, in order to resolve the dependencies CURL implies for generated PROJ config.  Cf. https://github.com/OSGeo/PROJ/pull/3169#issuecomment-1104846458.
(There might be better decoupling at least for dynamic linkage, but it would need to start at CURL.)

rouault:
> (There might be better decoupling at least for dynamic linkage, but it would need to start at CURL.)

coulnd't we just make the find_dependency(CURL) conditioned by BUILD_SHARED_LIBS being set ?

dg0yt:
> > (There might be better decoupling at least for dynamic linkage, but it would need to start at CURL.)
> 
> coulnd't we just make the find_dependency(CURL) conditioned by BUILD_SHARED_LIBS being set ?

First you have to update CURL, to make it link `CURL_LIBS` privately:
https://github.com/curl/curl/blob/8dcbb0f19176da937ddb3c85509b8c84086348f1/lib/CMakeLists.txt#L98
But now the public dependencies are already out in the wild.

PR #3173, opened 2022-04-21
Sync with geographiclib-c version 2.0

cffk:
- [ ] Closes #xxxx
 - [x] Tests added
 
For changes see https://geographiclib.sourceforge.io/C/doc/changes.html


cffk:
I assume the MacOS build failure is some transient problem.  What's the best way to get this check rerun?

mwtoews:
Yes, likely a transient networking issue. I've triggered to re-run failed jobs, so I'll expect this one to complete.

cffk:
@mwtoews, do you want to review this?  I'm comfortable going ahead with the merge but I'm happy to wait for your input.

mwtoews:
Sure, I can offer some input. Was it necessary to rename the test files from `.c` to `.cpp`?

cffk:
> Sure, I can offer some input. Was it necessary to rename the test files from `.c` to `.cpp`?

`geodtest.c` was renamed to `geodtest.cpp` back when some C++ migration was being done (by someone else).  I just followed suit with the new test (not wanting to rock the boat).

mwtoews (commented):
I have a few suggestions:

- Rename the C files back to their C extension, so they mirror their [source files](https://github.com/geographiclib/geographiclib-c)
- Remove `bin_geodtest.cmake`, since these are not installed binaries (like the other `bin_*.cmake` files), and move some of the contents to `src/tests/CMakeLists.txt` with the suggested source:
```cmake
if(NOT BUILD_TESTING)
  # stop here and return to including file
  return()
endif()

add_executable(geodtest geodtest.c)
target_link_libraries(geodtest ${PROJ_LIBRARIES})

add_executable(geodsigntest geodsigntest.c)
# geodsigntest includes geodesic.c directly so it doesn't need to link
# against the library.
if(HAVE_LIBM)
  target_link_libraries(geodsigntest m)
endif()

# Do not install, instead run as a tests
add_test(NAME geodesic-test COMMAND geodtest)
add_test(NAME geodesic-signtest COMMAND geodsigntest)
```
- At the top, it seemed that the check with `BUILD_GEOD` was unnecessary, as that's a separate CLI tool
- I've removed `source_group()` for the two tests source files, since `source_group()` appear to be used for the main source files that go to the library and installed CLI tools, but not for any other test files
- Only `geodsigntest` needs to be explicitly linked to libm (my guess it it was implicitly linked as a `.cpp` file, which is why it wasn't needed)

cffk:
Thanks, these are good suggestions.  I made one modification and that was to move the `if(BUILD_TESTING)` to the parent `src/CMakeLists.txt` so people don't have to look into `src/tests/CMakeLists.txt` to see that it may not be invoked.  I also see that `multistresstest.cpp` is an orphan.  I'll post an issue about this once this PR is merged.

PR #3180, opened 2022-05-03
Make it possible to determine which grid files were actually used during a transformation (fixes #3176)

### LINKED ISSUE #3176: Add API to determine which grid files were actually used during a transformation
As a PROJ user integrating the library into my company’s software I would like to be able to inform my users when a requested grid file in a transformation has resolved to a different file.

For example, currently if the file NTv2_0.gsb in the PROJ search path, but the transformation specified uses ca_nrc_ntv2_0.tif, PROJ will silently use NTv2_0.gsb with no simple way of discovering it has done so. I would like to be able to query the transformation object for the required grids and also what they were resolved to.

Something like the following (API just given as an example of what we’re looking for, it doesn’t need to be exactly this):

Added to proj.h

```c++
struct PROJ_GRID_RESOLVE_INFO {
    const char* requested_grid;
    const char* resolved_grid_path;
};
 
PROJ_GRID_RESOLVE_INFO** proj_resolved_grids(PJ* trans, int* out_result_count);
```

Example Usage:
```c++
const char* proj_string = "+proj=hgridshift +grids=ca_nrc_ntv2_0.tif";
 
PJ* transformation = proj_create(nullptr, proj_string);
PJ_COORD coordinate = proj_coord(-123.10876, 49.27782, 0, 0);
proj_trans(transformation, coordinate);
 
PROJ_GRID_RESOLVE_INFO** resolved_grids = nullptr;
int num_resolve_info = 0;
 
resolved_grids = proj_resolved_grids(transformation, &num_resolve_info);
 
for (int i = 0; i < num_resolve_info; ++i) {
    if (resolved_grids[i]->resolved_grid_path != nullptr) {
        std::cout << "Requested grid " << resolved_grids[i]->requested_grid; 
        std::cout << " resolved to " << resolved_grids[i]->resolved_grid_path << std::endl;
    } else {
        std::cout << "Requested grid " << resolved_grids[i]->requested_grid << " not found" << std::endl;
    }
}
```

Which for my example would print:

`Requested grid ca_nrc_ntv2_0.tif resolved to /path/to/NTv2_0.gsb`

If networking were enabled and there was no matching grid locally it would print the CDN URL that was used:

`Requested grid ca_nrc_ntv2_0.tif resolved to https://cdn.proj.org/ca_nrc_ntv2_0.tif`

And of course if it was not found then it would print:

`Requested grid ca_nrc_ntv2_0.tif not found`
 
This was motivated by one of our developers trying to put together an example of a transformation that needed networking, and was surprised to discover that the transformation worked even without networking by using a completely different (though equivalent) grid file.

rouault:


docs/source/development/reference/functions.rst
```diff
@@ -273,6 +273,21 @@ Coordinate transformation
     :returns: :c:type:`PJ_COORD`
 
 
+
+.. c:function:: PJ* proj_trans_get_last_used_operation(PJ *P)
+
+    .. versionadded:: 5.1.0
```
snowman2:
```suggestion
    .. versionadded:: 9.1.0
```

rouault: fixed

PR #3200, opened 2022-05-14
Implement Vertical Offset and slope transformation method

rouault:
@kbevers  I was not sure how to map the latitude and longitude of the "evaluation point" of the method, which is the 'origin' of the inclinated plane, to PROJ string parameters. I've used the classic lon_0 and lat_0 parameters, but not completely sure it is appropriate (I have to re-add lon_0 in the forward method, since generic code in PROJ, substracts it, as it is what is convenient for map projections)


docs/source/operations/transformations/vertoffset.rst
```diff
@@ -0,0 +1,61 @@
...
+| **Domain**          | 3D                                                       |
++---------------------+----------------------------------------------------------+
+| **Input type**      | Geodetic coordinates (horizontal), meters (vertical)     |
++---------------------+----------------------------------------------------------+
+| **output type**     | Geodetic coordinates (horizontal), meters (vertical)     |
++---------------------+----------------------------------------------------------+
+
+It is documented as coordinate operation method code 1046 in the EPSG dataset (:cite:`IOGP2018`).
+It is typically used in Europe, to relate national vertical systems to
+pan-European vertical systems (EVRF200, EVRF2007).
```
kbevers, line 25: I would like to see a more in-depth description of the transformation here. It's too much of a faf to have to look up what 1046 means in the EPSG-world.

src/transformations/vertoffset.cpp
```diff
@@ -0,0 +1,90 @@
...
+    double nu0;
+};
+} // anonymous namespace
+
+// Cf EPSG Dataset coordinate operation method code 1046 "Vertical Offset and Slope"
+
+static PJ_XYZ forward_3d(PJ_LPZ lpz, PJ *P) {
+    const struct pj_opaque_vertoffset *Q = (const struct pj_opaque_vertoffset *) P->opaque;
+    PJ_XYZ xyz;
+    xyz.x = lpz.lam + P->lam0;
```
kbevers, line 60: Since this is a bit unusual a comment explaining what's going on would be nice

kbevers (commented):
> I was not sure how to map the latitude and longitude of the "evaluation point" of the method, which is the 'origin' of the inclinated plane, to PROJ string parameters. I've used the classic lon_0 and lat_0 parameters, but not completely sure it is appropriate (I have to re-add lon_0 in the forward method, since generic code in PROJ, substracts it, as it is what is convenient for map projections)

I think the use of `+lat_0` and `+lon_0` fits the bill here. The meaning is of course a little bit different but I think it works alright in this case.

docs/source/operations/transformations/vertoffset.rst
rouault: there's isn't much text to borrow from the description of the EPSG dataset. I've added a Formula section which is the "objective" description of what this method does

src/transformations/vertoffset.cpp
rouault: comment added

docs/source/operations/transformations/vertoffset.rst
kbevers: Thanks, I think that puts the docs in a good place

kbevers (approved):

rhuijben:
Can you update the in c declaration header to include the required arguments of the projection, like we have in the other projection definitions?

In my Proj use I try to automatically create some test scenario's based on those. And it would also be nice for the built-in help from the Proj tools.

rouault:
> Can you update the in c declaration header to include the required arguments

done, although as I mentioned in a commit comment, we haven't done that for transformation operations until now, just map projections

PR #3221, opened 2022-06-05
PROJJSON parser: do not error out if a datum ensemble member is unknown in the database

rouault:
Fix issue raised in https://github.com/opengeospatial/geoparquet/discussions/110


rouault:
unrelated doc build issue (cf https://github.com/OSGeo/PROJ/pull/3222)

Issue #2348, opened 2020-09-05
Suboptimal (?) result when going from "WGS 84" to "GDA2020 + AHD height" 


rouault:
```
echo -33 151 5 | PROJ_NETWORK=ON PROJ_DEBUG=3 PROJ_LIB=data src/cs2cs -d 8 "WGS 84" "GDA2020 + AHD height"
```
returns
```
-32.99998729	151.00000557 -20.85300064
```
and takes the "Using coordinate operation Inverse of Null geographic offset from WGS 84 (geog2D) to WGS 84 (geog3D) + Inverse of AGD66 to WGS 84 (17) + AGD66 to GDA94 (11) + Inverse of Null geographic offset from GDA94 (geog3D) to GDA94 (geog2D) + GDA94 to AHD height (49) + GDA94 to GDA2020 (1)" root, using au_ga_AUSGeoid09_V1.01.tif

whereas
```
echo -33 151 5 | PROJ_NETWORK=ON PROJ_DEBUG=3 PROJ_LIB=data src/cs2cs -d 8 GDA2020 "GDA2020 + AHD height"
```
returns
```
-33.00000000	151.00000000 -20.74200058
```
using
au_ga_AUSGeoid2020_20180201.tif

The later should ideally also be the result of  "WGS 84" to "GDA2020 + AHD height" since there is a (null) WGS 84 to GDA2020 transformation available, and there is a geoid transformation for AHD height to GDA2020 (using AUSGeoid2020). Going through older AGD66, GDA94 and AUSGeoid09 is clearly suboptimal

rouault:
Hum, actually this is tricky !

Looking at:
```
projinfo -s EPSG:4326 -t  EPSG:7856+5711  --spatial-test intersects --summary
Candidate operations found: 42
unknown id, WGS 84 to WGS 84 (G1762) + Inverse of Conversion from WGS 84 (G1762) (geocentric) to WGS 84 (G1762) (geog2D) + Inverse of GDA2020 to WGS 84 (G1762) (1) + Inverse of Conversion from GDA2020 (geog3D) to GDA2020 (geocentric) + GDA2020 to AHD height (1) + Map Grid of Australia zone 56, 2.23 m, Australia Christmas and Cocos - onshore
[...]
[...]

PR #3248, opened 2022-07-03
createOperations(): prefer simpler pipelines / affects WGS 84 to GDA94/GDA2020

rouault:
Prior to this change, transforming to WGS 84 to GDA2020 used in priority
transformation EPSG:9690 ("WGS 84 to GDA2020 (3)"), which is a Helmert
transformation assuming WGS 84 ~= GDA94.
But transforming from GDA2020 to WGS 84 used EPSG:8450 "GDA2020 to WGS
84 (2)" which is a no-op.
Both have the same advertized accuracy 3m
Similar case for WGS 84 <--> GDA94.

This non symetric behaviour is clearly non-desirable in scenarios where
data is transformed back and forth. I believe it is preferable to use in
that situation of transformations with the same accuracy to use the one
that involves the less transformation steps, ie the no-op one.

Seen when checking https://github.com/OSGeo/PROJ/issues/2348 again
(whose bulk issues happens to have been fixed by other previous commits)

CC @nyalldawson

### DIFF
--- a/include/proj/io.hpp
+++ b/include/proj/io.hpp
@@ -476,6 +476,8 @@ class PROJ_GCC_DLL PROJStringFormatter {
 
     PROJ_INTERNAL Convention convention() const;
 
+    PROJ_INTERNAL size_t getStepCount() const;
+
     //! @endcond
 
   protected:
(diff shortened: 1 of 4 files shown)


ValZapod:
> assuming WGS 84 ~= GDA94.

It is ITRF92, using reference epoch 1994.0 (so not 1988.0, it was moved). So the only difference with G730 (which was only 10 cm absolute accurate) is the use of GRS80 ellipsoid.

rouault:
@nyalldawson Does this change look appropriate to you ?

nyalldawson:
@rouault 

This looks safe to me. +1 to merge!

PR #3251, opened 2022-07-05
proj.pc/proj-config.cmake generation: deal with CMAKE_INSTALL_xxxx being possibly an absolute path (fixes #3206)

### LINKED ISSUE #3206: CMake and pkg-config files are broken when CMAKE_INSTALL_{BIN,DATAROOT,INCLUDE,LIB}DIR is absolute
#### Example of problem

```
$ cd /tmp
$ mkdir prefix prefix-bin prefix-lib prefix-include prefix-share
$ git clone https://github.com/OSGeo/PROJ
[snip]
$ cd PROJ
$ mkdir build && cd build
$ cmake -DCMAKE_INSTALL_PREFIX=/tmp/prefix -DCMAKE_INSTALL_BINDIR=/tmp/prefix-bin -DCMAKE_INSTALL_INCLUDEDIR=/tmp/prefix-include -DCMAKE_INSTALL_LIBDIR=/tmp/prefix-lib -DCMAKE_INSTALL_DATAROOTDIR=/tmp/prefix-share ..
[snip]
$ make && make install
[snip]
$ grep -r '//tmp/prefix' /tmp/prefix-lib
/tmp/prefix-lib/pkgconfig/proj.pc:libdir=${prefix}//tmp/prefix-lib
/tmp/prefix-lib/pkgconfig/proj.pc:includedir=${prefix}//tmp/prefix-include
/tmp/prefix-lib/pkgconfig/proj.pc:datarootdir=${prefix}//tmp/prefix-share
/tmp/prefix-lib/cmake/proj/proj-config.cmake:set (PROJ_INCLUDE_DIRS "${_ROOT}//tmp/prefix-include")
/tmp/prefix-lib/cmake/proj/proj-config.cmake:set (PROJ_LIBRARY_DIRS "${_ROOT}//tmp/prefix-lib")
/tmp/prefix-lib/cmake/proj/proj-config.cmake:set (PROJ_BINARY_DIRS "${_ROOT}//tmp/prefix-bin")
/tmp/prefix-lib/cmake/proj4/proj4-config.cmake:set (PROJ4_INCLUDE_DIRS "${_ROOT}//tmp/prefix-include")
/tmp/prefix-lib/cmake/proj4/proj4-config.cmake:set (PROJ4_LIBRARY_DIRS "${_ROOT}//tmp/prefix-lib")
/tmp/prefix-lib/cmake/proj4/proj4-config.cmake:set (PROJ4_BINARY_DIRS "${_ROOT}//tmp/prefix-bin")
$ # oops, none of these will work to locate the corresponding files.
```

#### Problem description

Per the [documentation of GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables), the value of a `CMAKE_INSTALL_*DIR` variable “should typically be a path relative to the installation prefix [...]. However, an absolute path is also allowed.” In https://github.com/OSGeo/PROJ/pull/3150, the old `PROJ_*_SUBDIR` variables were eliminated in favour of the standard `CMAKE_INSTALL_*DIR` variables (which were used to populate the old ones in any case), but the assumption that they are relative (“`SUBDIR`”) remained.

#### Expected Output

The listed values should contain the absolute paths passed at configure time and no reference to `${prefix}` or `${_ROOT}`. Normally this is what the `CMAKE_INSTALL_FULL_*DIR` variables set by GNUInstallDirs are for, but [other environments](https://github.com/libjxl/libjxl/issues/1400#issuecomment-1122183126), which do *not* pass absolute paths at configure time, depend on the *presence* of references to `${prefix}` for relocatability. Thus a more complicated solution is required.

See “[Concatenating paths when building pkg-config files](https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files)” for a thorough discussion of the problem and a suggested fix, or [KDE’s extra-cmake-modules](https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/modules/ECMGeneratePkgConfigFile.cmake#L166) for a simpler approach.

#### Environment Information
 - PROJ version (`proj`) = 9.0.0 (pre #3150), but exists on master (post #3150) as well
 - Operation [sic] System Information = `Linux localhost 5.15.36 #1-NixOS SMP Wed Apr 27 12:39:02 UTC 2022 x86_64 GNU/Linux`

rouault:


mwtoews (approved):
Nice fix!

PR #3253, opened 2022-07-07
Introduce PROJ_DATA environment variable to deprecate PROJ_LIB

rouault:
Given its looong existence and to avoid disruption in existing workflows,
PROJ_LIB will still work for the time being, and should probably be kept
functional at least for the PROJ 10 series.
A runtime debug message is logged (PROJ_DEBUG>=2) if using PROJ_LIB to
point to the recommended PROJ_DATA.

The PROJ_LIB_ENV_VAR_TRIED_LAST CMake option is renamed consistently to
PROJ_DATA_ENV_VAR_TRIED_LAST and a CMake warning is emitted when the old
name is used (in which case the value of PROJ_LIB_ENV_VAR_TRIED_LAST is
used for PROJ_DATA_ENV_VAR_TRIED_LAST)

The rational for this change is:
- PROJ_LIB is a misleading name. Some users believe it is related to the
  loading of the PROJ executable library.
- to align nicely with GDAL_DATA since they are about the same thing.


kbevers (approved):
I'm good with this change and I think it is sufficiently friendly to those relying on the variable being `PROJ_LIB`. As far as I read the code only CMake makes noise about this. Would it be too annoying to put out a deprecation notice everytime one of the command line apps is run? Similar to what Python does on DeprecationWarnings. It's annoying but effective.

Once merged, I'd like to see a new issue on removing support for `PROJ_LIB`. Put it in the 10.0.0 milestone (I think it's far enough in the future).

mwtoews:
Great change, as PROJ_LIB is a misleading name for a data directory. Any reason why reference to PROJ_LIB are not updated in README.md? Or should these be deferred to later (as to not confuse users that land on the first GH page and think they are seeing information for the current release).

rouault:
> Any reason why reference to PROJ_LIB are not updated in README.md? Or should these be deferred to later (as to not confuse users that land on the first GH page and think they are seeing information for the current release).

I had missed that reference, but your justification is a good one :-)

PR #3257, opened 2022-07-09
Database: resync ESRI objects with https://github.com/Esri/projection-engine-db-doc/tree/v3.0.0

rouault:


scripts/build_db_from_esri.py
```diff
@@ -1786,6 +1786,10 @@ def import_hvcoordsys():
             authority = row[idx_authority]
             esri_name = row[idx_name]
 
+            if authority == 'Esri' and code in ('9897',):
```
nyalldawson, line 1789: Sorry, missed this before merge -- but there's other instances of this I've found too. Eg 9459. I tried to get a comment from ESRI about it in this ticket (https://github.com/Esri/projection-engine-db-doc/issues/34) but their repo looks like a "dump and ignore" type job.

I'm curious on what your thoughts are on the best way to handle this. Should we continue building a mega list here of the incorrectly attributed ones?

rouault: > Should we continue building a mega list here of the incorrectly attributed ones?

For the 9459,9459,"PE_GT_ATRF_2014_TO_GDA_2020_AT2020" case, and if there are indeed numerous instances of that, we could probably automate the mis-attribution by checking that srid==latestWkid < 32767 and checking in EPSG that there's a similar record (e.g based on names, but we might need to massage names a bit by stripping underscores, some suffixes, etc. Or maybe just checking that the source & target SRS are the same in the ESRI and EPSG records)

PR #3275, opened 2022-08-02
Update ellipsoids.rst: typo

mdsumner:
- [x] change typo 'laton' to 'latlon'

todo/questsion: 

- [x]  YES: should lines be fully qualified proj params e.g. '+proj=latlon +R=7000000'
- [x] NO: should family be "lonlat"

### DIFF
--- a/docs/source/usage/ellipsoids.rst
+++ b/docs/source/usage/ellipsoids.rst
@@ -137,23 +137,23 @@ Spherical earth with radius 7000km:
 
 ::
 
-    proj=latlon R=7000000
+    +proj=latlon +R=7000000
 
 Using the GRS80 ellipsoid:
 
 ::
 
-    proj=latlon ellps=GRS80
+    +proj=latlon +ellps=GRS80
 
 Expressing ellipsoid by semi-major axis and reverse flattening (:math:`1/f`):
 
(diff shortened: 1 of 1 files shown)


kbevers:
> should lines be fully qualified proj params e.g. '+proj=latlon +R=7000000'

I'd prefer that, yes. The plusses aren't necessary in all cases (e.g. in a gie file or when using the API) but this could easily be confusing when trying things out on he command line where the plusses are mandatory.

> should family be "lonlat"

They are all synonyms so it doesn't really matter. The latitude/longitude order is the ISO standard for representing geographical coordinates though so I have a slight preference for `latlon`.

mdsumner:
great, thanks - updated now

mwtoews (approved):
Thanks for catching this!

mdsumner:
there is still a dangling +R_V token in the final example, I just wasn't sure how that's meant to work - I was hoping to figure that out and include it here but wasn't able to get to it

kbevers:
Looking at the example it looks alright to me. The `+R_V` flag is used to tell PROJ that you want a sphere of the same volume as the ellipsoid determined by the given parameters (`+a` and `+rf`).

PR #3299, opened 2022-08-27
Avoid using upper case project name in doc install dir

### LINKED ISSUE #3296: docfiles installed in wrong location
#### Example of problem
PROJ 9.1.0-rc1 installs the docfiles in the wrong location:
```
/usr/share/doc/PROJ/AUTHORS
/usr/share/doc/PROJ/COPYING
/usr/share/doc/PROJ/NEWS
```
This used to be `/usr/share/doc/proj` but `CMAKE_INSTALL_DOCDIR` uses the project name which is in uppercase.

The following patch works for me:
```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,8 +320,9 @@ if(BUILD_TESTING)
 endif()
 
 set(docfiles COPYING NEWS AUTHORS)
+string(TOLOWER "${CMAKE_INSTALL_DOCDIR}" DOCDIR)
 install(FILES ${docfiles}
-        DESTINATION ${CMAKE_INSTALL_DOCDIR})
+        DESTINATION ${DOCDIR})
 
 ################################################################################
 # pkg-config support
```

kbevers:
Closes #3296

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,7 +293,7 @@ link_directories(${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 include(ProjInstallPath)
 
 # By default GNUInstallDirs will use the upper case project name
-# for the DOC_INSTALL_DIR, resulting in something like share/doc/PROJ
+# for CMAKE_INSTALL_DOCDIR, resulting in something like share/doc/PROJ
 # instead of share/doc/proj which historically have been the path used
 # by the project. 
 # Here force the use of a lower case project name and reset after
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,7 +291,20 @@ link_directories(${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 # Installation
 ################################################################################
 include(ProjInstallPath)
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -291,7 +291,20 @@ link_directories(${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 # Installation
 ################################################################################
 include(ProjInstallPath)
+
+# By default GNUInstallDirs will use the upper case project name
+# for the DOC_INSTALL_DIR, resulting in something like share/doc/PROJ
```
mwtoews:
```suggestion
# for CMAKE_INSTALL_DOCDIR, resulting in something like share/doc/PROJ
```

mwtoews (approved):
Thanks, and I can both confirm the reported issue and that this PR fixes it. Also, [looking at the module](https://github.com/Kitware/CMake/blob/v3.24.0/Modules/GNUInstallDirs.cmake), I can confirm that `DOCDIR` is the only variable that has this case-sensitive issue with `${PROJECT_NAME}`.

Just one minor suggestion, and it should be ready to merge/backport.

PR #3310, opened 2022-09-05
Update another current year reference in CITATION.

rhuijben:
Fix other year value. Should help for the year bump next year.


kbevers:
Good call. Perhaps this should be reflected in `HOWTORELEASE` as well? It is clearly not doing a good enough job of telling me what to update since I've missed it two years in a row...

rouault:
superseded per https://github.com/OSGeo/PROJ/commit/24c7996db18282115a1d4abcc6c7f73b30e9952d

PR #3312, opened 2022-09-06
Docs: update doc of grid-based methods to mention/advertize more GeoTIFF

rouault:


docs/source/operations/transformations/deformation.rst
```diff
@@ -159,6 +133,33 @@ Parameters
 
     .. note:: :option:`+dt` is mutually exclusive with :option:`+t_epoch`
 
+
+Parameters (deprecated)
```
kbevers: There is also the [`:deprecated:` directive ](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated), which would be a more correct way to mark these parameters as deprecated.

rouault: done

kbevers (approved):

PR #3313, opened 2022-09-06
GeoTIFF grids: support UNITTYPE=arc-seconds per year for temporal hgridshift

rouault:


kbevers (commented):
This looks like an extensenion to the GTG format. Are we keeping track of version numbers somewhere? And should it be raised?

rouault:
> This looks like an extensenion to the GTG format. Are we keeping track of version numbers somewhere? And should it be raised?

there is no version numbering. I could potentially add a mention this is only supported since PROJ 9.1.1

kbevers:
>  there is no version numbering

Would it make sense to have that? We do it for the PROJJSON spec already. It's been fairly stable so far but I think it would be good to be transparent about the evolution of the format.

rouault:
> Would it make sense to have that?

ok, I've added a Revisions section at the bottom of the .rst

kbevers:
>  ok, I've added a Revisions section at the bottom of the .rst

Thanks. Good to go from my point of view!

PR #3319, opened 2022-09-13
Implement alterCSLinearUnit for DerivedProjectedCRS

jjimenezshaw:
- [x] Tests added

Working with `DerivedProjectedCRS`, having this functionality would be useful (in combination with `promoteTo3D`)


jjimenezshaw:
I can implement for CompoundCRS in a different PR if it makes sense.

rouault:
> I can implement for CompoundCRS in a different PR if it makes sense.

probably. I guess that it should be implemented as altering the linear units of the horizontal (when the horizontal CRS has a CS with linear units) and vertical members of the compound CRS

Issue #3168, opened 2022-04-19
Remove PROJ4 CMake name


mwtoews:
PROJ 7.0 changed the CMake project name from PROJ4 to PROJ (#1885, #1910).

PROJ 9.1 made `find_package(PROJ4)` show a deprecation warning (#3094, #3165).

PROJ 10.0 should finally remove this support.

rouault:
to solve https://github.com/OSGeo/gdal/issues/5646, maybe we should offer a CMake variable to opt to disable exporting PROJ4, as something doable in PROJ 9.x series ?

mwtoews:
Fine with me to bump this earlier in the 9.x series, but I also like to look further at the GDAL issue.

dg0yt:
~~~
... target "proj" that is not in this export set, but in multiple other export sets: lib/cmake/proj/proj-targets.cmake, lib/cmake/proj/proj4-targets.cmake, lib/cmake/proj4/proj-targets.cmake, lib/cmake/proj4/proj4-targets.cmake
~~~

Instead of creating multiple exports of `proj`, it should be exported only once under the canonical name, and every alternative name should have been created as an interface library which depends on the canonical name.

PR #3324, opened 2022-09-14
CMake: add a INSTALL_LEGACY_CMAKE_FILES option (fixes OSGeo/gdal#5646)

rouault:
to control whether the PROJ4 legacy target should be exported (refs #3168)

The default is ON, unless PROJ is built as part of a CMake subproject.

I've verified this fixes the following reproducer (and that I hit the issue without this PR)
```
add_subdirectory(proj)
add_subdirectory(gdal)
```

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -90,7 +90,17 @@ else ()
   set (CMAKE_CROSSCOMPILING_STR "OFF")
 endif ()
 
-foreach (PROJECT_VARIANT_NAME ${PROJECT_NAME} ${PROJECT_LEGACY_NAME})
+# If this is a regular build (PROJECT_SOURCE_DIR == CMAKE_SOURCE_DIR), export legacy files by default
+# Otherwise, this is embedded in another project, only export PROJ target
+# cf https://github.com/OSGeo/gdal/issues/5646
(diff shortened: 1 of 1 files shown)


cmake/CMakeLists.txt
```diff
@@ -90,7 +90,21 @@ else ()
...
 
-foreach (PROJECT_VARIANT_NAME ${PROJECT_NAME} ${PROJECT_LEGACY_NAME})
+if ("${PROJECT_SOURCE_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
+    # If this is a regular build, export legacy files by default
+    set(DEFAULT_VAL ON)
+else ()
+    # If this is embedded in another project, only export PROJ target
+    # cf https://github.com/OSGeo/gdal/issues/5646
+    set(DEFAULT_VAL OFF)
+endif ()
```
dg0yt:
~~~cmake
string(COMPARE EQUAL "${PROJECT_SOURCE_DIR}" "${CMAKE_SOURCE_DIR}" DEFAULT_VAL)

dg0yt (commented):
This PR look like adding more complexity to solve problems created from ... complexity.

cmake/CMakeLists.txt
rouault: done

mwtoews (approved):
I've not tested this locally, but trust that Even's verifications work. While it does pile a bit more complexity, it seems the best approach forward for the 9.x series.

PR #3325, opened 2022-09-14
implement alterCSLinearUnit for CompoundCRS

jjimenezshaw:
- [x] Tests added

For completeness, this PR implements `alterCSLinearUnit` for `CompoundCRS`, applying `alterCSLinearUnit` on every component.

### DIFF
--- a/test/unit/test_crs.cpp
+++ b/test/unit/test_crs.cpp
@@ -6551,11 +6551,25 @@ TEST(crs, crs_alterCSLinearUnit) {
     }
 
     {
-        // Not implemented on compoundCRS
         auto crs =
             createCompoundCRS()->alterCSLinearUnit(UnitOfMeasure("my unit", 2));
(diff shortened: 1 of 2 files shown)


test/unit/test_crs.cpp
```diff
@@ -6551,10 +6551,19 @@ TEST(crs, crs_alterCSLinearUnit) {
     }
 
     {
-        // Not implemented on compoundCRS
         auto crs =
             createCompoundCRS()->alterCSLinearUnit(UnitOfMeasure("my unit", 2));
-        EXPECT_TRUE(createCompoundCRS()->isEquivalentTo(crs.get()));
+        auto compoundCRS = dynamic_cast<CompoundCRS *>(crs.get());
```
rouault, line 6556: for the sake of completness, and test line 393 of crs.cpp, we'd need a test on a type of CRS that has no implementation of alterCSLinearUnit(). For example a temporal or parametric CRS

jjimenezshaw: done

PR #3331, opened 2022-09-20
Fix typos/formatting in tutorials

AlexBass05:
Thanks for adding the tutorials! They are a great way to get started with PROJ.

While reading through them I spotted a couple of typos and formatting inconsistencies.

### DIFF
--- a/docs/source/tutorials/EUREF2019/answers.rst
+++ b/docs/source/tutorials/EUREF2019/answers.rst
@@ -3,7 +3,7 @@
 Answers to the PROJ tutorial exercises
 ++++++++++++++++++++++++++++++++++++++++
 
-Projections.1: UTM
+Projections1: UTM
 -------------------
 
 .. code-block::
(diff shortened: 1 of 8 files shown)


docs/source/tutorials/EUREF2019/answers.rst
```diff
@@ -13,7 +13,7 @@ Projections.1: UTM
     3.      +proj=utm +zone=35 +ellps=intl
 
 
-Projections2.: LCC
+Projections.2: LCC
```
kbevers: Shouldn't the dot just be removed all together?

AlexBass05: I don't know... maybe? The format with the dot is used above in `Projection.1` and below in `Projection.3`.
This change is only meant to align the formatting.

kbevers: I think they were supposed to reflect the naming of the gie files. Så `projection1` etc would be correct

AlexBass05: I see. Makes sense. I removed the dots.

PR #3338, opened 2022-09-22
Support for more generic Space Oblique Mercator (SOM) projection

### LINKED ISSUE #3337: Support for more generic Space Oblique Mercator (SOM) projection
Currently, the PROJ library supports the Space Oblique Mercator (SOM) projection used by the MISR instrument aboard the Terra spacecraft via the `misrsom` projection implemented at `src/projections/misrsom.cpp` (despsite the name, this projection is also used by other instruments aboard the Terra platform).

Unfortunately, this projection is a repeating 233 path based projection, and the Terra spacecraft will no longer be able to maintain that repeating path based orbit due to fuel constraints. [More info](https://terra.nasa.gov/news/terra-begins-to-drift-in-time)

In preparation for distribution of new data products, which will continue to use the SOM projection but will no longer be repeating paths, we would like to request support for a more generic SOM projection (one which will vary orbit to orbit) and which has less strong assumptions about the height / period of the orbit.

In particular, we would like to add to PROJ support for the projection supported in [GCTP as "Space Oblique Merc A"](https://github.com/tdanckaert/GCTP/blob/github/doc/appendixA.txt) which parameterizes the SOM projection by ascending longitude, inclination angle, and orbital period.

sval-dev:
- [X] Closes #3337 
 - [X] Tests added


rouault:
To avoid the code duplication, if the only difference is hard-coded parameters vs user-provided one, a better way would be to have a common setup() method to misrom and genericsom, and have the entry points of the 2 functions call it. Like "utm" and "tmerc" in src/projections/tmerc.cpp (there are other similar instances in sconics.cpp, putp6.cpp, etc.)

sval-dev:
I think I understand.

From my point of view, it looks like misrsom is a more specialized version of genericsom.

Would it work for you to keep the currently duplicated code in genericsom (with modification if necessary to support re-use) and modify misrsom to use the functions from genericsom rather than be duplicated?

rouault:
> Would it work for you to keep the currently duplicated code in genericsom (with modification if necessary to support re-use) and modify misrsom to use the functions from genericsom rather than be duplicated?

that's exactly what I meant :-)  Basically you'd remove misrom.cpp entirely, and just move its entry point ``PJ *PROJECTION(misrsom)`` inside genericsom.cpp and make it reuse genericsom code but with the few needed parameters hardcoded.

sval-dev:
Thanks! I'll make the change and push once more.

src/projections/genericsom.cpp
```diff
@@ -173,29 +186,11 @@ static PJ_LP misrsom_e_inverse (PJ_XY xy, PJ *P) {          /* Ellipsoidal, inve
...
-        return pj_default_destructor(P, PROJ_ERR_INVALID_OP_ILLEGAL_ARG_VALUE);
-    }
-
-    P->lam0 = DEG_TO_RAD * 129.3056 - M_TWOPI / 233. * path;
-    alf = 98.30382 * DEG_TO_RAD;
-    Q->p22 = 98.88 / 1440.0;
-
-    Q->sa = sin(alf);
-    Q->ca = cos(alf);
+static PJ *setup(PJ *P, SOMProjectionType somProjType) {
```
rouault: this somProjType parameter isn't use, so it could be removed and the SOMProjectionType enumeration itself too

src/projections/genericsom.cpp
```diff
@@ -221,8 +216,62 @@ PJ *PROJECTION(misrsom) {
...
+
+PJ *PROJECTION(genericsom) {
+    struct pj_opaque *Q = static_cast<struct pj_opaque*>(calloc (1, sizeof (struct pj_opaque)));
+    if (nullptr==Q)
+        return pj_default_destructor (P, PROJ_ERR_OTHER /*ENOMEM*/);
+    P->opaque = Q;
+
+    // ascending longitude (radians)
+    P->lam0 = pj_param(P->ctx, P->params, "dasc_lon").f;
+    if (P->lam0 < -M_TWOPI || P->lam0 > M_TWOPI)
```
rouault: it is quite untypical for PROJ angular projection parameters to be specified in radians. They are rather specified in degrees. If you use "rasc_lon" it will expect degrees as the input but will return the result converted in radians
Same remark for inclination angle

test/cli/tv_out.dist
```diff
@@ -424,6 +424,12 @@ Test misrsom forward projection
 Test misrsom inverse projection
 7461300.0 528000.0 0.0	48.6496622	66.2263207 0.0000000
 ##############################################################
+Test genericsom forward projection
+-120.001014 38.999576 0.        15758237.5060846        427373.6032512 0.0000000
+##############################################################
+Test genericsom inverse projection
+15758237.5060846 427373.6032512 0.0     -120.0010140    38.9995760 0.0000000
+##############################################################
```
rouault:
```suggestion
```

test/cli/testvarious
```diff
@@ -862,6 +862,29 @@ $EXE -f '%.7f' \
...
+echo  "Test genericsom inverse projection" >> ${OUT}
+#
+$EXE -f '%.7f' \
+    +proj=genericsom +inc_angle=1.7157253262878522 +ps_rev=0.06866666666666667 \
+    +asc_lon=1.1242171183417042 +ellps=WGS84 +units=m +to +proj=latlong \
+    +datum=WGS84 -f '%0.7f' \
+ -E >> ${OUT} <<EOF
+15758237.5060846 427373.6032512 0.0
+EOF
+##############################################################################
```
rouault: I would drop those changes. Testing with builtins.gie is enough and more portable (testvarious is a shell script and can't run on Windows)

```suggestion
```

src/lib_proj.cmake
```diff
@@ -68,7 +68,7 @@ set(SRC_LIBPROJ_PROJECTIONS
   projections/gall.cpp
   projections/labrd.cpp
   projections/lsat.cpp
-  projections/misrsom.cpp
+  projections/genericsom.cpp
```
rouault: we should probably just drop the "generic" in the file name and projection method id. Generic is kind of implied

src/projections/genericsom.cpp
sval-dev: I'm currently trying to unify with lsat.cpp as well, where this will be in use. (Assuming that is okay)

rouault: > I'm currently trying to unify with lsat.cpp as well, where this will be in use. (Assuming that is okay)

That would be fantastic! Indeed the side-to-side comparison between lsat.cpp and misrom.cpp shows a lot of similarities !

src/projections/genericsom.cpp
sval-dev: It is stored internally and displayed to users as radians in the data products, so that's the direction I went in since it saves users a conversion. If it isn't expected behavior here though, I can change it (and the tests) to match.

test/cli/testvarious
sval-dev: It's okay with me. I was matching misrsom behavior here though, so perhaps those tests should be removed as well for consistency?

src/lib_proj.cmake
sval-dev: Will do.

test/cli/testvarious
rouault: yes, likely.
one thing to be aware is that the tests in builtins.gie have been all generated for the same input coordinates, which don't necessarily tests all the relevant code paths of the given projection method. So sometimes it can be appropriate to insert input coordinates that are dependent on the projection method.

src/projections/genericsom.cpp
rouault: Maybe @kbevers want to chime in ?

src/projections/genericsom.cpp
sval-dev: I moved init earlier so that this parameter was no longer required (and neither was the enum).

test/cli/tv_out.dist
sval-dev: Removed as requested.

test/cli/testvarious
sval-dev: Removed as requested

src/lib_proj.cmake
sval-dev: Changed name and methods to remove generic prefix.

src/projections/genericsom.cpp
sval-dev: As some additional context, I'll mention that GCTP takes the values in radians which might be why this was adopted for the MISR output products as well (i.e. so that the projection parameters specified in the products could be directly used to init GCTP projections).
The [MISR DPS](https://asdc.larc.nasa.gov/documents/misr/DPS_v50_RevS.pdf) mentions:
"[GCTP's] for_init takes the projection parameters from step a. as arguments" where "step a" refers to the result of [HDF-EOS GDprojinfo()](https://hdfeos.org/examples/c_grid_lonlat.php) used on the MISR data products.

That being said, I'm still happy to change it to degrees and document the same if that's the preferred convention to use here.

kbevers: I think it would be confusing to use radians as input here. At least as the only option. You could possibly implement options for both degrees (`+asc_lon`) and radians (`+asc_lon_rad`). They would of course have to be mutally exclusive.

rouault: > You could possibly implement options for both degrees (`+asc_lon`) and radians (`+asc_lon_rad`).

Actually looking at the dmstor_ctx() function (which is used by pj_param("r{parameter_name}")), I've discovered a hidden feature (at least to me, but I see from commit history that it dates back to the initial commit of PROJ4!). If the value is immediately followed by 'r' or 'R' , it mean radians (you may put anything after the r

```
$ echo 0 0 | bin/cs2cs +proj=longlat +to +proj=longlat +lon_0=1.5707963267948966r
90dW	0dN 0.000

$ echo 0 0 | bin/cs2cs +proj=longlat +to +proj=longlat +lon_0=1.5707963267948966radwhatever
90dW	0dN 0.000

$ echo 0 0 | bin/cs2cs +proj=longlat +to +proj=longlat +lon_0=90
90dW	0dN 0.000
```

So there's nothing to do but grabbing the parameter with pj_param(P->ctx, P->params, "rasc_lon").f to trigger the use of dmstor_ctx()

kbevers: Well isn't that just lovely! Solves the issue at hand it would seem. Just make sure to mention it in the docs for the projection so that users are made aware of that option.

Related question: Where do we document this generic feature? Somewhere in https://github.com/OSGeo/PROJ/blob/master/docs/source/usage/projections.rst perhaps?

rouault: > Somewhere in https://github.com/OSGeo/PROJ/blob/master/docs/source/usage/projections.rst perhaps?

would be a good idea.  Formats like ``1d45'60''W`` are not documented too

sval-dev: Thanks for the feedback! I have updated the radian input arguments to the "r" (degrees) parameter type and updated the docs to match as well as adding tests for both decimal degrees and radian inputs (using the "r" suffix to the input value).

Please let me know if this is what you were thinking or if you'd like anything further.

docs/source/operations/projections/som.rst
```diff
@@ -0,0 +1,71 @@
...
++---------------------+----------------------------------------------------------+
+| **Alias**           | som                                                      |
++---------------------+----------------------------------------------------------+
+| **Domain**          | 2D                                                       |
++---------------------+----------------------------------------------------------+
+| **Input type**      | Geodetic coordinates                                     |
++---------------------+----------------------------------------------------------+
+| **Output type**     | Projected coordinates                                    |
++---------------------+----------------------------------------------------------+
+
```
kbevers, line 33: It would be great to have some background info here (see https://proj.org/operations/projections/merc.html for a good example). You seem to know what this projection is used for and possibly also the history behind it. It would be great to capture that knowledge here so the next person doesn't start from scratch.

I guess this also goes for the misrsom projection.

sval-dev: I've given a shot at a bit of background and linked to the relevant references which hopefully should cover the curious.

kbevers: Looks great, thanks!

src/projections/genericsom.cpp
kbevers: Looks good to me

rouault (approved):

PR #3346, opened 2022-09-26
Add preliminary documentation of suffix based specification of units

### LINKED ISSUE #3345: Add documentation for unit specification of input parameter values
There is some automatic conversion that can happen dmstor_ctx() (in `src/dmstor.cpp`) such that the input values for some parameters can be interpreted as either decimal degrees or radians, and converted appropriately into radians (i.e. converted into radians for the value that is returned in PROJVALUE.f).

This feature request adds some preliminary documentation for this suffix-based unit specification and the convention that decimal degrees is the default interpretation.

sval-dev:
Add preliminary documentation of suffix based specification of units for input parameter values.

 - [X] Closes #3345

### DIFF
--- a/docs/source/usage/projections.rst
+++ b/docs/source/usage/projections.rst
@@ -38,6 +38,8 @@ documenting the individual :doc:`projections<../operations/projections/index>`.
 
 In the sections below most of the parameters are explained in details.
 
+.. _projection_units:
+
 Units
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
--- a/docs/source/usage/projections.rst
+++ b/docs/source/usage/projections.rst
@@ -85,9 +85,9 @@ suffix to input values.
     | R              |                     |
     +----------------+---------------------+
 
-Example of use.  The longitude ``long=90``, can also be expressed more explictly
-with units of decimal degrees as ``long=90d`` and can also be expressed in radian
-units as ``long=1.5708r`` (approximately).
+Example of use.  The longitude of the central meridian ``+lon_0=90``, can also be expressed more explictly
+with units of decimal degrees as ``+lon_0=90d`` or in radian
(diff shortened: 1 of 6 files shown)


docs/source/usage/projections.rst
```diff
@@ -64,6 +64,32 @@ Scaling of output units can be done by applying the ``+k_0`` argument. The
...
+    | °              |                     |
+    +----------------+---------------------+
+    | r              | Radians             |
+    +----------------+                     +
+    | R              |                     |
+    +----------------+---------------------+
+
+Example of use.  The longitude ``long=90``, can also be expressed more explictly
+with units of decimal degrees as ``long=90d`` and can also be expressed in radian
+units as ``long=1.5708r`` (approximately).
```
kbevers: Reworded to use a well-known projection parameter:

```suggestion
Example of use.  The longitude of the central meridian ``+lon_0=90``, can also be expressed more explictly
with units of decimal degrees as ``+lon_0=90d`` or in radian
units as ``+lon_0=1.5708r`` (approximately).
```

kbevers (changes requested):
Thanks for adding this! I've suggested a different wording of the example using the central meridian.

It would be great to mention this feature in the docs for each of the generic options where it is useful. Perhaps just a simple example of how to give the value as radian and a link back to the new section of the docs provided in this PR. I believe the relevant options are `+lon_0`, `+lat_0`, `+lat_1`, `lat_2` and `+lat_ts`. They described here: https://github.com/OSGeo/PROJ/tree/master/docs/source/operations/options

kbevers (approved):
This looks great, thanks for brushing the docs up a bit

sval-dev:
Added some examples to the suggestion operation options.

sval-dev:
> This looks great, thanks for brushing the docs up a bit

Glad to help!

PR #3361, opened 2022-10-05
createOperations(): emulate PROJ < 6 behavior when doing geocentric <--> geographic transformation between datum with unknown transformation (fixes #3360)

### LINKED ISSUE #3360: Incorrect propagation of +R parameter in transform pipeline
#### Example of problem
There is an inconsistency in behaviour between PROJ4 and PROJ9 for the transform from EPSG:4978 (ECEF) to ESRI:104199.

Here are the definitions of the two CRS:

```
$ projinfo epsg:4978 -o PROJ
PROJ.4 string:
+proj=geocent +datum=WGS84 +units=m +no_defs +type=crs

$ projinfo esri:104199 -o PROJ
Warning: object is deprecated

PROJ.4 string:
+proj=longlat +R=6378137 +no_defs +type=crs
```

When transforming a point using PROJ4 (version 4.9.3), I get the expected result:

```
$ echo "4057449.87922115 1236185.10760693 4747287.50008112" | cs2cs +proj=geocent +datum=WGS84 +units=m +no_defs +type=crs +to +proj=longlat +R=6378137 +no_defs +type=crs -f "%.10f"
16.9444298151   48.4111246681 -74.6128296684
```

However, when using PROJ9 (version 9.1.0), I get a different (incorrect) result:
```
$ echo "4057449.87922115 1236185.10760693 4747287.50008112" | cs2cs +proj=geocent +datum=WGS84 +units=m +no_defs +type=crs +to +proj=longlat +R=6378137 +no_defs +type=crs -f "%.10f"
16.9444298151   48.2199854103 -11992.8176014163
```

Even using Authority Codes with PROJ9 results in the same incorrect result:
```
$ echo "4057449.87922115 1236185.10760693 4747287.50008112" | cs2cs epsg:4978 esri:104199 -f "%.10f"
48.2199854103   16.9444298151 -11992.8176014163
```

#### Problem description

Inspecting the PROJ9 transform pipeline shows:
```
$ projinfo -s epsg:4978 -t esri:104199 -o PROJ
Candidate operations found: 1
Operation No. 1:

unknown id, Ballpark geocentric translation from WGS 84 to GCS_WGS_1984_Major_Auxiliary_Sphere (geocentric) + Conversion from GCS_WGS_1984_Major_Auxiliary_Sphere (geocentric) to GCS_WGS_1984_Major_Auxiliary_Sphere, unknown accuracy, World, has ballpark transformation

PROJ string:
+proj=pipeline
  +step +inv +proj=cart +R=6378137
  +step +proj=longlat +R=6378137
```

I think the problem is the first step incorrectly contains an unexpected +R parameter:
`+step +inv +proj=cart +R=6378137`

This transform step should use the ellipsoid definition from EPSG:4978 (WGS84), and _not_ include the +R value, which should only apply to the ESRI:104199 transform step. I expected this first step to be something like:
`+step +inv +proj=cart +ellps=WGS84`

#### Expected Output
I would expect to see a transform pipeline similar to:
```
+proj=pipeline
  +step +inv +proj=cart +ellps=WGS94
  +step +proj=longlat +R=6378137
```
In fact, if I manually construct a transform with this first step corrected (and remove the axisswap), then I get an output that exactly matches the expected output from PROJ4:
```
$ echo "4057449.87922115 1236185.10760693 4747287.50008112" | cct -d 10 +proj=pipeline +step +inv +proj=cart +ellps=WGS84 +step +proj=longlat +R=6378137 +step +proj=unitconvert +xy_in=rad +xy_out=deg
 16.9444298151   48.4111246681  -74.6128296675           inf
```

It looks to me that there is some logic in the pipeline builder that is incorrectly propagating the +R parameter between steps, when it should not.

Is this behaviour expected? Or is there some other way to get behaviour more consistent with PROJ4?

Many thanks

#### Environment Information
 - PROJ version (9.1.0)
 - Operation System: Ubuntu Linux

#### Installation method
 - Compiled from source

rouault:
We switch from one arbitrary to the other one. The new behavior is perhaps slightly better because, besides matching PROJ <6 results, it matches also the behavior when operating in the geographic coordinate space between datums without known transformations.
@kbevers opinions ? I considered at that point to add some API switch to select which behavior is desired, but that seems overkill and nobody will figure that they need to use such potential switch, right ?

Likely not appropriate for 9.1.x backport

### DIFF
--- a/src/iso19111/operation/coordinateoperationfactory.cpp
+++ b/src/iso19111/operation/coordinateoperationfactory.cpp
@@ -4037,6 +4037,10 @@ void CoordinateOperationFactory::Private::createOperationsGeodToGeod(
             res.emplace_back(
                 Conversion::createGeographicGeocentric(sourceCRS, targetCRS));
         } else if (isSrcGeocentric && geogDst) {
+#if 0
+            // The below logic was used between PROJ >= 6.0 and < 9.2
+            // It assumed that the geocentric origin of the 2 datums
+            // matched.
             std::string interm_crs_name(geogDst->nameStr());
             interm_crs_name += " (geocentric)";
             auto interm_crs =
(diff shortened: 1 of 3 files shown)


kbevers:
I don't see why this solution is arbitrary. To me it is obviously correct to do the `+proj=cart +inv` step using the WGS84 ellipsoid and not the radius of a sphere. This might be easier for human eyes to see than for a computer program that follows strict rules.

In fact, the `+proj=longlat +R=6378137` is not really doing anything in the pipeline, as seen below where I've removed it in the last example. 
```
$ echo "4057449.87922115 1236185.10760693 4747287.50008112" | cct -t0 +proj=pipeline +step +inv +proj=cart +ellps=WGS84 +step +proj=longlat +R=6378137 +step +proj=unitconvert +xy_in=rad +xy_out=deg +step +proj=axisswap +order=2,1
 48.4111246681   16.9444298151      -74.6128        0.0000
```

```
$ echo "4057449.87922115 1236185.10760693 4747287.50008112" | cct -t0 +proj=pipeline +step +inv +proj=cart +ellps=WGS84 +step +proj=unitconvert +xy_in=rad +xy_out=deg +step +proj=axisswap +order=2,1     
 48.4111246681   16.9444298151      -74.6128        0.0000
```

Had there been a a projection in play on the output CRS the `+R` would of course play a role but the `+proj=latlong +R=6378137` step would still be unnecessary.

So in conclusion, looks more like a bug fix than a breaking change in behaviour to me.

rouault:
> To me it is obviously correct

well, I'm sill a bit perplexed by that. anyway... Adding the 9.1 backport label then

ValZapod:
> is obviously correct to do the `+proj=cart +inv` step using the WGS84 ellipsoid and not the radius of a sphere.

Geocentric (ECEF) does not have an ellipsoid... Right... BTW, did you check what other do? Like Global Mapper? Or ArcGIS.

kbevers:
> Geocentric (ECEF) does not have an ellipsoid... Right...

But WGS84 as a datum does, and that needs to be taken into consideration when converting from cartesian coordinates to geodetic coordinates.

PR #3362, opened 2022-10-05
CMake: add a uninstall target

rouault:
updated and enhanced from GDAL


kbevers:
Perhaps this should be added to the docs too? Somewhere around here https://proj.org/install.html#compilation-and-installation-from-source-code

rouault:
> Perhaps this should be added to the docs too?

done

PR #3377, opened 2022-10-10
Update users.rst

morandiaye:


docs/source/users.rst
```diff
@@ -43,6 +43,7 @@ Private organizations
 ---------------------
 
 (add yourself here if you are a direct user of PROJ API or utilities)
+Mor Ndiaye
```
rouault, line 46: I believe we'd prefer organizations than individuals to be listed. so here looking at your github profile, probably SAED would be more appropriate

PR #3379, opened 2022-10-10
Add link to proj Rust crate

urschrei:
Update `users.html` with link to crate.


kbevers:
I think this is better suited for the language bindings list: https://proj.org/development/bindings.html

urschrei:
Ah, my mistake, and I already added it.

rouault:
in https://github.com/OSGeo/PROJ/pull/3382, I've created a sub paragraph to link to bindings.rst

PR #3402, opened 2022-10-16
Add CITATION.cff

mwtoews:
This PR replaces .zenodo.json with a [CITATION.cff](https://citation-file-format.github.io/), which is used by Zenodo, GitHub, and others. The YAML-based file can be processed into several other formats.

Two entries for "Danish Agency for Data Supply and Efficiency" are updated to the current agency name "Danish Agency for Data Supply and Infrastructure".

This PR assumes both CITATION and CITATION.cff could co-exist, however this is an open question. Should CITATION be removed? If this file serves any purpose, it is more concise to reference.


.zenodo.json
```diff
@@ -1,79 +0,0 @@
...
-        "cartography",
-        "coordinates",
-        "projections",
-        "conversion"
-    ],
-    "license": {
-        "id": "MIT"
-    },
-    "title": "PROJ"
-}
```
mwtoews, line 79: this file is re-generated using `cffconvert -f zenodo`

rouault:
> The YAML-based file can be processed into several other formats.

yes, but I assume that zenodo.org automatic creation of DOI from github releases needs the .zenodo.json file to be in the repository, doesn't it ?

rouault:
ok, forget my previous comment: https://citation-file-format.github.io/ clarifies that zenodo can use CITATION.cff

rouault (approved):

CITATION.cff
```diff
@@ -0,0 +1,69 @@
+cff-version: 1.2.0
+message: Please cite this software using these metadata or in the CITATION file.
+type: software
+title: PROJ
+version: 9.0.1
```
kbevers: Aren't we at 9.1.0 now?

mwtoews: Yes, thanks!

PR #3406, opened 2022-10-20
Simplify INSTALL to a short document

### LINKED ISSUE #3400: INSTALL file seems irrelevant
#### Problem description

As far as I can tell, the INSTALL file in the source code dist is completely irrelevant — it is the generic GNU autotools INSTALL file, but the build process seems to have changed to use cmake.

#### Environment Information
 - PROJ version 9.1.0

#### Installation method
 - from source

mwtoews:
An alternative option is to delete the file. I don't mind keeping it, so long as it is short and points to the main online documentation.

Closes #3400


kbevers (approved):
Looks good to me

PR #3421, opened 2022-10-31
Replaced most uses of sprintf with the safer snprintf

seanm:


seanm:
Thanks @rouault !  There are 4 calls to `sprintf` remaining that would be nice to eliminate, but they are trickier due to some pointer math, and I don't know this codebase at all.  Could you maybe take a look?

rouault:
> Could you maybe take a look?

please file a ticket about that

PR #3424, opened 2022-11-02
Misc warnings

seanm:


src/iso19111/c_api.cpp
```diff
@@ -289,7 +289,6 @@ PJ_OBJ_LIST::~PJ_OBJ_LIST() = default;
  * @param ctx Ignored
  * @param autoclose Ignored
  * @since 6.2
- * @deprecated Since 8.1
```
rouault, line 292: Please keep it as a regular Doxygen warning then, without the '@'

src/list.cpp
```diff
@@ -1,10 +1,10 @@
 /* Projection System: default list of projections
-** Use local definition of PJ_LIST_H for subset.
+** Use local definition of PJ_LIST_H (should this be USE_PJ_LIST_H ??) for subset.
 */
 
 #include "proj.h"
 
-#define USE_PJ_LIST_H 1
+#define USE_PJ_LIST_H 1 // -Wunused-macro here. Indeed "USE_PJ_LIST_H" appears nowhere else in code
```
rouault: could be removed
```suggestion
```

src/list.cpp
```diff
@@ -1,10 +1,10 @@
 /* Projection System: default list of projections
-** Use local definition of PJ_LIST_H for subset.
+** Use local definition of PJ_LIST_H (should this be USE_PJ_LIST_H ??) for subset.
```
rouault:
```suggestion
```

src/iso19111/c_api.cpp
seanm: Alternatively, does PROJ have a mechanism to mark things deprecated?  For example, is the [C++14 `[[deprecated]]`](https://en.cppreference.com/w/cpp/language/attributes/deprecated) too new?

rouault: > For example, is the [C++14 [[deprecated]]](https://en.cppreference.com/w/cpp/language/attributes/deprecated) too new?

yes, we stick to C++11 for now

src/list.cpp
seanm: Removed.

src/iso19111/c_api.cpp
seanm: OK, I just removed the `@` character then.

seanm:
Oh, didn't expect you to merge yet, one of the commits was still marked WIP...

rouault:
> Oh, didn't expect you to merge yet, one of the commits was still marked WIP...

ah, didn't notice the WIP (best to use to use the github feature to tag the pull request as Draft then). Our CI has Windows workers. What is unknown if that works with UWP for which we don't have CI. @gorlak does https://github.com/OSGeo/PROJ/pull/3424/commits/fecca6f2a1619347a9026cdb21a3576a1e550ef6 look good to you ?

gorlak:
That change makes sense to me 👍

PR #3425, opened 2022-11-02
Replaced platform-specific mutex implementations with std::recursive_…

seanm:
…mutex

This fixes a clang -Wreserved-id-macro warning by no longer ever needing to #define _GNU_SOURCE.


rouault:
The following snippet in src/lib_proj.cmake 
```
option(USE_THREAD "Build libproj with thread/mutex support " ON)
if(NOT USE_THREAD)
  add_definitions(-DMUTEX_stub)
endif()
find_package(Threads QUIET)
if(USE_THREAD AND Threads_FOUND AND CMAKE_USE_WIN32_THREADS_INIT)
  add_definitions(-DMUTEX_win32)
elseif(USE_THREAD AND Threads_FOUND AND CMAKE_USE_PTHREADS_INIT)
  add_definitions(-DMUTEX_pthread)
elseif(USE_THREAD AND NOT Threads_FOUND)
  message(FATAL_ERROR
    "No thread library found and thread/mutex support is "
    "required by USE_THREAD option")
endif()
```
should also be reworked to something like:
```
find_package(Threads QUIET)
if(NOT Threads_FOUND)
  message(FATAL_ERROR  "No thread library found")
endif()
```
and include/proj/internal/mutex.hpp should be removed, and its users modified to use std::mutex

seanm:
>and include/proj/internal/mutex.hpp should be removed, and its users modified to use std::mutex

So the comment about "mingw32-win32 doesn't implement std::mutex" is no longer a concern?

In the meantime, I've updated the CMake files...

rouault:
> So the comment about "mingw32-win32 doesn't implement std::mutex" is no longer a concern?

I was wondering, but we have no CI config testing that, and I'm not sure we should care abou a platform not implementing std::mutex

seanm:
> I was wondering, but we have no CI config testing that, and I'm not sure we should care abou a platform not implementing std::mutex

Well, as you require C++11 already, you are essentially already requiring `std::mutex` :)

PR #3430, opened 2022-11-03
Various -Wreserved-id-macro warning fixes

seanm:


rouault:
Could you rebase on top of latest master to solve the conflict in src/projections/s2.cpp ?

hobu:
These changes are a bit worrisome because there's a small chance external users of PROJ are depending on these macro names to control their behavior. Would you have that same worry @rouault?

rouault:
> Would you have that same worry @rouault?

no worry on my side. Those changes only affect .cpp files and internal headers. I don't see why/how someone would need to mess with those internal #define. This might only affect people who have a fork of PROJ with additional projection methods, but that's their responsibility to update their code with those changes

seanm:
I've done similar fixes to header guard names in several other projects and it hasn't been a problem in practice, though I did have the same worry at first.

PR #3431, opened 2022-11-03
Remove remaining sprintf() uses, and deprecate proj_rtodms() by newly added proj_rtodms2()

### LINKED ISSUE #3423: Replace remaining 4 calls to `sprintf` with `snprintf`
After https://github.com/OSGeo/PROJ/pull/3421 there are still 4 uses of `sprintf` that should be replaced with `snprintf`.

rouault:
Fixes #3423

### DIFF
--- a/docs/source/development/reference/functions.rst
+++ b/docs/source/development/reference/functions.rst
@@ -828,8 +828,8 @@ Various
 
     Convert radians to string representation of degrees, minutes and seconds.
 
-    Note: starting with PROJ 9.2, this function is deprecated by
-    :cpp:func:`proj_rtodms2`
+    .. deprecated:: 9.2
+       Use :cpp:func:`proj_rtodms2` instead.
 
     :param s: Buffer that holds the output string
     :type s: `char *`
--- a/docs/source/development/reference/functions.rst
+++ b/docs/source/development/reference/functions.rst
@@ -828,8 +828,30 @@ Various
 
     Convert radians to string representation of degrees, minutes and seconds.
 
+    Note: starting with PROJ 9.2, this function is deprecated by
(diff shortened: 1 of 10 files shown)


docs/source/development/reference/functions.rst
```diff
@@ -828,8 +828,30 @@ Various
 
     Convert radians to string representation of degrees, minutes and seconds.
 
+    Note: starting with PROJ 9.2, this function is deprecated by
+    :cpp:func:`proj_rtodms2`
```
mwtoews: Shouldn't the [deprecated directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) be used used? (does it work?)
```suggestion
    .. deprecated:: 9.2
       Use :cpp:func:`proj_rtodms2` instead.
```

seanm:
github didn't notify me about this PR until just now, when you closed it, so I'm only reviewing the code now, but I think there is a mistake. You did not account for the pointer arithmetic on `ss` that I mentioned in #3423. 

```
char *ss = s;
...
if (!pos) { *ss++ = '-'; sign = 0; }
```

After this point, `s` and `ss` are not always pointing to the same place.

Then you do:

```
snprintf(ss,sizeof_s,...
```

What you need is a `sizeof_ss` that starts the same as `sizeof_s` but gets decremented when `ss` gets incremented.

Otherwise you have a buffer overflow.  If I'm not mistaken...

rouault:
> You did not account for the pointer arithmetic on `ss` that I mentioned in #3423.

doh! thanks for the review. Should hopefully be fixed with https://github.com/OSGeo/PROJ/pull/3441

PR #3441, opened 2022-11-07
rtodms(): fix potential buffer overflow not dealt with with PR #3431 (master only)

rouault:


src/rtodms.cpp
```diff
@@ -45,10 +45,19 @@ rtodms(char *s, size_t sizeof_s, double r, int pos, int neg) {
...
 
 	if (r < 0) {
 		r = -r;
-		if  (!pos) { *ss++ = '-'; sign = 0; }
+		if  (!pos) {
+		    if( sizeof_s == 1 ) {
+		        *s = 0;
+		        return s;
+		    }
+		    sizeof_s --;
```
seanm: `sizeof_ss--` :)

rouault: sigh... thanks again. commit amended with the fix and more extensive tests, to be faithful with my motto "untested code is broken code"

seanm: LGTM now. thanks!

PR #3453, opened 2022-11-11
cs2cs: better validate value of -W option (fixes #3451)

### LINKED ISSUE #3451: -W option: Valid range: -W0 through -W8 only
W9 = -W3, and -W10 = crash

```
==== cs2cs -W7 ...
43d05'18.7380000"E	79d04'11.1590000"S 0.000
==== cs2cs -W8 ...
43d05'18.73800000"E	79d04'11.15900000"S 0.000
==== cs2cs -W9 ...
43d5'18.738"E	79d4'11.159"S 0.000
==== cs2cs -W10 ...
Rel. 9.1.0, September 1st, 2022
<cs2cs>:
invalid option: -0
program abnormally terminated
```
I only tested -W, not -w.

rouault:


src/apps/emess.cpp
```diff
@@ -27,8 +27,19 @@ emess(int code, const char *fmt, ...) {
...
+        const char* pszDisplayProgramName = getenv("PROJ_DISPLAY_PROGRAM_NAME");
+        if( pszDisplayProgramName && strcmp(pszDisplayProgramName, "NO") == 0 )
+        {
+            // do nothing
+        }
+        else
+        {
+            (void)fprintf(stderr,"%s\n<%s>: ",pj_get_release(),
+                                  emess_dat.Prog_name);
+        }
```
kbevers: Is there a reason this expression can't be negated? Something like 
```suggestion
        if( !pszDisplayProgramName && strcmp(pszDisplayProgramName, "NO") != 0 )
        {
            (void)fprintf(stderr,"%s\n<%s>: ",pj_get_release(),
                                  emess_dat.Prog_name);
        }
```

rouault: > Is there a reason this expression can't be negated ?

no, just that humain brain tends to have issues with negation of clauses with and operation. Your negation proposal was wrong ;-).
I've changed to ``if( !(pszDisplayProgramName && strcmp(pszDisplayProgramName, "NO") == 0) )``

kbevers: > Your negation proposal was wrong ;-). 

Hence the "something like" :-) Improved readibility is a valid reason for your original code. Just looked a bit odd having a code branch that does nothing

busstoptaktik: Yet another reason to import the perl "unless" (negated if) keyword into C/C++ 🙂

Issue #3428, opened 2022-11-02
BOUNDCRS with a derived geographic as SOURCECRS doesn't work as expected


hernando:
#### Example of problem

I'm trying to express vertical offsets over different types of CRSs as well defined WKT2 strings. This time is the turn of a bound CRS. It seems that a derived of a bound is not accepted by the WKT2 grammar (to be confirmed independently), so I tried to turn the definition around and write a bound of a base derived and conventional target.

```
TGT='
BOUNDCRS[
    SOURCECRS[
        GEOGCRS["CH1903+ with height offset",
            BASEGEOGCRS[
                "CH1903+",
                DATUM["CH1903+",
                    ELLIPSOID["Bessel 1841",6377397.155,299.1528128,
                        LENGTHUNIT["metre",1]]],
                PRIMEM["Greenwich",0,
                    ANGLEUNIT["degree",0.0174532925199433]]],
            DERIVINGCONVERSION["Ellipsoidal to gravity related height",
                METHOD["Geographic3D offsets",
                    ID["EPSG",9660]],
                PARAMETER["Latitude offset",0,
                    ANGLEUNIT["degree",0.0174532925199433],
                    ID["EPSG",8601]],
                PARAMETER["Longitude offset",0,
                    ANGLEUNIT["degree",0.0174532925199433],
                    ID["EPSG",8602]],
                PARAMETER["Vertical Offset",10,
                    LENGTHUNIT["metre",1],
                    ID["EPSG",8603]]],
            CS[ellipsoidal,3],
                AXIS["geodetic latitude (Lat)",north,
                    ORDER[1],
                    ANGLEUNIT["degree",0.0174532925199433,
                        ID["EPSG",9122]]],
                AXIS["geodetic longitude (Lon)",east,
                    ORDER[2],
                    ANGLEUNIT["degree",0.0174532925199433,
                        ID["EPSG",9122]]],
                AXIS["ellipsoidal height (h)",up,
                    ORDER[3],
                    LENGTHUNIT["metre",1]]]],
   TARGETCRS[
       GEOGCRS["WGS 84",
           ENSEMBLE["World Geodetic System 1984 ensemble",
               MEMBER["World Geodetic System 1984 (Transit)"],
               MEMBER["World Geodetic System 1984 (G730)"],
               MEMBER["World Geodetic System 1984 (G873)"],
               MEMBER["World Geodetic System 1984 (G1150)"],
               MEMBER["World Geodetic System 1984 (G1674)"],
[...]

PR #3454, opened 2022-11-11
PROJJSON: bump to 0.6 with an additional optional source_crs property…

rouault:
… in abridged_transformation (refs #3428)


rouault:
@kbevers does that look reasonable to you ? This is an extension of the BoundCRS of WKT adding an explicit sourceCRS to the transformation embedded in the BoundCRS (as it is sometimes a bit fuzzy to know exactly to which CRS the transformation applies too when the sourceCRS of the BoundCRS is something more complex than a regular geographicCRS), but specific to PROJJSON as I don't want to mess with WKT

kbevers:
I'm not sure about this. Is this the first time PROJJSON conceptually diverges from WKT?

rouault:
> I'm not sure about this. Is this the first time PROJJSON conceptually diverges from WKT?

no, we allow BOUNDCRS[] to be used as one of the horizontal or vertical component of a CompoundCRS for example, and we actually allow that in the C++ code, WKT and PROJJSON, wheras it isn't allowed by the WKT spec. Actually we treat a BoundCRS as a CRS, whereas BoundCRS is only defined in WKT and its status is not clearly the one of a CRS but some object in-between a CRS and a transformation.

kbevers:
> wheras it isn't allowed by the WKT spec

Does that mean we produce incorrect WKT? 

I'm not sure I've got a strong opinion on this matter. Ideally PROJJSON is a just a JSON representation of WKT but if WKT is insufficiently specced it makes sense to do something smarter with PROJJSON.

rouault:
> we produce incorrect WKT?

yes, or "extended" in commercial terms :-)

hernando:
As I understand, this change won't introduce anything invalid in a WKT2, it just extends PROJJSON with information that is known to the BoundCRS C++ class but can't be expressed in WKT2. The lack of this information renders the WKT2 ambiguous in the presence of a derived CRS as the base.
I would prefer to see a solution in the WKT standard, but despite I've raised the question in the OGC's coordtrans mailing list, the discussion didn't make any progress for almost 2 weeks.

kbevers:
> yes, or "extended" in commercial terms :-)

Well, we are know to be quite commercial around here :-)

> I would prefer to see a solution in the WKT standard, but despite I've raised the question in the OGC's coordtrans mailing list, the discussion didn't make any progress for almost 2 weeks.

Adding to the standard is going to be a slow process. PROJJSON can possibly be used as a sort of beta test for new concepts that are candidates to be included in WKT. I think that's a good enough reason to merge this PR.

PR #3456, opened 2022-11-12
Update transformation.rst

jidanni:
Note: I have only mentioned that the +'s went away.

Apparently the author is using +init files here, and showing only their contents.

However even if the user placed the example snippets seen in the boxes into files in the current directory, there is no complete working example!

I am saying that earlier in this book, the user was able to run the examples shown.

But now if the user wishes to continue running them, he has to put the +'s back,

or use some special syntax that you haven't shown yet in a complete working example!

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/docs/source/usage/transformation.rst
+++ b/docs/source/usage/transformation.rst
@@ -70,7 +70,9 @@ solutions for a wide spectrum of geodetic tasks.
 
 As a first example, let us take a look at the iconic
 *geodetic → Cartesian → Helmert → geodetic* case (steps 2 to 4 in the example in
-the introduction). In PROJ it can be implemented as
+the introduction). 
+(Note, for the rest of this book the "+"s are removed from the examples.)
+In PROJ it can be implemented as
 
 ::
 


jidanni:
Looking at https://proj.org/resource_files.html apparently in +init files the pluses are still there.

That means the author has simply removed all the pluses,
or is running in some unexplained non Unix environment etc.
Anyway it is a compete mystery to the reader.

jidanni:
So my change should also mention "Be sure to put the pluses back in order to run these examples."

jidanni:
Also regarding

> step proj=helmert convention=coordinate_frame
> x=-81.0703 y=-89.3603 z=-115.7526
> rx=-0.48488 ry=-0.02436 rz=-0.41321 s=-0.540645
> step proj=cart inv ellps=GRS80

do mention if it matters or not if we instead write e.g.,

> step proj=helmert
> convention=coordinate_frame
> x=-81.0703 y=-89.3603
> z=-115.7526
> rx=-0.48488
> ry=-0.02436
> rz=-0.41321
> s=-0.540645 step
> proj=cart inv
> ellps=GRS80

i.e., newline characters are significant or not.

kbevers:
A few comments:

- It's confusing to call the docs a book. The same material is presented as a website, a PDF-file (which I assume is what you are looking at) and man-files. It's not written as a cronological story and should probably not be read that way either.
- The plusses are only a requirement on the command line. You can leave them out in API-calls and init-files.
- I agree that it is confusing that sometimes the plusses are there and other times they are not

cf. the last point, in my opinion the correct fix here is to align the docs so that they use the +-notation everywhere (maybe with a few exceptions).

jidanni:
Haha so that's the mystery. API calls. Maybe at this point in the "book" just mention that's what's going on in it, and have a link to show people how to do one. 

As far as init files, the ones that I encountered with 
+init= needed the pluses in in them too. So there are probably two kinds of init files...

kbevers:
> Maybe at this point in the "book" just mention that's what's going on in it, and have a link to show people how to do one.

You are the one doing the PR so maybe just fix it yourself

jidanni:
I find some API documentation in
https://proj.org/development/reference/cpp/cpp_general.html
but it is way too advanced.

So perhaps use a sed script to add "+"s on all the examples, as you mentioned.

PR #3457, opened 2022-11-12
Update proj.1

jidanni:
Got rid of weird formatting, and flipped words around too.

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/man/man1/proj.1
+++ b/man/man1/proj.1
@@ -59,10 +59,8 @@ and allows bypassing formatting operations.
 .INDENT 0.0
 .TP
 .B \-d <n>
+Specify the number of decimals in the output. (New in version 5.2.0.)
 .UNINDENT
-.sp
-New in version 5.2.0: Specify the number of decimals in the output.
-
 .INDENT 0.0
 .TP
 .B \-i


rouault:
This file is automatically generated from docs/source/apps/proj.rst . Only the later should be modified

jidanni:
I thought I was smart by editing this, instead of raw nroff output, but as you say I wasn't smart enough.

Perhaps warnings could be placed in comments at the top of these files so others don't make the same mistake.

They could say be sure you're editing a .rst file instead of a .1 etc.file.

rouault:
the first line is ".\" Man page generated from reStructuredText." . "generated" implies "do not modify it directly". I'm not sure we have much control on this first line. It is written by Sphinx.

jidanni:
Seems like my eyeballs missed that obvious warning sign!

jidanni:
All I know is in https://github.com/OSGeo/PROJ/edit/master/docs/source/apps/cs2cs.rst directly using that ..versionadded in the middle of the item makes ugly output... But for now I don't have the entire testbed installed.

```
.. option:: -d <n>

    .. versionadded:: 5.2.0

    Specify the number of decimals in the output.
```

So I'll drop the issue.

PR #3465, opened 2022-11-14
Update cct.rst

jidanni:
The cct man page doesn't mention what input cct expects.

I'm making a first wild stab at fixing the problem. Please adjust. Thanks.

(Yes, I expect the lines will get reformatted to width 80 when read.)

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/docs/source/apps/cct.rst
+++ b/docs/source/apps/cct.rst
@@ -48,8 +48,13 @@ file referenced by the {object_reference} must contain a valid
 {object_definition}.
 
     .. versionadded:: 8.0.0
-
-
+    
+Input should be lines of consisting of:
+longitude (in degrees),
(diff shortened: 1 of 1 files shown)


docs/source/apps/cct.rst
```diff
@@ -48,8 +48,13 @@ file referenced by the {object_reference} must contain a valid
 {object_definition}.
 
     .. versionadded:: 8.0.0
-
-
+    
+Input should be lines of consisting of:
```
rouault, line 52: unfortunately that is not all that simple. It could be longitude+latitude, or latitude+longitude, or easting+northing, or northing+easting. The units might be anything.  And for time it is not in seconds but in decimal year (like 2022.9)

PR #3477, opened 2022-11-18
Implement normalizeForVisualization for DerivedProjected

jjimenezshaw:
- [x] Tests added

Inspecting the code I found that the functions `normalizeForVisualization` and `applyAxisOrderReversal` are not implemented for DerivedProjected.

### DIFF
--- a/src/iso19111/crs.cpp
+++ b/src/iso19111/crs.cpp
@@ -1031,8 +1031,8 @@ CRSNNPtr CRS::applyAxisOrderReversal(const char *nameSuffix) const {
             return props;
         };
 
-    const CompoundCRS *compoundCRS = dynamic_cast<const CompoundCRS *>(this);
-    if (compoundCRS) {
+    if (const CompoundCRS *compoundCRS =
+            dynamic_cast<const CompoundCRS *>(this)) {
         const auto &comps = compoundCRS->componentReferenceSystems();
         if (!comps.empty()) {
             std::vector<CRSNNPtr> newComps;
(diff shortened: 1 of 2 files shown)


jjimenezshaw:
Tests are pending. I do not know if I should implement them in `test_crs.cpp` or `test_operation.cpp`. Any preference?

I have tested it locally with `projinfo` using the _undocumented_ option `--normalize-axis-order`. Is it not documented in purpose?

jjimenezshaw:
I would prefer to include the variable definition inside the `if` scope, like
```C++
    if (const DerivedProjectedCRS *derivedProjCRS =
            dynamic_cast<const DerivedProjectedCRS *>(this)) {
        const auto &axisList = derivedProjCRS->coordinateSystem()->axisList();
        if (mustAxisOrderBeSwitchedForVisualizationInternal(axisList)) {
            return applyAxisOrderReversal(NORMALIZED_AXIS_ORDER_SUFFIX_STR);
        }
    }
```
But I have not done it for consistency with the previous. Should I change it in these two functions?
(I was almost using `projCRS` instead of `derivedProjCRS`, and it compiled)

rouault:
> in test_crs.cpp

that would be the better place. Strangely I don't find direct testing of CRS::normalizeForVisualization() on other CRS types. It must be currently done indirectly through operations that involve them, but direct unit testing would be better practice

> Is it not documented in purpose?

looking at the comment of 88426c18e43538edc6075d47b3b6829ada7e9a76 , yes apparently :-) Probably because I'm moderately comfortable with the concept of "normalized axis order" which isn't super well defined.

> I would prefer to include the variable definition inside the if scope

yes please change other occurences if you wish

PR #3482, opened 2022-11-19
Implement  alterGeodeticCRS and stripVerticalComponent for DerivedProjected

jjimenezshaw:
- [x] Tests added

After #3477 I went through `crs.cpp` looking for functions that may miss the implementation for class `DerivedProjected`. I found these two (there could be more). In previous PRs I did it when I was needing them. I hope this is the last one ;) This PR is done for completeness.

### DIFF
--- a/src/iso19111/crs.cpp
+++ b/src/iso19111/crs.cpp
@@ -275,20 +275,27 @@ createPropertyMap(const common::IdentifiedObject *obj) {
 
 //! @cond Doxygen_Suppress
 CRSNNPtr CRS::alterGeodeticCRS(const GeodeticCRSNNPtr &newGeodCRS) const {
-    auto geodCRS = dynamic_cast<const GeodeticCRS *>(this);
-    if (geodCRS) {
+    if (dynamic_cast<const GeodeticCRS *>(this)) {
         return newGeodCRS;
     }
 
-    auto projCRS = dynamic_cast<const ProjectedCRS *>(this);
-    if (projCRS) {
+    if (auto projCRS = dynamic_cast<const ProjectedCRS *>(this)) {
         return ProjectedCRS::create(createPropertyMap(this), newGeodCRS,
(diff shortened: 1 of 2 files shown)


src/iso19111/crs.cpp
```diff
@@ -275,20 +275,29 @@ createPropertyMap(const common::IdentifiedObject *obj) {
...
+    if (auto projCRS = dynamic_cast<const ProjectedCRS *>(this)) {
         return ProjectedCRS::create(createPropertyMap(this), newGeodCRS,
                                     projCRS->derivingConversion(),
                                     projCRS->coordinateSystem());
     }
 
-    auto compoundCRS = dynamic_cast<const CompoundCRS *>(this);
-    if (compoundCRS) {
+    if (auto derivedProjCRS = dynamic_cast<const DerivedProjectedCRS *>(this)) {
+        auto newProjCRS = std::dynamic_pointer_cast<ProjectedCRS>(
```
jjimenezshaw: Maybe there is a nicer way to do this cast, not needing later `NN_NO_CHECK`,... or maybe not.

rouault: you could probably use the slight variation that works directly on a non-null pointer, which avoids the .as_nullable() cast to a nullable intermediate shared pointer:
```
auto newProjCRS = NN_CHECK_ASSERT(util::nn_dynamic_pointer_cast<ProjectedCRS>(derivedProjCRS->baseCRS()->alterGeodeticCRS(newGeodCRS))
```
and remove the NN_NO_CHECK() cast in the create() call

I use here NN_CHECK_ASSERT as it is not completely trivial that derivedProjCRS->baseCRS()->alterGeodeticCRS() returns something that is a ProjectedCRS...

jjimenezshaw: Thanks.  Done.

PR #3486, opened 2022-11-24
Add bench_proj_trans benchmarking utility

rouault:


kbevers:
Is this only part of the test suite?

rouault:
> Is this only part of the test suite?

I hesitated if I should put it under src/apps or under test. I believe src/apps is only/mostly for utilities that are going to be installed, whereas this utility is more for development / non-production-related uses.

kbevers:
It makes sense. A few people might have a use to test the efficiency of various pipelines but if that's the case they can just build the code themselves.

PR #3488, opened 2022-11-25
add PJ_TYPE_DERIVED_PROJECTED_CRS

jjimenezshaw:
The function `PJ_TYPE proj_get_type(const PJ *obj)` does not cover the case of a `DerivedProjectedCRS`, so it returns the generic type `PJ_TYPE_OTHER_CRS`. This PR defined the type `PJ_TYPE_DERIVED_PROJECTED_CRS`, and returns it in that case.

 - [x] Closes https://lists.osgeo.org/pipermail/proj/2022-November/010829.html
 - [x] Tests added

### DIFF
--- a/src/proj.h
+++ b/src/proj.h
@@ -803,6 +803,8 @@ typedef enum
     PJ_TYPE_TEMPORAL_DATUM,
     PJ_TYPE_ENGINEERING_DATUM,
     PJ_TYPE_PARAMETRIC_DATUM,
+
+    PJ_TYPE_DERIVED_PROJECTED_CRS,
 } PJ_TYPE;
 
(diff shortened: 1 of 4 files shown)


src/proj.h
```diff
@@ -803,6 +803,8 @@ typedef enum
     PJ_TYPE_TEMPORAL_DATUM,
     PJ_TYPE_ENGINEERING_DATUM,
     PJ_TYPE_PARAMETRIC_DATUM,
+
+    PJ_TYPE_DERIVED_PROJECTED_CRS,
```
jjimenezshaw, line 807: I added it at the end of the enum for backward compatibility. If it is better  anywhere else, let me know.

PR #3489, opened 2022-11-28
Misc warnings2

seanm:


rouault:
Thanks for those cleanups!

rouault:
It could make sense to add those compiler flags to the defaults one used by PROJ. This could be added in top CMakeLists.txt, and logic from https://github.com/OSGeo/gdal/blob/master/gdal.cmake#L133 could be borrowed to add detection for the availability of those flags

PR #3496, opened 2022-12-05
Wcast-qual

seanm:
This fixes most, but not all, -Wcast-qual warnings.

There are some API changes, not sure how problematic that is...

### DIFF
--- a/src/strtod.cpp
+++ b/src/strtod.cpp
@@ -36,8 +36,6 @@
 #include "proj_config.h"
 #include "proj_internal.h"
 
-#define PJ_STRTOD_WORK_BUFFER_SIZE 64
-
 /************************************************************************/
 /*                              pj_atof()                               */
 /************************************************************************/
@@ -49,7 +47,7 @@
  * by nptr to double floating point representation. The behavior is the
  * same as
  *
- *   pj_strtod(nptr, (char **)NULL);
+ *   pj_strtod(nptr, nullptr);
  *
  * This function does the same as standard atof(3), but does not take
  * locale in account. That means, the decimal delimiter is always '.'
@@ -69,8 +67,7 @@ double pj_atof( const char* nptr )
 /*                     replace_point_by_locale_point()               */
 /************************************************************************/
 
-static char* replace_point_by_locale_point(const char* pszNumber, char point,
-                                              char* pszWorkBuffer)
+static char* replace_point_by_locale_point(const char* pszNumber, char point)
 {
 #if !defined(HAVE_LOCALECONV)
 
@@ -92,23 +89,19 @@ static char* replace_point_by_locale_point(const char* pszNumber, char point,
         const char* pszPoint = strchr(pszNumber, point);
         if (pszPoint)
         {
-            char* pszNew;
-            if( strlen(pszNumber) < PJ_STRTOD_WORK_BUFFER_SIZE )
-            {
-                strcpy(pszWorkBuffer, pszNumber);
-                pszNew = pszWorkBuffer;
-            }
-            else {
-                pszNew = pj_strdup(pszNumber);
-                if (!pszNew)
-                    return NULL;
-            }
+            char* pszNew = pj_strdup(pszNumber);
+            if (!pszNew)
+                return nullptr;
             pszNew[pszPoint - pszNumber] = byPoint;
             return pszNew;
         }
     }
+
+    return nullptr;
+
 #else
-    struct lconv *poLconv = localeconv();
+
+    const struct lconv *poLconv = localeconv();
     if ( poLconv
          && poLconv->decimal_point
          && poLconv->decimal_point[0] != '\0' )
@@ -121,17 +114,9 @@ static char* replace_point_by_locale_point(const char* pszNumber, char point,
             const char* pszPoint = strchr(pszNumber, point);
             if (pszPoint || pszLocalePoint)
             {
-                char* pszNew;
-                if( strlen(pszNumber) < PJ_STRTOD_WORK_BUFFER_SIZE )
-                {
-                    strcpy(pszWorkBuffer, pszNumber);
-                    pszNew = pszWorkBuffer;
-                }
-                else {
-                    pszNew = pj_strdup(pszNumber);
-                    if (!pszNew)
-                        return nullptr;
-                }
+                char* pszNew = pj_strdup(pszNumber);
+                if (!pszNew)
+                    return nullptr;
                 if( pszLocalePoint )
                     pszNew[pszLocalePoint - pszNumber] = ' ';
                 if( pszPoint )
@@ -140,8 +125,10 @@ static char* replace_point_by_locale_point(const char* pszNumber, char point,
             }
         }
     }
+
+    return nullptr;
+
 #endif
-    return (char*) pszNumber;
 }
 
 /************************************************************************/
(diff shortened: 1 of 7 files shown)


rouault:
I don't think we can drop the localeconv() check. GDAL does Android builds, that include doing PROJ builds , and if you look at the raw log of https://github.com/OSGeo/gdal/actions/runs/3614821944/jobs/6091514275 , you'll see 
```
2022-12-04T19:57:18.1906597Z -- Looking for localeconv
2022-12-04T19:57:18.2112350Z -- Looking for localeconv - not found
```
Cf https://stackoverflow.com/questions/44736135/ndk-clang-error-undefined-reference-to-localeconv also

src/proj.h
```diff
@@ -672,7 +672,7 @@ PJ_COORD PROJ_DLL proj_geod (const PJ *P, PJ_COORD a, PJ_COORD b);
 #define PROJ_ERR_OTHER_NETWORK_ERROR                     (PROJ_ERR_OTHER+3)             /* failure when accessing a network resource */
 
 /* Set or read error level */
-int  PROJ_DLL proj_context_errno (PJ_CONTEXT *ctx);
+int  PROJ_DLL proj_context_errno (const PJ_CONTEXT *ctx);
```
kbevers: As far as I can tell this is the only slightly concerning change. I'm unsure if we can keep the `const` promise but we probably can. I think we've talked about doing something similar before (for other objects?) without really getting to the bottom of it. 

What are the consequences of this going forwards? If any. I suspect downstream code will run just fine in most cases but I'm unsure if it holds true in *all* cases.

dg0yt: Doesn't it impact ABI which implies a soversion change?
If it does, must this change target a particular PROJ version change?

kbevers: > Doesn't it impact ABI which implies a soversion change?

That's what I'm thinking but I don't fully understand the implications. If this breaks things we'll have to go to version 10 and that's at least five years into the future.

rouault: This change is safe from an ABI point of view. The constness of arguments is not captured in the C ABI (contrary to C++ symbols). And from an API point of view, changing from ``foo*`` to ``const foo*`` is also safe as it doesn't trigger any compiler warnings.

src/ctx.cpp
```diff
@@ -43,7 +43,7 @@
 /*                             pj_get_ctx()                             */
 /************************************************************************/
 
-PJ_CONTEXT* pj_get_ctx( PJ *pj )
+PJ_CONTEXT* pj_get_ctx( const PJ *pj )
```
dg0yt: This is fine for the compiler, but is it also logically right from an API design POV? 
`pj_get_ctx` may hand out a pointer to `pj->ctx`. And this pointer allows to modify the object pointed to by `pj`. In this sense, the function needs a pointer to a non-const object, even if the compiler is happy with a pointer to a const object.
IOW if you want a mutable `PJ_CONTEXT *`, you must bring a mutable `PJ *`.

rouault: @dg0yt I'm not sure I share this analysis. ``pj_get_ctx( const PJ *pj )`` means that this particular function doesn't modify the passed PJ* object.  The fact that the returned PJ_CONTEXT* might be later modified, and those modifications could affect how the PJ* behaves is another concern.
If you look at the C standard library, there is a ``char *strchr(const char *s, int c);``that returns a char* pointer on the input const char*. But that's more because in C you can't have both ``char* strchr(char* s, int c);`` and ``const char* strchr(const char* s, int c);``
As pj_get_ctx() is in the internal C++ API, we could probably have 2 versions ``PJ_CONTEXT* pj_get_ctx(PJ* pj);`` and  ``const PJ_CONTEXT* pj_get_ctx(const PJ* pj);``

seanm: >pj_get_ctx( const PJ *pj ) means that this particular function doesn't modify the passed PJ* object. The fact that the returned PJ_CONTEXT* might be later modified, and those modifications could affect how the PJ* behaves is another concern.

This is my understanding/expectation/interpretation too.

>But that's more because in C you can't have both char* strchr(char* s, int c); and const char* strchr(const char* s, int c);

In fact, C23 has fixed exactly that problem with `strchr`: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3020.pdf

> As pj_get_ctx() is in the internal C++ API, we could probably have 2 versions PJ_CONTEXT* pj_get_ctx(PJ* pj); and const PJ_CONTEXT* pj_get_ctx(const PJ* pj);

I could try that, see if it compiles...

dg0yt: `strchr` seems to really illustrate the point I'm making. Other than C, C++ offers two different overloads, for obvious reasons: https://en.cppreference.com/w/cpp/string/byte/strchr

Back in context. It is not about the inside (implementation) of the function, but about the outside (usage):
~~~c++
const PJ *my_const_pj = get_a_pj();
assert(has_some_property(my_const_pj));
do_something_fancy(pj_get_ctx(my_const_pj)); // Feeling safe, my_const_pj is const with no mutable properties.
assert(has_some_property(my_const_pj)); // Oops!
~~~

I don't have a strong opinion about it. I just would like to ensure that new compiler warnings don't translate too quickly into API changes.

seanm: Whereas today, if you try to write a `do_something_fancy` that takes a const, you end up having to cast away the const when calling it, like: https://github.com/OSGeo/PROJ/pull/3496/commits/4dd98f60b73c2ef7c73de2762e413a113b013596

rouault: Weirdness can already happen without involving pj_get_ctx():
```c++
PROJ_CONTEXT* ctxt = proj_context_create();
const PJ* pj = proj_create(ctxt, "bla");
assert(has_some_property(my_const_pj));
modify_some_state_of_context_that_affects_PJ_objects_linked_to_it(ctxt);
assert(has_some_property(my_const_pj)); // Oops!
```

Or:
```c++
std::vector<int> v(1);
const std::vector<int>* p_v = &v;
assert(p_v->size() == 1);
v.resize(2);
assert(p_v->size() == 1); // Oops!
```

Basically const promise in C/C++ is a very weak one... Far from immutability

src/proj.h
kbevers: Thanks for clearing that up

src/ctx.cpp
dg0yt: @rouault These examples miss the point because they introduce direct access to the underlying object (`ctxt`, `v`) while I was explicitly illustrating a scope where there is only a pointer to the const object. FTR, I also didn't discuss concurrenct access.

@seanm The pointer to the actual usage patterns is useful. It seems like a common pattern is logging and setting errors, e.g.
~~~
proj_context_errno_set (pj_get_ctx ((PJ *) P), err);
~~~~~
Logging and error state change may be acceptable changes of observable state on an otherwise const object.

The calling scope has a `const PJ *`, and there is a compiler warning for `((PJ *) P)`. Let us enumerate all alternatives: 
- Make `pj_get_ctx` accept `const PJ *`.
  - Change the signature, and just benefit from the pointer to non-const (Your PR).
  - Add another overload.
- Permit this particular ignorance of `const` by an explicit C++ `const_cast<PJ *>(P) `.
  (In my example, it would make *visible* that something non-const happens between the two asserts.)
- Make `P` non-const.

Again, my point is not about const correctness on the compiler level. It ends at pointers. The compiler doesn't know our reasons for making member variables pointers instead of acutal types. My question is API correctness beyond such  implementation decisions.
And I don't know the answer. I do not know enough about the coupling between `PJ_CONTEXT` and `PJ`.

(I have seen big PRs guided by compiler warnings in other projects, and some of the initial changes were wrong.)

seanm:
> I don't think we can drop the localeconv() check.

It's now back.  And I rebased on newest master.

src/ctx.cpp
seanm: >>As pj_get_ctx() is in the internal C++ API, we could probably have 2 versions PJ_CONTEXT* pj_get_ctx(PJ* pj); and const PJ_CONTEXT* pj_get_ctx(const PJ* pj);

>I could try that, see if it compiles...

It doesn't readily compile.

```C
int proj_errno_set (const PJ *P, int err) {
/******************************************************************************
    Set context-errno, bubble it up to the thread local errno, return err
******************************************************************************/
    /* Use proj_errno_reset to explicitly clear the error status */
    if (0==err)
        return 0;

    /* For P==0 err goes to the default context */
    proj_context_errno_set (pj_get_ctx (P), err); // <---------------- ••• error •••
    errno = err;

    return err;
}
```

The const-returning variant of `pj_get_ctx` gets passed to `proj_context_errno_set`, and it can't be changed to take const because it mutates the given `PJ_CONTEXT`.

seanm:
So, any recent thoughts on this PR?  It has 8 commits, I can always remove objectionable ones, like https://github.com/OSGeo/PROJ/pull/3496/commits/3eca2866f173cea877f73eea5200a8b74f9c640e perhaps.

rouault:
> I can always remove objectionable ones, like [3eca286](https://github.com/OSGeo/PROJ/commit/3eca2866f173cea877f73eea5200a8b74f9c640e) perhaps.

yes if you remove that one, I believe the rest is non controversial

seanm:
I removed the controversial commit, and another that depended on it.

src/strtod.cpp
```diff
@@ -167,25 +154,32 @@ double pj_strtod( const char *nptr, char **endptr )
...
-        *endptr = (char *)nptr + (*endptr - pszNumber);
+		free( pszNumber );
 
-    if (pszNumber != (char*) nptr && pszNumber != szWorkBuffer )
-        free( pszNumber );
+		errno = nError;
 
-    errno = nError;
-    return dfValue;
+        return dfValue;
```
rouault: indentation is odd. Please use 4-space per indentation level

seanm: Fixed

PR #3499, opened 2022-12-07
Bugfix in CRS::alterCSLinearUnit for DerivedProjectedCRS

hernando:
The implementation was calling `alterUnit` on the CS of the base projected CRS, but this CS could be different than that of the derived projected CRS being modified (e.g. the axes could be swapped).

The fix consist on altering the unit of the derived projected CRS, but only if the CS can be cast to Cartesian. It's doesn't look like other CS types need to be considered.

 - [x] Tests added

### DIFF
--- a/src/iso19111/crs.cpp
+++ b/src/iso19111/crs.cpp
@@ -378,10 +378,14 @@ CRSNNPtr CRS::alterCSLinearUnit(const common::UnitOfMeasure &unit) const {
     {
         auto derivedProjCRS = dynamic_cast<const DerivedProjectedCRS *>(this);
         if (derivedProjCRS) {
+            auto cs = derivedProjCRS->coordinateSystem();
+            auto cartCS = util::nn_dynamic_pointer_cast<cs::CartesianCS>(cs);
+            if (cartCS) {
+                cs = cartCS->alterUnit(unit);
+            }
             return DerivedProjectedCRS::create(
                 createPropertyMap(this), derivedProjCRS->baseCRS(),
-                derivedProjCRS->derivingConversion(),
-                derivedProjCRS->baseCRS()->coordinateSystem()->alterUnit(unit));
+                derivedProjCRS->derivingConversion(), cs);
         }
     }
 
--- a/test/unit/test_crs.cpp
+++ b/test/unit/test_crs.cpp
@@ -6543,6 +6543,24 @@ TEST(crs, crs_alterCSLinearUnit) {
     }
 
     {
+        auto crs = createDerivedProjectedCRSNorthingEasting();
+        auto alteredCRS = crs->alterCSLinearUnit(UnitOfMeasure("my unit", 2));
+
+        auto leftHandedDerivedCRS =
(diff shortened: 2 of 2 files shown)


src/iso19111/crs.cpp
```diff
@@ -378,10 +378,16 @@ CRSNNPtr CRS::alterCSLinearUnit(const common::UnitOfMeasure &unit) const {
     {
         auto derivedProjCRS = dynamic_cast<const DerivedProjectedCRS *>(this);
         if (derivedProjCRS) {
+            auto cs = derivedProjCRS->coordinateSystem();
+            // DerivedProjectedCRS's factory accepts a generic CoordinateSystem,
+            // but does it make sense to consider any type other than Cartesian?
+            auto cartCS = util::nn_dynamic_pointer_cast<cs::CartesianCS>(cs);
```
rouault, line 382: [http://docs.opengeospatial.org/is/18-010r7/18-010r7.html lis](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html#119) shows an example where the coordinate system of the DerivedProjectedCRS is a OrdinalCS. So this test looks good

src/iso19111/crs.cpp
```diff
@@ -390,7 +396,7 @@ CRSNNPtr CRS::alterCSLinearUnit(const common::UnitOfMeasure &unit) const {
         if (compoundCRS) {
             std::vector<CRSNNPtr> components;
             for (const auto &subCrs :
-                 compoundCRS->componentReferenceSystems()) {
+                compoundCRS->componentReferenceSystems()) {
```
rouault: change to revert

rouault (commented):
CC @jjimenezshaw

jjimenezshaw:
> CC @jjimenezshaw

Thanks @hernando . Good catch about the axis order. Good that `createDerivedProjectedCRSNorthingEasting` test helper is used again.
And thanks @rouault for the OrdinalCS reference.

src/iso19111/crs.cpp
hernando: OK, but then there's no other possibility that requires the unit to be altered, right?

rouault: According to https://docs.opengeospatial.org/as/18-005r4/18-005r4.html "Table 43: Defining elements of CoordinateSystems::DerivedProjectedCS class", the valid CS for a DerivedProjectedCRS are CartesianCS, CylindricalCS, OrdinalCS, PolarCS, SphericalCS.
CylindricalCS is a "three-dimensional coordinate system in Euclidean space in which position is specified by two linear coordinates and one angular coordinate", so in theory it could be affected, but there's no corresponding implementation of it in PROJ currently. The other CS use angular or non-linear units.

hernando: Thanks for the reference. I was just looking at the classes in coordinatesystem.hpp and I didn't find any other with an `alterUnit` member function that was applicable, but I wasn't sure.

PR #3509, opened 2022-12-16
Mention the proj_create function

oleg-alexandrov:
Following the PROJ doc is like jumping through many hoops. Or if you wish, many reprojections across documents are needed for something coherent to emerge. This adds a blurb mentioning the proj_create() function, which is really one of the two workhorses of this library, together with proj_create_crs_to_crs. 

If this is accepted, I plan to also add an example of using this function, with every step spelled out.

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/docs/source/development/quickstart.rst
+++ b/docs/source/development/quickstart.rst
@@ -46,11 +46,11 @@ Next we create the :c:type:`PJ` transformation object ``P`` with the function
   :dedent: 4
 
 Here we have set up a transformation from geographic coordinates to UTM zone
-32N. In general, this is a transform between two different projections
+32N. In general, this is a transformation between two different coordinate reference systems
 (one of which is here in geographic coordinates).  The related function 
-:c:func:`proj_create` can be used with a single 
-projection, and can transform from that projection to geographic
-coordinates for the same geodetic datum, and vice versa.
+:c:func:`proj_create` can be used set up transformations that are not available through 
+the PROJ database, for instance for converting geodetic coordinates to a custom definition
+of a map projection.
 
 :c:func:`proj_create_crs_to_crs` takes as its arguments:
 
--- a/docs/source/development/quickstart.rst
+++ b/docs/source/development/quickstart.rst
@@ -48,7 +48,7 @@ Next we create the :c:type:`PJ` transformation object ``P`` with the function
 Here we have set up a transformation from geographic coordinates to UTM zone
 32N. In general, this is a transformation between two different coordinate reference systems
 (one of which is here in geographic coordinates).  The related function 
-:c:func:`proj_create` can be used set up transformations that are not available through 
+:c:func:`proj_create` can be used to set up transformations that are not available through 
 the PROJ database, for instance for converting geodetic coordinates to a custom definition
 of a map projection.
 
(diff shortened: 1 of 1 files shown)


docs/source/development/quickstart.rst
```diff
@@ -46,7 +46,11 @@ Next we create the :c:type:`PJ` transformation object ``P`` with the function
   :dedent: 4
 
 Here we have set up a transformation from geographic coordinates to UTM zone
-32N.
+32N. In general, this is a transform between two different projections
+(one of which is here in geographic coordinates).  The related function 
+:c:func:`proj_create` can be used with a single 
+projection, and can transform from that projection to geographic
+coordinates for the same geodetic datum, and vice versa.
```
kbevers: This is a not a correct representation of the differences between `proj_create` and `proj_create_crs_to_crs`.  What comes out of the `proj_create*`-functions are what we call "coordinate operations". Coordinate operations (quite often referred to as "transformations", although not entirely correct) are operations that change coordinates from one coordinate reference system (CRS) to another. A coordinate operation can be a projection, i.e. a mathematical mapping of longitudes and latitudes on the sphere to a flat plane (a map). Or it can be a transformation which deals with the physical reality of geodesy, i.e. continents moving about and stuff like that. It's important to not mix up the terminology. A coordinate operation can be a combination of both a transformation and a projection.

Among GIS-users there's a bit of a bad habit of calling CRS's "projections". You've done it here. It's a common mistake rooted in decades of inconsistent terminology and ill-defined standards. All of the above to say that

> :c:func:`proj_create` can be used with a single projection

is not a particularly good way to describe what `proj_create` does. It *can* output a projection, it doesn't take a projection as input. `proj_create` takes a description of a coordinate operation in either the PROJ notation (+operaters) or WKT. In most cases `proj_create` will be used in cases where a certain coordinate operation can't be pulled from the PROJ database and a custom made one is necessary. That can be as simple as a projection with custom parameters, e.g `+proj=merc +lon_0=12.42342 +k_0=1.5`, or a more complicated transformation pipeline with many steps. In fact `proj_create_from_crs_to_crs` constructs such a transformation pipeline based on the source and target CRS's that's fed to it.

Here's a go at rephrasing your suggestion in the light of the above:

```suggestion
32N. In general, this is a transformation between two different coordinate reference systems
(one of which is here in geographic coordinates).  The related function 
:c:func:`proj_create` can be used set up transformations that are not available through 
the PROJ database, for instance for converting geodetic coordinates to a custom definition
of a map projection.
```

oleg-alexandrov:
I agree with your suggestion. Thank you for the review.

docs/source/development/quickstart.rst
```diff
@@ -46,7 +46,11 @@ Next we create the :c:type:`PJ` transformation object ``P`` with the function
   :dedent: 4
 
 Here we have set up a transformation from geographic coordinates to UTM zone
-32N.
+32N. In general, this is a transformation between two different coordinate reference systems
+(one of which is here in geographic coordinates).  The related function 
+:c:func:`proj_create` can be used set up transformations that are not available through
```
kbevers:
```suggestion
:c:func:`proj_create` can be used to set up transformations that are not available through 
```

PR #3515, opened 2022-12-16
Add example for how to use proj_create().

oleg-alexandrov:
This is a proposed example for the proj_create() function. This function has its own peculiarities, which can't be inferred from the example illustrating proj_create_crs_to_crs(). It is shown how to populate the longitude and latitude, convert to radians, how to transform to projected units, then transform back, and how to convert back to degrees. 

The code was compiled and tested. It is suggested to verify it compiles and prints what is expected.

### DIFF
--- a/docs/source/development/quickstart.rst
+++ b/docs/source/development/quickstart.rst
@@ -4,10 +4,12 @@
 Quick start
 ================================================================================
 
-This is a short introduction to the PROJ API. In the following section we
-create a simple program that transforms a geodetic coordinate to UTM and back
-again. The program is explained a few lines at a time. The complete program can
-be seen at the end of the section.
+This is a short introduction to the PROJ API. In the following section
+we create two simple programs that illustrate how to transform points
+between two different coordinate systems, and how to convert between
+projected and geodetic (geographic) coordinates for a single
+coordinate system. Explanations for individual code sniplets and the
+full programs are provided.
 
 See the following sections for more in-depth descriptions of different parts of
 the PROJ API or consult the :doc:`API reference <reference/index>` for specifics.
@@ -201,4 +203,12 @@ A complete compilable version of the example code can be seen below:
   :language: c
   :linenos:
   :lines: 39-
+  
+The following example illustrates how to convert between a CRS and
+geodetic coordinates for that CRS.
+
+.. literalinclude:: ../../../examples/crs_to_geodetic.c
+  :language: c
+  :linenos:
+  :lines: 9-
 
--- a/examples/crs_to_geodetic.c
+++ b/examples/crs_to_geodetic.c
@@ -1,0 +1,61 @@
+/*******************************************************************************
+    Example code that illustrates how to convert between a CRS and
+    geodetic coordnates for that CRS.
+
+    Note: This file is in-lined in the documentation. Any changes must be
+    reflected in docs/source/development/quickstart.rst
+
+*******************************************************************************/
+#include <proj.h>
+#include <stdio.h>
+#include <math.h>
+
+int main (void) {
+  
+  /* Create the context. */
+  /* You may set C=PJ_DEFAULT_CTX if you are sure you will     */
+  /* use PJ objects from only one thread                       */
+  PJ_CONTEXT *C = proj_context_create();
+
+  /* Create a projection. */
+  PJ *P = proj_create(C, "+proj=utm +zone=32 +datum=WGS84 +type=crs");
+
+  if (0 == P) {
+    fprintf(stderr, "Failed to create transformation object.\n");
+    return 1;
+  }
+
+  /* Get the geodetic CRS for that projection. */
+  PJ *G = proj_crs_get_geodetic_crs(C, P);
+
+  /* Create the transform from geodetic to projected coordinates.*/
+  PJ_AREA *A = NULL;
+  const char *const *options = NULL;
+  PJ *G2P = proj_create_crs_to_crs_from_pj(C, G, P, A, options);
+  
+  /* Longitude and latitude of Copenhagen, in degrees. */
+  double lon = 12.0, lat = 55.0;
+
+  /* Prepare the input */
+  PJ_COORD c_in;
+  c_in.lpzt.z = 0.0;
+  c_in.lpzt.t = HUGE_VAL; // important only for time-dependent projections
+  c_in.lp.lam = lon;
+  c_in.lp.phi = lat;
+  printf ("Input longitude: %g, latitude: %g (degrees)\n", c_in.lp.lam, c_in.lp.phi);
+
+  /* Compute easting and northing */
+  PJ_COORD c_out = proj_trans(G2P, PJ_FWD, c_in);
+  printf ("Output easting: %g, northing: %g (meters)\n", c_out.enu.e, c_out.enu.n);
+
+  /* Apply the inverse transform */
+  PJ_COORD c_inv = proj_trans(G2P, PJ_INV, c_out);
+  printf ("Inverse applied. Longitude: %g, latitude: %g (degrees)\n", c_inv.lp.lam, c_inv.lp.phi);
+
+  /* Clean up */
+  proj_destroy(P);
+  proj_destroy(G);
+  proj_destroy(G2P);
+  proj_context_destroy(C); /* may be omitted in the single threaded case */
+  return 0;
+}


kbevers:
Sorry for not getting back on this earlier. Christmas vacation got in the way.

I don't think this examples brings anything new to the table. Quite the contrary, it mostly looks like code written for PROJ.4 that's being shoe-horned to fit the API of PROJ 9. I get that it is probably exactly what you are doing at the moment and that it is frustrating to not have a clearly defined path to follow when moving from the old style API. There's probably a lot of that going on out there since many people have been through the transition but it's not behaviour we want to promote. 

What's done in the proposed example program could just as easily be achieved by altering the input to `proj_create_crs_to_crs()`. In fact, the two example programs does the exact same conversion from geodetic coordinates to UTM zone 32. The difference is that in the `proj_create_crs_to_crs` example we tell PROJ *what* we want (a transformation from source to target) and in the `proj_create` example PROJ is told *how* to do the wanted transformation.

If we are to have a parallel example showcasing `proj_create` I would to see it do something that is not immediately possible by using `proj_create_crs_to_crs` and annotated like the current example so the reader is helped along the way. Additionally, the example should be put in a separate file similar to the current example, so that it can be build and run using our test suite. The latter is meant to ensure that the examples actually work with the current version of the software.

oleg-alexandrov:
Given an arbitrary projection for a certain datum, the current documentation does not offer a way of doing transforms from geodetic coordinates for that projection to projected coordinates for it. (You'll tell me "projection" is the wrong word, but it it works well enough for my point here.)

Yes, proj_create_crs_to_crs can be made to do that. One would read in the CRS, then form a second CRS by replacing UTM or what not with geodetic coordinates in the existing CRS. Then create a projection among the two CRS. How to remove the UTM from a projection while keeping the datum is not documented.

Hence my example, which avoids these undocumented complications, and uses another function from the codebase which does exactly what is needed.

oleg-alexandrov:
To perhaps clarify things more, if the user has the strings ""+proj=longlat +datum=WGS84" and "+proj=utm +zone=32 +datum=WGS84", you are right that one can use proj_create_crs_to_crs() to convert from lonlat to utm.

If the user has to do a lonlat to utm conversion for anything passed in, such as the string EPSG:32632, one can't use proj_create_crs_to_crs(), as this function requires two input CRS. Then one has to use proj_create(), and that's what the proposed example is meant to show.

rouault:
> Then one has to use proj_create(), and that's what the proposed example is meant to show.

perhaps using the pipeline ``+proj=pipeline
        +step +inv +proj=utm +zone=32
        +step +proj=cart +ellps=intl
        +step +proj=helmert +x=-81.1 +y=-89.4 +z=-115.8
        +rx=0.485 +ry=0.024 +rz=0.413 +s=-0.54 +convention=position_vector
        +step +inv +proj=cart +ellps=GRS80
        +step +proj=utm +zone=33`` from https://proj.org/tutorials/EUREF2019/answers.html would be more convincing as an example of where proj_create() is needed

oleg-alexandrov:
Any thoughts if this example may be helpful in the doc? If it is accepted, I can have it in a separate compilable file as expected.

kbevers:
> If the user has to do a lonlat to utm conversion for anything passed in, such as the string EPSG:32632, one can't use proj_create_crs_to_crs(), as this function requires two input CRS. Then one has to use proj_create(), and that's what the proposed example is meant to show.

Sure you can, just put in EPSG:4326 as the source CRS and EPSG:32632 as the target CRS. The WGS84 datum ensemble (EPSG:4326) is about as vaguely defined as can be and is generally a good starting point if all you know about you input data is that they are geodetic coordinates.

> Any thoughts if this example may be helpful in the doc?

It certainly showcases a use-case that is more in the line of what `proj_create` is meant for.

oleg-alexandrov:
> Sure you can, just put in EPSG:4326 as the source CRS and EPSG:32632 as the target CRS. The WGS84 datum ensemble (EPSG:4326) is about as vaguely defined as can be and is generally a good starting point if all you know about you input data is that they are geodetic coordinates.

And if a user is presented with an arbitrary projection string, how can one look up the second corresponding string? 

> It certainly showcases a use-case that is more in the line of what `proj_create` is meant for.

Does this mean my pull request will be accepted if submitted according to the style guide? My motivation here is that I spent many hours figuring out things that the existing document does not explain, and the hope is that folks will find this example helpful.

kbevers:
> And if a user is presented with an arbitrary projection string, how can one look up the second corresponding string?

If it is indeed a projection (i.e. `+proj +mercator +lat_ts=23.12`) you can use `+proj=latlon` as the source CRS. Well, you need to append `+type=crs` to both strings for it to work with  `proj_create_crs_to_crs`. Example using `projinfo` which is basically a wrapper for `proj_create_crs_to_crs`:

```
$ projinfo -s "+proj=latlon +type=crs" -t "+proj=merc +lat_ts=23.12 +type=crs"                
Candidate operations found: 1
Operation No. 1:

unknown id, unknown, 0 m, unknown domain of validity

PROJ string:
+proj=pipeline
  +step +proj=merc +lat_ts=23.12 +lon_0=0 +x_0=0 +y_0=0 +ellps=WGS84

WKT2:2019 string:
CONVERSION["unknown",
    METHOD["Mercator (variant B)",
        ID["EPSG",9805]],
    PARAMETER["Latitude of 1st standard parallel",23.12,
        ID["EPSG",8823]],
    PARAMETER["Longitude of natural origin",0,
        ID["EPSG",8802]],
    PARAMETER["False easting",0,
        LENGTHUNIT["metre",1],
        ID["EPSG",8806]],
    PARAMETER["False northing",0,
        LENGTHUNIT["metre",1],
        ID["EPSG",8807]]]
```

This generally emulates PROJ.4 behaviour quite well although you will run into some differences once in a while due to various issues in the old code that has been fixed.

> Does this mean my pull request will be accepted if submitted according to the style guide?

Yes

oleg-alexandrov:
Done. I synched-up with master and added this on top. I built the code locally. That doesn't build the code examples directory, but I built that separately. The doc is not built either, so I can't verify any generated html. 

It looks that the doc at https://proj.org/index.html reflects the 9.1 branch, and not the master. Is the master branch doc created anywhere? I almost made the mistake of including the wrong .c file in the .rst, and it would be nice if one can check the final result. 

I don't know why (after 2 hours of building) one gets the error: 

/__w/PROJ/PROJ/docs/source/development/quickstart.rst.rst:210:non-whitespace stripped by dedent.

The file I put in seems clean, with everything being ascii and no extra whitespace. 

Should this be pushed to branch 9.1 as well?

oleg-alexandrov:
Now it looks that there's a failure with GitHub's Sphinx.

Running Sphinx v5.3.0

  File "/home/docs/checkouts/readthedocs.org/user_builds/osgeo-proj/envs/3515/lib/python3.7/site-packages/sphinx/config.py", line 350, in eval_config_file
    exec(code, namespace)
  File "/home/docs/checkouts/readthedocs.org/user_builds/osgeo-proj/checkouts/3515/docs/source/conf.py", line 25, in <module>
    import bibstyle
  File "/home/docs/checkouts/readthedocs.org/user_builds/osgeo-proj/checkouts/3515/docs/source/bibstyle.py", line 30, in <module>
    from pybtex.style.formatting.unsrt import Style as UnsrtStyle
ModuleNotFoundError: No module named 'pybtex'

kbevers:
> I built the code locally. That doesn't build the code examples directory, but I built that separately. 

I was certain that the existing example was built and run in the testsuite but that doesn't appear to be the case, so I think what you have now is good enough. 

> The doc is not built either, so I can't verify any generated html.

It's a bit of a faf to get going but you can build the docs locally by running `make html` in `docs/`. This won't be necesary after https://github.com/OSGeo/PROJ/pull/3538 is merged (probably soon).

> Is the master branch doc created anywhere? 

The docs are build in GitHub Actions that run on PRs but it's not pushed anywhere. It might be possible to dig out the results from the artifacts produced by the run but I'm not sure how that works. 

> Should this be pushed to branch 9.1 as well?

No, once it's merged we'll add the `backport 9.1` label and let the backport bot do the work.

rouault:
> It might be possible to dig out the results from the artifacts produced by the run but I'm not sure how that works.

Just look at the bottom of https://github.com/OSGeo/PROJ/actions/runs/3842624376 ("Summary" link accessed from https://github.com/OSGeo/PROJ/actions/runs/3842624376/jobs/6544094895, which is accessed from the above "Details" link in front of the Docs / Plots job)

oleg-alexandrov:
I checked the generated pdf, it looks good. I guess this can be pulled in, and then back-ported to 9.1.  Thank you for your help throughout.

examples/proj_create_example.c
```diff
@@ -0,0 +1,66 @@
...
+  PJ_CONTEXT *C;
+  PJ *P;
+
+  /* Create the context. */
+  /* You may set C=PJ_DEFAULT_CTX if you are sure you will     */
+  /* use PJ objects from only one thread                       */
+  C = proj_context_create();
+
+  /* Create the projection. */
+  P = proj_create(C, "+proj=utm +zone=32 +datum=WGS84");
```
kbevers: This doesn't reflect the discussion in the PR. Please adjust according to @rouault's suggestion https://github.com/OSGeo/PROJ/pull/3515#issuecomment-1369250539

Once the pipeline is changed adjustments are needed in the following code, as we are now transforming between two projected CRS's. Comments inlined where I believe changes are necessary.

examples/proj_create_example.c
```diff
@@ -0,0 +1,66 @@
...
+
+  if (0 == P) {
+    fprintf(stderr, "Failed to create transformation object.\n");
+    return 1;
+  }
+
+  /* Longitude and latitude of Copenhagen, in degrees. */
+  double lon = 12.0, lat = 55.0;
```
kbevers: Something along the lines of 
```suggestion
  double x = 6080642.11, y = 1886936.97;
```

examples/proj_create_example.c
```diff
@@ -0,0 +1,66 @@
...
+
+  /* Prepare the input */
+  PJ_COORD c_in;
+  c_in.lpzt.z = 0.0;
+  c_in.lpzt.t = HUGE_VAL; // important only for time-dependent projections
+  c_in.lp.lam = lon;
+  c_in.lp.phi = lat;
```
kbevers:
```suggestion
  c_in.xyzt.z = 0.0;
  c_in.xyzt.t = HUGE_VAL; // important only for time-dependent projections
  c_in.xyzt.x = x;
  c_in.xyzt.y = y;
```

examples/proj_create_example.c
```diff
@@ -0,0 +1,66 @@
...
+  printf ("Input longitude: %g, latitude: %g (degrees)\n", c_in.lp.lam, c_in.lp.phi);
```
kbevers: easting/northing instead of lat/lon

examples/proj_create_example.c
```diff
@@ -0,0 +1,66 @@
...
+
+  /* Convert from degrees to radians */
+  c_in.lp.lam = proj_torad(c_in.lp.lam);
+  c_in.lp.phi = proj_torad(c_in.lp.phi);
+
```
kbevers: Not needed anymore

examples/proj_create_example.c
```diff
@@ -0,0 +1,66 @@
...
+  PJ_COORD c_out = proj_trans(P, PJ_FWD, c_in);
+  printf ("Output easting: %g, northing: %g (meters)\n", c_out.enu.e, c_out.enu.n);
+
+  /* Apply the inverse transform */
+  PJ_COORD c_inv = proj_trans(P, PJ_INV, c_out);
+
+  /* Convert from radians to degrees. */
+  c_inv.lp.lam = proj_todeg(c_inv.lp.lam);
+  c_inv.lp.phi = proj_todeg(c_inv.lp.phi);
+
```
kbevers: Not needed

examples/proj_create_example.c
```diff
@@ -0,0 +1,66 @@
...
+  printf ("Inverse applied. Longitude: %g, latitude: %g (degrees)\n", c_inv.lp.lam, c_inv.lp.phi);
```
kbevers: easting/northing instead of lat/lon

docs/source/development/quickstart.rst
```diff
@@ -5,9 +5,10 @@ Quick start
 ================================================================================
 
 This is a short introduction to the PROJ API. In the following section we
-create a simple program that transforms a geodetic coordinate to UTM and back
-again. The program is explained a few lines at a time. The complete program can
-be seen at the end of the section.
+create two simple programs that illustrate how to transform points between
+two different coordinate systems, and how to convert between projected and
+geodetic coordinates for a single coordinate system. Explanations for individual
```
kbevers: The second example, using a pipeline with `proj_create()` transforms between ED50/UTM32 and ETRS89/UTM33 and not between geodetic and projected coordinates. See exercise from the tutorial for more https://proj.org/tutorials/EUREF2019/exercises/pipelines.html#ed50-utm32-etrs89-utm33

The text here should reflect that.

kbevers (changes requested):
The `proj_create()` example hasn't been updated to use the pipeline suggested by @rouault. I've provided some pointers in in-lined comments on how to update the example so it demonstrates the kind of scenario `proj_create()` is meant to handle. The PR is ready to be merged once that is changed.

examples/proj_create_example.c
oleg-alexandrov: Never mind, I have another comment below to reflect this whole thing.

oleg-alexandrov:
Never mind. I don't agree. I explain below.

examples/proj_create_example.c
oleg-alexandrov: I don't agree. This defeats the purpose. I want an example from lon-lat to northing-easting. We already have an example which deals with transforms among two CRS.

oleg-alexandrov:
Dear folks, and I would like to hear the ideas of other people than just @kbevers, though his patience here is also appreciated.

I am looking to add to this documentation a very simple example of converting long-lat to projected coordinates. This is necessary logic, not illustrated by examples so far. It shows how you start with long-lat, go to radians, apply the projection, and then go back.

It looks to me that @kbevers is turning this into an illustration of how clever some people can be, creating an esoteric application of the proj_create() function between two easting-northing systems, while removing the very thing I care for, which is how to go from long-lat to projected coordinates. 

Any thoughts of how to give the user a basic example of going from long-lat to easting-northing?

rouault:
Merging. Thanks!

mwtoews (approved):
Looks good, thanks!

Note that the RTD error can be ignored (this commit is before the configuration was created)

PR #3522, opened 2022-12-21
Implement the ellipsoidal gnomonic projection

cffk:
This implements the ellipsoidal gnomonic projection given in Karney (2013).  It is the limit of a doubly azimuthal projection, a projection where the azimuths from 2 points are preserved, as the two points merge into the center point.  Geodesics project to approximately straight lines (these are exactly straight if the geodesic includes the center point).

This change will potentially break some user code because the previous implementation of the projection silently converted an ellipsoid into a sphere with the same equatorial radius.  Now, it'll be treated as an ellipsoid.  The previous (inconsistent!) behavior can be recovered by setting the flattening to zero with `+f=0`.

The boundary test in `test_c_api.cpp` needed to be updated because it relied on this inconsistent treatment of the ellipsoid parameters.  The change is explained in the file.

All the gnomonic tests in `builtins.gie` were supplemented by ellipsoidal versions (with flattening = 1/200).

The documentation was updated to reflect this addition and the gnomonic projection is listed as azimuthal (instead of, incorrectly, as pseudocylindrical) and the defined area has been changed from "global" to "within a quarter circumference of the center point".

Because the azimuthal equidistant projection uses the same geodesic machinery as the new gnomonic projection, I'm simplified some of the code in `aeqd.cpp` to parallel `gnom.cpp`; in particular, aeqd no longer does the scaling/unscaling by `P->a` and the shifting/unshifting by `P->lam0`.

 - [x] Tests added


rouault:
There might be a subtelty with this change related how we map ESRI's Gnomonic method (https://pro.arcgis.com/en/pro-app/latest/help/mapping/properties/gnomonic.htm) to WKT / PROJ string. I guess we should add a +f=0 parameter now when mapping to a PROJ string.

Could @melitakennedy confirm if, for example, CRS ESRI:102036 "South_Pole_Gnomonic" which uses the WGS 84 ellipsoid is intended to use the spherical or ellipsoidal formulation of Gnomonic ?  Seeing in https://raw.githubusercontent.com/Esri/projection-engine-db-doc/master/csv/pe_list_projcs.csv that is uses PROJECTION["Gnomonic"] and not PROJECTION["Gnomonic_Ellipsoidal"], I assume this should be the spherical formulation using the semi-major radius as the radius.

cffk:
Arguably "ellipsoidal gnomonic" is a contradiction on an ellipsoidal
there's no projection which map geodesics to straight lines.  If, on the
other hand, the gnomonic projection is defined as the limit of a double
azimuthal projection, then generalizing it to an ellipsoid is
straightforward (and, because geodesics are nearly straight, still
useful).

My guess is that this generalization has not been considered by ESRI, so
that what it calls "gnomonic" is the spherical projection and the only
question is how pick the radius of the sphere.

Perhaps it's worth soliciting feedback on the PROJ mailing list?

cffk:
Ah, I see that ESRI says:
> The gnomonic ellipsoidal variant is available in ArcGIS Pro 1.2 and later and in ArcGIS Desktop 10.4 and later. This is the only variant of the three that correctly supports projection of ellipsoids.

But, I don't know what "correctly supports projection of ellipsoids" means.

cffk:
Full disclosure:

I know of three possible generalizations of the gnomonic projection to
ellipsoids; see [Karney (2013)](https://doi.org/10.1007/s00190-012-0578-z) for the citations:

* In separate papers, Bowring and Williams (1997) describe the central
  projection onto a plane tangent to the ellipsoid.

* Letoval’tsev (1963) gives a projection involving a conformal
  projection to a sphere followed by a spherical gnomonic projection.
  (Not totally sure about this; it's been a while since I looked at this
  in detail.)

* Karney (2013) uses the limit of a double azimuthal projection.

These are ranked (from worst to best) in how much projected geodesics
deviate from straight lines.

src/projections/aeqd.cpp
```diff
@@ -202,25 +202,21 @@ static PJ_LP aeqd_e_inverse (PJ_XY xy, PJ *P) {          /* Ellipsoidal, inverse
...
-
-        x2 = xy.x * P->a;
-        y2 = xy.y * P->a;
         lat1 = P->phi0 / DEG_TO_RAD;
-        lon1 = P->lam0 / DEG_TO_RAD;
-        azi1 = atan2(x2, y2) / DEG_TO_RAD;
-        s12 = sqrt(x2 * x2 + y2 * y2);
+        lon1 = 0;
+        azi1 = atan2(xy.x, xy.y) / DEG_TO_RAD;
+        s12 = hypot(xy.x, xy.y);
```
rouault: s12 quantity has been already computed at line 207 in variable c. Some cleanup is possible here

src/projections/gnom.cpp
```diff
@@ -122,25 +124,105 @@ static PJ_LP gnom_s_inverse (PJ_XY xy, PJ *P) {           /* Spheroidal, inverse
...
+                    nullptr, &azi0, nullptr, &m, &M, nullptr, nullptr);
+    if (M <= 0) {
+        proj_errno_set(P, PROJ_ERR_COORD_TRANSFM_OUTSIDE_PROJECTION_DOMAIN);
+        xy.x = xy.y = HUGE_VAL;
+    }
+    else {
+        double rho = m/M;
+        azi0 *= DEG_TO_RAD;
+        xy.x = rho * sin(azi0);
+        xy.y = rho * cos(azi0);
```
rouault, line 148: I've the feeling that geod_geninverse_int() returns those sin() and cos() values, so their re-computation could potentially be saved. Actually same pattern in aeqd

src/projections/aeqd.cpp
```diff
@@ -202,25 +202,21 @@ static PJ_LP aeqd_e_inverse (PJ_XY xy, PJ *P) {          /* Ellipsoidal, inverse
...
     if (Q->mode == OBLIQ || Q->mode == EQUIT) {
-
-        x2 = xy.x * P->a;
-        y2 = xy.y * P->a;
         lat1 = P->phi0 / DEG_TO_RAD;
-        lon1 = P->lam0 / DEG_TO_RAD;
-        azi1 = atan2(x2, y2) / DEG_TO_RAD;
-        s12 = sqrt(x2 * x2 + y2 * y2);
+        lon1 = 0;
+        azi1 = atan2(xy.x, xy.y) / DEG_TO_RAD;
```
rouault: usually atan2() is called with atan2(y, x) arguments. This must be a reversed angle here. Perhaps some comment indicating this is not an error could be added

src/projections/gnom.cpp
```diff
@@ -122,25 +124,105 @@ static PJ_LP gnom_s_inverse (PJ_XY xy, PJ *P) {           /* Spheroidal, inverse
...
+    }
+    return xy;
+}
+
+
+static PJ_LP gnom_e_inverse (PJ_XY xy, PJ *P) { /* Ellipsoidal, inverse */
+    static const int numit_ = 10;
```
rouault:
```suggestion
    constexpr int numit_ = 10;
```

src/projections/gnom.cpp
```diff
@@ -122,25 +124,105 @@ static PJ_LP gnom_s_inverse (PJ_XY xy, PJ *P) {           /* Spheroidal, inverse
...
+    static const double eps_ = 0.01 * sqrt(DBL_EPSILON);
+
+    PJ_LP lp = {0.0,0.0};
+    struct pj_opaque *Q = static_cast<struct pj_opaque*>(P->opaque);
+
+    double
+        lat0 = P->phi0 / DEG_TO_RAD,
+        lon0 = 0,
+        azi0 = atan2(xy.x, xy.y) / DEG_TO_RAD,
```
rouault: same remark as above regarding atan2(x,y) vs atan2(y,x)

src/projections/gnom.cpp
```diff
@@ -122,25 +124,105 @@ static PJ_LP gnom_s_inverse (PJ_XY xy, PJ *P) {           /* Spheroidal, inverse
...
+        // If little, solve rho(s) = rho with drho(s)/ds = 1/M^2
+        // else solve 1/rho(s) = 1/rho with d(1/rho(s))/ds = -1/m^2
+        double ds = little ? (m - rho * M) * M : (rho * m - M) * m;
+        s -= ds;
+        // Reversed test to allow escape with NaNs
+        if (!(fabs(ds) >= eps_))
+            ++trip;
+    }
+    if (trip) {
+        lp.phi = lat1 * DEG_TO_RAD; lp.lam = lon1 * DEG_TO_RAD;
```
rouault:
```suggestion
        lp.phi = lat1 * DEG_TO_RAD;
        lp.lam = lon1 * DEG_TO_RAD;
```

src/projections/aeqd.cpp
```diff
@@ -96,7 +96,7 @@ static PJ_XY aeqd_e_forward (PJ_LP lp, PJ *P) {          /* Ellipsoidal, forward
     struct pj_opaque *Q = static_cast<struct pj_opaque*>(P->opaque);
     double  coslam, cosphi, sinphi, rho;
     double azi1, azi2, s12;
-    double lam1, phi1, lam2, phi2;
+    double lat1, lon1, lat2, lon2;
```
rouault, line 99: unrelated to your changes, but looking at that code, I see that coslam, cosphi and sinphi are not used in the EQUIT and OBLIQ cases. I was wondering if the compiler is smart enough to realize that, and in fact it is only 50% smart. It doesn't compute coslam, but probably due to a heuristics when seeing cos(lp.phi) and sin(lp.phi), it calls the sincos() function in all code paths. cf this disassembly of the method where one sees there's no conditional jump before the call to sincos()

```
0000000000118fb0 <aeqd_e_forward(PJ_LP, PJconsts*)>:
  118fb0:       f3 0f 1e fa             endbr64 
  118fb4:       41 55                   push   %r13
  118fb6:       41 54                   push   %r12
  118fb8:       55                      push   %rbp
  118fb9:       53                      push   %rbx
  118fba:       48 89 fb                mov    %rdi,%rbx
  118fbd:       48 83 ec 58             sub    $0x58,%rsp
  118fc1:       4c 8d 64 24 28          lea    0x28(%rsp),%r12
  118fc6:       4c 8d 6c 24 20          lea    0x20(%rsp),%r13
  118fcb:       f2 0f 11 0c 24          movsd  %xmm1,(%rsp)
  118fd0:       64 48 8b 04 25 28 00 00 00      mov    %fs:0x28,%rax
  118fd9:       48 89 44 24 48          mov    %rax,0x48(%rsp)
  118fde:       31 c0                   xor    %eax,%eax
  118fe0:       4c 89 ee                mov    %r13,%rsi
  118fe3:       4c 89 e7                mov    %r12,%rdi
  118fe6:       f2 0f 11 44 24 08       movsd  %xmm0,0x8(%rsp)
  118fec:       e8 ef 4a f7 ff          callq  8dae0 <sincos@plt>
```

This is true of gcc 9.4 in -O2 mode, and clang 10.0 in -O2 (clang is actually worse and calls cos(), cos() and sin() !)

rouault: cf https://github.com/OSGeo/PROJ/pull/3523 for a cleanup done in the spherical forward case which suffers from the same issue (didn't touch the ellipsoidal case to avoid conflicts with this PR)

src/projections/aeqd.cpp
rouault: not a blocker for this PR.

src/projections/aeqd.cpp
cffk: Done, thanks for the catch.

src/projections/aeqd.cpp
cffk: The angle is measured clockwise from north.  I've added a comment here and in gnom.cpp

src/projections/gnom.cpp
cffk: See above...

src/projections/gnom.cpp
cffk: Done!

src/projections/gnom.cpp
cffk: Done

src/projections/gnom.cpp
rouault: this remark is more as a potential later optimization than something in the scope of this PR

src/projections/aeqd.cpp
cffk: I moved the assignments for cosphi+sinphi to the `S_POLE` case which should take care of this.

If it were up to me (and you and I probably differ here), I would remove the special treatment for `{S,N}_POLE` and let the general code handle all cases.  Pluses: shorter code, easier to maintain, the results in the special cases serve as a check for the general case.  Minus: slower is what might be relatively common special cases.

src/projections/gnom.cpp
cffk: `geod_geninverse_int` is currently not exposed and I'm reluctant to do so to capture this particular optimization. At some future point (who knows when), the accuracy and speed of all PROJ code could probably be improved by routinely passing angles around as a `[sin/cos]` pair.  I started experimenting with this with the sample code for my paper on auxiliary latitudes.  See [the documentation on AuxAngle](https://geographiclib.sourceforge.io/C++/2.1.2/classGeographicLib_1_1AuxAngle.html).  (Ah, and I see your note about "potential future optimization".  So we are on the same page here.)

PR #3527, opened 2022-12-30
Avoid cppcheck master warnings

rouault:


src/transformations/helmert.cpp
```diff
@@ -450,7 +450,8 @@ static void helmert_forward_4d (PJ_COORD &point, PJ *P) {
         build_rot_matrix(P);
     }
 
-    point.xyz = helmert_forward_3d (point.lpz, P);
+    const auto xyz = helmert_forward_3d (point.lpz, P);
+    point.xyz = xyz;
```
dg0yt, line 457: Can you explain these changes?
The next benevolent contributor might just revert them.

rouault: Latest cppcheck versions complain about "Overlapping read/write of union is undefined behavior" when doing "point.xyz = foo(point.lpz)", which is indeed a legitimate warning as, formally, the C++ standard disallows this. But this is fine in C!, so compilers do not do crazy things on this. Doing the assignment in 2 steps is enough to avoid the corresponding cppcheck to trigger (but probably still illegal from a pedantic point of view).
I've just added comments in the code w.r.t that

dg0yt: Thanks for the clarification. IIUC the point is that `point` is used with different members on rhs (read) and lhs (write), and this topic is covered by the "Member lifetime" section of https://en.cppreference.com/w/cpp/language/union.
With this knowledge, I would have suggested to simly choose an identifier such as `out_of_union_xyz` instead of `xyz`. But of course it is okay to leave the comments.

I'm afraid that the current usage of `PJ_COORD` comes with a lot of UB wrt to "active member".

PR #3529, opened 2022-12-31
replace the Wagner 7 projection with a new version that has an inverse

GrahamAsher:
The Wagner 7 projection had no inverse and thus was of no use for applications that needed an inverse projection, such as those which allow a user to click on a map and obtain the latitude and longitude. These new formulae are adapted from the D3 library, which has a permissive license, but more accurate constants, taken from a different source, are used. The new code has been tested and seems to work well.

### DIFF
--- a/src/projections/wag7.cpp
+++ b/src/projections/wag7.cpp
@@ -5,30 +5,58 @@
 #include "proj.h"
 #include "proj_internal.h"
 
-PROJ_HEAD(wag7, "Wagner VII") "\n\tMisc Sph, no inv";
+PROJ_HEAD(wag7, "Wagner VII") "\n\tMisc Sph";
 
+/*
+These formulae are from https://github.com/d3/d3-geo-projection/blob/main/src/wagner7.js.
+The original code is under the ISC License, which is permissive.
+See https://github.com/d3/d3/blob/main/LICENSE.
+The constants are explained at
+http://www.boehmwanderkarten.de/kartographie/is_netze_wagner_123456789_inversions.html.
 
+Required license text:
+
+Copyright 2013-2021 Mike Bostock
+
+Permission to use, copy, modify, and/or distribute this software for any purpose
+with or without fee is hereby granted, provided that the above copyright notice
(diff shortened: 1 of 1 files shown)


rouault:
The tests in test/gie/builtins.gie should be adapted (lines 7416 to 7422), and tests for the inverse direction added.

Digging a bit I see the formulas currently used in PROJ come from https://pubs.usgs.gov/pp/1453/report.pdf , page 233

I've no opinion if a compatibility mode to use the old formulas should be added (apparently, looking at the failed test, the difference between the old and new formulas is ~ half-meter at 1-2 degree away from the center of the projection)

src/projections/wag7.cpp
```diff
@@ -5,30 +5,58 @@
...
+    return xy;
 }
 
+static PJ_LP wag7_s_inverse (PJ_XY xy, PJ *P) {
+    (void)P;
+    PJ_LP lp = { 0, 0 };
+    double t1 = xy.x / (5.3344669029266510 / 2.0);
+    double t2 = xy.y / (2.4820727672498521 / 2.0);
+    double p = sqrt(t1 * t1 + t2 * t2);
+    double c = 2 * asin(p / 2);
```
rouault, line 51:
```suggestion
    if (p / 2 > 1.) {
        proj_errno_set(P, PROJ_ERR_COORD_TRANSFM_OUTSIDE_PROJECTION_DOMAIN);
        return xy;
    }
    double c = 2 * asin(p / 2);
```

GrahamAsher:
I wonder if the discrepancy was because of the difference between some of the constants used. One thing to try would be to use the old constants instead of my new ones.

Please feel free to adapt or change the code I provided if you need to; I might not have time to take it all the way through the process myself, because of pressure of work, although I will have another look after the weekend. My intention was not to give you or anybody else extra unwanted work, so please ignore the pull request if it is inconvenient or if there is no demand for the feature.

I am not a user of PROJ, but use the older Proj4 library in my code. I discovered the problem when trying to create a sample map of the world, and after some digging found the D3 implementation.

PR #3531, opened 2023-01-01
Add link to projsync

oleg-alexandrov:
The page https://proj.org/resource_files.html mentions projsync but doesn't link to it. This pull requests addresses that.

### DIFF
--- a/docs/source/resource_files.rst
+++ b/docs/source/resource_files.rst
@@ -172,7 +172,7 @@ The ``proj-data`` package is a collection of all the resource files that are
 freely available for use with PROJ. The package is maintained on
 `GitHub <https://github.com/OSGeo/PROJ-data>`_ and the contents of the package
 are show-cased on the `PROJ CDN <https://cdn.proj.org/>`_. The contents of the
-package can be installed using the :program:`projsync` package or by downloading
+package can be installed using the :ref:`projsync` program or by downloading
 the zip archive of the package and unpacking in the :envvar:`PROJ_DATA` directory.
 
 proj-datumgrid


kbevers:
The proposed fix changes the formatting from the black bold text of the `program`-directive to blue, regular font of clickable links. That's probably not a big deal in and of itself but it's inconsistent to just about all other references to the PROJ CLI apps used in the docs. I'm not sure this is the best fix to the problem, as this just scratces a particular itch of yours instead of introducing a general improvement across all of the documentation that keeps the look and feel of the docs consistent.

I agree that it would be much better to have the program-directive link to the actual doc-page for that application but that is not possible and is a well-known limitation of Sphinx. A better solution *might* land in Sphinx at some point cf https://github.com/sphinx-doc/sphinx/issues/880.

hobu:
> That's probably not a big deal in and of itself but it's inconsistent to just about all other references to the PROJ CLI apps used in the docs.

Why shouldn't all of the links to the PROJ CLI apps be clickable links?

oleg-alexandrov:
> this just scratces a particular itch of yours

I don't appreciate being second-guessed in my intentions. 

The proposed change offers a link to a relevant tool. I would think that is more important than look-and-feel. But that is for you folks to figure out.

snowman2:
I really like links when navigating documentation and always appreciate it when they exist as it makes for a better experience, for me at least.

Thoughts about updating this PR with a general search and replace to change projsync references to links for consistency sake?

rouault:
cf https://github.com/OSGeo/PROJ/pull/3534 for an attempt at a generic solution for linkable program directive

kbevers:
>  Why shouldn't all of the links to the PROJ CLI apps be clickable links?

That'd be lovely, which is why I was asking for a general solution in favour of the local fix. Seems @rouault how found a way to make that way.

rouault:
superseded per https://github.com/OSGeo/PROJ/pull/3534

PR #3534, opened 2023-01-02
Doc: make 'program' an alias of 'ref'

rouault:
Potential general resolution for https://github.com/OSGeo/PROJ/pull/3531


kbevers:
Looks promising! Do you mind grabbing a screenshot that shows how the text looks after this change?

rouault:
> Do you mind grabbing a screenshot that shows how the text looks after this change?

it looks just like a link of a regular :ref: (so blue text in HTML) . No bold font

kbevers:
Ah, okay. Would be cool to also have special formatting that hints that we're talking about an application but I can see how that might not be easy. As I understand the code it basically just replaces all occurences of `:program:` with `:ref:`, right?

rouault:
> Would be cool to also have special formatting that hints that we're talking about an application but I can see how that might not be easy.

I've just tried the following extra change, where I use innernodeclass=addnodes.literal_strong, which is the class that 'program' normally uses, but unfortunately that doesn't help. I'm afraid improving that is beyond my current level of mastering of Sphinx:
```
$ git diff
diff --git a/docs/source/_extensions/program_with_link.py b/docs/source/_extensions/program_with_link.py
index 62b31ef0..fe23530a 100644
--- a/docs/source/_extensions/program_with_link.py
+++ b/docs/source/_extensions/program_with_link.py
@@ -1,3 +1,5 @@
+from sphinx import addnodes
+from sphinx.roles import XRefRole
 from sphinx.domains.std import StandardDomain
 
 class MyStandardDomain(StandardDomain):
@@ -9,6 +11,6 @@ class MyStandardDomain(StandardDomain):
 
 def setup(app):
     # Override the "program" role to be an alias of "ref"
-    MyStandardDomain.roles["program"] = StandardDomain.roles["ref"]
+    MyStandardDomain.roles["program"] = XRefRole(lowercase=True, innernodeclass=addnodes.literal_strong, warn_dangling=True)
     app.add_domain(MyStandardDomain, override=True)
     return { 'parallel_read_safe': True, 'parallel_write_safe': True }
```

> As I understand the code it basically just replaces all occurences of :program: with :ref:, right?

Yes, :program: will behave exactly as :ref:

kbevers (approved):

kbevers:
I had a hunch it wouldn't be a simple change like that. Ultimately clickable links are better than fancy formatting without links, so a thumbs up from me on this one. With a bit of luck some day someone figures out a way to also do the formatting

PR #3538, opened 2023-01-03
Implement ReadTheDocs 

hobu:
ReadTheDocs now supports everything our custom documentation setup has plus more:

* versioned docs
* convenient per-PR doc rendering
* branch-based doc builds

This PR implements the plot generation, doxygen output, and HTML/PDF/epub builds. 

Once we are happy with the builds, we can explore using RTD as the host of proj.org


mwtoews:
would RTD be under a subdomain like docs.proj.org or doc.proj.org? Because the JSON specs need to be published separately in proj.org

hobu:
> would RTD be under a subdomain like docs.proj.org or doc.proj.org? Because the JSON specs need to be published separately in proj.org

It is up to the team to sort out. I was hoping to eliminate a bunch of hand-rolled infrastructure that is no longer needed and see if we liked the outcome. At the very least, the ability to see rendered doc output inside of PRs (not yet active) is a big win so people don't have to render things themselves to see edits.

hobu:
PR builds now activated. Click the link to see the docs rendered at this sha.

image

kbevers:
> would RTD be under a subdomain like docs.proj.org or doc.proj.org?

Is using proj.org not an option?

mwtoews:
xref January 2020 on the PROJ mail list: https://lists.osgeo.org/pipermail/proj/2020-January/thread.html ("proj manual, versioning, download location")

In particular, @rouault had a detailed RTD checklist https://lists.osgeo.org/pipermail/proj/2020-January/009268.html

Some of these checks were for versions of projjson schema, e.g.:

- https://proj.org/schemas/v0.1/projjson.schema.json
- https://proj.org/schemas/v0.2/projjson.schema.json
- ...
- https://proj.org/schemas/triangulation.schema.json

but similarly, I don't see these published via the `make html`. There could be a `post_build` job to do this ([docs](https://docs.readthedocs.io/en/stable/build-customization.html)). But I'm unsure how versioned schemas would work with versioned docs (would each docs version have their own set of schema versions??).

As for subdomains, I recall somewhere that RTD prefers subdomains rather than naked domains. Also, might help with using (e.g.) gh-pages to publish json schemas, and RTD to publish docs under a subdomain. But perhaps there are cleaver folks out there that know how to both host RTD from proj.org and allow the published schemas to function as they are now.

mwtoews:
`.github/workflows/doc_build.yml` could be removed and re-worked into `pre_build` / `post_build` jobs, e.g. Lint and Misspelled steps.

hobu:
> but similarly, I don't see these published via the make html. There could be a post_build job to do this ([docs](https://docs.readthedocs.io/en/stable/build-customization.html)). But I'm unsure how versioned schemas would work with versioned docs (would each docs version have their own set of schema versions??).

We have a couple of options. RTD has a somewhat sophisticated Redirect capability, and we could use that to point to a stable set of schema docs no matter the relative URL of the documentation build. Whatever we we would chose to do, we can control it all with a `post_build` event as you suggest.

> As for subdomains, I recall somewhere that RTD prefers subdomains rather than naked domains. Also, might help with using (e.g.) gh-pages to publish json schemas, and RTD to publish docs under a subdomain. But perhaps there are cleaver folks out there that know how to both host RTD from proj.org and allow the published schemas to function as they are now.

pdal.io is running RTD at the naked domain. I think this was true in the past but has been relaxed in the last year+.

hobu:
BTW, PSC members should register with RTD and let me know your UIDs so I can add you as admins for the project there.

kbevers:
> PSC members should register with RTD and let me know your UIDs so I can add you as admins for the project there.

Same username as here.

mwtoews:
> pdal.io is running RTD at the naked domain. I think this was true in the past but has been relaxed in the last year+.

Awesome, keeping docs at proj.org would obviously be preferred. And it seems that many workarounds can now be accomplished with `post_build`.

> PSC members should register with RTD and let me know your UIDs so I can add you as admins for the project there.

I'm not a PSC member, but would be happy to help (e.g.) maintain backports, same username as here too.

hobu:
https://osgeo-proj--3538.org.readthedocs.build/en/3538/schemas/v0.5/projjson.schema.json finally being included after a bunch of flailing

mwtoews:
For some reason, PDF and Epub formats are not built.

jjimenezshaw:
> PR builds now activated. Click the link to see the docs rendered at this sha.

Having a look at the docs produced in this PR, I saw something strange. For instance in 

https://osgeo-proj--3538.org.readthedocs.build/en/3538/development/reference/cpp/operation.html#_CPPv4N5osgeo4proj9operation10Conversion48createHotineObliqueMercatorTwoPointNaturalOriginERKN4util11PropertyMapERKN6common5AngleERKN6common5AngleERKN6common5AngleERKN6common5AngleERKN6common5AngleERKN6common5ScaleERKN6common6LengthERKN6common6LengthE 

there is an "explicit link" (not formatted) like 
```
Instantiate a conversion based on the <a href=”https://proj.org/operations/projections/omerc.html”.

>Hotine Oblique Mercator Two Point Natural Origin projection method.
```

Is this because the base domain is different (`proj.org` vs `osgeo-proj--3538.org.readthedocs.build`), or because the link is wrongly generated?

hobu:
at one point during all my flailing, the PDF builders were working, but now they don't seem to be 😦

hobu:
Maybe it is getting mangled by our `post_build` run?

mwtoews:
that's my guess too, but I can't find any docs to help with the scenario.

mwtoews:
@jjimenezshaw thanks for reading, these errors are fixed. It seems that Doxygen [prefers HTML's href](https://doxygen.nl/manual/autolink.html) over MarkDown-style links. I've also updated many links from https://www.epsg-registry.org to https://epsg.org

A trial run disabling `post_build` did not build the other formats, and I can't see any previous builds that have these, so this will remain a mystery for now.

hobu:
Still no PDFs, right?

mwtoews:
Found it: pdfs are not built with PRs, but will be built when merged https://docs.readthedocs.io/en/stable/pull-requests.html#limitations

<s>Should .github/workflows/doc_build.yml be dismantled in this PR?</s> Lets do that later, since gh-pages is still the host for proj.org

jjimenezshaw:
I saw in the built example, hosted in readthedocs, some advertisements at the end of each page. Will that stay there in the final version?

proj4docs:
OSGeo could pay $60/yr to remove adds if we wanted to do that.

mwtoews:
Another last question to @hobu is why use mambaforge and not pip?

hobu:
> Another last question to @hobu is why use mambaforge and not pip?

So we can support the plotting activity in the same run, as all the packages we need to do that are in conda-forge and won't clash like they can when installed as wheels from pip.

mwtoews (approved):
Let's merge this. Next steps:

1. Probably some PSC decisions around changing host for proj.org to use RTD, and consider membership options for PROJ and/or OSGeo to remove ads
2. After hosting has changed, dismantle gh-pages hosting and workflow steps, remove hard-links to PDF in docs

mwtoews:
See https://osgeo-proj.readthedocs.io/en/latest/ for html, pdf and epub outputs for the master branch.

kbevers:
Looks great. One minor detail that needs to be fixed is the PDF-link in the last line of the front page. It still points at the gh-pages PDF

mwtoews:
There are two PDF links still there (the other was on the Download page). I'll clean these up after proj.org is redirected to RTD.

jjimenezshaw:
> after proj.org is redirected to RTD.

redirected or aliased? I mean, if I type proj.org in the browser, is it going to become https://osgeo-proj.readthedocs.io/en/latest/ ?

jjimenezshaw:
Ok, I see that we "just have to" change the DNS: https://docs.readthedocs.io/en/stable/custom-domains.html

rouault:
we should probably get rid off the ad banner before making it official. One possibility would be that I pay and get then reimbursed through NumFocus. It looks like a legit expense that would fall under the "Cloud Services & Hosting: Cloud service and storage subscriptions, domain hosting platforms and services" category. What do you think @hobu ?

mwtoews:
Another change to note is that I've created an [exact redirect for a sub path](https://docs.readthedocs.io/en/stable/user-defined-redirects.html#redirecting-a-whole-sub-path-to-a-different-one): `/schemas/$rest` -> `/en/latest/schemas/`

This will (e.g.) redirect:
- From here: https://osgeo-proj.readthedocs.io/schemas/v0.6/projjson.schema.json
- To here: https://osgeo-proj.readthedocs.io/en/latest/schemas/v0.6/projjson.schema.json

which means that `master` would publish the schemas, including future ones that are not released.

kbevers:
>  One possibility would be that I pay and get then reimbursed through NumFocus

Another possibility is to let OSGeo handle it.

jjimenezshaw:
One question about the versions of PROJ in readthedocs (after playing a bit with it)
I see that the available versions (there is a selector at the bottom left corner) are:
   `latest`, `9.1`, `9.0`, `8.2`, `8.1`, `8.0`, `7.2`
where `latest` is the default.

In the current documentation in proj.org, the title of the left pane is "PROJ 9.1.1", while now in readthedocs is "PROJ latest". Do we want this? Should `latest` be the default, or the current stable release (with or without bugfix number)?

Opening the PDF of `latest`, says `Release 9.2.0`

https://docs.readthedocs.io/en/stable/versions.html#redirects-on-root-urls

mwtoews:
Normally the default version would be `stable`, which will only be available when [9.2 is released in March 2023](https://github.com/OSGeo/PROJ/milestone/36).

We could set the default to `9.1`.

PR #3550, opened 2023-01-11
Check about consistent celestial body: relax tolerance for Mars use cases…

rouault:
… and add a PROJ_IGNORE_CELESTIAL_BODY=YES environment variable


snowman2:
Thoughts about using a similar strategy for this option that is used in #3535 ?

rouault:
> Thoughts about using a similar strategy for this option that is used in #3535 ?

hum, that one is slightly more esoteric. I may add the setting, but after #3535 is merged to avoid conflicts as this would touch the same areas in files.

PR #3554, opened 2023-01-12
Implement ReadTheDocs for 9.0, disable other CI workflows

mwtoews:
This is a manual backport of #3553 to the 9.0 branch.

This also removes GitHub Actions build of docs, and PDF links in HTML outputs.

See result at https://osgeo-proj.readthedocs.io/en/9.0/


mwtoews:
@rouault while I'm backporting down versions, could I disable other CI checks? If so, which ones to keep or loose?

rouault:
> If so, which ones to keep or loose?

we could probably remove all jobs except the readthedoc one, as those branches are no longer active

mwtoews:
All CI workflows are disabled by simply relocating them to a `/disabled_workflows` directory.

PR #3578, opened 2023-01-21
Implement EPSG:9621 'Similarity transformation' and import related tranformations from EPSG

rouault:
This is used for a number of projectedCRS -> projectedCRS coordinate transformations.


src/iso19111/operation/singleoperation.cpp
```diff
@@ -2883,6 +2883,36 @@ bool SingleOperation::exportToPROJStringGeneric(
...
+                .value();
+        const double M =
+            parameterValueMeasure(
+                EPSG_CODE_PARAMETER_SCALE_FACTOR_FOR_SOURCE_CRS_AXES)
+                .value();
+        const double q = parameterValueNumeric(
+            EPSG_CODE_PARAMETER_ROTATION_ANGLE_OF_SOURCE_CRS_AXES,
+            common::UnitOfMeasure::RADIAN);
+
+        // Do not mess with axis unit and order for that transformation
```
jjimenezshaw, line 2903: Does it mean that if I use it in a DERIVEDPROJCRS in the DERIVINGCONVERSION, the axis unit and order of the CS are ignored?

rouault: > Does it mean that if I use it in a DERIVEDPROJCRS in the DERIVINGCONVERSION, the axis unit and order of the CS are ignored?

yes I wasn't too sure about what to do here, so I did the same as for Affine

PR #3588, opened 2023-01-23
C API: add proj_get_domain_count(), proj_get_area_of_use_ex() and proj_get_scope_ex()

rouault:
Up to now, the C API only handled one single domain/usage par object, but there are objets in the database that have several ones.

refs #3246

### DIFF
--- a/scripts/reference_exported_symbols.txt
+++ b/scripts/reference_exported_symbols.txt
@@ -1005,6 +1005,7 @@ proj_errno_string
 proj_factors
 proj_geod
 proj_get_area_of_use
+proj_get_area_of_use_ex
 proj_get_authorities_from_database
 proj_get_celestial_body_list_from_database
 proj_get_celestial_body_name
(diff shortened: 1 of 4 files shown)


kbevers:
Looks alright to me. I'm not sure how the `_ex` should be understood. Extended? Also, remember to update `docs/source/development/reference/functions.rst` before merging.

rouault:
yes, _ex is for extended. We have a proj_create_vertical_crs_ex() in proj_experimental.h

There's no need to explictly extend the documentation, as those functions below to the iso19111_functions Doxygen group that is ingested through Breathe, with the ".. doxygengroup:: iso19111_functions" directive at end of docs/source/development/reference/functions.rst

rouault:
Regarding automatic documentation, cf https://osgeo-proj--3588.org.readthedocs.build/en/3588/development/reference/functions.html#c.proj_get_domain_count

kbevers:
> There's no need to explictly extend the documentation, a

Right, I forgot about that. Everything's good then.

PR #3590, opened 2023-01-23
Whole code-base reformatting

rouault:
- Add a .pre-commit-config.yaml pre-commit configuration file
- Doc: add a dev_practices.rst page
- Whole code base reformatting
- Add .git-blame-ignore-revs


rouault:
Doc how to setup pre-commit hook at https://github.com/rouault/PROJ/blob/reformat_whole/docs/source/development/dev_practices.rst
This mimmicks recent similar move in GDAL and libtiff

docs/source/development/index.rst
```diff
@@ -11,6 +11,7 @@ PROJ project or using the library in their own software.
 .. toctree::
    :maxdepth: 1
 
+   dev_pratices
```
hobu: typo

hobu (approved):

docs/source/development/index.rst
rouault: thanks. fixed

busstoptaktik (commented):
If I understand the meaning of this item correctly, I think the commit some years back, where you renamed all .c files to .cpp should probably also be ignored? (there may be others as well, which I have lost track of?)

rouault:
> I think the commit some years back, where you renamed all .c files to .cpp should probably also be ignored? (there may be others as well, which I have lost track of?)

I don't believe this is necessary as git is already smart enough to figure out file renames.

busstoptaktik:
> I don't believe this is necessary as git is already smart enough to figure out file renames.

I think you're right: "git blame" appears to work fine - it's just the "contributors"-banner dispayed by Github over each file that only includes people contributing after the renaming. Perhaps more of a Github UI/UX bug-let...

scripts/reformat_cpp.sh
```diff
@@ -15,12 +15,12 @@ esac
 
 TOPDIR="$(readlink -f $SCRIPT_DIR/..)"
 
-if ! (clang-format --version 2>/dev/null | grep 10 >/dev/null); then
-    echo "clang-format 10 not available. Running it from a Docker image";
-    docker build -t proj_clang_format -f "$TOPDIR"/scripts/proj_clang_format/Dockerfile "$TOPDIR"/scripts/proj_clang_format
+if ! (clang-format --version 2>/dev/null | grep 15 >/dev/null); then
```
jjimenezshaw: If here you directly call `clang-format-15` (instead of the symbolic link `clang-format`) it is not mandatory to have the symbolic link to _that_ specific version. That allows the developer to have different versions of clang-format installed nicely. That should imply changes also in the script that finally calls clang-format.

rouault: done in extra commit

jjimenezshaw:
I had some difficulties installing clang-format-15 in my Ubuntu 22.04 (things are too old or too new)
So this is a tip if anybody has the same problem. I installed manually the deb files 

- `libllvm15_15.0.6-3_ubuntu0.22.04.2_amd64.deb` 
- `libclang-cpp15_15.0.6-3_ubuntu0.22.04.2_amd64.deb`
- `clang-format-15_15.0.6-3_ubuntu0.22.04.2_amd64.deb`
 
from http://security.ubuntu.com/ubuntu/pool/universe/l/llvm-toolchain-15/

rouault:
> I had some difficulties installing clang-format-15 in my Ubuntu 22.04 (things are too old or too new)

unless you want to manually run reformat_cpp.sh, you don't need to do that. Just follow instructions at https://github.com/rouault/PROJ/blob/reformat_whole/docs/source/development/dev_practices.rst#commit-hooks

PR #3598, opened 2023-02-01
implement normalizeForVisualization for BoundCRS

jjimenezshaw:
Just adding this missing implementation 

- [x] Tests added

### DIFF
--- a/src/iso19111/crs.cpp
+++ b/src/iso19111/crs.cpp
@@ -1169,6 +1169,12 @@ CRSNNPtr CRS::normalizeForVisualization() const {
         }
     }
 
+    if (const BoundCRS *boundCRS = dynamic_cast<const BoundCRS *>(this)) {
+        auto baseNormCRS = boundCRS->baseCRS()->normalizeForVisualization();
+        return BoundCRS::create(baseNormCRS, boundCRS->hubCRS(),
+                                boundCRS->transformation());
+    }
+
     return NN_NO_CHECK(
         std::static_pointer_cast<CRS>(shared_from_this().as_nullable()));
 }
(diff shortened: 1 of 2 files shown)


rouault:
Did you check that the transformation associated with the boundCRS still works OK after normalizeForVisualization() ? I'd think so, but worth checking if not done

jjimenezshaw:
> Did you check that the transformation associated with the boundCRS still works OK after normalizeForVisualization() ? I'd think so, but worth checking if not done

I added a check of `exportToPROJString`. Is that what you mean? the output of `exportToPROJString` is the same before and after `normalizeForVisualization`.

rouault:
> I added a check of `exportToPROJString`. Is that what you mean? the output of `exportToPROJString` is the same before and after `normalizeForVisualization`.

I was thinking more to the transformation pipeline returned by createOperation() from the bound CRS to EPSG:4326.  exportToPROJString() on the boundCRS itself isn't very significant as PROJ CRS strings don't capture axis order.

jjimenezshaw:
>  I was thinking more to the transformation pipeline returned by createOperation()

That was my other option. I'll do it later.

jjimenezshaw:
MacOS CI is failing on something that I guess is not related with my changes.

rouault:
> MacOS CI is failing on something that I guess is not related with my changes.

agreed. PR merged

src/iso19111/crs.cpp
```diff
@@ -1169,6 +1169,12 @@ CRSNNPtr CRS::normalizeForVisualization() const {
         }
     }
 
+    if (const BoundCRS *boundCRS = dynamic_cast<const BoundCRS *>(this)) {
+        auto baseNormCRS = boundCRS->baseCRS()->normalizeForVisualization();
+        return BoundCRS::create(baseNormCRS, boundCRS->hubCRS(),
+                                boundCRS->transformation());
+    }
```
hernando, line 1176: Thanks for this fix. You forgot to simplify the code that calls this function assuming it doesn't handle BoundCRS.

hernando: Ignore this comment, I got confused by other code base.

PR #3604, opened 2023-02-03
Doc: Prevent double hypen (--) to be replaced by Unicode long dash character

rouault:


mwtoews:
an alternative is to re-format the usage section of these man pages as [literal blocks](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#literal-blocks), e.g.:
```
**projinfo**::

    [-o formats] [-k crs|operation|datum|ensemble|ellipsoid] [--summary] [-q]
...
```
but I don't mind `smartquotes=False` as an easy solution for now.

PR #3608, opened 2023-02-05
proj_trans(): restore behaviour of PROJ 9.1 with transforming points …

### LINKED ISSUE #3607: Difference in results after only best change added
From: https://github.com/pyproj4/pyproj/issues/1230

#### Example of problem

Note: PROJ_NETWORK is OFF. With PROJ_NETWORK=ON, the results are the same.

PROJ master (9.2)
```
echo -111.5 45.25919444444 | ./cs2cs +proj=latlong +datum=WGS84 +to +proj=utm +zone=10 +datum=NAD27
1402291.08	5076289.59 0.00
```
PROJ 9.1
```
echo -111.5 45.25919444444 | cs2cs +proj=latlong +datum=WGS84 +to +proj=utm +zone=10 +datum=NAD27
1402288.54	5076296.64 0.00
```
#### Problem description

Was not expecting this change when the only best option was added. Wanting to verify the behavior change is expected.

git bisect points to this change:
https://github.com/OSGeo/PROJ/commit/d11c9d57668697904ce95e02a121cd9389ebb2c2

#### Expected Output
No change with previous PROJ versions.

#### Installation method
 - from source

rouault:
…in northern USA states from NAD27 to WGS 84 when no grid is available

Fixes #3607 and regression of d11c9d5

### DIFF
--- a/src/4D_api.cpp
+++ b/src/4D_api.cpp
@@ -209,8 +209,8 @@ static bool isSpecialCaseForNAD83_to_NAD83HARN(const PJCoordOperation &op) {
 /**************************************************************************************/
 int pj_get_suggested_operation(PJ_CONTEXT *,
                                const std::vector<PJCoordOperation> &opList,
-                               const int iExcluded[2], PJ_DIRECTION direction,
-                               PJ_COORD coord)
+                               const int iExcluded[2], bool skipNonInstantiable,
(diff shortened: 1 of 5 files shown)


kbevers:
This is a refinement of #3533, right? No need for special treatment in release notes?

rouault:
yes, this is a refinement of #3533

PR #3615, opened 2023-02-07
Doc: typo - missing parenthesis

jjimenezshaw:
Simply that. I had to read twice to figure out where was the missing `)`.


kbevers:
The man-files are automatically generated from the Sphinx docs, so this needs to be fixed upstream in `docs\source\apps\projinfo.rst`

jjimenezshaw:
> The man-files are automatically generated from the Sphinx docs, so this needs to be fixed upstream in `docs\source\apps\projinfo.rst`

Sorry. Now it is in the right file.

PR #3621, opened 2023-02-12
pj_get_suggested_operation(): tune for GDA94 <--> WGS84 and GDA2020 <…-> WGS84

rouault:
Before this change, the au_icsm_GDA94_GDA2020_conformal_and_distortion.tif was considered as the best operation for GDA94 <--> WGS84 and GDA2020 <--> WGS84 from cs2cs / proj_crs_to_crs() point of view. This change makes the null transformation be the prefered one instead. Applying the GDA94 <--> GDA2020 grid was highly questionable GDA2020 <--> WGS84. Not completely sure for GDA94 <--> WGS84 though...

Relates to https://github.com/qgis/QGIS/issues/51792

@nyalldawson Opinion greatly welcome

### DIFF
--- a/src/4D_api.cpp
+++ b/src/4D_api.cpp
@@ -222,6 +222,18 @@ static bool isSpecialCaseForNAD83_to_NAD83HARN(const PJCoordOperation &op) {
            op.name.find("NAD83 to NAD83(HARN) (50)") != std::string::npos;
 }
 
+// Returns true if the passed operation uses "GDA94 to WGS 84 (1)", which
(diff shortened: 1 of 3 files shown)


nyalldawson:
I'll weigh in next week -- this is too tricky for a Sunday morning 🤣

nyalldawson:
Yep, this looks correct by my understanding:

gda2020 <-> gda94 -- not null
gda2020 <-> wgs94 = null
gda94 <-> wgs94 = null

PR #3630, opened 2023-02-22
NEWS: move bug fix mention from 9.1.1 to 9.2.0

rouault:
Not critical but FYI @kbevers


kbevers:
Good catch! It'll get included in the final release. Or RC2 if there's a reason make one of those.

PR #3645, opened 2023-03-06
Adds Pseudo Mercator to build_db_from_iau.py

AndrewAnnex:
Note there is some inconsistency with hyphenation of "Pseudo-Mercator" so this may need a double check, but I believe what I have is consistent with EPSG:3857 WKT2:2019 representation. Related to https://github.com/pdssp/csvforwkt/pull/5


rouault:
Settting it as a draft seeing remark from @thareUSGS in https://github.com/pdssp/csvforwkt/issues/4#issue-1588444022. In particular the XXXX90 code needs to be confirmed for interoperability with other implementations

thareUSGS (commented):
@AndrewAnnex  Why "Popular Visualization Pseudo Mercator" (EPSG: 1024) and not simply  "Mercator (Spherical)" (EPSG: 1026). https://epsg.io/1026-method

I have never fully understood the whole Mercator / Web Mercator issue. I would rather use a name that isn't so odd (e.g., "Popular Visualization...") if possible. Looks like it is the same equation used in Mercator (Spherical).

rouault:
> Looks like it is the same equation used in Mercator (Spherical).

There's a subtlety if it is applied on a non-spherical ellipsoid. EPSG:1024 will always use the semi-major axis as the R value, whereas EPSG:1026 will use the radius of the conformal sphere, at the latitude of origin, which will be a * sqrt(1 - e^2) for latitude of origin = 0

AndrewAnnex:
@rouault thanks for that explanation, I hope that 1024 is the best choice of "mercator" in order to make it slightly more straight forward to adapt/extend standards and front ends that assume 3857 for Earth, so that the correct semi-major axis can just be swapped in.

thareUSGS:
@rouault This update for "Web Mercator" has been approved and merged. https://github.com/pdssp/csvforwkt/pull/5  It would be great to see this pull request also approved for PROJ. thanks!

PR #3648, opened 2023-03-07
Update clang static analyzer CI and docs for ubuntu-22.04

mwtoews:
Recent Ubuntu has clang's `scan-build` in their [`clang-tools` package](https://packages.ubuntu.com/jammy/clang-tools) (jammy depends on [`clang-tools-14`](https://packages.ubuntu.com/jammy/clang-tools-14)), so this PR changes to the packaged version.

With this change, the CI script has a new error:
> jq: error (at <stdin>:809): null (null) and string ("_generated...) cannot have their containment checked

which is (maybe?) resolved via [this Q/A](https://stackoverflow.com/a/70188269/) to use `index` instead of `contains`.

This also upgrades to `ubuntu-22.04` and `actions/checkout@v3`. (I have a follow-up PR to upgrade other overdue CI components elsewhere).

### DIFF
--- a/.github/workflows/clang_static_analyzer.yml
+++ b/.github/workflows/clang_static_analyzer.yml
@@ -15,11 +15,11 @@ concurrency:
 jobs:
 
   clang_static_analyzer:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
     if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]')"
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Run
         run: .github/workflows/clang_static_analyzer/start.sh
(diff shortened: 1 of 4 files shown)


mwtoews:
The new warning/error raised here is:
```
/home/runner/work/PROJ/PROJ/src/filemanager.cpp:894:21: warning: The parameter must not be null [cplusplus.StringChecker]
    if (starts_with(filename, "http://") || starts_with(filename, "https://")) {
                    ^~~~~~~~
```
I don't have a fix for this, but @rouault is welcome to add a commit to this PR (or in a subsequent PR).

rouault:
> The new warning/error raised here is:

@mwtoews fix in https://github.com/mwtoews/PROJ/pull/4

rouault:
actually I've just pushed the commit to this PR

mwtoews:
Given that the `ubuntu-18.04` actions runner image will be fully unsupported by 2023/04/01, how should this be applied to the 9.2 branch? As an easy option, the clang static analyzer job could simply be disabled for the `9.2` branch.

rouault:
> As an easy option, the clang static analyzer job could simply be disabled for the `9.2` branch.

probably the simpler way.

PR #3649, opened 2023-03-07
CI: update ubuntu images, actions/checkout, actions/cache, unique job IDs

mwtoews:
This PR upgrades a few components and resolves a few warning messages from GitHub actions, i.e.:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see `https://github.com/actions/virtual-environments/issues/6002`

<s>Ubuntu versions in docker containers like .github/workflows/mingw_w64.yml are upgraded for consistency.</s>

Some job IDs were renamed so they are unique, which is important for tools like [act](https://github.com/nektos/act) (`act -l`).

Other upgrades are handled with #3648.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   backport:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     name: Backport
     steps:
       - name: Backport Bot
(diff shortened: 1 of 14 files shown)


mwtoews:
I'll manually backport this to `9.2` along with parts of #3648. Keeping the ubuntu-18.04 docker image for mingw_w64 for now, since it needs further work (and has lower priority). New cppcheck failures appear to be harmless false positives.

rouault (approved):

PR #3661, opened 2023-03-10
[Performance regression fix] Fix slowness on proj_trans() on WGS 84 <--> NAD83 conversions

rouault:
Fix performance regression of 9.2.0 due to
proj_coordoperation_is_instantiable() being potentially called at each point transformation and being a slow operation, due to calling exportToPROJString()

### DIFF
--- a/src/proj_internal.h
+++ b/src/proj_internal.h
@@ -316,6 +316,7 @@ typedef void (*PJ_OPERATOR)(PJ_COORD &, PJ *);
 #define PJD_WGS84 4 /* WGS84 (or anything considered equivalent) */
 
 struct PJCoordOperation {
+  public:
     int idxInOriginalList;
     double minxSrc = 0.0;
     double minySrc = 0.0;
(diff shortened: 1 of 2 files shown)


src/proj_internal.h
```diff
@@ -390,6 +391,13 @@ struct PJCoordOperation {
     }
 
     ~PJCoordOperation() { proj_destroy(pj); }
+
+    bool isInstantiable();
+
+  private:
+    static constexpr int INSTANTIABLE_STATUS_UNKNOWN =
+        -1; // must be different from 0(=false) and 1(=true)
+    int isInstantiableCached = INSTANTIABLE_STATUS_UNKNOWN;
```
jjimenezshaw: Would it make sense to set it as `mutable` and avoid removing all those `const`s? A cache variable is a typical example to be mutable. (I do not know if there is any multithreading issue in this case)

rouault: I had the same feeling when committing. I will change to mutable and restore the const's.
const is in no way a hint that somethin is multithread safe. All those objects can only be run from one PROJ_CONTEXT*

jjimenezshaw (approved):

PR #3668, opened 2023-03-17
vgridshift/gridshift: accept hydroid_height as valid band description

rouault:


rouault:
cf https://github.com/OSGeo/PROJ-data/pull/97

rouault:
CC @kbevers @Jochem-L

PR #3673, opened 2023-03-22
Coord. operation factory: count identified concatenated operations as a single step

rouault:


rouault:
appveyor failure unrelated

PR #3685, opened 2023-03-28
EngineeringCRS: make proj_create_engineering_crs() set a datum name, …

rouault:
…and relax isEquivalentTo() comparisons to deal with case of https://github.com/r-spatial/sf/issues/2049#issuecomment-1486600723

### DIFF
--- a/test/unit/test_crs.cpp
+++ b/test/unit/test_crs.cpp
@@ -5968,6 +5968,51 @@ TEST(crs, engineeringCRS_WKT1) {
 
 // ---------------------------------------------------------------------------
 
+TEST(crs, engineeringCRS_unknown_equivalence) {
+
+    // Test equivalent of CRS definition got from GPKG (wkt2) and its equivalent
+    // from GeoTIFF (wkt1)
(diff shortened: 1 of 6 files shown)


test/unit/test_crs.cpp
```diff
@@ -5968,6 +5968,51 @@ TEST(crs, engineeringCRS_WKT1) {
 
 // ---------------------------------------------------------------------------
 
+TEST(crs, engineeringCRS_unknown_equivalence) {
+
+    // Test equivalent of CRS definition got from GPKG (wkt2) and its equivalent
+    // from GeoTIFF (wkt1)
+    // Cf https://github.com/r-spatial/sf/issues/2049#issuecomment-1486600723
+    auto wkt1 = "ENGCRS[\"Undefined Cartesian SRS with unknown unit\",\n"
```
jjimenezshaw, line 5976: Why is it an `unknown unit`?  There says `metre`.

jjimenezshaw: Oh, I guess it is coming from the example in the comment.

rouault: > Oh, I guess it is coming from the example in the comment.

yes

PR #3697, opened 2023-04-04
CMake: avoid imbalanced cmake_policy push/pop if TIFF or CURL dependency cannot be found (fixes #3696)

### LINKED ISSUE #3696: CMake build issues (GDAL latest) with PROJ 9.2.0 Docker Image
From: https://github.com/rasterio/rasterio/pull/2788

Failed build with more detailed build logs: https://github.com/rasterio/rasterio/actions/runs/4603105992/jobs/8132791454?pr=2788

This seemed like the culprit, but could be something else:
```
CMake Error in /usr/lib/x86_64-linux-gnu/cmake/proj/proj-config.cmake:
  cmake_policy PUSH without matching POP
  cmake/helpers/CheckDependentLibraries.cmake:348 (find_package)
  gdal.cmake:265 (include)
  CMakeLists.txt:246 (include)
```

Building GDAL latest worked fine on the PROJ 9.1.0 image, the only change was updating to 9.2.0.

If you need more details, please let me know and I would be happy to provide them.

Thanks!

rouault:


jjimenezshaw (approved):

dg0yt:
Given that `CMP0012` came with 2.8.0, I wonder if is this should be carried around at all. It is needed in one of the following cases:
- Not calling `cmake_minimum_required()` before `project()`.
- Actually setting a minimum version less than 2.8.0. 
- Actually setting `CMP0012` to `OLD`.

The first case is a user error. For the other cases, it needs to be decided: supported or not?

When building package which require a lower version but don't explicitly set `CMP0012` (second case), I can use `-DCMAKE_POLICY_DEFAULT_CMP<NNNN>=NEW`.
https://cmake.org/cmake/help/v3.20/variable/CMAKE_POLICY_DEFAULT_CMPNNNN.html

rouault:
> Given that `CMP0012` came with 2.8.0, I wonder if is this should be carried around at all.

yeah, I wonder why I introduced that at the beginning, but I assume I hit some error in some config otherwise I wouldn't have never put that, but I don't recall the situation. Hopefully the new code should be safe, even if mostly unnecessary.

kloczek:
I think that it would be good to pysh new release only because thay fix 🤔

rouault:
The fix here just avoids the somewhat misleading error message about imbalanced cmake_policy. But the real fix is that users need to make sure PROJ dependencies are available. That said, seeing that, I'm a bit ambivalent about the effect of the original change done in https://github.com/OSGeo/PROJ/pull/3374 that brought the need for that new requirement, which is useless if libtiff or libcurl are dynamic libraries. But I'm not sure if there's an elegant way of doing that. For Unix, testing for .so or .dylib extensions should be OK. For Windows, the extension is always .lib for both static or dynamic import lib.

kloczek:
Generally speaking this is why I hate cmake dependencies because cmake uses procedural description instead declarative one like pkgconfig.
IMO as OSGeos projests are provifing .pc files all cmake chedks of external libraries PROj and other peojects should be rewtited by rrplace use cmake modules by single lines `pkg_search_module()` calls.
With that it would be possible to remove tons of (IMO completly useless) cmake code .. all to be KISS pronciple comppliant.

PR #3705, opened 2023-04-09
Add C++ API to transform coordinate and a C++ quickstart

### LINKED ISSUE #3593: Examples for the C++ API
Hello,

I would like to use the C++ API instead of the C api but the entry cost is high : there are no examples for using the C++ API and the documentation is sparse.

Best regards,

Victor

rouault:
Fixes #3593


src/iso19111/operation/singleoperation.cpp
```diff
@@ -384,6 +407,92 @@ CoordinateOperationNNPtr CoordinateOperation::shallowClone() const {
...
+    c.v[3] = HUGE_VAL;
+    c = transformer->transform(c);
+    EXPECT_NEAR(c.v[0], 426857.98771728, 1e-8); // easting in metre
+    EXPECT_NEAR(c.v[1], 5427937.52346492, 1e-8); // northing in metre
+    proj_context_destroy(ctx);
+\endcode
+ */
+PJ_COORD CoordinateTransformer::transform(PJ_COORD coord) {
+    return proj_trans(d->pj_, PJ_FWD, coord);
+}
```
mitjap, line 492: Sorry for late feedback after this is already merged. Would be nice to be able to pass `proj_direction` as parameter.

rouault: > Would be nice to be able to pass `proj_direction` as parameter.

you can call inverse() on a CoordinateOperation

PR #3712, opened 2023-04-27
CI: upgrade vcpkg tag used for AppVeyor; use --triplet option

mwtoews:
Resolves two issues:

1. AppVeyor used an old tag of vcpkg that failed to download zlib, so upgrade to a newer tag
2. Resolve a warning: ""Starting with the September 2023 release, the default triplet for vcpkg libraries will change from x86-windows to the detected host triplet (x64-windows). To resolve this message, add --triplet x86-windows to keep the same behavior". This is applied to two CI jobs and to docs.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -46,9 +46,7 @@ jobs:
         shell: cmd
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
-            vcpkg install sqlite3[core,tool]:${{ env.ARCH }}-windows
-            vcpkg install tiff:${{ env.ARCH }}-windows
-            vcpkg install curl:${{ env.ARCH }}-windows
+            vcpkg install sqlite3[core,tool] tiff curl --triplet=${{ env.ARCH }}-windows
 
       - name: Build
         shell: cmd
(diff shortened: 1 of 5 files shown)


mwtoews:
### Errors from 4D_api.cpp

```
..\src\4D_api.cpp(328): error C2220: warning treated as error - no 'object' file generated
..\src\4D_api.cpp(328): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
```
https://github.com/OSGeo/PROJ/blob/700721eeafd239112f289bba12d08fc8844c376d/src/4D_api.cpp#L328

### Errors from c_api.cpp
```
..\src\iso19111\c_api.cpp(9067): error C2220: warning treated as error - no 'object' file generated
..\src\iso19111\c_api.cpp(9067): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
..\src\iso19111\c_api.cpp(9106): warning C4800: 'const int': forcing value to bool 'true' or 'false' (performance warning)
```
https://github.com/OSGeo/PROJ/blob/700721eeafd239112f289bba12d08fc8844c376d/src/iso19111/c_api.cpp#L9067
https://github.com/OSGeo/PROJ/blob/700721eeafd239112f289bba12d08fc8844c376d/src/iso19111/c_api.cpp#L9106

PR #3743, opened 2023-05-28
Doc/CI: Transition to ReadTheDocs, clean-up previous methods

mwtoews:
See [this ref](https://lists.osgeo.org/pipermail/proj/2023-May/011042.html) to the transition of proj.org to ReadTheDocs. There are a few updates and clean-ups to follow on:

- Remove HTML references to PDF documentation
- Update HOWTO-RELEASE with new steps
- Clean-up GitHub doc workflow
- <s>Clean-up Travis CI scripts and files related to publishing docs</s>
- <s>Clean-up files associated with the proj-docs docker image</s>

Is there too much clean-up? Should anything be kept, either be commenting-out or relocating to another directory?

A subset of these changes will be manually backported, as necessary (mostly the HTML references to PDF documentation).


.github/workflows/docker.yml
```diff
@@ -22,13 +22,8 @@ jobs:
...
     strategy:
       matrix:
-        container: ["proj","proj-docs"]
-        dockerfile: ["./Dockerfile", "./docs/docbuild/Dockerfile"]
-        exclude:
-          - container: "proj"
-            dockerfile: "./docs/docbuild/Dockerfile"
-          - container: "proj-docs"
-            dockerfile: "./Dockerfile"
+        container: ["proj"]
```
rouault: I would revert this: proj-docs is used by (at least) GDAL

docs/docbuild/Dockerfile
```diff
@@ -1,32 +0,0 @@
-FROM ubuntu:20.04
```
rouault: I would revert this: proj-docs is used by (at least) GDAL

docs/docbuild/README.md
```diff
@@ -1,8 +0,0 @@
-# Building PROJ documentation with proj-docs docker image
```
rouault: I would revert this: proj-docs is used by (at least) GDAL

.github/workflows/doc_build.yml
```diff
@@ -1,124 +0,0 @@
-name: Docs
```
rouault, line 1: Isn't that job potentially useful to check the output of a pull request touching docs, thanks to the artifacts ? (unless there's an equivalent with RTD)

.travis.yml
```diff
@@ -14,15 +14,6 @@ matrix:
   fast_finish: true
   include:
 
-    - os: linux
-      dist: xenial
```
rouault: Would it be possible to isolate all Travis related changes in a single commit, so we can revert it in a isolated way if we had to go back ?

jjimenezshaw:
Maybe this is not the place to ask.

Let's say there is a webpage or a pdf with a link to https://proj.org/operations/transformations/gridshift.html
(That was any link to PROJ documentation until yesterday)

Now it is returning a 404

Is there a way to redirect automatically to somewhere meaningful? For instance to https://proj.org/en/latest/operations/transformations/gridshift.html

Webpages will need some work to update (if the owner is aware and cares). Pdfs probably not possible.

Thanks

hobu:
Yes we probably need to set up some forwards.

hobu:
> Let's say there is a webpage or a pdf with a link to https://proj.org/operations/transformations/gridshift.html
(That was any link to PROJ documentation until yesterday)

Redirect added

https://proj.org/operations/transformations/gridshift.html should properly forward now

jjimenezshaw:
> 
> Redirect added
> 


Great! 
Thanks.

.github/workflows/doc_build.yml
mwtoews: The RTD check should fail in a similar way, but I'll check in this PR with a dummy commit (soon).

.travis.yml
mwtoews: See #3744

mwtoews:
Thanks for reviews, Travis CI related changes are split off to #3744.

The [`proj-docs`](https://hub.docker.com/r/osgeo/proj-docs) Docker content is restored. Perhaps this could be considered to be moved into a separate repo like `OSGeo/proj-docs`?

On the subject of redirects, note that schema URLs like this:
> https://proj.org/schemas/v0.4/projjson.schema.json

now redirect to:
> https://proj.org/en/9.2/schemas/v0.4/projjson.schema.json

.github/workflows/doc_build.yml
rouault: ah, I was thinking more on downloading the artifacts to check they look nice, but I see that the RTD job has an even nicer output that can be directly visualized, so my point is moot

mwtoews:
I've tested a "bad commit" a8c0caf9568f1ce7bf0a86f821eed63ef4cad326 to recreate [this recent failure](https://github.com/OSGeo/PROJ/actions/runs/4387922346/jobs/7684058230) that originated from doc_build.yml to see what it looks like with a [RTD build](https://readthedocs.org/projects/osgeo-proj/builds/20847175/). Indeed the RTD build log is useful, and the GitHub checks show ❌ "Read the Docs build failed!" 
image
So I don't see any need to keep the doc_build.yml workflow.

hobu:
> The [`proj-docs`](https://hub.docker.com/r/osgeo/proj-docs) Docker content is restored. Perhaps this could be considered to be moved into a separate repo like `OSGeo/proj-docs`?

Only GDAL is left using the `proj-docs` Dockerfile, with the long term intent that GDAL move forward to RTD at some point as well. I think it should just be pushed over into GDAL's repo.

hobu:
We can use `latest` to ensure that public URLs have an indirection:

> https://proj.org/en/latest/schemas/v0.4/projjson.schema.json
> https://proj.org/en/latest/schemas/v0.5/projjson.schema.json

snowman2:
I recommend using `stable` for redirection so it defaults to the latest release: https://proj.org/en/stable/schemas/v0.5/projjson.schema.json

rouault:
> Only GDAL is left using the `proj-docs` Dockerfile, with the long term intent that GDAL move forward to RTD at some point as well. I think it should just be pushed over into GDAL's repo.

I've created a GDAL ticket about that: https://github.com/OSGeo/gdal/issues/7880

mwtoews:
A similar subset of this PR was manually backported:

- `9.2`: 21166b23fa7710ee27b90431f18d052d28f0b036
- `9.1`: 127583f178fd4d864759f150d32b622801eb9481

PR #3747, opened 2023-06-13
CMake: remove useless cross-compiling related checks (fixes #3746)

### LINKED ISSUE #3746: PROJ encoding of cross-compilation mode stops valid usage of package
#### Example of problem

Currently PROJ encodes of it is being built with a cross compilation toolchain in the `project-config-version.cmake`:
https://github.com/OSGeo/PROJ/blob/master/cmake/project-config-version.cmake.in#L17

So for example if PROJ is built on an x86 machine for ARM it will encode the fact that PROJ was cross compiled.

If this version of PROJ is installed onto an ARM machine it will not be usable via CMake since the consumers will not be cross compiling.

#### Problem description

The current behavior of PROJ encoding of cross compilation is simply encoding how PROJ was built which doesn't matter. If it should encode anything it would be about the target platform ( 32/64bit, x86/arm/power/, OS ).

Likewise since PROJ is simply encoding that consumers must also be cross compiling it creates false positive use cases.
Consider when PROJ is cross compiled from x86 to ARM we get the following truth table:

| host | target | can 'use' proj ( current ) | actually usable |
| - | - | - | - |
| x86 | arm | yes | yes |
| arm | arm | no | yes |
| x86 | x86 | yes | no |
| arm | x86 | yes | no  |
| x86 | power | yes | no |

#### Expected Output

We should drop all encoding of how the state of PROJ cross compilation in the `project-config-version.cmake`

#### Environment Information
 - PROJ version 9.2.1
 - Linux x86, arm, power, .....

#### Installation method
 - conda

rouault:


hobu:
Would seem reasonable to backport this to 9.2.x as well if we have another.

rouault:
The next planned version seems to be 9.3.0 beginning of september

PR #3748, opened 2023-06-14
GeographicBoundingBox::intersection(): avoid infinite recursion and stack overflow on invalid bounding boxes

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57328

### DIFF
--- a/src/iso19111/metadata.cpp
+++ b/src/iso19111/metadata.cpp
@@ -421,6 +421,13 @@ GeographicBoundingBox::Private::intersection(const Private &otherExtent) const {
             return nullptr;
         }
 
+        // Bail out on longitudes not in [-180,180]. We could probably make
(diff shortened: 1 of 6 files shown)


src/iso19111/metadata.cpp
```diff
@@ -422,6 +422,9 @@ GeographicBoundingBox::Private::intersection(const Private &otherExtent) const {
         }
 
         // Return larger of two parts of the multipolygon
```
jjimenezshaw: Looks like the comment is for the if statement, and I think it is for the code after it. I would add the if statement _before_ the comment.

rouault: indeed. also added a new comment to explain the new check

PR #3754, opened 2023-06-19
Update proj.rst example to modern usage

jidanni:
Here we give the best current version of the same example.

### DIFF
--- a/docs/source/apps/proj.rst
+++ b/docs/source/apps/proj.rst
@@ -192,7 +192,7 @@ The following script
 
 .. code-block:: console
 
-    proj +proj=utm +zone=12 -r <<EOF
+    proj +proj=utm +zone=12 +ellps=GRS80 -r <<EOF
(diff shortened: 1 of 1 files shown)


jidanni:
Yes if I had edited this only once it would have worked out perfectly.

docs/source/apps/proj.rst
```diff
@@ -192,18 +192,18 @@ The following script
 
 .. code-block:: console
 
-    proj +proj=utm +lon_0=112w +ellps=clrk66 -r <<EOF
+    proj +proj=utm +zone=12 -r <<EOF
```
rouault:
```suggestion
    proj +proj=utm +zone=12 +ellps=GRS80 -r <<EOF
```

jidanni: But GRS80 is the default.
No real need to list it.

PR #3755, opened 2023-06-19
Simply eqc.rst example

jidanni:
Use all defaults.

### DIFF
--- a/docs/source/operations/projections/eqc.rst
+++ b/docs/source/operations/projections/eqc.rst
@@ -63,7 +63,7 @@ The following table gives special cases of the cylindrical equidistant projectio
 
 Example using EPSG 32662 (WGS84 Plate Carrée)::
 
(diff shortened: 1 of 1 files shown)


docs/source/operations/projections/eqc.rst
```diff
@@ -63,7 +63,7 @@ The following table gives special cases of the cylindrical equidistant projectio
 
 Example using EPSG 32662 (WGS84 Plate Carrée)::
 
-    $ echo 2 47 | proj +proj=eqc +lat_ts=0 +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +ellps=WGS84 +units=m
+    $ echo 2 47 | proj +proj=eqc
```
rouault: I would let the ellipsoid for rigorous (although there's no difference between GRS80 and WGS84 for that projection which has no ellipsoidal formulation):
```suggestion
    $ echo 2 47 | proj +proj=eqc +ellps=WGS84
```

Issue #829, opened 2018-02-28
GRS80 ellipsoid seems to be equal to WGS84


GwennaelM:
Hi

I'm not an expert in PROJ.4 usage, so I apologize if this issue is not a bug.

As conversions using WGS84 and GRS80 ellipsoids give the same results, it seems their parameters are equals when they should be different (~0.1mm for minor semi-axis).

Here an example:
```
echo "2.5 43.5 10" | cs2cs -f %5.15f +proj=lonlat +ellps=WGS84 +towgs84=0,0,0 +to +proj=lonlat +ellps=GRS80 +towgs84=0,0,0
2.500000000000000	43.500000000000007 10.000000000000000
```

This works fine with Clarke 80:
```
echo "2.5 43.5 10" | cs2cs -f %5.15f +proj=lonlat +ellps=WGS84 +towgs84=0,0,0 +to +proj=lonlat +ellps=clrk80 +towgs84=0,0,0
2.500000000000000	43.503140817107521 63.170659539289773
```

But I don't really understand why I need to add the +towgs84 parameters... My mistake is maybe here. Same example as above but without +towgs84:
```
echo "2.5 43.5 10" | cs2cs -f %5.15f +proj=lonlat +ellps=WGS84 +to +proj=lonlat +ellps=clrk80
2.500000000000000	43.500000000000007 10.000000000000000
```

Tested on v4.9.3.

Thanks

busstoptaktik:
@GwennaelM: By design the ``pj_transform()``, and hence ``cs2cs``, considers WGS84 and GRS80 ellipsoids identical: See the [comment in these lines](https://github.com/OSGeo/proj.4/blob/master/src/pj_transform.c#L560-566) of ``pj_compare_datums``.

So this is the intention in the old API, where the aim of the datum handling code is to make it easy to use, even for people without much knowledge of reference frames and geodesy in general.

The new API (in ``proj.h``) provides much more fine grained control over datum shifts (reference frame transformations), but is also somewhat more demanding wrt. the geodetic knowledge required.

The thing you are attempting can be done using ``cct``, the command line interface to the new API:

    $ echo 2.5 43 10 0 | cct +proj=pipeline +step +proj=cart +ellps=WGS84 +step +inv +proj=cart +ellps=GRS80
    > 2.5000000000   43.0000000009       10.0000        0.0000

Where you see the difference in the last decimal of the latitude.

[...]

PR #3764, opened 2023-06-23
Always use the cleanest examples.

jidanni:
Move repetitive default into text explanation.

P.S., maybe one day even consider changing the
default +ellps=GRS80 to +ellps=WGS84... but no big deal...

### DIFF
--- a/docs/source/apps/proj.rst
+++ b/docs/source/apps/proj.rst
@@ -192,13 +192,14 @@ The following script
 
 .. code-block:: console
 
-    proj +proj=utm +zone=12 +ellps=GRS80 -r <<EOF
+    proj +proj=utm +zone=12 -r <<EOF
     45d15'33.1" 111.5W
     45d15.551666667N -111d30
     +45.25919444444 111d30'000w
     EOF
 
 will perform UTM zone 12 forward projection.
+The default +ellps=GRS80 is used as no +ellps was specified.
 The geographic values of this example are equivalent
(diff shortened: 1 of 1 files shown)


kbevers:
> default +ellps=GRS80 to +ellps=WGS84... but no big deal...

The WGS84 ellipsoid is specific to the WGS84 reference frame of the GPS system - GRS80 is generally the recommended ellipsoid to use, and basically everything else is using it. So no, it shouldn't be the default.

jidanni:
Indeed, we see my suggestion would actually be going backwards 
https://proj.org/en/9.2/usage/differences.html#version-6-0-0

ValeZAA:
>is generally the recommended ellipsoid to use, and basically everything else is using it. 

No, everyone uses correct code, in particular NRO/IGS and Space Force, tiny difference of 0.105 mm in the semi-minor axis matters. Who cares about those others who use wrong GRS 80? The USA spy network uses correct ellipsoid, WGS 84. This reference code that implements IGS20/ITRG2020 GPS uses since 27 Novermber 2022 uses WGS 84: https://github.com/noaa-ngs/HTDP/commit/1dbd6a1513bbfa067fd9f2c977f8c42e5d8cfbc4 

See also new API 
in proj.h makes them different: https://github.com/OSGeo/PROJ/issues/829#issuecomment-369188790

How are you going to support more complex ellipsoinds and soon new standard Height Reference Frame?

PR #3767, opened 2023-06-23
Remove redundant optional default parameters!

jidanni:
We get the exact same answer without the clutter that doesn't make things any more clearer.

No, even the  +ellps=WGS84 becoming GRS80 doesn't affect the answer, and just adds to the clutter.

### DIFF
--- a/docs/source/operations/projections/gall.rst
+++ b/docs/source/operations/projections/gall.rst
@@ -49,12 +49,12 @@ Unlike the Mercator, the Gall shows the poles as lines running across the top an
 
 Example using Gall Stereographic  ::
 
-    $ echo 9 51 | proj +proj=gall +lon_0=0 +x_0=0 +y_0=0 +ellps=WGS84 +units=m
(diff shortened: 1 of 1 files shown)


ValeZAA:
>No, even the +ellps=WGS84 becoming GRS80 doesn't

False. Calculations are different, all assembly that is being executed is different, NAK. Tiny difference of 0.105 mm in the semi-minor axis is important, of course.

jidanni:
(I meant in the answer shown on the page. Yes, different if one turns on more digits.)

ValeZAA:
> (I meant in the answer shown on the page. Yes, different if one turns on more digits.)

No, the calculations are done with high precision and so it will be different assembly opperations. Sure, it prints only some of numbers, but that does not matter.

jidanni:
OK. Anyway, if there is something there that still can be trimmed without making it less clear in the document, that be my guest to trim it. Thanks.

rouault:
already merged

PR #3770, opened 2023-06-23
Remove one repeated default parameter

jidanni:
Shorten example by removing repeated default parameter.


jidanni:
https://github.com/OSGeo/PROJ/pull/3774 is even better.

PR #3771, opened 2023-06-23
Can remove default parameter.

jidanni:
Shorten string by removing redundancies.


jidanni:
Anybody have a look at this?

hobu:
> Anybody have a look at this?

I think these are sitting here because no one wants to go through and gather all of your documentation patches into a single PR. If you can do that it will get merged very quickly.

jidanni:
Many years ago I installed git on Linux.


However these years I just use the GitHub web interface.

Any more tinkering by me will be certain disaster.

rouault:
already merged

PR #3782, opened 2023-06-23
Update vertoffset.rst

jidanni:
GRS80 already implied and documented.


rouault:
@jidanni  Please hold on with all those changes for now, at least until we have some concensus within the maintainers of the project. I personnaly prefer explicit +ellps=GRS80 to be kept . It doesn't hurt.  @kbevers opinion ?

And if we decide to do so, then do all those changes in a single pull request to decrease the "cost" of review.

jidanni:
OK. No more GRS80. But I would say that just burdens the documentation...
And it looks like back in the old days, when things really didn't work without it.

Also I'm just using the web interface, so it is hard to group them.
Also I fix other stuff at the same time.

PR #3790, opened 2023-06-23
Update nzmg.rst

jidanni:
Here I am taking a guess at which grid this is referring to and thus trying to make it clearer.


jidanni:
https://proj.org/en/9.2/operations/projections/ortho.html has another way of mentioning the EPSG number on the page.
> This projection method corresponds to EPSG:...

 Maybe do like that instead.

hobu:
whoops, I was going to merge all these into a single branch first 😒 

I will do so with the rest

PR #3792, opened 2023-06-23
Update tmerc.rst

jidanni:
Trimming defaults reduces horizontal scrolling.

### DIFF
--- a/docs/source/operations/projections/tmerc.rst
+++ b/docs/source/operations/projections/tmerc.rst
@@ -64,12 +64,12 @@ The following table gives special cases of the Transverse Mercator projection.
 
 Example using Gauss-Kruger on Germany area (aka EPSG:31467) ::
 
-    $ echo 9 51 | proj +proj=tmerc +lat_0=0 +lon_0=9 +k_0=1 +x_0=3500000 +y_0=0 +ellps=bessel +units=m
+    $ echo 9 51 | proj +proj=tmerc +lon_0=9 +x_0=3500000 +ellps=bessel
     3500000.00  5651505.56
 
 Example using Gauss Boaga on Italy area (EPSG:3004) ::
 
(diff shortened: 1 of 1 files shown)


hobu:
I causes me a little anxiety to be removing these defaults. Having them in a copy-paste'able example means that users can see what to expect without regard to any default changing with any version. I know these practically never change and maybe this fear is unwarranted.

kbevers:
@hobu It's likely not a big deal to change them, but the change also doesn't bring anything new to the table. It's mostly change for the sake of change and that irks me the wrong way.

@jidanni I appreciate the effort to try to improve the docs but the amount of small and similar PR's are getting out of hand. It's causing more noise than I as a maintainer like I'm inclined to just them all verbatim, simply based on the messy git history they create. I'm happy to consider your suggestions but please present them in a manner that's easy on those of us who maintain this library. That means you need to bundle all your changes in one pull request, with similar edits combined in one commit. I realise that you use the "Edit on GitHub" link from the online docs, which is fine for smaller one off edits, but it's not feasible for this amount of changes. This may involve you familiarizing yourself with git and other tooling used by this project.

hobu:
> It's causing more noise than I as a maintainer like I'm inclined to just them all verbatim, simply based on the messy git history they create. I'm happy to consider your suggestions but please present them in a manner that's easy on those of us who maintain this library. 

All true, but thanks for the attention! Documentation love is appreciated.

One big branch with a slug of doc changes, tweaks, and updates is indeed going to be cleaner to manage, but we the maintainers can do that pretty easily. If you're trying to pad your merged PR stats we might not be so excited 😉

jidanni:
- I used to use git on my home computer, but now have gone back to the web interface.
- My torrent of changes are over, as I have already made all the changes I can think of, so there should be only a few more scattered weeks apart.
- Also you shouldn't approve of more than one of my changes at one time, as the quality of each one varies widely.

PR #3802, opened 2023-06-29
Add stale GitHub action, as replacement for old stale GitHub integration

kbevers:
Previously we had the "stale bot" which posted reminders in inactive issues and PR's. The stale bot used now removed functionality of GitHub and it hasn't worked for about a year. This commit introduces the same functionality based on GitHub Actions. The action setup is grabbed from the QGIS project and adapted to the PROJ project.

With the stale action in place, PR's and issues are marked "stale" after 60 days of inactivity. If 7 more days passes without any activity the issue or PR is closed.


.github/workflows/stale.yml
```diff
@@ -0,0 +1,93 @@
...
+    if: github.repository_owner == 'osgeo'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v8
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-pr-message: >
+          The PROJ project highly values your contribution and would love to see
+          this work merged!
+          Unfortunately this PR has not had any activity in the last 14 days and
```
mwtoews:
```suggestion
          Unfortunately this PR has not had any activity in the last two months and
```
because `days-before-pr-stale: 60`

mwtoews (approved):
Thanks, looks good, only need to adjust the words to match one of the parameters. Here's a link to [actions/stale](https://github.com/actions/stale) for reference.

.github/workflows/stale.yml
kbevers: Thanks. I knew I would have missed at least one of them!

PR #3806, opened 2023-07-02
Update unitconvert.rst

jidanni:
Mention defaults.
I only listed the ones I know.
The rest I didn't list.


jidanni:
Or maybe the defaults are "the units of the previous step, else meters."

PR #3810, opened 2023-07-04
ReadTheDocs: add htmlzip to offline formats; remove ePub

### LINKED ISSUE #3808: Offer Offline HTML documentation download
In 
https://github.com/GenericMappingTools/gmt/issues/7597 we see Proj doesn't have a HTML Offline document set. Only PDF...

We also see a way of mentioning a wget way users could use to get an HTML set.

Maybe the Proj website should also mention such a method. Linux distributions can't always be counted on to include docs. And the user might in fact not even have any distribution installed, and just want the HTML documents, for offline perusal in the North Woods.

mwtoews:
Closes #3808

I know PDFs are not rendered with PRs, but I'm unsure about htmlzip, so it might need to wait until merge to take effect.

Is there any value to keeping ePub? Now is an opportunity to remove it.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -30,8 +30,8 @@ build:
     - libcurl4-openssl-dev
 
 formats:
+  - htmlzip
   - pdf
-  - epub
 
 conda:
   environment: docs/environment.yml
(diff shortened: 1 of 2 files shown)


kbevers:
> Is there any value to keeping ePub? Now is an opportunity to remove it.

I've got no idea what it is, so for me it's a +1 on removing it.

PR #3811, opened 2023-07-05
plot.py: rewrite using shapely>=2

mwtoews:
This PR rewrites plot.py using shapely 2.0 features, which should enable building docs to be much faster.

Also, the source code has been "modernized" via pyupgrade / isort / black.

The images look similar, although I haven't quantified this beyond a few visual checks.

Some unused scripts have been cleaned up too.


kbevers (approved):
a change for the better!

mwtoews:
On RTD, running plot.py is between 5-7 times faster, which should drop the whole doc build down considerably. I'd like to backport this just to 9.2, but anyone can object by closing the automated PR.

Issue #3819, opened 2023-07-11
concatenated operations


barry-gallagher:
Hello,

We are making a lot of vertical datums, crs and transforms which seem to work thanks to #3616 (and #3617).  We wanted to add some concatenated operations and those seem to work unless I add a geographic 3d to geographic 3d step at the end.  Is the problem in trying to add a 3d transform at the end of what was mostly vertical, related to our custom authority or missing a step or method code?  Thanks in advance for any help or pointers on what I have done wrong, I've attached my prototype [proj.db.zip](https://github.com/OSGeo/PROJ/files/12017707/proj.db.zip).

For example, we have tidal datums of Mean Lower Low Water (MLLW) and Local Mean Sea Level (LMSL) and a transform between them.  We also have a transform from LMSL to Geoid09.  I've tried to create two concatenated operations, one to go from MLLW to EPSG:4893 NAD83(NSRS2007) and the other to go from MLLW to EPSG:6319 NAD83(2011).  The first one works but the second fails.  

Both transforms start with the same three steps.
1) MLLW (NOAA:752) -> LMSL (NOAA:348) using grid_transformation NOAA:754
2) LMSL (NOAA:348) -> Geoid09(Tried both EPSG:5500 and EPSG:5703) using grid_transformation NOAA:350
3) Geoid09 -> NAD83(NSRS2007) (EPSG:4893) using grid_transform EPSG:9173

The second transform adds one more step.
4) NAD83(NSRS2007) (EPSG:4893) to NAD83(2011) (EPSG:6319) using grid_transform EPSG:8559

Running from our tidal EPSG:4759+NOAA:752 NAD83(NSRS2007) 2D+MLLW to EPSG:4893 NAD83(NSRS2007) works
Trying to extend to NAD83(2011) fails.
``` 
projinfo
Rel. 9.2.1, June 1st, 2023


cs2cs "EPSG:4759+NOAA:752" "EPSG:4893" < coords.txt
26d36'N 97dW -22.923


cs2cs "EPSG:4759+NOAA:752" "EPSG:6319" < coords.txt
Rel. 9.2.1, June 1st, 2023
<cs2cs>:
cannot initialize transformation
cause: (null)
program abnormally terminated
```

[...]

PR #3820, opened 2023-07-11
Various fixes related to #3819

rouault:


src/iso19111/factory.cpp
```diff
@@ -7228,24 +7259,36 @@ AuthorityFactory::createFromCRSCodesWithIntermediates(
...
-        listTmp.emplace_back(
-            operation::ConcatenatedOperation::createComputeMetadata(
-                {op1, op2->inverse()}, false));
+            listTmp.emplace_back(
+                operation::ConcatenatedOperation::createComputeMetadata(
+                    {op1, op2->inverse()}, false));
+        } catch (const std::exception &e) {
+            // Mostly for debugging purposes when using an inconsistent
+            // database
+            if (getenv("PROJ_IGNORE_INSTANCIATION_ERRORS")) {
```
busstoptaktik: I suggest to use the English spelling "instantiation" rather than the French(?) "instaciation"

busstoptaktik (commented):
Just nitpicking on some spelling

src/iso19111/factory.cpp
rouault: thanks. fixed

PR #3825, opened 2023-07-16
Add option in proj CLI to use a CRS

### LINKED ISSUE #3824: proj_factors produces wrong results with EPSG:2222 (ft)
Looking for the behaviour of proj_factors with with a CRS in feet, I found that it returns a huge value

#### Example of problem
I guess there is no option in `proj` CLI to pass a CRS identifier. So I modified the file `gie_self_tests.cpp`. Sorry for the hack.

```diff
diff --git a/test/unit/gie_self_tests.cpp b/test/unit/gie_self_tests.cpp
index 434822b0..9d7ef11e 100644
--- a/test/unit/gie_self_tests.cpp
+++ b/test/unit/gie_self_tests.cpp
@@ -509,9 +509,11 @@ TEST(gie, info_functions) {
 
     // Test with a projected CRS
     {
-        P = proj_create(PJ_DEFAULT_CTX, "EPSG:3395");
-
+        P = proj_create(PJ_DEFAULT_CTX, "EPSG:2222");
+        a.lp.lam = proj_torad(-110.1666);
+        a.lp.phi = proj_torad(31);
         const auto factors2 = proj_factors(P, a);
+        std::cout << "******" << factors2.meridional_scale << std::endl;
 
         EXPECT_EQ(factors.angular_distortion, factors2.angular_distortion);
         EXPECT_EQ(factors.meridian_parallel_angle,

```

With this input (using a system in ft) the meridional_scale is `3.51837e+07`. However, with EPSG:26948 (the equivalent CRS in m), the escale is `0.9999` (the expected value)

#### Problem description
The huge meridional_scale value makes no sense. The other factors are nonsense as well.
Debuging the code, I saw that with EPSG:2222 (and NOT with EPSG:26948) it gets into the function `fwd_prepare` in `fwd.cpp`. In the line 108
```c
coo.lp.lam = (coo.lp.lam - P->from_greenwich) - P->lam0;
```
it changes `coo.lp.lam` from a small value (near zero) to something like 1.9 radians.

I guess that the problem is that it was previously "adjusted" in factors.cpp, line 65
```
lp.lam -= P->lam0;
```
Reached this point I do not know how to fix it without breaking anything else.

#### Expected Output
`proj_factors` produce meaningful values for any CRS, including CRS in feet.

#### Environment Information
 - Operation System Information: Ubuntu 22.04

#### Installation method
 - compiled

jjimenezshaw:
Following #3824 , to easily apply `proj` on the projection of a projected CRS, this PR adds the option `-C`.
It follows the same logic as in function `proj_factors`, that is to compute the projection from the CRS to the geographic underlying CRS.

This PR does NOT fix #3824.

Example of usage:

```
echo -110.16666 31 | proj -C EPSG:26948 -V 
#Transverse Mercator
#	Cyl, Sph&Ell
#	approx
# +proj=tmerc +lat_0=31 +lon_0=-110.166666666667 +k=0.9999 +x_0=213360 +y_0=0
# +ellps=GRS80
#Final Earth figure: ellipsoid
#  Major axis (a): 6378137.000
#  1/flattening: 298.257222
#  squared eccentricity: 0.006694380023
Longitude: 110d9'59.976"W [ -110.16666 ]
Latitude:  31dN [ 31 ]
Easting (x):   213360.64
Northing (y):  0.00
Meridian scale (h) : 0.99990000  ( -0.01 % error )
Parallel scale (k) : 0.99990000  ( -0.01 % error )
Areal scale (s):     0.99980001  ( -0.02 % error )
Angular distortion (w): 0.000
Meridian/Parallel angle: 90.00000
Convergence : 0d0'0.012" [ 0.00000343 ]
Max-min (Tissot axis a-b) scale error: 0.99990 0.99990
```

 - [x] Tests added


jmckenna:
thanks @jjimenezshaw very useful!

jjimenezshaw:
BTW, I chose `-C` just due to "CRS". If there is a more consistent option among the CLI programs, I can use a different one. Just let me know.

busstoptaktik:
Perhaps skip the flag and just parse whether it looks like a CRS, rather than a PROJ definition?

rouault:
> Perhaps skip the flag and just parse whether it looks like a CRS, rather than a PROJ definition?

I was just going to suggest the same thing.  cs2cs has logic to be able to deal with things like "cs2cs EPSG:foo EPSG:bar" or "cs2cs +proj=... +to + proj=...." automatically.

Here I believe that the criterion would be "argvVector.empty() && eargc >= 1", and consider eargv[0] as the CRS code instead of being the first file

rouault:
Hum while fixing #3824, I'm wondering what we should do if allowing to pass CRS by codes when they have northing/easting axis order:
- should we normalize to easting/northing order ?  That way in verbose mode, the "Easting (x)" and "Northing (y)" labels will be OK
- or should we output with axis axis order of the CRS ? If so, then we should switch the values in verbose mode.
At the very least this will deserve documentation.

Note that the proj utility didn't honour the (now deprecated) +axis= modifier which was a cs2cs / pj_transform() thing and ignored by the pj_fwd() logic .

jjimenezshaw:
> what we should do if allowing to pass CRS by codes when they have northing/easting axis order

I was making myself the same question. Somehow I have a preference to keep the CRS order (to be consistent with the printed pipeline). But it is not very strong opinion.

See that the problem is not only with CRS codes (or wkt). If you pass the pipeline with the units or axis order (or anything that is not the "standar" way probably) #3824 still fails (probably you realized that already).

jjimenezshaw:
In case it is not clear: the CRS definition can be anything that works with "from user input". Can be a code, a WKT, a PROJJSON, etc.
I want to study later how it works with a derived projected (enabling that case in the if statement).

jjimenezshaw:
I applied the same criterion as in #3826 for the proj_factors calls (not for the projection calls).
I am still thinking about the axis order. Is there any function to easily know if the `PJ*` object has swapped axis?

rouault:
> Is there any function to easily know if the `PJ*` object has swapped axis?

With the C++ API: CRS::mustAxisOrderBeSwitchedForVisualization()

jjimenezshaw:
> mustAxisOrderBeSwitchedForVisualization

Unfortunately the linker does not find it. Anyhow, the check was easy with the C++ object.

The output of a northing-easting CRS is something like this:
```
echo 9 0 | ./bin/proj EPSG:3044 -V
#Transformation pipeline manager
# +proj=pipeline +step +proj=utm +zone=32 +ellps=GRS80 +step +proj=axisswap
# +order=2,1
#Final Earth figure: ellipsoid
#  Major axis (a): 6378137.000
#  1/flattening: 298.257222
#  squared eccentricity: 0.006694380023
Longitude: 9dE [ 9 ]
Latitude:  0dN [ 0 ]
Northing (y):  0.00
Easting (x):   500000.00
Meridian scale (h) : 0.99960000  ( -0.04 % error )
Parallel scale (k) : 0.99960000  ( -0.04 % error )
Areal scale (s):     0.99920016  ( -0.07998 % error )
Angular distortion (w): 0.000
Meridian/Parallel angle: 90.00000
Convergence : 0d [ -0.00000000 ]
Max-min (Tissot axis a-b) scale error: 0.99960 0.99960
```

jjimenezshaw:
Is there anything missing in this PR? (behavior, feature or documentation)

docs/source/apps/proj.rst
```diff
@@ -172,6 +172,13 @@ also used for supporting files like datum shift files.
     Usage of *+opt* varies with projection and for a complete description
     consult the :ref:`projection pages <projections>`.
 
+.. versionadded:: 9.3.0
+
+    *{crs}* is one of the possibilities accepted by proj_create(), provided it
```
kbevers:
```suggestion
    *{crs}* is one of the possibilities accepted by :c::func:`proj_create()`, provided it
```

kbevers: Further down the page an example using proj with a CRS would be nice as well

kbevers (approved):
Looks good to me, but a few minor things should be improved in the docs

docs/source/apps/proj.rst
jjimenezshaw: Done

PR #3826, opened 2023-07-17
proj_factors(): make it work with projected CRS with non-metre unit and/or northing/easting axis order (fixes #3824)

### LINKED ISSUE #3824: proj_factors produces wrong results with EPSG:2222 (ft)

rouault:


jjimenezshaw:
thanks @rouault

PR #3834, opened 2023-07-19
Avoid C++ exceptions to be thrown (and caught) when parsing strings like '+proj=longlat +datum=WGS84 +type=crs'

rouault:
This makes it more friendly with debuggers under Windows, and should help workarounding https://github.com/MapServer/MapServer/issues/6915 (note however that the issue in that ticket is more a 'system level' issue than a PROJ bug)


jmckenna:
Tested with MSVC and works great.

PR #3835, opened 2023-07-19
PROJ_DEBUG: make ON an alias of 2, and OFF of 1 (fixes #3832) 

### LINKED ISSUE #3832: PROJ_DEBUG=ON?
A minor nit, but `PROJ_DEBUG`'s behavior isn't aligned with things like `CPL_DEBUG` for the situation of `PROJ_DEBUG=ON`. Instead, you should set `PROJ_DEBUG=2` to set a verbosity level AND debug output. 

The behavior is fine, but I would like to propose that `PROJ_DEBUG=ON` be accepted and it default to `PROJ_DEBUG=2` verbosity level.

rouault:
and emits a stderr message when the value specified isn't understood

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -35,6 +35,7 @@
 #include <new>
 
 #include "filemanager.hpp"
+#include "proj/internal/internal.hpp"
 #include "proj/internal/io_internal.hpp"
 #include "proj_experimental.h"
 #include "proj_internal.h"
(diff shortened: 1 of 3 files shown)


src/ctx.cpp
```diff
@@ -88,11 +89,25 @@ pj_ctx pj_ctx::createDefault() {
...
     if (projDebug != nullptr) {
-        const int debugLevel = atoi(projDebug);
-        if (debugLevel >= -PJ_LOG_TRACE)
-            ctx.debug_level = debugLevel;
-        else
-            ctx.debug_level = PJ_LOG_TRACE;
+        if (NS_PROJ::internal::ci_equal(projDebug, "ON")) {
+            ctx.debug_level = PJ_LOG_DEBUG;
+        } else if (NS_PROJ::internal::ci_equal(projDebug, "OFF")) {
+            ctx.debug_level = PJ_LOG_ERROR;
```
jjimenezshaw, line 95: In the documentation file it says
"and ``OFF`` as an alias for ``0``."
Should this be `PJ_LOG_NONE`? Or is doc file above wrong?
```suggestion
            ctx.debug_level = PJ_LOG_NONE;
```

rouault: > Or is doc file above wrong?

ah yes, I intended OFF to be as the default, that is hide all debug and traces messages, but keep errors. Fixing the odc

Issue #2640, opened 2021-04-04
Add --version


jidanni:
As there is no way to get the version, mentioned on the proj, cs2cs, etc. man pages. The current method is to trigger an error:
```
$ proj //////
Rel. 7.2.1, January 1st, 2021
...
```

So a better solution would be
```
$ proj --version
proj 7.2.1, January 1st, 2021
```

Note the program name is better than "Rel." too.

PR #3836, opened 2023-07-20
CLI: add --version long argument to return PROJ version

### LINKED ISSUE #3980: Add `--version` CLI arg to all commands (e.g., `proj`, `projsync`, etc.)
It appears that, at least as of v8 and maybe v9, these tools don't have a `--version` flag. They should. It makes checking installs and paths and such way easier, especially when working with systems like Docker.

lbartoletti:
The proposed `--version` option will enable users to easily retrieve the application's version without having to refer to documentation or resort to external means. This feature is essential for testing various behaviors and fulfilling a specific user requirement.

I have tried to adhere to the existing code conventions while implementing this feature. As of now, the changes have been applied only to one utility, but if it's OK, I'll extend the same modifications to other apps.

cc @guts

 - [ ] Closes #3980, #2640
 - [ ] Tests added


kbevers:
I think this is a good idea. Can you please include an example here? Also, remember to adjust the docs as well so that this feature isn't a secret.

rouault:
The gie and cct utilities already implement --version with the following output:
```
$ bin/gie --version
gie: Rel. 9.3.0, September 1st, 2023
$ bin/cct --version
cct: Rel. 9.3.0, September 1st, 2023
```
I guess with your proposed change, you'd reduce this to ``9.3.0``
I don't think that change would be critically breaking. Perhaps @busstoptaktik has an opinion regarding this

You could also change proj_info() in src/4D_api.cpp to change "info.version = version;" to "info.version = pj_get_version()"

kbevers:
I wouldn't mind it if the date disappeared from the `--version` outputs of the apps. It's an annoying thing to maintain in releases.

rouault:
> It's an annoying thing to maintain in releases.

proj_info() reports 2 fields  ".version=X.Y.Z" and ".release=Rel. X.Y.Z, date"

Guts:
Thanks @lbartoletti!

To give context, I need to identify the installed PROJ version to adapt some parts of my logic. After some research, I've ended on this:

```python
import re
import subprocess

proj_output = check_output(["proj"], stderr=STDOUT, text=True)
version_match = re.search(r"Rel\. ([0-9.]+)", proj_output):
if version_match:
    print(version_match.group(1))
```

But I did not know cie and cct!

See also:

- [thread on Mastodon](https://mapstodon.space/@geojulien/110741460535184493)
- [article on geotribu](https://preview-pullrequest-954--geotribu-preprod.netlify.app/articles/2023/2023-07-25_python-obtenir-la-version-de-proj-avec-gdal-pyproj-binaire/#avec-les-bindings-gdal) (preview) :fr:

busstoptaktik:
> Perhaps @busstoptaktik has an opinion regarding this

@rouault this is one of the few things I cannot talk myself into having an opinion of 🙂 - but thanks for pinging me on the subject!

snowman2:
There are quite a few implementations for this depending on the program. One option to consider could be to provide a simple version (easier to parse for the use case of this PR) and a more verbose version like Python does.

```bash
$ python --version
Python 3.10.10
$ python -V
Python 3.10.10
$ python -VV
Python 3.10.10 | packaged by conda-forge | (main, Mar 24 2023, 20:08:06) [GCC 11.3.0]
```

snowman2:
Side note:
```
$ proj
```
```
Rel. 9.1.1, December 1st, 2022
usage: proj [-bdeEfiIlmorsStTvVwW [args]] [+opt[=arg] ...] [file ...]
```

snowman2:
[Code](https://github.com/pyproj4/pyproj/blob/1452ba404be58c14a6b64d4551c320022f5aafcf/setup.py#L33-L53) pyproj uses to get the `proj` version.

jmckenna:
This is so needed, thank-you @lbartoletti for pushing this.

jmckenna:
(initially reported at #2640 )

lbartoletti:
> There are quite a few implementations for this depending on the program. One option to consider could be to provide a simple version (easier to parse for the use case of this PR) and a more verbose version like Python does.
> 
> ```shell
> $ python --version
> Python 3.10.10
> $ python -V
> Python 3.10.10
> $ python -VV
> Python 3.10.10 | packaged by conda-forge | (main, Mar 24 2023, 20:08:06) [GCC 11.3.0]
> ```

Indeed, `my_soft --version` should return `my_soft X.Y.Z` and not only `X.Y.Z`. Long version, could the actual one aka `pg_get_release()` : `app_name: Rel. 9.3.0, September 1st, 2023`

However, in our case, -VV is not possible, since it's used for verbosity in proj app.

May I suggest:

- app_name --version: `app_name X.Y.Z`, so `proj --version`: `proj 9.3.0`
- app_name --release (or --long-version?): `app_name: Rel. 9.3.0, September 1st, 2023`
- app_name --short-version (or --version-number? --version-semver?): `9.3.0`

rouault:
Personally, I would just stick with below for the sake of simplicity (most more popular utilities than the PROJ ones have just --version, so it might be over-engineered to offer 3 variants for PROJ...):

> * app_name --version: `app_name X.Y.Z`, so `proj --version`: `proj 9.3.0`

The long version with the release date doesn't bring that much, given the release date is something that is a 1:1 match with the version number.

jmckenna:
As a packager, this release date causes confusion to users (the date is often in the future, which causes confusion for the user, and grief for the packager).   It would be good to not include this, in any utility.

kbevers:
> It would be good to not include this, in any utility.

Can someone give an educated guess as to what the impact of changing the output of `cct --version` and `gie --version` is? Have people been relying on these two to get the version number in a simpler way than using `proj`? If we think a limited number of users rely on this I am happy to change it but we shouldn't do it if it forces a lot of people to change existing workflows.

lbartoletti:
Not staled

Guts:
Still interested in this feature and homogenization work.

kbevers:
I agree that this is a sane and useful change but since I haven't received any feedback to this

> Can someone give an educated guess as to what the impact of changing the output of cct --version and gie --version is? Have people been relying on these two to get the version number in a simpler way than using proj? If we think a limited number of users rely on this I am happy to change it but we shouldn't do it if it forces a lot of people to change existing workflows.

I can't really make a decision on how to proceed. For now I am inclined to not merge this PR and instead streamline the version options of the various CLI utilities in a future 10.0 release.

jmckenna:
@kbevers can you leave this PR open and possibly add a label to it of "10.0 release" ?  I am sure that @lbartoletti can adapt the changes to anything that you need for 10.0        

Thanks for all of this effort @lbartoletti, it is very useful to the community.

DeflateAwning:
Well that's funny. Just opened a feature request issue for this exact feature. Happy to see it's implemented! 

Fixes #3980

PR #3859, opened 2023-08-26
Docker: update base image to ubuntu:22.04

jjimenezshaw:
Upgrade the Ubuntu-based docker images to the latest Ubuntu LTS, with end of standard support in April 2027 and EOL 2032.

Specific motivation for Dockerfile is to use it with `pyproj` that requires python 3.9, which is no longer supported upstream and by many packages in use. Ubuntu 22.04 includes support for Python 3.10.

Images and docs (html) were built locally, and look fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ##
 # OSGeo/PROJ
 
-FROM ubuntu:20.04 as builder
+FROM ubuntu:22.04 as builder
 
 MAINTAINER Howard Butler <howard@hobu.co>
 
@@ -27,7 +27,7 @@ RUN cd /PROJ \
 
 
 
-FROM ubuntu:20.04 as runner
+FROM ubuntu:22.04 as runner
 
 RUN date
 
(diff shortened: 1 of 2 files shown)


jjimenezshaw:
fyi @snowman2

jjimenezshaw:
I set `[ci skip]` inspired by #3043 , but I do not know if there is any way to test the changes (apart from what I tested in my computer)

jjimenezshaw:
Uhmmm, there are some failing tests in master after merging this PR. Sorry about that.
One in `proj-docs Container` that should be related, but I could not reproduce during the development.
Other tree `Conda`, `Windows build` and ` MacOS build` that I do not understand how are related to the change in docker. They are about a serialization in JSON.

~Should I do a PR undoing the changes of this PR?~ There is a `Revert` button. Nice.

jjimenezshaw:
> One in `proj-docs Container` that should be related, but I could not reproduce during the development.

If you run `docker build .` twice, the first fails but the second it works. I changed something... and the second time it worked!

That's not good, and it hides a problem. The option `--no-cache` ensures that you build from scratch.

Anyhow I do not understand what is failing exactly in the line `python3 -m pip install ...`. The error message is not clear to me.

PR #3861, opened 2023-08-27
Fix doc docker ubuntu 22.04

jjimenezshaw:
Trying to fix #3859 (the doc build part)

It removes `awscli` python dependency that was producing some problems. It also limits the version of `sphinxcontrib-spelling`.


kbevers:
@jjimenezshaw are you happy with this PR as is? Checks seem to be fine so I reckon all is good :-)

jjimenezshaw:
> @jjimenezshaw are you happy with this PR as is? Checks seem to be fine so I reckon all is good :-)

I am happy, but checks are not running the `proj-docs Container` job in the `Docker` github action that was failing. I do not know why.

kbevers:
I don't understand it either. As far as I can tell there's no reason for the GitHub Actions to not run in this case. I'll merge it and believe all checks should run against master then.

PR #3867, opened 2023-08-31
ITRF2008: fix wrong sign for 'dry' parameter of EURA and EURA_T

rouault:


jjimenezshaw:
Hi @rouault , how was the impact of this bug? When is this Helmert transformation used? Thanks.

rouault:
> how was the impact of this bug?

only when using "+init=ITRF2008:EURA"

Issue #2644, opened 2021-04-04
geod is lat,lon; proj is lon,lat


jidanni:
proj's default output order is lon,lat:
```
$ echo 121 24|proj +proj=tmerc|proj +proj=tmerc -I
121dE   24dN
```
proj also has
```
       -r     This  options reverses the order of the expected input from lon‐
              gitude-latitude or x-y to latitude-longitude or y-x.

       -s     This options reverses the order of the output from x-y or longi‐
              tude-latitude to y-x or latitude-longitude.
```

geod's **only** input order is the opposite: lat,lon.

Nor does geod have any switches like proj, to deal with even proj's
default output order. So one needs proj -s...

One would think programs packaged together would be more coordinated.

(Yes we can use perl etc. if needed to fix the order before piping it
into geod.) 

Anyway perhaps acknowledge the situation on the geod man page.

jidanni:
Doesn't anybody have any opinions?

jidanni:
By the way, don't add a -r to geod, as it would then be the opposite of proj's.
Add a --lola, for longitude/latitude.

busstoptaktik:
> Doesn't anybody have any opinions?

It's hard to argue against *historical reasons*.

We could correct the unconventional input order of `proj` to fit the conventional one of `geod`, but it would probably be quite annoying for a lot of people using `proj` the way it was originally conceived by Jerry Evenden, i.e. as a Unix style filter, typically embedded deeply into the dark caves of shell scripts, not touched by human hands for decades.

One (unrecommended) way to resolve this would be to make `proj` default to the conventional input order (i.e. as if invoked as `proj -r`) if invoked as *some other name* (the anadrome `jorp` would be the logical choice).

This would mimic exactly the existing case, where `invproj` is just `proj` behaving as `proj -I`, if `argv[0]` matches the string constant **"invproj"**.

jidanni:
Definitely don't rip up the defaults,
of any program, because that would break people's scripts, and create a loss of trust that future scripts will continue to work etc.

So all that needs fixing, is adding a little switch to geod.

The switch,  which should **not** be called -r, would simply add the ability to geod to accept lon lat.

jidanni:
(In fact, minus r could be added to geod, but it should have parentheses there on the man page saying default.)

And then proj, and cs2cs, could have the same new option added as what is coming to geod, with parentheses on their Man pages saying default.

jidanni:
[...]

PR #3895, opened 2023-09-20
Update faq.rst to mention geod

jidanni:
See https://github.com/OSGeo/PROJ/issues/2644

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -135,7 +135,9 @@ input ordered as latitude/longitude  (typically with the EPSG dataset), however,
 internally PROJ expects an longitude/latitude ordering for all projections. This
 is generally hidden for users but in a few cases it is exposed at the surface
 level of PROJ, most prominently in the :program:`proj` utility which expects
-longitude/latitude ordering of input date (unless :option:`proj -r` is used).
+longitude/latitude ordering of input data (unless :option:`proj -r` is used).
+(And `geod`, which on the contrary expects latitude/longetude, and lacks any
+`-r` style workarounds.)
 
 In case of doubt about the axis order of a specific CRS :program:`projinfo` is
 able to provide an answer. Simply look up the CRS and examine the axis specification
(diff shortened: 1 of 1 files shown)


docs/source/faq.rst
```diff
@@ -135,7 +135,9 @@ input ordered as latitude/longitude  (typically with the EPSG dataset), however,
 internally PROJ expects an longitude/latitude ordering for all projections. This
 is generally hidden for users but in a few cases it is exposed at the surface
 level of PROJ, most prominently in the :program:`proj` utility which expects
-longitude/latitude ordering of input date (unless :option:`proj -r` is used).
+longitude/latitude ordering of input data (unless :option:`proj -r` is used).
+(And `geod`, which on the contrary expects latitude/longetude, and lacks any
```
busstoptaktik: \`geod\` should probably be :program:\`geod\`.

Also, putting an entire sentence in parenthesis looks strange, especially as it bumps oddly against the parenthesis at the end of the previous sentence. I suggest something like
```rst
... (unless :option:`proj -r` is used), while :program:`geod`,
on the contrary, expects latitude/longitude, and does not support the `-r` option.
```

Also note the correction longetude -> longitude

busstoptaktik (changes requested):
Needs some linguistic TLC

jidanni:
I'm very grateful for anybody to take this anywhere they want to.

Regarding minus r. Note that if such an option was added to geod, it should be a no-op. And then a different option added to accept lon lat.

And then after adding these options, the article might just complain about contrasting defaults, without mentioning  lack of options.

jidanni:
In https://github.com/OSGeo/PROJ/issues/2644#issuecomment-1731100116 I mention a scheme, where adding options to all the programs offered, could wallpaper over the worry about different defaults for different programs (that the PROJect distributes... Internal contradictions)

In fact the FAQ article could mention this scheme as a cherry note.

Of course implementing it: I'll leave that up to you guys :-)

docs/source/faq.rst
rouault:
```suggestion
(And :program:`geod`, which on the contrary expects latitude/longitude, and lacks any
```

PR #3906, opened 2023-09-27
Update lcc.rst

jidanni:
I am sorry.
I don't know how to eliminate the horizontal scrolling. I looked at https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#grid-tables and cannot figure it out.
The table is causing horizontal scrolling, instead of just doing the `<P>` formatting.
Maybe one can put the P tags right in to the MarkDown text?

### DIFF
--- a/docs/source/operations/projections/lcc.rst
+++ b/docs/source/operations/projections/lcc.rst
@@ -20,8 +20,10 @@ justifies adding this otherwise non-standard projection.
 +---------------------+----------------------------------------------------------+
 | **Classification**  | Conformal conic                                          |
 +---------------------+----------------------------------------------------------+
-| **Available forms** | Forward and inverse, spherical and ellipsoidal          .|
+| **Available forms** | Forward and inverse, spherical and ellipsoidal.          |
+|                     |                                                          |
 |                     | One or two standard parallels (1SP and 2SP).             |
+|                     |                                                          |
 |                     | "LCC 2SP Michigan" form can be used by setting           |
 |                     | ``+k_0`` parameter to specify ellipsoid scale.           |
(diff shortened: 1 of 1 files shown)


mwtoews:
I wonder if [field lists](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#field-lists) would work better than tables for these sections. It could look something like this:
```rst
:Classification: Conformal conic
:Available forms: - Forward and inverse, spherical and ellipsoidal.
                  - One or two standard parallels (1SP and 2SP).
                  - "LCC 2SP Michigan" form can be used by setting
                    ``+k_0`` parameter to specify ellipsoid scale.
:Defined area: Best for regions predominantly east–west in extent and
               located in the middle north or south latitudes.
:Alias: lcc
:Domain: 2D
:Input type: Geodetic coordinates
:Output type: Projected coordinates
```

mwtoews:
_Update:_ Here is the table from this PR:
Screenshot from 2023-09-28 22-10-24

And here is the field lists example:
Screenshot from 2023-09-28 22-10-53

The downside is that the PDF formatting is much longer, as it uses much more vertical spacing.

kbevers:
> I wonder if [field lists](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#field-lists) would work better than tables for these sections

I think it looks like a good solution. I believe I came up with the tables way back when. They were a simple way of organising information in a fairly standardised way but I agree that they don't look particularly good where there are lots of text.

jidanni:
Go for it, dudes.
All I know is I'm used to Unix man pages: no horizontal scrolling.

jidanni:
Too bad.

Issue #3922, opened 2023-10-16
Inconsistent results on ppc64le on conda-forge with gcc 12


djhoese:
#### Example of problem

First example works fine. Second transformation does not.

```bash
$ echo 0 0 | cs2cs EPSG:3034 +to EPSG:3035
363301.33       546152.98 0.00
$ echo 0 0 | cs2cs EPSG:3034 +to EPSG:3035
363301.33       546152.98 0.00
$ echo 569704.57 4269024.67 | cs2cs EPSG:26915 +to EPSG:26715
589057.58       -3.54 0.00
$ echo 569704.57 4269024.67 | cs2cs EPSG:26915 +to EPSG:26715
569722.40       4268814.27 0.00
```

#### Problem description

Reference issue: https://github.com/conda-forge/proj.4-feedstock/pull/137

As mentioned in the many issues related to this on the conda-forge feedstock repositories, there seems to be an issue when some transformations happen, but so far only seems to effect ppc64le systems (perhaps the only little endian platform tested on conda-forge). This was originally discovered with pyproj where ppc64le builds were added, but a single test (yes, just one) would fail with very incorrect results. However, it only seems to happen on the first execution of this command and only with specific CRSes (UTM EPSG codes so far).

Note in the above example how the first non-UTM transformation gives the same results the first and second time. The second transformation with the 26915 and 26715 EPSG codes produces an incorrect result in the first execution and then gives valid result in the second instance.

Note that I've tested this with running a `proj` command before the `cs2cs` and `cs2cs` still has the same behavior. I've also done it with `cs2cs` with no arguments to show the usage/help message and the same behavior. In the above example running `cs2cs` with the first transformation also doesn't seem to "initialize" something that the second set of transformations are dependent on.

So far I've only tested this with conda-forge's build system (local docker-based emulated builds) mainly because getting ppc64le emulated builds working is a pain otherwise.

#### Expected Output

The same result both times:

```
$ echo 569704.57 4269024.67 | cs2cs EPSG:26915 +to EPSG:26715
569722.40       4268814.27 0.00
$ echo 569704.57 4269024.67 | cs2cs EPSG:26915 +to EPSG:26715
569722.40       4268814.27 0.00
```

#### Environment Information
 - PROJ version (`proj`): Seen with 9.3.0 and 9.2.1. Haven't tested beyond that.
 - Operation System Information: Power PC 64-bit Little Endian (ppc64le)

#### Installation method
- conda with conda-forge

[...]

PR #3923, opened 2023-10-16
[DO NOT MERGE ME] Test/debug #3922

rouault:


djhoese:
Does this format of tests start a new process and/or PROJ context for each test? Or is there a place to define shell-level `cs2cs` tests in the PROJ test system? As mentioned in the related issue I still see issues when run on travis and I think I'm running natively.

rouault:
> Does this format of tests start a new process and/or PROJ context for each test?

yes, as I've written it, cs2cs is restarted between the 2 attempts.

PR #3936, opened 2023-10-30
SSL_OPTIONS: set SSL options on Curl library

ollira:
I would like to suggest a build-time option for adjusting Curl SSL options on PROJ.  The default Curl SSL options appear to be more pedantic on certificate revocation checks than web browsers are. It could be discussed, if fail-hard checking is needed, when grid data downloads are concerned. See https://en.wikipedia.org/wiki/Certificate_revocation for some general info.

PROJ has the environment variable PROJ_UNSAFE_SSL for turning off all SSL checks. However, I would like to be able to use CURLSSLOPT_NO_REVOKE option in some environments, and keep all the other SSL checks enabled.

After the code change suggested in this pull request, CURLSSLOPT_NO_REVOKE could be selected at build-time by this kind of addition to PROJ cmake command:
  -DCMAKE_CXX_FLAGS="-DSSL_OPTIONS=2"

Curl SSL options are described in https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,6 +228,14 @@ if(ENABLE_CURL)
         INTERFACE_LINK_LIBRARIES      "${CURL_LIBRARIES}"
       )
     endif()
+
+    # Curl SSL options are described in
+    #   https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html
+    #set(CURLSSLOPT_NO_REVOKE 2)
+    #set(SSL_OPTIONS ${CURLSSLOPT_NO_REVOKE})
+    #add_compile_definitions(SSL_OPTIONS=${SSL_OPTIONS})
+    #mark_as_advanced(SSL_OPTIONS)
+
   else()
     message(SEND_ERROR "curl dependency not found!")
   endif()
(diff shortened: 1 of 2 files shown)


rouault:
I believe it would be desirable to have a (mark_as_advanced) CMake option in CMakeLists.txt to enable this, as a minimum documentation for that functionality

ollira:
I added SSL_OPTIONS example to CMakeLists.txt (commented out).  If there is a better way to present numeric options in cmake, let me know.

CMakeLists.txt
```diff
@@ -228,6 +228,14 @@ if(ENABLE_CURL)
         INTERFACE_LINK_LIBRARIES      "${CURL_LIBRARIES}"
       )
     endif()
+
+    # Curl SSL options are described in
+    #   https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html
+    #set(CURLSSLOPT_NO_REVOKE 2)
+    #set(SSL_OPTIONS ${CURLSSLOPT_NO_REVOKE})
+    #add_compile_definitions(SSL_OPTIONS=${SSL_OPTIONS})
+    #mark_as_advanced(SSL_OPTIONS)
```
rouault:
```suggestion
```

PR #3949, opened 2023-11-10
proj_factor: fix when input is a compound CRS of a projected CRS, and…

rouault:
… a projected CRS whose datum has a non-Greenwich prime meridian


jjimenezshaw:
Should it be applied also in `apps/proj.cpp`?

rouault:
> Should it be applied also in `apps/proj.cpp`?

good catch, commit added

PR #3966, opened 2023-11-24
DatabaDatabase: update to EPSG 10.098

rouault:


kbevers:
This is good timing. I plan on doing 9.3.1RC1 later today.

PR #3973, opened 2023-11-30
Database: update to EPSG 11.001, and update PROJ_DATA.VERSION to 1.17

rouault:
@kbevers / @busstoptaktik  This update adds records for dvr90_2002.tif and dvr90_2023.tif that are missing in PROJ-data and grid_alternatives.sql


kbevers:
>  This update adds records for dvr90_2002.tif and dvr90_2023.tif that are missing in PROJ-data and grid_alternatives.sql

I am well aware. I'll add them to PROJ-data soon.

rouault:
and "Database: reference RAC23.mnt --> fr_ign_RAC23.tif, and update PROJ_DATA.VERSION to 1.17"

PR #3974, opened 2023-12-04
if(WIN32) add shell32 and ole32 libraries in lib_proj.cmake

### LINKED ISSUE #2983: (Windows. Win32) Problem to add proj lib version 8.1.1 to Visual Studio 2019 Project
I made console C++ app in Visual Studio 2019.
I include proj lib 8.1.1 to my project. I configure CMake to make library VS project to x86 platform.
Lib was make as static lib. 
When I try build my project, I have such errors that linker can not find  __imp__CoTaskMemFree@4 and _SHGetKnownFolderPath@16 symbol of functions.

As I understand, this functions used in filemanager of proj lib (core of lib). 
Tell me please, what settings of VS 2019 console project should have,
so that the linker can find the paths in the static library to these functions.
As additional dependecies I define curl, tiff, sqlite3. Such dependencies as ole32 and shell32 defined by default.

Thank you for help
Leo

gitartpiano:
to fix missing symbols at link time for windows target as described in https://github.com/OSGeo/PROJ/issues/2983

Closes #2983

### DIFF
--- a/src/lib_proj.cmake
+++ b/src/lib_proj.cmake
@@ -438,6 +438,13 @@ set_target_properties(proj
 set(PROJ_LIBRARIES proj)
 # hack, required for test/unit
 set(PROJ_LIBRARIES ${PROJ_LIBRARIES} PARENT_SCOPE)
+if(WIN32)
+target_link_libraries (proj
+  PRIVATE
+    shell32.lib
+    ole32.lib
+)
+endif()
 if(UNIX)
   find_library(M_LIB m)
   if(M_LIB)


src/lib_proj.cmake
```diff
@@ -438,6 +438,13 @@ set_target_properties(proj
 set(PROJ_LIBRARIES proj)
 # hack, required for test/unit
 set(PROJ_LIBRARIES ${PROJ_LIBRARIES} PARENT_SCOPE)
+if(WIN32)
+target_link_libraries (proj
+  PRIVATE
+    shell32.lib
+    ole32.lib
```
dg0yt, line 445: This change shouldn't be needed: Both libs are members of `CMAKE_C_STANDARD_LIBRARIES_INIT` for Window. CMake will link them anyways.  
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_STANDARD_LIBRARIES.html  
https://github.com/Kitware/CMake/blob/316bb67ff74e89a87e50370d809429f122689e3d/Modules/Platform/Windows-MSVC.cmake#L256

PR #3979, opened 2023-12-08
cmake/Ccache.cmake: raise cmake_minimum_required() as newer CMake versions warn that support for CMake < 3.5 is going to be dropped at some later point

rouault:


kbevers:
Will this make CMake 3.9 a requirement? If so, RFC3 states that

> Changes for software package dependencies can be introduced with minor version releases of PROJ. Changing the version requirements for a dependency needs to be approved by the PSC.

rouault:
> Will this make CMake 3.9 a requirement?

it is already :-) :-) Cf root CMakeLists.txt: https://github.com/OSGeo/PROJ/blob/master/CMakeLists.txt#L12

kbevers:
> it is already :-)

Well, great :-) I guess RFC3 is not up to date then. It says the minimum CMake requirement is 3.5

rouault:
> I guess RFC3 is not up to date then

was overriden by a PSC vote in Feb 2020: https://lists.osgeo.org/pipermail/proj/2020-February/009324.html

kbevers:
Yes, that was clear from the previous message. My point was that it would be nice to have the current requirements reflected in the RFC - I don't regard them as static documents

mwtoews:
It could also be removed (e.g. this was done for geos https://github.com/libgeos/geos/commit/3d4870f7970dfac1b090e8a0d05d9b969a9d1180). This statement is only needed per-project, whereas I might have thought this was a per-module spec. But this merge fix is also fine.

PR #3981, opened 2023-12-20
Fix `projsync --system-directory` command in install guide

DeflateAwning:
Also added a comma

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -187,10 +187,10 @@ Tests are run with::
 
     ctest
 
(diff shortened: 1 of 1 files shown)


DeflateAwning:
This should also get pulled into `master`. Not certain what target I was supposed to open the PR for. I may have done it wrong.

rouault:
> Not certain what target I was supposed to open the PR for

that's fine for PR targetting features of released versions. I've added the "backport master" label so that a PR targetting master is automatically filed once this one is merged

PR #3987, opened 2024-01-05
Link to explicit SQLite3 target

ErixenCruz:

This change originates from Conan https://github.com/conan-io/conan-center-index/pull/21512#issuecomment-1873516363.

I create an explicit SQLite::SQLite3 target library and then add that target. The target has the include directory property set inside it, so we no longer have to do a `target_include_directories`. This makes it easier for package managers to control the SQLite3 dependency, and it aligns with more recent Cmake version [sqlite3 usage](https://github.com/Kitware/CMake/blob/1f66051983ef0bdefa5de139fa9013830a4c3047/Modules/FindSQLite3.cmake#L58-L67). 

Both [vcpkg](https://github.com/microsoft/vcpkg/blob/master/ports/proj/fix-proj4-targets-cmake.patch) and [conan](https://github.com/conan-io/conan-center-index/blob/master/recipes/proj/all/patches/0001-use-cmake-targets-9.2.1.patch) are forced to patch the PROJ CMakeLists.txt so that they can inject their own Sqlite3 right now, but if we can instead link a `SQLite::SQLite3` target, package managers can just make sure that that target is defined before trying to build PROJ.

Another alternative is requiring a more recent Cmake version and using built-in FindSQLite3. I think that may be a more involved change though.


ErixenCruz:
I believe this is failing because we need to indicate to consumers of the library that they need to find SQLite3 the same way that we do for [TIFF](https://github.com/OSGeo/PROJ/blob/ef276d39d982611ebbd3b3c715f4cbdec6d28976/cmake/project-config.cmake.in#L31) and [CURL](https://github.com/OSGeo/PROJ/blob/ef276d39d982611ebbd3b3c715f4cbdec6d28976/cmake/project-config.cmake.in#L36). The c_app build test does not find it on its own. I think this is kind of tricky because that means we need to pass "FindSqlite3.cmake" in the variable CMAKE_MODULE_PATH. its not reasonable for people integrating PROJ with cmake to have to copy in FindSqlite3. I think it works as it is now because c_app uses whatever sqlite that PROJ found when it was installed. This is the crux of the problem when a package manager wants to control the sqlite that PROJ uses.

so im leaning towards requiring Cmake version 3.14 so that we can just use built-in [FindSQLite3](https://cmake.org/cmake/help/latest/module/FindSQLite3.html). ubuntu 20.04 LTS ships 3.16, so I think 3.14 isn't crazy. Bumping Cmake version requires PSC approval ([RFC3](https://proj.org/en/9.3/community/rfc/rfc-3.html#summary)).

Is this something thats possible?

PR #3989, opened 2024-01-07
Fix most Coverity Scan performance warnings

rouault:


jjimenezshaw:
Now I am curious. Were there performance warnings due to the lack of `const`?

rouault:
> Now I am curious. Were there performance warnings due to the lack of `const`?

not directly. But Coverity Scan insists on using std::move() as much as possible. But in some places, I find it a bit dangerous, because it is not immediately obvious that the variable might not be used afterwards. Adding const makes the variable non-moveable, hence std::move() can't be used and the copy assignment operator is used.

I don't believe any of those "fixes" are going to have measurable perf improvements. The main goal is to shut off Coverity Scan.

jjimenezshaw:
Than makes sense. Thanks!

PR #3997, opened 2024-01-10
Increase CMake minimum version from 3.9 to 3.16

mwtoews:
This PR upgrades the CMake minimum version from 3.9 to 3.16. It also removes old workaround and hacks that were needed to support older CMake versions.

<s>Also, fetching the "internal" googletest is modified according to [this doc](https://github.com/google/googletest/blob/main/googletest/README.md) that needs CMake 3.14.</s> (to be resubmitted for a separate PR)

See [discussion on mail list](https://lists.osgeo.org/pipermail/proj/2024-January/011210.html).

Related to #3987 (ping @ErixenCruz).
Refer to #1907 for previous CMake bump four years ago.


rouault:
The error with CIFuzz is unrelated and will be fixed if you rebase on top of latest master

dg0yt:
Do you also want to migrate from `cmake/FindSqlite3.cmake` to the official `SQLite3`?
https://cmake.org/cmake/help/latest/module/FindSQLite3.html 

Note that it uses a different set of variable names, and cache variables are used for user input.

mwtoews:
Thanks for the suggestion @dg0yt . It seems that we'd need to translate (and possibly deprecate) these [documented](https://proj.org/en/9.3/install.html#cmdoption-arg-SQLITE3_INCLUDE_DIR) options:

- `SQLITE3_INCLUDE_DIR` → `SQLite3_INCLUDE_DIRS`
- `SQLITE3_LIBRARY` → `SQLite3_LIBRARIES`

ErixenCruz:
Vote for CMake 3.16 passed https://lists.osgeo.org/pipermail/proj/2024-January/011235.html. This PR can update FindSQLite3, or we can do it separately in https://github.com/OSGeo/PROJ/pull/3987. It may make sense to do it all here in one "3.16 upgrade" PR, but it could go either way

.github/workflows/mingw_w64/start.sh
```diff
@@ -64,7 +64,7 @@ WINE_SYSDIR=$WINE_PREFIX/drive_c/windows
 wine64 cmd /c dir
 ln -s $MINGW_PREFIX/libstdc++-6.dll $WINE_SYSDIR
 ln -s $MINGW_PREFIX/libgcc_s_seh-1.dll $WINE_SYSDIR
-ln -s $MINGW_PREFIX/libgcc_s_sjlj-1.dll $WINE_SYSDIR
+# ln -s $MINGW_PREFIX/libgcc_s_sjlj-1.dll $WINE_SYSDIR
```
ErixenCruz, line 67: im curious/clueless why this change? not familiar with how windows is done

test/unit/CMakeLists.txt
```diff
@@ -9,65 +9,52 @@ endif()
...
-  ${PROJ_SOURCE_DIR}/test/googletest/CMakeLists.txt.in
-  ${PROJ_BINARY_DIR}/googletest-download/CMakeLists.txt)
-
-if(CYGWIN)
-# needed at least with gcc 11.3.0 Cygwin, otherwise build errors in
-# in googletest-src/googletest/include/gtest/internal/gtest-port.h related to
-# fileno() and other functions, when building gtest or including it
-add_definitions(-D_GNU_SOURCE)
-endif()
+  # Add Google Test
```
ErixenCruz: Do you know why we fetch directly in CMake instead of relying on something like a git submodule? I think git submodule could be simpler/cleaner

.github/workflows/mingw_w64/start.sh
mwtoews: It was commented-out since it is unavailable from `gcc-mingw-w64-x86-64` (this was probably up until 18.04). These are [exception handling methods](https://stackoverflow.com/questions/15670169/what-is-difference-between-sjlj-vs-dwarf-vs-seh), and this one is for setjmp/longjmp.

mwtoews:
Thanks for the update @ErixenCruz

I'm going to simplify this PR, and split off FetchContent/googletest and FindSqlite3 upgrades into separate PRs, since these are slightly more complicated and probably deserve a bit more focus/conversation. This PR should cover the ubuntu upgrades and the low-hanging CMake upgrades already done (are there any others?)

PR #4001, opened 2024-01-14
CMake: add a policy_max to cmake_minimum_required(), and remove obsolete cmake_policy() statements

rouault:


mwtoews:
With regard to the googletest changes, wouldn't it make more sense to use FetchContent described [here](https://github.com/google/googletest/blob/main/googletest/README.md)? Some of the relevant changes were previously in [commits from PR 3997](https://github.com/OSGeo/PROJ/compare/56acc4f2b369c136805dd01f514b8d24056912ae..7bc07ff6b6bf94637db2131beab4f5d17190a988) (but some of these changes were backed out). The method used in PROJ was based on an [old version of the guide](https://github.com/google/googletest/blob/e6e2f3f455996b0d9d227cb1ed4ac4adf137c6fc/googletest/README.md) before FetchContent was available.

rouault:
> With regard to the googletest changes, wouldn't it make more sense to use FetchContent described [here](https://github.com/google/googletest/blob/main/googletest/README.md)?

yes probably. I didn't want to change that, but changing cmake_minimum_required(VERSION 3.16) to cmake_minimum_required(VERSION 3.16...3.28)  caused warnings in the googletest stuff. And now there's this mysterious error with MacOS build. I'm going to close this PR for now to let other changes slip in before

PR #4002, opened 2024-01-14
 Fix Coverity performance inefficiencies warnings

rouault:


src/4D_api.cpp
```diff
@@ -2347,18 +2347,15 @@ PJ *proj_create_crs_to_crs_from_pj(PJ_CONTEXT *ctx, const PJ *source_crs,
                     // PROJ_GRID_AVAILABILITY_DISCARD_OPERATION_IF_MISSING_GRID
                     // returned 0.
                     if (op_count == 1 &&
-                        (errorIfBestTransformationNotAvailable ||
-                         warnIfBestTransformationNotAvailable)) {
+                        errorIfBestTransformationNotAvailable) {
```
jjimenezshaw, line 2350: if I understand correctly, this is a change in behavior, not just about warning removal, right? Same a few lines below.

rouault: > if I understand correctly, this is a change in behavior

not really, if warnIfBestTransformationNotAvailable = true, nothing would be actually done.

PR #4007, opened 2024-01-17
CMake: Replace custom FindSqlite3 with FindSQLite3 built-in

### LINKED ISSUE #4004: Clean up sqlite3 further in CMake
https://github.com/OSGeo/PROJ/pull/3987 changed the CMake to link to the explicit target `SQLite::SQLite3` and find SQLite3, which makes use of CMake's built-in FindSQLite3 script instead of PROJ's. 

@mwtoews mentioned in the CMake 3.16 upgrade PR that there are some Sqlite3 configuration variables that should probably be adjusted or deprecated. https://github.com/OSGeo/PROJ/pull/3997#issuecomment-1886679622. Reading about [`find_path`](https://cmake.org/cmake/help/latest/command/find_path.html), which is called in the built-in FindSQLite3 script, the configuration variables might just have to be mapped to `SQLite3_ROOT`. A CMake expert may chime in here.

The other thing is that we can probably remove PROJ's [FindSqlite3](https://github.com/OSGeo/PROJ/blob/master/cmake/FindSqlite3.cmake) and rely entirely on built-in CMake.

mwtoews:
This PR replaces the custom FindSqlite3 module with the [FindSQLite3](https://cmake.org/cmake/help/latest/module/FindSQLite3.html) built-in.

<s>Upon further inspection, the variables `SQLITE3_INCLUDE_DIR` and `SQLITE3_LIBRARY` cannot be easily mapped to modern variables, so these variables are removed.</s> Users are encouraged to use [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html) instead, which is added to the install docs.

Closes #4004


rouault:
SQLite3_INCLUDE_DIR (https://github.com/Kitware/CMake/blob/fa4a499238502dead01c482a168eb52b6599dc22/Modules/FindSQLite3.cmake#L37) and SQLite3_LIBRARY (https://github.com/Kitware/CMake/blob/fa4a499238502dead01c482a168eb52b6599dc22/Modules/FindSQLite3.cmake#L41) can still be used

mwtoews:
Ah, thanks for pointer. I can map these, and provide a deprecation message. I'd like to still remove these from the docs, as these are advanced variables that aren't even in the CMake docs.

mwtoews:
A small deprecation notice is added to docs here:
Screenshot from 2024-01-18 11-51-45

although I might be misusing the `deprecated::` directive, so the formatting looks a bit off. It is normally indented under an option.

For consistency, we may want to also undocument `CURL_` and `TIFF_` variables for `LIBRARY` and `INCLUDE_DIR` too, as these should ideally be handled by `CMAKE_PREFIX_PATH`. I'm also fine to put back `SQLITE3_INCLUDE_DIR` and `SQLITE3_LIBRARY` docs with their `deprecated::` too, if that's preferred.

rouault (approved):

CMakeLists.txt
```diff
@@ -176,15 +176,22 @@ if(NOT EXE_SQLITE3)
...
-if(NOT SQLITE3_FOUND)
-  message(SEND_ERROR "sqlite3 dependency not found!")
+# Deprecated variables since PROJ 9.4.0
+if(DEFINED SQLITE3_INCLUDE_DIR)
+  message(DEPRECATION "Use SQLite3_INCLUDE_DIR instead of SQLITE3_INCLUDE_DIR")
+  set(SQLite3_INCLUDE_DIR ${SQLITE3_INCLUDE_DIR})
 endif()
+if(DEFINED SQLITE3_LIBRARY)
+  message(DEPRECATION "Use SQLite3_LIBRARY instead of SQLITE3_LIBRARY")
+  set(SQLite3_LIBRARY ${SQLITE3_LIBRARY})
```
dg0yt, line 186: FTR the target variables are (meant to be) cache variables. But it probably works as is.

Issue #4013, opened 2024-01-27
Allow +inv to work like -I


jidanni:
This restriction makes zero sense.

```
: This works fine:
echo -88 42|proj \
+proj=pipeline \
+step +proj=utm +zone=16 \
+step +proj=utm +zone=16 +inv \
+step +proj=utm +zone=16
417181.93	4650259.85
: This also works fine:
echo -88 42|proj \
+proj=pipeline \
+step +proj=utm +zone=16 \
|proj +proj=utm +zone=16 -I
88dW	42dN
: So why not allow this to also work fine?:
echo -88 42|proj \
+proj=pipeline \
+step +proj=utm +zone=16 \
+step +proj=utm +zone=16 +inv
Rel. 9.3.1, December 1st, 2023
<proj>:
can't initialize operations that produce angular output coordinates
program abnormally terminated
: And indeed this should work too:
echo -88 42|proj \
+proj=pipeline \
+step +proj=utm +zone=16 \
|proj +proj=utm +zone=16 +inv
Rel. 9.3.1, December 1st, 2023
<proj>:
can't initialize operations that take non-angular input coordinates
program abnormally terminated
```

rouault:
proj is a legacy utility that was designed to only deal with map projections, and that is now frozen.
If you use pipeline and +inv, you need to use cct instead.

PR #4015, opened 2024-01-28
Update pipeline.rst

jidanni:
For #4013

### DIFF
--- a/docs/source/operations/pipeline.rst
+++ b/docs/source/operations/pipeline.rst
@@ -127,12 +127,11 @@ Optional
 
 .. option:: +inv
 
-    Invert a step in a pipeline. Note, cct handles more situations:
+    Invert a step in a pipeline.
 
-::
+    ::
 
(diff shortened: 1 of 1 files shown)


docs/source/operations/pipeline.rst
```diff
@@ -127,7 +127,12 @@ Optional
 
 .. option:: +inv
 
-    Invert a step in a pipeline.
+    Invert a step in a pipeline. Note, cct handles more situations:
+
+::
+
+    echo -88 42    |proj +zone=16 +proj=pipeline +step +proj=utm +step +proj=utm +inv #fails
+    echo -88 42 0 0|cct  +zone=16 +proj=pipeline +step +proj=utm +step +proj=utm +inv #succeeds
```
rouault: I've removed the comment about "cct handles more situations" as I don't think it belongs there, and the example with "proj" that fails as it's going to create more confusion

```suggestion
    Invert a step in a pipeline.

    ::

        echo -88 42 0 0|cct +zone=16 +proj=pipeline +step +proj=utm +step +proj=utm +inv
```

PR #4016, opened 2024-01-28
Update proj.cpp

jidanni:
For #4013.


dg0yt:
I wish PRs would have telling names and messages. So that e-mail notification are immediately useful.

kbevers:
> I wish PRs would have telling names and messages. So that e-mail notification are immediately useful.

That would certainly be nice, but that responsibility lies with whoever opens the PR. Some people are more diligent with that stuff than others. I tend to ignore what doesn't immediately make sense to me

jidanni:
> I wish PRs would have telling names and messages. So that e-mail notification are immediately useful.

Oops. Sorry.

PR #4018, opened 2024-01-29
Database (ESRI): use the deprecation table instead of the supersession table for deprecated ESRI CRS (as done with EPSG objects)

rouault:


jjimenezshaw:
Maybe you should mention `ESRI` in the PR title. I had to go to the changes to understand what was it about.

Out of curiosity, why are you changing this now? (is this a bug or a feature? ;)

Related to this topic, 3 years ago I made this url https://jjimenezshaw.github.io/Esri-codes-analyzer/deprecated.html to list all the deprecated CRSs from ESRI and their equivalent in EPSG. It should make it easy for a final user to know why an ESRI code disappears from a CRSs list (because it was deprecated). You can get there from the README of https://github.com/OSGeo/PROJ-CRS-Explorer

rouault:
> Maybe you should mention ESRI in the PR title

done

> Out of curiosity, why are you changing this now? (is this a bug or a feature? ;)

Basically upstreaming changes that ESRI has done in a fork (they've agreed to that upstreaming), and which make sense

jjimenezshaw (approved):

PR #4020, opened 2024-01-30
CI: add a M1 ARM64 MacOS runner

rouault:


rouault:
Looks like we have a blocker with conda-incubator/setup-miniconda action not supporting yet arm64: https://github.com/conda-incubator/setup-miniconda/issues/312

rouault:
M1 arm64 MacOS runners are quite amazing: the job runs in 12 min 22 sec, vs 41m 43s  for x64 Mac !

PR #4021, opened 2024-01-31
Add +proj=mod_krovak projection method for Modified Krovak that applies to S-JTSK/05 in the Czech Republic

rouault:
Validated with the test point given in EPSG Guidance Note 7-2

Also validated with CUZK online calculator at https://geoportal.cuzk.cz/(S(g4ipaut0ckzb5lellu153lvf))/Default.aspx?head_tab=sekce-01-gp&mode=TextMeta&text=wcts&menu=19

Unfortunately this calculator does not offer 'elementary' transformations, and cannot just validate Modified Krovak alone. The best we can do is to use ETRS89 <--> S-JTSK/05 / Krovak modified, which involves a Helmert transformation as well for the datum transformation.

* Validation of reverse modified Krovak:

  Online calculator:
  - Souřadnice/input: -5568990.91 -6050538.71 100
  - Source CRS: S-JTSK/05 + Bpv (-Y-XH /east-north)
  - Destination CRS: ETRS89 (BLh / DEG)
  - Výsledek/result: 50.208297081 16.848326835 143.762

  With PROJ:
```shell
  echo -5568990.91 -6050538.71 100 | bin/cs2cs -d 9 EPSG:5516 EPSG:4937 --3d
  50.208297081 16.848326833 143.143233569
```

  Note the different on longitude is totally neglectable: 0.1 mm, as computed by:
```shell
  $ echo 50.208297081 16.848326835 50.208297081 16.848326833 | bin/geod -F "%.4f" -I +ellps=GRS80
  -90d 90d 0.0001
```
  and that when outputing Krovak modified, the online calculator displays Y/X with
  a precision of 1 cm.

  The difference on Z is significant, but can be explained from the fact that
  the online calculator using Baltic Height for S-JTSK/05 and thus apply a
  Baltic Height<-->ETRS89 geoid that we don't apply here.

* Validation of forward Krovak using same point:

  Online calculator:
  - Souřadnice/input: 50.208297081 16.848326835 143.762
  - Source: ETRS89 (BLh / DEG)
  - Destination: S-JTSK/05 + Bpv (-Y-XH /east-north)
  - Výsledek/result: -5568990.91 -6050538.71 100

  With PROJ:
```shell
  $ echo 50.208297081 16.848326835 143.762 | bin/cs2cs -d 2 EPSG:4937 EPSG:5516 --3d
  -5568990.91 -6050538.71 100.62
```


jjimenezshaw (approved):

mwtoews:
There is an inconsistency between "Krovak Modified" and "Modified Krovak".

Guidance Note 7-2 mostly uses "Krovak Modified", with a few exceptions (e.g. "Modified Krovak East North")

An advantage of `krovak_mod` is that it nicely alpha-sorts after `krovak`.

Another more convoluted method is to keep only `+proj=krovak` with `+modified` option.

rouault:
> There is an inconsistency between "Krovak Modified" and "Modified Krovak".

yeah, I considered krovak_mod at first, but there was the risk of not noticing the _mod suffix, so I switched the mod_krovak to make things obvious. That said, if you feel that krovak_mod is preferable, I could switch to that

> Another more convoluted method is to keep only +proj=krovak with +modified option.

I also considered that, but same reason, it is easy to miss it (and if people would copy&paste PROJ strings with +proj=krovak +modified to older PROJ releases, the +modified would be silently ignored, whereas +proj=mod_krovak will just error out). Furthermore, technically/formally, modified krovak takes extra parameters (the correction coefficients), which are currently hard-coded, because there's no point in requiring users to type them, given that the S-JTSK/05 is the single instance where this projection method is typed.

I was also wondering if we should set the default values of x_0 and y_0 to 5000000, which is what is used for S-JTSK/05, but was undecided, so I kept the 0 default value.

docs/source/operations/projections/mod_krovak.rst
```diff
@@ -0,0 +1,77 @@
+.. _mod_krovak:
+
+********************************************************************************
+Krovak Modified
```
mwtoews:
```suggestion
Modified Krovak
```

docs/source/operations/projections/mod_krovak.rst
```diff
@@ -0,0 +1,77 @@
...
+| **Input type**      | Geodetic coordinates                                     |
++---------------------+----------------------------------------------------------+
+| **Output type**     | Projected coordinates                                    |
++---------------------+----------------------------------------------------------+
+
+
+.. figure:: ./images/krovak.png
+   :width: 500 px
+   :align: center
+   :alt:   Krovak Modified
```
mwtoews: This is a repeat of an existing figure. Does it make sense to create a new `mod_krovak.png`? It would need to be added to docs/plot/plotdefs.json

src/pj_list.h
```diff
@@ -104,6 +104,7 @@ PROJ_HEAD(merc, "Mercator")
 PROJ_HEAD(mil_os, "Miller Oblated Stereographic")
 PROJ_HEAD(mill, "Miller Cylindrical")
 PROJ_HEAD(misrsom, "Space oblique for MISR")
+PROJ_HEAD(mod_krovak, "Krovak Modified")
```
mwtoews:
```suggestion
PROJ_HEAD(mod_krovak, "Modified Krovak")
```

src/projections/krovak.cpp
```diff
@@ -78,10 +78,13 @@
 #include <errno.h>
 #include <math.h>
 
+#include <algorithm>
+
 #include "proj.h"
 #include "proj_internal.h"
 
 PROJ_HEAD(krovak, "Krovak") "\n\tPCyl, Ell";
+PROJ_HEAD(mod_krovak, "Krovak Modified") "\n\tPCyl, Ell";
```
mwtoews:
```suggestion
PROJ_HEAD(mod_krovak, "Modified Krovak") "\n\tPCyl, Ell";
```

src/projections/krovak.cpp
```diff
@@ -97,41 +100,107 @@ struct pj_krovak_data {
...
+constexpr double C3 = 1.193845912E-07;
+constexpr double C4 = -4.668270147E-07;
+constexpr double C5 = 9.233980362E-12;
+constexpr double C6 = 1.523735715E-12;
+constexpr double C7 = 1.696780024E-18;
+constexpr double C8 = 4.408314235E-18;
+constexpr double C9 = -8.331083518E-24;
+constexpr double C10 = -3.689471323E-24;
+
+// Correction terms to be applied to regular Krovak to obtain Krovak Modified.
```
mwtoews:
```suggestion
// Correction terms to be applied to regular Krovak to obtain Modified Krovak.
```

mwtoews (commented):
Re: "Krovak Modified" and "Modified Krovak". Looking further, there is no obvious "correct" form.

- In Czech it is "modifikovaném Křovákově", never the other way around
- English would normally prefer "Modified Krovak" too

So sure, let's stick with "Modified Krovak". See related suggested changes...

src/proj_constants.h
```diff
@@ -173,6 +173,13 @@
 #define EPSG_NAME_METHOD_KROVAK "Krovak"
 #define EPSG_CODE_METHOD_KROVAK 9819
 
+#define EPSG_NAME_METHOD_KROVAK_MODIFIED "Krovak Modified"
+#define EPSG_CODE_METHOD_KROVAK_MODIFIED 1042
+
+#define EPSG_NAME_METHOD_KROVAK_MODIFIED_NORTH_ORIENTED                        \
+    "Krovak Modified (North Orientated)"
+#define EPSG_CODE_METHOD_KROVAK_MODIFIED_NORTH_ORIENTED 1043
+
```
mwtoews, line 182: Keep these untouched, as they are referenced this way in EPSG Guidance Note 7-2

docs/source/operations/projections/mod_krovak.rst
rouault: > Does it make sense to create a new `mod_krovak.png`?

Not really. The difference between both is sub-metric...

docs/source/operations/projections/mod_krovak.rst
rouault: done

src/pj_list.h
rouault: done

src/projections/krovak.cpp
rouault: done

PR #4033, opened 2024-02-02
+proj=push/pop: add a +bank=<name> parameter to interleave push/pop operations

rouault:
The motivation is to be able to do transformation that first do a geoid transformation followed by a Helmert one.

The following example shows a ETRS89 to S-JTSK/05 + Baltic 1957 height transformation that chains a geoid model registered against ETRS89 with a Helmert transformation.

```
       +proj=pipeline \
            +step +proj=push +v_3 +omit_inv \                         # Save ETRS89 ellipsoidal height
            +step +inv +proj=vgridshift +grids=CR2005.tif +multiplier=1 \  # Apply geoid
            +step +proj=push +v_3 +bank=baltic_height \               # Save Baltic 1957 height
            +step +proj=pop +v_3 +omit_inv \                          # Restore ETRS89 ellipsoidal height
            +step +proj=cart +ellps=GRS80 \                           # Helmert transformation
            +step +inv +proj=helmert +x=572.213 +y=85.334 +z=461.94 \
                +rx=-4.9732 +ry=-1.529 +rz=-5.2484 +s=3.5378 +convention=coordinate_frame \
            +step +proj=pop +v_3 +bank=baltic_height                  # Restore Baltic 1957 height
```

Without that, the current alternative is a ugly workaround involving using the v_4 component and axisswap
```
       +proj=pipeline \
            +step +proj=push +v_4 \                                   # Save v_4 as we are going to use it as a temp variable
            +step +proj=push +v_3 +omit_inv \                         # Save ETRS89 ellipsoidal height
            +step +inv +proj=vgridshift +grids=CR2005.tif +multiplier=1 \  # Apply geoid
            +step +proj=push +v_3 +omit_fwd \                         # On reverse path, restore initial Baltic height
            +step +proj=axisswap +order=1,2,4,3 +omit_inv \           # On forward path, save Baltic height in v_4 component...
            +step +proj=pop +v_3 +omit_inv \                          # On forward parth, restore initial ellipsoidal height
            +step +proj=cart +ellps=GRS80 \                           # Helmert transformation
            +step +inv +proj=helmert +x=572.213 +y=85.334 +z=461.94 \
                +rx=-4.9732 +ry=-1.529 +rz=-5.2484 +s=3.5378 +convention=coordinate_frame \
            +step +proj=axisswap +order=1,2,4,3 +omit_inv \           # On forward path, restore Baltic height from v_4 component...
            +step +proj=pop +v_3 +omit_fwd \                          # On reverse path, save initial Baltic height
            +step +proj=pop +v_4
```

### DIFF
--- a/src/pipeline.cpp
+++ b/src/pipeline.cpp
@@ -97,6 +97,8 @@ Thomas Knudsen, thokn@sdfe.dk, 2016-05-20
 *
 ********************************************************************************/
 
+#include <array>
+#include <map>
 #include <math.h>
 #include <stack>
 #include <stddef.h>
@@ -137,13 +139,17 @@ struct Pipeline {
     char **current_argv = nullptr;
     std::vector<Step> steps{};
     std::stack<double> stack[4];
+    std::map<std::string, int> bank_names_to_idx{};
+    std::vector<std::array<double, 4>> bank{};
 };
 
 struct PushPop {
-    bool v1;
-    bool v2;
-    bool v3;
-    bool v4;
+    std::string bank_name{};
+    int bank_index = -1;
+    bool v1 = false;
+    bool v2 = false;
+    bool v3 = false;
+    bool v4 = false;
 };
 } // anonymous namespace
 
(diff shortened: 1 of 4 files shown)


src/pipeline.cpp
```diff
@@ -642,55 +648,113 @@ static void push(PJ_COORD &point, PJ *P) {
...
+        const auto iter = pipeline->bank_names_to_idx.find(pushpop->bank_name);
+        if (iter != pipeline->bank_names_to_idx.end()) {
+            // Yes, then store its known index, so we don't have to fetch
+            // it anymore
+            pushpop->bank_index = iter->second;
+        } else {
+            // Unknown bank name. Add a new one, and register it in the
+            // pipeline.
+            const int bank_index = static_cast<int>(pipeline->bank.size());
+            pipeline->bank.resize(pipeline->bank.size() + 1);
```
jjimenezshaw, line 665: Should we initialize the new `bank` element, an std::array<4> (with a consistent value, like 0 or inf or whatever)? I know it should not affect with a proper usage, but if the user is latter popping the wrong index, then the value will be random "garbage" that is difficult to debug. If it is consistently something, it is easier.

src/pipeline.cpp
```diff
@@ -642,55 +648,113 @@ static void push(PJ_COORD &point, PJ *P) {
...
+        }
     }
 
-    if (pushpop->v2 && !pipeline->stack[1].empty()) {
-        point.v[1] = pipeline->stack[1].top();
-        pipeline->stack[1].pop();
-    }
+    if (pushpop->bank_index >= 0) {
+        if (pushpop->v1)
+            point.v[0] = pipeline->bank[pushpop->bank_index][0];
```
jjimenezshaw, line 714: If I understand it correctly, each named bank has a single element (std::array<4>), it is not a stack. Right? The element is not deleted from the bank after a `pop`. (I say this because the unnamed pop and push are going to a "real" stack). I am not asking that it should be a stack. Just that it was my first impression. Maybe it should be documented.

src/pipeline.cpp
rouault: My understanding of https://stackoverflow.com/questions/66875565/zero-initialization-of-stdvectorstdarraydouble-10 is that a ``std::array<double, N>`` is zero-initialized on construction.

And given the following test program:
```c++
#include <array>

std::array<double, 4> foo()
{
    return std::array<double, 4>();
}
```

``g++ -std=c++11 -c test.cpp -O2 -g && objdump -dw test.o`` shows:
```
0000000000000000 <_Z3foov>:
   0:	f3 0f 1e fa          	endbr64 
   4:	48 c7 07 00 00 00 00 	movq   $0x0,(%rdi)
   b:	48 89 f8             	mov    %rdi,%rax
   e:	48 c7 47 08 00 00 00 00 	movq   $0x0,0x8(%rdi)
  16:	48 c7 47 10 00 00 00 00 	movq   $0x0,0x10(%rdi)
  1e:	48 c7 47 18 00 00 00 00 	movq   $0x0,0x18(%rdi)
  26:	c3                   	retq   
```
which shows the zeo initialization of the array

src/pipeline.cpp
rouault: > If I understand it correctly, each named bank has a single element (std::array<4>), it is not a stack

yes, good point. That's also something that makes me slightly uneasy as it goes a bit against the push/pop stack philosophy. Perhaps instead of hijacking push/pop, we should introduce a save/load as I proposed in one of my emails?

``+proj=save +v_1 +bank=foo`` / ``+prop=load +v_1 +bank=foo``

busstoptaktik: That is way better. But I think I have an idea for an even more clean design, which will be entirely orthogonal to the current unfortunate design of the push/pop mechanism, and handle both stack and register based "side storage" in a better abstracted way.

It will allow us to deprecate the current design, and have a stack mechanism which is **actually** a stack mechanism, and a register mechanism which is **actually** a register mechanism.

I'm writing this up as a proposal right now - will post here and on the related mailing list thread in a few hours.

**The stack/register mish-mash in this PR, I'm not happy with** - it's papering over an already unfortunate design with even more unfortunate design.

Personally, I do not see the need for register-based side storage: Repairing the stack mechanism would do. But I understand that there is a perceived need for this, so I'm going for a design including both, but with an extra layer of abstraction which allows for a cleaner syntax, better extensibility, and full three way orthogonality between the new stack mechanism, the new register mechanism, and the existing (and hopefully soon deprecated) push/pop implementation.

rouault: > I'm writing this up as a proposal right now

eagerly waiting for it.

Thinking at the topic, I had a look back at the ISO-19111 "Pass-through coordinate operation" (https://docs.ogc.org/as/18-005r4/18-005r4.html#103) concept, that is somehow related, but probably not quite advanced enough (I didn't implement PassthroughOperation during the ISO-19111'ification of PROJ), because just a list of the index of components modified by an operation doesn't fully address the need to do the geoid operation and Helmert one "in parallel", and merge the results at a later stage, because it requires that the 2 operations operate on separate coordinate indices. That is it would work for a vertical-only + horizontal-only operation. But here the horizontal part is actually a 3D Helmert for which we only keep the horizonal-part. So something more convoluted...

busstoptaktik: > So something more convoluted...

Yes, this is really one of the problems with 19111: Real geodesy is always more convoluted than what the standard is able to cover. 19111 can be drastically simplified by dropping the geodetically unrealistic misconception of "a CRS defined by internal state", accepting that a CRS is merely a label, enabling us to look up related transformations in a register, and that 19111 should focus on an improved representation of transformations by essentially codifying the workings of the pipeline operator

busstoptaktik: > eagerly waiting for it.

@rouault - here it is

The main problem with PROJ's current push/pop functionality is that the four-stack implementation makes it impossible to push one dimension of a coordinate-tuple and pop it into another, without an elaborate axisswap dance at both ends of the push/pop-pair.

This is a consequence of the `v_1 ... v_4` syntax used: Since the order of operator arguments is not necessarily preserved during the parsing process, a pair specified as `v_1 v_2` may appear as `v_2 v_1` at parsing time.

A more axisswap-like syntax would have eliminated this - e.g.
```
proj=push items=1,3
```
to push dimensions 1 and 3 of the coordinate tuple onto the
stack, and
```
proj=pop items=3,1
```
to pop them back into swapped dimensions.

But since the obvious names of push and pop are now taken, **let's make a virtue out of necessity** by adding a little more encapsulation by introducing the "stack subsystem", giving this syntax:
```
proj=stack push=1,3
```
and
```
proj=stack pop=3,1
```
The implementation will still mimick that of push/pop (i.e being handled as part of the pipeline machinery, rather than as ordinary proj=foo-operators), but it will not stomp on the existing unfortunate `push/pop`-implementation, and it leaves room for extension, e.g. by introducing the swap and flip operators mused over earlier in this thread:
```
proj=stack push=1,3
proj=stack swap
proj=stack pop=1,3
```
to essentially do the same as above and:
```
proj=stack push=1,3
proj=stack flip=3,1
```
although I believe the need for these will be limited, once we have a push/pop pair actually working like a stack.

The wish for named registers mentioned earlier in the thread may be introduced using a similar syntax, where:
```
proj=reg save=1,3 into=first,third
proj=reg restore=1,3 from=third,first
```
will do the same swapping of first and third dimension as above, but using named registers instead of stack slots.

I see the potential for improved clarity when using named registers instead of stack slots, by using meaningful names for the registers (e.g. `ellipsoidal_height` and `orthometric_height`, to take the obvious use case).

The problem, however, is when trying to interpret the workings of a pipeline running in inverse mode:

In the inverse stack-case, we only need to flip the abstract concepts of stack slots, and swap push for pop.

Whereas in the register case, we also need to "abstractly swap the concrete concepts behind the concrete register names" which, at least to my imperfect brain, is much harder.

But while not personally seeing the need for registers, I acknowledge the wish for them, and to make this work in a fairly backwards compatible way, I suggest the solution outlined above, i.e. two new, and well encapsulated subsystems in the pipeline implementation. And once they are well established: Deprecation of the current push/pop pair.

busstoptaktik: And with a resulting pipeline somewhat like this (with the caveat that I am not sure I understand the functionality completely):

```
proj=pipeline 
    step proj=stack push=3                      # Save ETRS89 ellipsoidal height
    step inv proj=vgridshift grids=CR2005.tif   # Transform to orthometric height? (why inv here?)
    step proj=stack flip=3                      # Save Baltic 1957 height, restore ETRS89
    step proj=cart ellps=GRS80                  # Helmert transformation
    step inv proj=helmert
             x=572.213  rx=-4.9732
             y=85.334   ry=-1.529
             z=461.94   rz=-5.2484
             s=3.5378   convention=coordinate_frame
    step inv proj=cart ellps=bessel 
    step proj=stack pop=3                       # Restore Baltic 1957 height
```

rouault: >  Transform to orthometric height? (why inv here?)

 By default "proj=vgridshift" subtracts the grid value (the default value of multiplier is -1), which is a bit unusual. A geoid grid contains values to add to orthometric height to get ellipsoidal heights. So here as we want to go from ellipsoidal height to orthometric, that is subtract the grid value, we need to use either "proj=vgridshift" or "proj=vgridshift +multiplier=1 +inv" which are equivalent

> And with a resulting pipeline somewhat like this (with the caveat that I am not sure I understand the functionality completely):

That pipeline would likely work in the forward path, but not in the reverse path.   Since the inversion of the first step " step proj=stack push=3" would restore a "approximate" ETRS89 ellipsoidal height computed by helmert.  So in the reverse path, we need to restore the Baltic 57 height before the execution of vgridshift

rouault:
Related discussion thread at https://lists.osgeo.org/pipermail/proj/2024-February/thread.html#11251

src/pipeline.cpp
busstoptaktik: > By default "proj=vgridshift" subtracts the grid value (the default value of multiplier is -1), which is a bit unusual

I would not call that unusual, it's just the common geodetic convention: The geoid model represents the geoid separation, i.e. the height of the geoid wrt. the ellipsoid, so to make geometrical heights, referred to the ellipsoid, compatible with heights obtained by levelling in a different, orthometric system, we subtract the value of the geoid model, expressed in the geometrical system (in this case ETRS89).

And as PROJ is geometry-centric, the natural forward direction operation of vgridshift is to go from geometric to orthometric - so according to the comments indicating ETRS->Baltic57, i expected `proj=vgridshift grids=...`: No `inv` and no `multiplier`, just the plain forward operation - and with and `inv`, I would have expected the multiplier to be `-1`. So just a bit surprised here over the sign convention.

src/pipeline.cpp
jjimenezshaw: I did this test (trying to "pollute" the stack, but you never know how the compiler is optimizing)
```
$ cat test.cpp 
#include <iostream>
#include <array>

void with_init() {
    int i = 428;
    std::array<double, 4> a;
    a[0] = 55;
    std::cout << i << " " << a[0] << std::endl;
}
void not_init() {
    int i;
    std::array<double, 4> a;
    std::cout << i << " " << a[0] << std::endl;
}

int main()
{
    with_init();
    not_init();

    return 0;
}
```

that produces different outputs depending on the compiler options
```
$ c++ test.cpp  && ./a.out
428 55
428 55

$ c++ test.cpp -O2  && ./a.out
428 55
0 6.95277e-310
```

The result makes me very suspicious.

rouault: > I did this test

That's different because ``std::array<double, 4> a`` itself doesn't cause initialization. It will do if you for example add ``{}`` at the end.
```
$ cat test.cpp
#include <array>

std::array<double, 4> foo()
{
    std::array<double, 4> x{};
    return x;
}


std::array<double, 4> foo2()
{
    std::array<double, 4> x;
    return x;
}

$ g++ -std=c++11 -c test.cpp -O2 -g && objdump -dw test.o

0000000000000000 <_Z3foov>:
   0:	f3 0f 1e fa          	endbr64 
   4:	66 0f ef c0          	pxor   %xmm0,%xmm0
   8:	48 89 f8             	mov    %rdi,%rax
   b:	0f 11 07             	movups %xmm0,(%rdi)
   e:	0f 11 47 10          	movups %xmm0,0x10(%rdi)
  12:	c3                   	retq   
  13:	66 66 2e 0f 1f 84 00 00 00 00 00 	data16 nopw %cs:0x0(%rax,%rax,1)
  1e:	66 90                	xchg   %ax,%ax

0000000000000020 <_Z4foo2v>:
  20:	f3 0f 1e fa          	endbr64 
  24:	48 89 f8             	mov    %rdi,%rax
  27:	c3                   	retq   
```

 This is the fact that you resize a std::vector<> will causes initialization of std::array<> to be done. This is actually the same for a simple ``double`` variable.  ``double d;`` doesn't initialize it, but extending/resizing a ``std::vector<double>`` will default (=zero)-initialize the new elements. Yes, C++ is a horrible language

jjimenezshaw: :facepalm: oh, yes, those `{}` that sometimes are doing things, and sometimes not.
The backwards compatibility of C++ (into C as well) creates strange syntax rules.

rouault:
Superseded per https://github.com/OSGeo/PROJ/pull/4036

PR #4034, opened 2024-02-03
PROJString formatter optimizer: simplify pipelines doing [Modified]Krovak (South West) <--> [Modified]Krovak (East North)…

rouault:
… by just doing an axis swap

### DIFF
--- a/src/iso19111/io.cpp
+++ b/src/iso19111/io.cpp
@@ -9241,6 +9241,51 @@ const std::string &PROJStringFormatter::toString() const {
                 }
             }
 
+            // Optimize patterns like Krovak (South West) to Krovak East North
+            // (also applies to Modified Krovak)
+            //   +step +inv +proj=krovak +axis=swu +lat_0=49.5
+            //   +lon_0=24.8333333333333
+            //     +alpha=30.2881397527778 +k=0.9999 +x_0=0 +y_0=0 +ellps=bessel
+            //   +step +proj=krovak +lat_0=49.5 +lon_0=24.8333333333333
+            //     +alpha=30.2881397527778 +k=0.9999 +x_0=0 +y_0=0 +ellps=bessel
+            // as:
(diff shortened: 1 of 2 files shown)


src/iso19111/io.cpp
```diff
@@ -9241,6 +9241,68 @@ const std::string &PROJStringFormatter::toString() const {
...
+                curStep.name == prevStep.name &&
+                curStepParamCount + 1 == prevStepParamCount &&
+                prevStep.paramValues[0].equals("axis", "swu")) {
+                bool allSame = true; // skipping prevStep[0] (axis=swu)
+                for (size_t j = 0; j < curStepParamCount; j++) {
+                    if (curStep.paramValues[j] != prevStep.paramValues[j + 1]) {
+                        allSame = false;
+                        break;
+                    }
+                }
```
jjimenezshaw: Something like this (not compiled) would avoid to have the symmetrical code later, right?
```suggestion
            if (curStep.inverted != prevStep.inverted &&
                curStep.name == prevStep.name &&
                abs(curStepParamCount - prevStepParamCount) == 1 &&
                (prevStep.paramValues[0].equals("axis", "swu") || curStep.paramValues[0].equals("axis", "swu"))) {
                const auto& sw = (curStepParamCount - prevStepParamCount) > 0 ? curStep : prevStep;
                const auto& en = (curStepParamCount - prevStepParamCount) < 0 ? curStep : prevStep;
                bool allSame = true; // skipping prevStep[0] (axis=swu)
                for (size_t j = 0; j < min(curStepParamCount, prevStepParamCount); j++) {
                    if (en.paramValues[j] != sw.paramValues[j + 1]) {
                        allSame = false;
                        break;
                    }
                }
```

rouault: PR ammended with something close to your suggestion

jjimenezshaw (approved):

Issue #3197, opened 2022-05-13
hgridshift for projected CRS


yurish:
There are many CRS that require hgridshift, but the actual grid is in projected coordinates. An example of such grid would be Greece GGRS87 grid and Hungary Vitel grid. As far as I know, hgridshift works only with geodetic coordinates. Is there any possibility to apply grid shift for projected coordinates?

ErixenCruz:
This was discussed in the mailing list too https://lists.osgeo.org/pipermail/proj/2022-March/010577.html for EPSG:5513, Korvak S-JTSK

PR #4038, opened 2024-02-06
+proj=gridshift: enhance to support grids referenced in projected CRS, and with easting_offset/northing_offset corrections

rouault:
PROJ-data's check_gtif_grid.py script also improved per https://github.com/OSGeo/PROJ-data/pull/117 to support that new possibility


ErixenCruz:
Closes https://github.com/OSGeo/PROJ/issues/3197

PR #4039, opened 2024-02-07
Update omerc.rst to add vital caveat about the two-point method

jidanni:
It is imperative that users be informed about the vast difference of output of the two point method vs. one-point, undocumented till today!

See also
https://gis.stackexchange.com/questions/323570/proj-oblique-mercator-specifying-two-points-on-central-line/337169#337169

### DIFF
--- a/docs/source/operations/projections/omerc.rst
+++ b/docs/source/operations/projections/omerc.rst
@@ -182,3 +182,20 @@ Optional
 .. include:: ../options/x_0.rst
 
 .. include:: ../options/y_0.rst
+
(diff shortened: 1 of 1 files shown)


busstoptaktik (approved):
This PR makes the point, and warns sufficiently, that the two point method should probably never be used. A more full explanation of "how to use it anyway" would be out of scope (=noise) for the PROJ end user documentation. So while the full story would have been "nice-to-have" (although out of scope), this PR provides the "need-to-have" warning.

Issue #4030, opened 2024-02-02
Prime meridian inconsistency: Copenhagen


mwtoews:
[Copenhagen (EPSG:1026)](https://epsg.org/prime-meridian_1026/Copenhagen.html) is defined as 12° 34' 39.9" E, which is different to PROJ's [12d34'40.35"E](https://github.com/OSGeo/PROJ/blob/50e8dd72bc569c2844ec1f52e9dee0446b809228/src/datums.cpp#L82) with a difference of 0.45" or [approximately](https://geographiclib.sourceforge.io/cgi-bin/GeodSolve?type=I&input=55.6761+12d34%2739.9%22E+55.6761+12d34%2740.35%22E&format=g&azi2=f&unroll=r&prec=3&radius=6378137&flattening=1%2F298.257223563&option=Submit) 7.864 m.

This was added to PROJ with #721 (@busstoptaktik). This was followed with a [change by IOGP](https://epsg.org/closed-change-request/browse/id/2022.078) on 10 May 2023, also by @busstoptaktik, but it's not clear if the meridian value was changed.

Is to worth changing or adding meridians for PROJ?

busstoptaktik:
I'll have to investigate this further - my memory, but not my reading, tells me that I wrote a very long explanatory commit message telling about both the difference between the old and the new Copenhagen Observatory (in principle built on the same meridian, to continue observations consistently), the new about 1 km due north of the old, to move it out of the light pollution of the sprawling 19th century Danish capital (today the "new" observatory is very much in the city center). 

The difference may be due to the 2 observatories actually differing in longitude, or it may be due to different observations in different systems (e.g. System GS, using the Andræ ellipsoid, f=1/300, and System GI, using the international (Hayford) ellipsoid, f=1/297. But (see below) it may also be due to totally prosaic, empirical, and more current reasons...

I believe in principle, the longitudes in System GS were referred to the old observatory (Rundetårn/Runde Taarn "The Round Tower"). In my lab notes from when working out the [new transformations](https://epsg.org/closed-change-request/browse/id/2022.078), I have this to tell the eternity:

```
Legacy Danish systems often reckoned longitude from the meridian of the old
Copenhagen Observatory "Rundetårn" (active 1641–1861) [Borre, 2014].

From a poster [GAII, u.å], originally compiled for use in the department for
operational geodesy (G.A.II - "Geodætisk Afdeling II") of the Danish Geodetic
[...]

Issue #4029, opened 2024-02-02
Prime meridian inconsistency: Madrid


mwtoews:
[Madrid (EPSG:8905)](https://epsg.org/prime-meridian_8905/Madrid.html) is defined as 3° 41' 14.55" W, which is different to PROJ's [3d41'16.58"W](https://github.com/OSGeo/PROJ/blob/50e8dd72bc569c2844ec1f52e9dee0446b809228/src/datums.cpp#L73) with a difference of 2.03" or [approximately](https://geographiclib.sourceforge.io/cgi-bin/GeodSolve?type=I&input=40.4168+3d41%2714.55%22W+40.4168+3d41%2716.58%22W&format=g&azi2=f&unroll=r&prec=3&radius=6378137&flattening=1%2F298.257223563&option=Submit) 47.859 m. It seems that [EPSG changed](https://epsg.org/closed-change-request/browse/id/2020.056) this value on 1 July 2020 with the comment:

> For meridian 8905, previous value in the EPSG record was 3°41'16.58"W. This was not used by any other EPSG entities.

I'm not sure what this means entirely, but is to worth changing or adding meridians for PROJ?

---

A related note for `madrid` is that it briefly had a typo of `3d41'16.48"W` which was fixed with 84b8264c62947bb73306be9c54487209939597dc but the old value has persisted in the [docs](https://github.com/OSGeo/PROJ/blob/50e8dd72bc569c2844ec1f52e9dee0446b809228/docs/source/usage/projections.rst) for a while. I'll fix these docs after resolving other prime meridian issues.

rouault:
FWIW, when using EPSG CRS, PROJ will use the new value when going through PROJ4 strings:
```
$ bin/projinfo EPSG:2062
PROJ.4 string:
+proj=lcc +lat_1=40 +lat_0=40 +lon_0=0 +k_0=0.9988085293 +x_0=600000 +y_0=600000 +a=6378298.3 +rf=294.73 +pm=-3.687375 +units=m +no_defs +type=crs
```
So the change of the hardcoded +pm=madrid would affect only CRS definitions coming from PROJ.4 strings with explicit +pm=madrid

mwtoews:
Closing, with the hard-coded `+pm=madrid` unchanged.

To use EPSG:8905, use `+pm=3d41'14.55"W`

PR #4042, opened 2024-02-08
Freeze hard-coded prime meridians and note they are no longer updated

### LINKED ISSUE #4031: Missing Paris RGS prime meridian?
[Paris (EPSG:8903)](https://epsg.org/prime-meridian_8903/Paris.html) defined as 2.5969213 grad (×9/10=2.3372291699999996 deg) is very close to [2d20'14.025"E](https://github.com/OSGeo/PROJ/blob/50e8dd72bc569c2844ec1f52e9dee0446b809228/src/datums.cpp#L71) (according to [GeodSolve](https://geographiclib.sourceforge.io/cgi-bin/GeodSolve?type=I&input=48.8566+2.3372291699999996+48.8566+2d20%2714.025%22E&format=g&azi2=f&unroll=r&prec=6&radius=6378137&flattening=1%2F298.257223563&option=Submit) the difference is 0.245 mm), so no real issue here.

[Paris RGS (EPSG:8914)](https://epsg.org/prime-meridian_8914/Paris-RGS.html) (RGS=Royal Geographical Society) defined as 2° 20' 13.95" E is 0.075" or [approximately](https://geographiclib.sourceforge.io/cgi-bin/GeodSolve?type=I&input=48.8566+2d20%2713.95%22E+48.8566+2d20%2714.025%22E&format=g&azi2=f&unroll=r&prec=3&radius=6378137&flattening=1%2F298.257223563&option=Submit) 1.529 m away from the Paris datum. According to the metadata, this datum was replaced by IGN in 1936 to the preferred EPSG:8903 Paris meridian.

Despite this note, should this meridian be added to PROJ as `paris_rgs`? Or ignored, as it's unlikely to be used?

mwtoews:
This PR "freezes" hard-coded prime meridians that are listed in `src/datums.cpp`'s `PJ_PRIME_MERIDIANS` struct. This has only had a few modifications since it was added 22 years ago (27c197f9da8005a2fd58bb8d902beb0864a2916b), but has some omissions (#4031) and some values conflict with other sources (#4029, #4030). Also, there are potentially [many other prime meridians](https://en.wikipedia.org/wiki/Prime_meridian#List_of_historic_prime_meridians_on_Earth) that users may want to choose from.

This PR doesn't go as far as deprecating `cs2cs -lm`, actually it adds this missing doc item. However, a message is sent to stderr:
> <s>Hard-coded prime meridians are a legacy PROJ.4 feature.</s>
> This list is no longer updated, and some values may conflict with other sources.

Suggestions in rewording or changing tone are welcome.

This also updates docs/source/usage/projections.rst with a few corrections. The example is changed from `madrid` (which "[has had various determinations](https://epsg.org/prime-meridian_8905/Madrid.html)", including 84b8264c62947bb73306be9c54487209939597dc) to `lisbon` which I assume has stayed put there for a while now.

Closes #4031, #4029

### DIFF
--- a/docs/source/apps/cs2cs.rst
+++ b/docs/source/apps/cs2cs.rst
@@ -104,6 +104,10 @@ The following control parameters can appear in any order:
 
     List of all ellipsoids that can be selected with the *+ellps* parameters.
 
+.. option:: -lm
+
+    List of hard-coded prime meridians that can be selected with the *+pm* parameter.
+
 .. option:: -lu
 
     List of all distance units that can be selected with the *+units* parameter.
--- a/docs/source/apps/cs2cs.rst
+++ b/docs/source/apps/cs2cs.rst
@@ -106,7 +106,9 @@ The following control parameters can appear in any order:
 
 .. option:: -lm
 
-    List of hard-coded prime meridians that can be selected with the *+pm* parameter.
+    List of hard-coded prime meridians that can be selected with the *+pm*
+    parameter.  Note that this list is no longer updated,
+    and some values may conflict with other sources.
 
 .. option:: -lu
 
--- a/docs/source/development/reference/datatypes.rst
+++ b/docs/source/development/reference/datatypes.rst
@@ -608,7 +608,8 @@ List structures
 
 .. c:type:: PJ_PRIME_MERIDIANS
 
-    Prime meridians defined in PROJ.
+    Hard-coded prime meridians defined in PROJ. Note that the structure is
+    no longer updated, and some values may conflict with other sources.
 
     .. code-block:: C
 
--- a/docs/source/usage/projections.rst
+++ b/docs/source/usage/projections.rst
@@ -70,10 +70,10 @@ The returned coordinates are scaled by the value assigned with the ``+k_0``
 parameter. This parameter is only used by projections that mention using it,
 and its exact effect is projection dependent.
 
-Input units for angular parameters (``+lon_0``, ``+lat_0``, etc.) are
+Input units for angular parameters (``+lon_0``, ``+lat_0``, ``+pm``, etc.) are
 interpreted to be decimal degrees by convention.
 Explicit specification of input angular units can be accomplished by adding the
-appropriate ruffix to input values.
+appropriate suffix to input values.
 
 
     +----------------+---------------------+
@@ -90,9 +90,23 @@ appropriate ruffix to input values.
     | R              |                     |
     +----------------+---------------------+
 
+The sign of an angle is taken from either a ``-`` or ``+`` prefix,
+or the last character specifying a cardinal direction,
+where ``e``/``E`` (East) or ``n``/``N`` (North) are positive
+and ``w``/``W`` (West) or ``s``/``S`` (South) are negative.
+
 Example of use.  The longitude of the central meridian ``+lon_0=90``, can also be expressed more explicitly
 with units of decimal degrees as ``+lon_0=90d`` or in radian
 units as ``+lon_0=1.570796r`` (approximately).
+
+Angles can be expressed in DMS notation, using ``'`` after arcminutes
+and ``"`` after arcseconds, ending with optional cardinal direction.
+For example, ``+pm=3d41'14.55"W``, but
+character escapes ``+pm=3d41\'14.55\"W`` may be required.
+
+Degree-minute notation does not require any quotation symbols.
+All of these are equivalent values: ``+pm=-17d40``, ``+pm=17D40W``,
+``+pm=17°40w`` or ``+pm=17d40'W`` (escaped as ``+pm=17d40\'W``).
 
 
 False Easting/Northing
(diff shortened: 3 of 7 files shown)


docs/source/development/reference/datatypes.rst
```diff
@@ -608,7 +608,8 @@ List structures
 
 .. c:type:: PJ_PRIME_MERIDIANS
 
-    Prime meridians defined in PROJ.
+    Hard-coded prime meridians. This is a legacy PROJ.4 feature,
```
rouault: PROJ.4 by itself is legacy. Prime meridians aren't more legacy than +datum, +ellps or +units, are they?

src/apps/cs2cs.cpp
```diff
@@ -580,8 +580,11 @@ int main(int argc, char **argv) {
                         }
                         proj_unit_list_destroy(units);
                     } else if (arg[1] == 'm') { /* list prime meridians */
+                        (void)fprintf(stderr,
```
rouault, line 583: I wouldn't emit a warning. If we want to emit one, then we should also do it for datum, ellipsoid, units, etc

rouault: I'm fine with the second sentence "This list is no longer updated, and some values may conflict with other sources"

mwtoews: Listing datums was already removed (e9605e41c87e532e7cc182c971e068de015adf2d).
Listing ellipsoids could potentially show a similar message.
Listing units uses `proj_get_units_from_database` so is fine.

rouault: Feel free to disregard my comments. I'm not that strong about them.

The only thing that is a bit weird is that internally PROJ still creates PROJ pipeline strings with those deprecated features, but one could argue that is an implementation detail. Well, actually not completely true, as we also export that to the sight of the advanced user...)
```
$ bin/projinfo -s "NTF (Paris)" -t "NTF" -o PROJ -q
+proj=pipeline
  +step +proj=unitconvert +xy_in=grad +xy_out=rad
  +step +inv +proj=longlat +ellps=clrk80ign +pm=paris
```

mwtoews: I think it's fine to have a short list of hard-coded of prime meridians (e.g. where `paris` ↔ 2.33722917) with the caveat that it doesn't change. It's a bit deprecated, but I wasn't planning to establish this as a deprecated feature. But yes, I can tone down the "legacy PROJ.4" wording throughout.

docs/source/development/reference/datatypes.rst
mwtoews: "legacy PROJ.4" wording removed.

rouault (approved):

docs/source/usage/projections.rst
```diff
@@ -135,27 +135,28 @@ Note that ``+over`` does **not** disable ``+lon_wrap``.
...
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-A prime meridian may be declared indicating the offset between the prime
-meridian of the declared coordinate system and that of greenwich.  A prime
-meridian is declared using the "pm" parameter, and may be assigned a symbolic
-name, or the longitude of the alternative prime meridian relative to greenwich.
+A prime meridian may be declared indicating the longitude offset between
+the prime meridian of the declared coordinate system and that of greenwich.
+A prime meridian is declared using the ``+pm`` parameter, and may be assigned
+an angle in DMS or decimal degrees format, or as a hard-coded name.
```
kbevers, line 155: I think an example of not using a hard-coded name would be useful. Something like

```
echo 13.5 45.0 | proj +proj=merc +pm=13.5         
0.00    5591295.92
```
could be placed in the example section below.

src/apps/cs2cs.cpp
kbevers: The above pipeline could also have `+pm` expressed in decimal degrees, so the PROJ internals could be changed to do that in the future, should we decide to drop support for these hard-coded lists.

docs/source/apps/cs2cs.rst
```diff
@@ -104,6 +104,10 @@ The following control parameters can appear in any order:
 
     List of all ellipsoids that can be selected with the *+ellps* parameters.
 
+.. option:: -lm
+
+    List of hard-coded prime meridians that can be selected with the *+pm* parameter.
```
kbevers: A note here stating that the list is hard-coded would be in order, I think. Basically the same as what is written for `PJ_PRIME_MERIDIANS`.

kbevers (approved):
I've added a few comments with suggestions but in general I am happy with this PR

docs/source/apps/cs2cs.rst
mwtoews: Done.

docs/source/usage/projections.rst
mwtoews: Done, but modified slightly. Further mods welcome.

mwtoews:
Thanks for feedback @kbevers. I've taken the moment to further expand the docs on angular inputs too, which I wouldn't mind a bit of review on too.

kbevers:
> which I wouldn't mind a bit of review on too.

Looks great. Good job!

PR #4059, opened 2024-02-20
Enable -Weffc++ for gcc and clang

rouault:


jjimenezshaw (approved):
I am curious how you ended up here, when there is not code change in the PR (so I assume that the current code works perfectly with that option).

In https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html I read "When selecting this option, be aware that the standard library headers do not obey all of these guidelines; use ‘grep -v’ to filter out those warnings." Are we affected?

rouault:
> I am curious how you ended up here, when there is not code change in the PR (so I assume that the current code works perfectly with that option).

Well, someone makes me resurrect, in a custom branch, autotools support for PROJ master, and autotools builds had -Weffc++ enabled, hence I noticed a few warnings, which I directly fixed in master per https://github.com/OSGeo/PROJ/commit/26fb427e8dce36d018a0417b1d0f31f9e4e96411 and https://github.com/OSGeo/PROJ/commit/81fecbc24329abf9d952bcdbfc7ebd31d376caf6 . So this is to avoid such occurrences to happen again

> In https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html I read "When selecting this option, be aware that the standard library headers do not obey all of these guidelines; use ‘grep -v’ to filter out those warnings." Are we affected?

I've not noticed that. At worse we can disable it if it becomes an annoyance but this is enabled in GDAL which draws much more stuff than PROJ, so I won't anticipate any issues for PROJ

mwtoews:
It seems that I forgot to check over these flags from https://github.com/OSGeo/PROJ/blob/8.2/configure.ac

Are there others? Here are some that were mentioned in the old `configure.ac` but not transferred into `CMakeLists.txt`:
- [ ] `-Winit-self`
- [ ] `-Wshorten-64-to-32`
- [ ] `-Wlogical-op`
- [ ] `-Werror=` various, including `vla`, `format-security`, `declaration-after-statement`
- [ ] `-Wdate-time`
- [ ] `-Wnull-dereference`
- [ ] `-Wduplicated-cond`
- [ ] `-Wduplicated-branches` (possibly)
- [ ] `-Wextra-semi`
- [ ] `-Wcomma`
- [ ] `-Wdocumentation -Wno-documentation-deprecated-sync`
- [ ] `-Wunused-private-field`
- [ ] `-Wnon-virtual-dtor`
- [ ] `-Wold-style-cast`
- [ ] `-Woverloaded-virtual`
- [ ] `-Wweak-vtables`
- [ ] `-Wdeprecated`
- [ ] `-Wabstract-vbase-init`
- [ ] `-Winconsistent-missing-destructor-override`
- [ ] `-Wzero-as-null-pointer-constant`

I realize that some of these may already be enabled by default, or enabled with (e.g.) `-Wall`, but there are so many different compiler options to be certain.

rouault:
@mwtoews  You could have a look/borrow from GDAL's gdal.cmake lines 100 to 200.
Looking at https://clang.llvm.org/docs/DiagnosticsReference.html and https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html, potential candidates could -Wcomma -Wextra-semi -Wold-style-cast -Wunused-private-field  -Wzero-as-null-pointer-constant -Wshorten-64-to-32 -Wlogical-op -Wduplicated-cond -Wduplicated-branches -Wdocumentation -Wdeprecated -Woverloaded-virtual -Wdate-time -Wweak-vtables -Werror for vla , format-security and declaration-after-statement
-Winit-self is now enabled by default on gcc by -Wall.
-Wnon-virtual-dtor and -Weffc++ are synonymous with clang

PR #4061, opened 2024-02-21
Add more compiler warning flags

mwtoews:
As discussed [here](https://github.com/OSGeo/PROJ/pull/4059#issuecomment-1955583676), this PR adds more compiler warning flags that were lost from the Autotools -> CMake transition. See the old [configure.ac](https://github.com/OSGeo/PROJ/blob/8.2/configure.ac) file for what was previously used.

A few warnings are resolved in this PR as separate commits.

Warnings that are not-yet enabled or resolved in this PR include:

- [ ] `-Wold-style-cast` - this might take a while to sort-out
- [ ] `-Wweak-vtables` - probably simple, but not obvious

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,6 @@ if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
     -Wlogical-op
   )
   set(PROJ_C_WARN_FLAGS ${PROJ_common_WARN_FLAGS}
-    -Werror=declaration-after-statement
     -Wmissing-prototypes
   )
   set(PROJ_CXX_WARN_FLAGS ${PROJ_common_WARN_FLAGS}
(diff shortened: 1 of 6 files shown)


rouault (approved):

CMakeLists.txt
```diff
@@ -61,8 +61,7 @@ if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
     -Wlogical-op
   )
   set(PROJ_C_WARN_FLAGS ${PROJ_common_WARN_FLAGS}
-    # -Werror=declaration-after-statement
-    -Wdeclaration-after-statement
+    -Werror=declaration-after-statement
```
rouault: looking at that more closely, we should drop that commit. We are now at C99 which allows declarations after statements, which isn't a bad practice

mwtoews: I was thinking to mention this. It was added in 2013 to support old MSVC compilers, before C99 was required for PROJ.

mwtoews:
@rouault if you have an easy fix for `-Wweak-vtables` I can add it here. Addressing `-Wold-style-cast` would probably need a separate PR (not by me) as it may involve choices of appropriate [cast alternatives](https://stackoverflow.com/a/5249436/).

PR #4064, opened 2024-02-25
CI: add support for Conda osx_arm64 builds

rouault:
Now that
https://github.com/conda-incubator/setup-miniconda/releases/tag/v3.0.2 has been released with support for them.


rouault:
merging directly, as it runs only on push

PR #4066, opened 2024-03-04
[Doc] Rename NEWS and HOWTO-RELEASE to *.md

jjimenezshaw:
On the last release I noticed that `NEWS` (and later `HOWTO-RELEASE`) files are formatted as Markdown, but not named with the extension `md`. That "disables" the automatic formatting in GitHub.

This PR is renaming these two files, and adjusting a bit the formatting where it is needed. The display in GitHub is nicer. The links to the issues in NEWS.md should work automatically, for example (or it should be).

The file `NEWS` was renamed to `NEWS.md`. I replaced all ` o ` with ` * ` (something suggested to create the rst, by the way). That are many lines, that makes GitHub to show them in the PR as different files, not a rename. If this is a problem I can make two PRs.

If there are other problems that I do not foresee now (including backwards compatibility), I can just discard this PR.


kbevers:
> On the last release I noticed that NEWS (and later HOWTO-RELEASE) files are formatted as Markdown

It's quite a stretch to regard the NEWS file as markdown-formatted. It's been around for a couple of decades before anyone thought of markdown :)

> I replaced all o with * (something suggested to create the rst, by the way). That are many lines, that makes GitHub to show them in the PR as different files, not a rename. If this is a problem I can make two PRs.

Regarding the formatting, it would be beneficial to also remove the indentation of the subsections. I'll make a suggestion to demonstrate what I mean.

When preparing the releases I probably spend the most time moving stuff from `NEWS` to `docs/source/news.rst`. The changes in this PR is a step in the right direction to bring that time down. Ideally it would be as simple as copying the text from `NEWS` to `docs/source/news.rst` but I think that would requite the NEWS file to be rst-formatted. It might be possible to do an auto-conversion but my experiences with that in past haven't been that good.

> If there are other problems that I do not foresee now (including backwards compatibility), I can just discard this PR.

I've been considering this for a long time but haven't bothered. I think it is a good change. It might be worthwhile to check in with a few of the packagers to see if it causes problems on their end.

GitHub can now prepare release notes automatically from PRs. I did that for PROJ-data this time around and with the right discipline it could work for PROJ as well, which would make preparing the release notes a brease: Click the button and copy/paste the Markdown to `NEWS.md`.

NEWS.md
```diff
@@ -0,0 +1,3141 @@
...
+-------------------
+
+ Updates
+ -------
+
+ * Add support for transformations involving coordinate epoch changes
+   (PointMotionOperation), specifically for Canadian NAD83(CSRS)(#3884)
+
+ * SSL_OPTIONS: set SSL options on Curl library (#3936)
+
```
kbevers: This indendation is unnecessary:

```suggestion
Updates
-------

* Add support for transformations involving coordinate epoch changes
  (PointMotionOperation), specifically for Canadian NAD83(CSRS)(#3884)

* SSL_OPTIONS: set SSL options on Curl library (#3936)
```

jmckenna:
thanks for these changes @jjimenezshaw, much appreciated by the users, especially on Windows as they cannot open a file with no extension easily.  Thanks!

hobu:
Renaming NEWS and README is likely to break a bunch of packagers scripts and workflows.

jmckenna:
(I'm one of those packagers, for the Windows community) This is a case of benefits for the users, versus a few changes required in a packager's workflow (as a packager, I see my role as focused on users, and the more pain I take in packaging, the easier it is for the user).  Other packagers my feel differently.

jjimenezshaw:
> Regarding the formatting, it would be beneficial to also remove the indentation of the subsections

Done in a single commit

jjimenezshaw:
> Renaming NEWS and README is likely to break a bunch of packagers scripts and workflows.

That was one of my fears. That is why I make the PR, so people can see the impact and comment the pros and cons. @jmckenna is already happy.
With more comments we can decide what we can/want do.
I changed only these two files because they are long and already with some format that Markdown is understanding (more or less). If having files without extension is a problem I can do something similar to `AUTHORS`, `CITATION`, `COPYING` and `INSTALL`.

jjimenezshaw:
> Renaming NEWS and README

`README` is already `README.md`. I am modifying `NEWS` and `HOWTO-RELEASE`.

busstoptaktik:
> `README` is already `README.md`. I am modifying `NEWS` and `HOWTO-RELEASE`.

...and IIRC, the change from `README` to `README.md` was actually surprisingly involved. These cases are probably simpler

mwtoews:
Note that the packaged README in the source dist is derived from README.md with some modifications done by [ProjReadme.cmake](https://github.com/OSGeo/PROJ/blob/master/cmake/ProjReadme.cmake)

jjimenezshaw:
What should we do now? Should I ask in the mailing list to other packagers?
So far only @jmckenna as packager said anything.

About renaming with cmake, would that be really needed? As Jeff mentioned, having a meaningful extension is convenient.

kbevers:
> Should I ask in the mailing list to other packagers?

Yeah, I think that would be good idea.

rouault:
It shouldn't be hopefully too hard for packagers to adapt their recipee for a filename change. We just need to communicate the change clearly in the release announcement.
@mwtoews do you remember why we did that step of "converting" README.md to README at generation time? For the sake of backward compatibility? Or perhaps this was done at a time where we had to support autotools, which has some builtin assumptions about some standard such files.
GDAL typically provides a NEWS.md file since quite a few releases, and it hasn't prevented anyone from packaging it as far as I know. This is handled in the Debian recipee at https://salsa.debian.org/debian-gis-team/gdal/-/blob/master/debian/rules?ref_type=heads#L132

mwtoews:
As far as I can tell (via #2690) the README conversation was to allow `make dist` to be backwards compatible with autotools. Forward looking we could just make things simpler and just add README.md to the source distribution.

jjimenezshaw:
> > Should I ask in the mailing list to other packagers?
> 
> Yeah, I think that would be good idea.

One week and no answer to https://lists.osgeo.org/pipermail/proj/2024-March/011314.html

NEWS.md
```diff
@@ -0,0 +1,3141 @@
+9.4.0 Release Notes
+-------------------
+
+Updates
+-------
+
```
kbevers: I think it would be better to use the standard markdown headings
```suggestion
# 9.4.0 Release Notes

## Updates

```

Without the indents and the same symbols used for denoting the heading it is not really clear that "Updates" is a subsection of "9.4.0 Release Notes".

Applies across the whole document.

NEWS.md
```diff
@@ -0,0 +1,3141 @@
...
+development.
+
+!!!
+!!! Therefore we have scheduled regular releases over the coming years
+!!! which will gradually remove the old programming interface.
+!!!
+!!! This will cause breaking changes with the next two major version
+!!! releases, which will affect all projects that depend on PROJ
+!!! (cf. section "deprecations" below).
+!!!
```
kbevers, line 2464: Perhas this can be expressed in a more markdown like manner?

kbevers: There's a similar thing further down

kbevers (changes requested):

NEWS.md
jjimenezshaw: Done. I hope I made all of them

NEWS.md
jjimenezshaw: I used a block. There is not so much options in markdown.

NEWS.md
jjimenezshaw: Cool! that way you can link to an specific section, like https://github.com/jjimenezshaw/PROJ/blob/md-files/NEWS.md#700-release-notes

kbevers (approved):

kbevers:
Looks good to me

mwtoews:
I think this PR should also remove README from the source dist (listed [here](https://github.com/OSGeo/PROJ/actions/runs/8307649102/job/22737106532?pr=4066#step:4:1618)), which can be done by removing [this block](https://github.com/OSGeo/PROJ/blob/60f20de1bc88d89c8eb95ac3a79bd9166804bb81/CMakeLists.txt#L421-L427) and deleting `cmake/ProjReadme.cmake`.

jjimenezshaw:
> I think this PR should also remove README from the source dist (listed [here](https://github.com/OSGeo/PROJ/actions/runs/8307649102/job/22737106532?pr=4066#step:4:1618)), which can be done by removing [this block](https://github.com/OSGeo/PROJ/blob/60f20de1bc88d89c8eb95ac3a79bd9166804bb81/CMakeLists.txt#L421-L427) and deleting `cmake/ProjReadme.cmake`.

Should it be in this PR, or in a later one? I am fine doing another PR.

kbevers:
> Should it be in this PR, or in a later one?

I think it fits quite well in this one

mwtoews (approved):
Thanks, looks good.

jmckenna:
thanks again @jjimenezshaw !

PR #4067, opened 2024-03-05
geodetic 27% faster

wblangdon:
Speed up geodetic(PJ_XYZ cart, PJ *P) 27%

In normal_radius_of_curvature() remove if (es == 0) as unneeded and actually 1.5% slower with it

Add invhypot()     so can replace division by multiplication with reciprocal, double precision multiplication is considerably faster than double precision division

Use invhypot() in geocentric_radius() and geodetic()

In geodetic() store result in intermediate variables rather than call a function multiple times.

Passes all 9.3.1 ctest
No changes to API
Magpie version cart.cpp_29-feb-2024


rouault:
> Use invhypot()

do you have a reference to a standard that defines that function? I can hardly find any reference to it. "grep -r invhypot /usr/include" doesn't get any match on my Ubuntu 20.04, there's no match for it in https://learn.microsoft.com/en-us/search/?scope=C%2B%2B&view=msvc-170&terms=invhypot . So I'm wondering if it is available on all platforms. Perhaps we should have a CMake check for its availability

rouault:
nevermind my above comment about invhypot. I missed it was is defined in your code...

src/conversions/cart.cpp
```diff
@@ -160,16 +163,16 @@ static PJ_LPZ geodetic(PJ_XYZ cart, PJ *P) {
 #else
     const double y_theta = cart.z * P->a;
     const double x_theta = p * P->b;
-    const double norm = hypot(y_theta, x_theta);
-    const double c = norm == 0 ? 1 : x_theta / norm;
-    const double s = norm == 0 ? 0 : y_theta / norm;
+    const double invnorm = invhypot(y_theta, x_theta);
```
rouault, line 181: why not just storing here 1. / hypot(y_theta, x_theta); (same below) ?
I'm uncomfortable with a function called invhypot() that is a lower accuracy version of 1. / hypot()

src/conversions/cart.cpp
```diff
@@ -121,8 +124,8 @@ static double geocentric_radius(double a, double b, double cosphi,
         This is from WP2, but uses hypot() for potentially better
         numerical robustness
     ***********************************************************************/
-    return hypot(a * a * cosphi, b * b * sinphi) /
-           hypot(a * cosphi, b * sinphi);
+    return hypot(a * a * cosphi, b *  b * sinphi) *
+       invhypot(a * cosphi, b * sinphi);
```
rouault, line 137: I guess that will cause both a multiplication and a division to happen here...

src/conversions/cart.cpp
wblangdon: As you have seen invhypot is simply doing 1.0/sqrt(x*x + y*y) this is all in double precision. Ok, glibc seems to be claiming better than 2ulp v. 1ulp for hypot(). Will investigate. Except ctest, are there other special cases I should look at?
Bill

rouault: I see that cart.cpp was mostly the work of @busstoptaktik . I let it comment about the appropriateness of this change
We don't have a better test suite than the one run by ctest.

src/conversions/cart.cpp
wblangdon: I did try gathering a*cosphi and b*sinphi togther but (at least with -O3) that did not make a consistent improvement.
BTW GCC -Ofast beats -O3 here but breaks the code elsewhere. I guess it would be possible to use -Ofast selectively but that seemed like it would complicate the build over much.
Will investigate lines 127-128 further.
Bill

src/conversions/cart.cpp
wblangdon: As an additional sanity check I have rerun the latest ctest (master test/ Mar  1 17:10),
with g++ (GCC) 10.2.1. As expected, the master and the pull request behave the same. Ie both pass all ctest (except test 2 - geodesic-signtest, were, as noted in the comments,  glibc's cos(810) gives 6.12323e-17, rather than 0, due to rounding error in remquo).

busstoptaktik: > I see that cart.cpp was mostly the work of @busstoptaktik .

While that is formally correct wrt. to the initial crude work, you @rouault did all the smart stuff speeding up the computations afterwards, so wrt. the statement

> I let it comment about the appropriateness of this change

I think you'll have more informed opinions on the speed and accuracy side of things.

busstoptaktik: ... however, I have a number of comments on the subject over in the Rust Geodesy version of [cart](https://github.com/busstoptaktik/geodesy/blob/main/src/ellipsoid/cartesians.rs).

The paper by Fukushima quoted there (full reference [here](https://github.com/busstoptaktik/geodesy/blob/main/src/bibliography/mod.rs)) stubbornly insists on using `sqrt(x*x+y*y)`, and I think I have seen timings elsewhere claiming it to be 30 times faster than `hypot(x, y)`, which on the other hand is ½ulp more accurate...

EDIT: Also, the implementation used in the [cart operator](https://github.com/busstoptaktik/geodesy/blob/main/src/inner_op/cart.rs) differs somewhat for the method implemented for the Ellipsoid. I do not remember why I switched to being based on the horrid (but admirably brief) Fortranesque code from Claessens' review paper.

rouault: > While that is formally correct wrt. to the initial crude work, you @rouault did all the smart stuff speeding up the computations afterwards,

doh, I had already almost forgotten about that. You must refer to https://github.com/OSGeo/PROJ/commit/78c1df51e0621a4e0b2314f3af9478627e018db3

@wblangdon Could you try the bench code used in that commit? Or maybe the "./bin/bench_proj_trans" that is built with "make bench_proj_trans" does the job too.
We have a test failure on Linux 32 bit: https://github.com/OSGeo/PROJ/actions/runs/8157690552/job/22299140174?pr=4067

wblangdon: sure:-)
I will be out of the office most of tomorrow (wednesday)
The system here defaults to 64bit, should be possible to switch it down to 32bit.
Bill

jjimenezshaw: Out of curiosity, I checked what the compiler is doing, and it was not helping me so much:
https://godbolt.org/z/en4Tend7x
Probably a `jmp` to `hypot` takes longer than a bunch of operations, including `sqrtsd`.

Have you tested the performance in in Windows?
https://godbolt.org/z/eKn7n6P3v

busstoptaktik: > Probably a `jmp` to `hypot` takes longer than a bunch of operations, including `sqrtsd`.

@jjimenezshaw yes, because in general `hypot` implementations [are littered with defensive coding](https://opensource.apple.com/source/Libm/Libm-47.1/ppc.subproj/hypot.c.auto.html) to squeeze any additional half ulp of accuracy out in the result. `hypot` is for accuracy, not for speed

src/conversions/cart.cpp
```diff
@@ -105,12 +105,15 @@ PROJ_HEAD(cart, "Geodetic/cartesian conversions");
 /*********************************************************************/
 static double normal_radius_of_curvature(double a, double es, double sinphi) {
     /*********************************************************************/
```
busstoptaktik, line 110: Could perhaps add `inline` here, now the code is touched anyway

src/conversions/cart.cpp
```diff
@@ -105,12 +105,15 @@ PROJ_HEAD(cart, "Geodetic/cartesian conversions");
...
-    if (es == 0)
-        return a;
+    /*Faster without this optimisation** if (es == 0)
+        return a;*/
     /* This is from WP.  HM formula 2-149 gives an a,b version */
     return a / sqrt(1 - es * sinphi * sinphi);
 }
 
+inline double invhypot(const double x, const double y) {
```
busstoptaktik, line 118: `invhypot` is used once only, so I do not think it is worth separating out, hence littering the namespace with a function name that is "not entirely obvious"

EDIT: Oops - I now see that it is actually used twice. I still think the name is more confusing than clarifying, however, so like @rouault I would really prefer manually inlining

src/conversions/cart.cpp
busstoptaktik: > I did try gathering a_cosphi and b_sinphi togther but (at least with -O3) that did not make a consistent improvement

No, because hopefully the optimizer will do that kind of folding even without being prompted to

wblangdon: Just a quick comment: I think from looking at GCC -O3 -S output, as you suggest the compiler does indeed inline and yip you are right that in lines 127-128 multiply by invhypot() is indeed converted to division in the assembly code.  (GCC  version 10.2.1, mock.s_divsd_mulsd_invhypot.cpp).

src/conversions/cart.cpp
wblangdon: As suggested I have tried 78c1df5,
https://github.com/OSGeo/PROJ/commit/78c1df51e0621a4e0b2314f3af9478627e018db3

I have run 78c1df5 as a benchmark (-O3 -DNDEBUG GCC 10.2.1, i7-4790 CPU @ 3.60GHz) 
with "fwd" the pull request is much the same (total, fwd+inv 9%) but with inv
alone the pull request is on average 17% faster.

I am away until Monday/Tuesday, plan next to investigate 32 bit.
Bill

wblangdon:
quick update: so far I am trying g++ -m32, will restart this afternoon. Hmm stuck on 32bit cmake. It looks like we may be missing a 32 bit system library libstdc++_nonshared.a. Have put a call to our tech support. Tech support say newest redhat/centos GCC does not support 32 bit runtime library. Neil suggests using old version of GCC. I am trying gcc 4.8.5 but now looking for compatible sqlite3 which is 32 bits. In the end could not get cmake here to produce 32 bit version, so have put new code inside conditional compilation so invhypot does not even get compiled if in 32bit mode. Naturally still passes 64bit version of ctest. Now running automated tests...

wblangdon:
Magpie version cart.cpp_14-mar-2024 appears to have pass 64bit tests but failed in 32bit mode just like the first version:-(
I have the log file, 0_linux_gcc_32bit.txt, is it possible to get more information?
  Is it possible to see g++ command lines?
  Is it possible to download .o and .so files?
  Is it possible to download 'target dist' ?
Bill

rouault:
See PR #4087 for an alternative implementation

rouault:
closing as superseded with now merged PR #4087

PR #4081, opened 2024-03-11
Update geod.rst to reduce head scratching

jidanni:


rouault:
@jidanni For your next doc pull requests, make sure to use a https://proj.org/en/9.4/ page, not a 9.3, now that we are at 9.4

jidanni:
It turns out, if the user simply goes to https://proj.org/ , and then choose the page he wants to edit and then presses  "Edit on GitHub" there in the upper right hand corner of the page, everything will work out fine.

The problem comes in if the user goes to Google and types in "proj geod". Then he will end up on the older version, and pressing "Edit on GitHub" will  end up editing the older version.

For some reason this is only a problem with Proj, not a problem at all with Gdal.

Something about having a different URL structure, even though both are on GitHub.

Yes, the user could always be sure to, in the menu on the left side of Proj pages, choose "latest" , (even if it's a mystery there in the menu what version number "latest"  actually equals. Well maybe it doesn't exactly equal a certain version.)
(However that is relying on user remembering to choose "latest".)

rouault:
> For some reason this is only a problem with Proj, not a problem at all with Gdal.

we use ReadTheDocs multi-versioning for PROJ. Ok so this is related to Google caching. Hopefully new 9.4 pages will be indexed and prefered over time. CC @mwtoews @kbevers

jjimenezshaw:
> For some reason this is only a problem with Proj, not a problem at all with Gdal.

PROJ documentation is in "readthedocs" where we have different versions, while GDAL is hosted in a different way, that has no "versions". In GDAL you can only change "master".

jidanni:
Maybe there should be a warning when the user attempts to edit an older version. That's what happens on Wikipedia...

rouault:
> Maybe there should be a warning when the user attempts to edit an older version

A potential solution would be to have html_context["github_version"] = "master", including in release branches, so that users always edit the master version. That may a bit surprising to them however as what they will be proposed to edit will not always match the version of the HTML they started from

jidanni:
Another problem is one users want to share one of these projects documentation pages 9 out of 10 times they just want to share a link that stays fresh.

Currently they need to do a lot of research too make it point to the latest version with the word latest in the url.

Now let's think about how Wikipedia handles sharing.
The user is always going to get a link that points the current version of a page. If they really want to link to an old version then they have to do the research, and poke around the side panels etc.

When it's true, for legal purposes, there always should be a way to link to a frozen version. But not out of 10 times that's not what you want. 

And you're really paying the price with Google, because it's not going to index the site more than oh, 45 days at a time.
And by that time you might be already ready to move on to 9.5.

Come to think of it Google is littered with links to various old versions of various Proj things.

Let's say some man page from version 6.1 had a lot of clicks (on the Google search results page).
Well as that is still a legitimate URL, it'll still stay way up at top of search results.

The cure is to not let Google get its hands on any more of these versioned urls.

Also maintaining 100 different copies of the website... is nice except... Well I would move it offline to avoid confusion.

Of course you might say what if somebody's running a slightly older version and would like to see some documents. Well that's sort of connects to my other bug report about making sure the documents are always available offline too... 

Anyway, with GDAL there's no such worries... Oops, except the offline documents...

You're probably curious how did I manage to type in such a long winded poorly researched comment here on Github? Chalk it up to the miracle of voice typing.

PR #4084, opened 2024-03-13
proj_normalize_for_visualization: check other left-handed orders, not…

jjimenezshaw:
… only N-E

A few CRSs use left-handed orders other than N-E, like S-W in [EPSG:5513](https://spatialreference.org/ref/epsg/5513/)

 - [x] Closes https://github.com/qgis/QGIS/issues/56802
 - [x] Tests added

### DIFF
--- a/src/iso19111/crs.cpp
+++ b/src/iso19111/crs.cpp
@@ -870,8 +870,14 @@ static bool mustAxisOrderBeSwitchedForVisualizationInternal(
     const std::vector<cs::CoordinateSystemAxisNNPtr> &axisList) {
     const auto &dir0 = axisList[0]->direction();
     const auto &dir1 = axisList[1]->direction();
-    if (&dir0 == &cs::AxisDirection::NORTH &&
-        &dir1 == &cs::AxisDirection::EAST) {
+    if ((&dir0 == &cs::AxisDirection::NORTH &&
+         &dir1 == &cs::AxisDirection::EAST) ||
+        (&dir0 == &cs::AxisDirection::EAST &&
+         &dir1 == &cs::AxisDirection::SOUTH) ||
+        (&dir0 == &cs::AxisDirection::SOUTH &&
+         &dir1 == &cs::AxisDirection::WEST) ||
+        (&dir0 == &cs::AxisDirection::WEST &&
+         &dir1 == &cs::AxisDirection::NORTH)) {
         return true;
     }
 
--- a/test/unit/test_c_api.cpp
+++ b/test/unit/test_c_api.cpp
@@ -4456,6 +4456,29 @@ TEST_F(CApi, proj_normalize_for_visualization_on_crs) {
 
 // ---------------------------------------------------------------------------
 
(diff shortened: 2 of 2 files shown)


src/iso19111/crs.cpp
```diff
@@ -881,8 +881,14 @@ static bool mustAxisOrderBeSwitchedForVisualizationInternal(
...
-    if (&dir0 == &cs::AxisDirection::NORTH &&
-        &dir1 == &cs::AxisDirection::EAST) {
+    if ((&dir0 == &cs::AxisDirection::NORTH &&
+         &dir1 == &cs::AxisDirection::EAST) ||
+        (&dir0 == &cs::AxisDirection::EAST &&
+         &dir1 == &cs::AxisDirection::SOUTH) ||
+        (&dir0 == &cs::AxisDirection::SOUTH &&
+         &dir1 == &cs::AxisDirection::WEST) ||
+        (&dir0 == &cs::AxisDirection::WEST &&
+         &dir1 == &cs::AxisDirection::NORTH)) {
```
rouault, line 880: I'm not convinced that just switching the axis is enough for those 3 coordinate system. We should also presumably transform southing to northing, westing to easting, and make sure the final order is easting, northing

jjimenezshaw: That is also changing other problem, that westing-southing is displayed in QGIS with Antartica at the top. In that case moving up in the screen will make the coordinates decrease. Is that normal?

I do not mean that it should not be changed (well, I got used to it in South Africa). But the initial problem in QGIS was the mirroring effect.

rouault: > In that case moving up in the screen will make the coordinates decrease. Is that normal?

I guess so.

I believe all those coordinate systems should have easting,northing axis order after normalize_for_visualization. The native orders are just conventions to avoid specifying negative values, or other esoteric reasons, but ultimately you want your map with increasing eastings towards the right of the screen, and increasing northings towards the top of the screen

Maybe @thareUSGS can confirm for planetary CRS how maps in a CRS like  IAU_2015:19901 "Mercury (2015) / Ographic" which has northing,westing  should be displayed. I presume this should be transformed as easting,northing ?

jjimenezshaw: > I believe all those coordinate systems should have easting,northing axis order after normalize_for_visualization. The native orders are just conventions to avoid specifying negative values, or other esoteric reasons, but ultimately you want your map with increasing eastings towards the right of the screen, and increasing northings towards the top of the screen

I have been thinking about this a bit more. Currently in QGIS (and also with this PR), using a WSU system, easting/westing is increasing towards the right of the screen, and northing/southing is increasing towards the top of the screen (for that reason Antarctica is at the top). I think in PROJ we cannot do anything to avoid it.

If QGIS wants to display Antarctica at the bottom with such CRS, it is something to be done in the display, QGIS.

jjimenezshaw:
As we were talking in the code sprint, I think we cannot go further than this. In QGIS the user can rotate the scene in the status bar. But the mirror effect cannot be undone in QGIS.

PR #4085, opened 2024-03-13
Database: update to EPSG v11.006

jjimenezshaw:

PR #4086, opened 2024-03-14
[Backport 9.4] Database: update to EPSG v11.006

rouault:
Backport of second commit of https://github.com/OSGeo/PROJ/pull/4085


rouault:
The slight doubt I had about the appropriateness of the backport is that the WGS84 datum ensemble has gained a new datum realization with "World Geodetic System 1984 (G2296)", which thus changes the WKT2:2019 output of all EPSG:4326 based CRS. This could potentially break code (particular unit tests) that would do strict text comparison. But as far as GDAL is concerned, I didn't have tou change any unit tests (perhaps we already bit that bullet in the past with prior additions of new realizations so our tests are now robust to that)
Furthermore the isEquivalentTo() method considers CRS WKT definitions that only differ by differences in members of datum ensembles to be the same. And the identify() method also works with slightly different members in a datum ensemble provided it has the same name
So all in all, I believe it is reasonably safe for backport to stable branch.
CC @jjimenezshaw @kbevers

kbevers (approved):
I think it is fine to backport this

PR #4087, opened 2024-03-14
Speed-up +proj=cart +inv

rouault:
Alternative implementation to #4067

### DIFF
--- a/src/conversions/cart.cpp
+++ b/src/conversions/cart.cpp
@@ -155,9 +155,19 @@ static PJ_LPZ geodetic(PJ_XYZ cart, PJ *P) {
     PJ_LPZ lpz;
 
     // Normalize (x,y,z) to the unit sphere/ellipsoid.
+#if (defined(__i386__) && !defined(__SSE__)) || defined(_M_IX86)
+    // i386 (actually non-SSE) code path to make following test case of
+    // testvarious happy
+    // "echo 6378137.00 -0.00 0.00 | bin/cs2cs +proj=geocent +datum=WGS84 +to
+    // +proj=latlong +datum=WGS84"
     const double x_div_a = cart.x / P->a;
     const double y_div_a = cart.y / P->a;
     const double z_div_a = cart.z / P->a;
+#else
+    const double x_div_a = cart.x * P->ra;
+    const double y_div_a = cart.y * P->ra;
+    const double z_div_a = cart.z * P->ra;
+#endif
 
     /* Perpendicular distance from point to Z-axis (HM eq. 5-28) */
     const double p_div_a = sqrt(x_div_a * x_div_a + y_div_a * y_div_a);
--- a/src/conversions/cart.cpp
+++ b/src/conversions/cart.cpp
@@ -125,11 +125,15 @@ static double geocentric_radius(double a, double b_div_a, double cosphi,
     // const double b = a * b_div_a;
     // return hypot(a * a * cosphi, b * b * sinphi) /
     //        hypot(a * cosphi, b * sinphi);
-
+    const double cosphi_squared = cosphi * cosphi;
+    const double sinphi_squared = sinphi * sinphi;
+    const double b_div_a_squared = b_div_a * b_div_a;
+    const double b_div_a_squared_mul_sinphi_squared =
+        b_div_a_squared * sinphi_squared;
     return a *
-           sqrt(cosphi * cosphi +
-                (b_div_a * b_div_a) * (b_div_a * b_div_a) * (sinphi * sinphi)) /
-           sqrt(cosphi * cosphi + (b_div_a * b_div_a) * (sinphi * sinphi));
+           sqrt(cosphi_squared +
+                b_div_a_squared * b_div_a_squared_mul_sinphi_squared) /
+           sqrt(cosphi_squared + b_div_a_squared_mul_sinphi_squared);
 }
 
 /*********************************************************************/
--- a/src/conversions/cart.cpp
+++ b/src/conversions/cart.cpp
@@ -130,10 +130,9 @@ static double geocentric_radius(double a, double b_div_a, double cosphi,
     const double b_div_a_squared = b_div_a * b_div_a;
     const double b_div_a_squared_mul_sinphi_squared =
         b_div_a_squared * sinphi_squared;
-    return a *
-           sqrt(cosphi_squared +
-                b_div_a_squared * b_div_a_squared_mul_sinphi_squared) /
-           sqrt(cosphi_squared + b_div_a_squared_mul_sinphi_squared);
+    return a * sqrt((cosphi_squared +
+                     b_div_a_squared * b_div_a_squared_mul_sinphi_squared) /
+                    (cosphi_squared + b_div_a_squared_mul_sinphi_squared));
 }
 
 /*********************************************************************/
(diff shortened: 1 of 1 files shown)


rouault:
@wblangdon  Assuming this PR pass the tests, can you give it a try on your side to measure the performance?

rouault:
On my system, with a RelWithDebInfo build (-O2), ``hyperfine --warmup 5 './bin/bench_proj_trans -p "+proj=pipeline +step +inv +proj=cart" 4189946.586317880544811 146316.158923541981494 4790634.218430089764297'`` runs in 441 ms, vs 497 ms with master, so a 13% improvement

wblangdon:
Passes (64 bit) tests here.  Speed up on  geodetic  37% (GCC 10.2.1 -O3, i7-4790 CPU @ 3.60GHz)

src/conversions/cart.cpp
```diff
@@ -121,8 +121,15 @@ static double geocentric_radius(double a, double b, double cosphi,
         This is from WP2, but uses hypot() for potentially better
         numerical robustness
     ***********************************************************************/
+#ifdef not_optimized
     return hypot(a * a * cosphi, b * b * sinphi) /
```
jjimenezshaw: This does not compile even when `not_optimized` is defined. Should it be a comment instead?

src/conversions/cart.cpp
```diff
@@ -121,8 +121,15 @@ static double geocentric_radius(double a, double b, double cosphi,
...
            hypot(a * cosphi, b * sinphi);
+#else
+    return a *
+           sqrt(cosphi * cosphi +
+                (b_div_a * b_div_a) * (b_div_a * b_div_a) * (sinphi * sinphi)) /
```
jjimenezshaw: Those parenthesis are not needed. Are for readability or helping the compiler?

src/conversions/cart.cpp
rouault: good point. commit ammended

src/conversions/cart.cpp
```diff
@@ -112,7 +112,7 @@ static double normal_radius_of_curvature(double a, double es, double sinphi) {
 }
 
 /*********************************************************************/
-static double geocentric_radius(double a, double b, double cosphi,
+static double geocentric_radius(double a, double b_div_a, double cosphi,
```
jjimenezshaw, line 115: Nice trick using `b/a`, that is always `<1`, to avoid overflows and improve performance.

src/conversions/cart.cpp
```diff
@@ -158,18 +180,33 @@ static PJ_LPZ geodetic(PJ_XYZ cart, PJ *P) {
...
-    const double c = norm == 0 ? 1 : x_theta / norm;
-    const double s = norm == 0 ? 0 : y_theta / norm;
+    const double b_div_a = 1 - P->f; // = P->b / P->a
+    const double p_div_a_b_div_a = p_div_a * b_div_a;
+    const double norm =
+        sqrt(z_div_a * z_div_a + p_div_a_b_div_a * p_div_a_b_div_a);
+    double c, s;
+    if (norm != 0) {
+        const double inv_norm = 1.0 / norm;
+        c = p_div_a_b_div_a * inv_norm;
```
jjimenezshaw, line 193: I had to write it down in paper to check that they are equivalent.

jjimenezshaw (approved):

src/conversions/cart.cpp
rouault: That might help the compiler. Otherwise if we do "a * a * a * a", I suspect he's supposed to evaluate it as "((a * a) * a) * a" given the left-to-right evaluation order mandated by C/C++. Grouping by pair will save one multiplication (since the optimizer should be smart enough to evaluate once b_div_a * b_div_a

rouault: actually I've just added a commit to further optimize things

wblangdon:
:-) I am out of the office next week, but do not hesitate to let me know if I can help when I get back, eg timings

rouault:
> :-) I am out of the office next week, but do not hesitate to let me know if I can help when I get back, eg timings

yes, you're feedback on timings would be helpful since my own measurements show only a modest 13% improvement

wblangdon:
sure:-) Please do not delay waiting for me, as I say I will be out-of-action for a while...

src/conversions/cart.cpp
busstoptaktik: Note that `b/a` is also called the *aspect ratio* of the ellipsoid, which may be more descriptive than `b_div_a`. I have noted another pair of identities [here](https://docs.rs/geodesy/latest/geodesy/struct.Ellipsoid.html#method.aspect_ratio). There's no reference to literature, but I remember adding the name and the referred ellipsoid method right after having seen it mentioned in an absolutely reliable source (perhaps Rapp, OSU)

rouault: Thanks. Do you know if there is a conventional short name/letter for the aspect ratio?

busstoptaktik: > Thanks. Do you know if there is a conventional short name/letter for the aspect ratio?

I do not remember one, and as I was too stupid to write down the reference in the [Rust Geodesy Bibliography](https://docs.rs/geodesy/latest/geodesy/enum.Bibliography.html) it will be hard to find again - but [this](https://www.researchgate.net/publication/32180639_Effective_stress_and_pore_pressure_in_the_Nankai_accretionary_prism_off_the_Muroto_Peninsula_Southwestern_Japan/figures?lo=1) paper uses alpha (and swaps the meaning of a and b!), while [Rapp](https://kb.osu.edu/bitstreams/ee6892dd-cfb7-5142-978a-a978ae792980/download) uses alpha for the *angular eccentricity*, and hence `b/a = arccos(alpha)`. So it does not look like there is any firmly established convention.

Also, I see from [this](https://en.wikipedia.org/wiki/Aspect_ratio) Wikipedia page, that it is also common to define the aspect ratio as the larger parameter divided by the smaller, so even in that aspect (!) there does not seem to be any clear convention.

rouault: after all, isn't b_div_a quite explicit ;-) ?

busstoptaktik: > after all, isn't b_div_a quite explicit ;-) ?

It can probably harder get any more explicit :-) so given the lack of common nomenclature, it is obviously the better choice!

busstoptaktik: > Nice trick using `b/a`, that is always `<1`, to avoid overflows and improve performance.

... although I think a few of Jupiter's moons are actually prolate, so this will not hold if we at some time accidentally blindly import IAU ellipsoids

jjimenezshaw: The algorithm is still stable. If it were thaaat eccentric, then using an ellipsoid model wouldn't be a good idea ;)
I'm curious what is the physics behind a spinning prolate moon.

busstoptaktik: > The algorithm is still stable.

Ah, yes, I now see that it is, since we may safely assume that b never equals the imaginary unit times a :-)

> I'm curious what is the physics behind a spinning prolate moon.

Physically, I believe it's as stable as an oblate, as long as the spin axis and the principal axes of inertia are reasonably aligned. Apparently it just happens to spin around the longest axis, for whichever reason `$DEITY` came up with during the creation of the solar system (and, in the case of Saturn, whichever perturbations the poor moon suffers from the debris of the ring it shepherds)

wblangdon:
geodetic() speed up cart.cpp_2d22cf6 relative PROJ-master 37% (GCC 10.2.1 -O3 i7-4790 3.60GHz)

PR #4089, opened 2024-03-15
Doc: add note warning that stere/sterea are different things and …

rouault:
…clarify their relationship with EPSG's Oblique Stereographic


docs/plot/plotdefs.json
```diff
@@ -1346,13 +1346,13 @@
...
-        "lonmax": 180,
-        "lonmin": -180,
+        "latmax": 70,
+        "latmin": 30,
+        "lonmax": 50,
+        "lonmin": -50,
         "name": "sterea",
-        "projstring": "+proj=sterea +lat_0=90",
-        "res": "low",
+        "projstring": "+proj=stere +lat_0=52.1561605555556 +lon_0=5.38763888888889 +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel",
```
mwtoews: was this suppose to be kept as `sterea`? This string is different than the one shown in the docs.
```suggestion
        "projstring": "+proj=sterea +lat_0=52.1561605555556 +lon_0=5.38763888888889 +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel",
```

rouault: good catch. fixed

PR #4096, opened 2024-03-20
DOC: consistency via markdownlint; move AUTHORS to AUTHORS.md

mwtoews:
This is yet more markdown churn (after #4066), mostly automated but with several manual edits.

This PR uses https://github.com/DavidAnson/markdownlint / https://github.com/igorshubovych/markdownlint-cli tools to check or create consistent documents. For example, this is my main method:

```sh
docker run -v $PWD:/workdir ghcr.io/igorshubovych/markdownlint-cli:latest --fix *.md
```

<s>A configuration file to control the rules is included with this PR, and a hook is added to the mostly unused pre-commit configuration (i.e. it does not run in CI - perhaps this should be switched on someday).</s>

Notes on linting and rules:

- A line-length rule is not enforced, but is "mostly" 80 to 100 characters, with many exceptions just left as-is
- Bare URLs are converted to use angle brackets, as this seems to be a common markdown mistake
- All rules are configured using [this doc](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md)

Other notes on edits:

- AUTHORS is moved/translated to AUTHORS.md
- With NEWS.md, a top-level header is set to "PROJ Release Notes", which necessitates all other sections dropped down a level (i.e. add another `#`). Each release is a second-level header, and renamed from (e.g.) "9.4.0 Release Notes" to "9.4.0". If anyone fells strongly about restoring "Release Notes" to each header, this can be re-added.
- With data/sql/README.md, rather than using a nested list, the top list items are simply converted to headers


AUTHORS.md
```diff
@@ -0,0 +1,33 @@
...
+- Kristian Evers <kreve@sdfe.dk>
+
+Members:
+
+- Frank Warmerdam <warmerdam@pobox.com>
+- Howard Butler <howard@hobu.co>
+- Charles Karney <charles.karney@sri.com>
+- Thomas Knudsen <thokn@sdfe.dk>
+- Even Rouault <even.rouault@spatialys.com>
+- Kurt Schwehr <schwehr@gmail.com>
```
kbevers:
```suggestion
- Kristian Evers <kreve@sdfi.dk>
- Even Rouault <even.rouault@spatialys.com>

## Project Steering Committee

Process and membership can be found at:
<https://proj.org/community/rfc/rfc-1.html>

Chair:

- Kristian Evers <kreve@sdfi.dk>

Members:

- Frank Warmerdam <warmerdam@pobox.com>
- Howard Butler <howard@hobu.co>
- Charles Karney <charles.karney@sri.com>
- Thomas Knudsen <thokn@sdfi.dk>
- Even Rouault <even.rouault@spatialys.com>
- Kurt Schwehr <schwehr@gmail.com>
```

kbevers (commented):
Looks alright to me. I would prefer if this was separated into a few different commits:

1. `AUTHORS` -> `AUTHORS.md`
2. `NEWS.md` headings
3. Markdown reformatting

I think that would make the changes in this PR more readable. As it is now it is hard to see the "real" changes for all the rewrapped lines

mwtoews:
Thanks for feedback @kbevers , I've split this into a few logical commits, mostly to separate the automated edits from the manual edits.

Also I've held-back on adding a pre-commit hook, for now, as that should be a separate PR/discussion.

kbevers (approved):
Thanks @mwtoews, I'm happy to see this PR merged.

As for the pre-commit hooks, I agree that it requires a separate discussion. In principle I am in favour of them but have in the past found them rather finicky and often causing more trouble than solving problems, unfortunately. It's been a while so things may have changed for the better since then.

PR #4100, opened 2024-03-24
Update cct.rst adding long options. But...

jidanni:
But somebody still needs to fix the SYNOPSIS, like gdaltransform's.


kbevers:
> But somebody still needs to fix the SYNOPSIS

Perhaps that somebody could be you?

jidanni:
It is too bad that this PR will be thrown away.

PR #4102, opened 2024-03-24
Update cct.rst possibly making it clearer

jidanni:


kbevers:
I disagree with this change. I don't think it is correct to say that cct is an enhancement of proj - it's its own thing with a different purpose. Incidentally you can achieve the same thing with cct as with proj...

jidanni:
The word "equivalent"  seemed even worse to me. 

So I'm glad I was going in the right direction to make them seem further apart. 

I was just too timid to make them too far apart. 

Perhaps you could step in and  word it even better. Thanks!

jidanni:
Well I guess this all will be swept away by the bot.

PR #4106, opened 2024-03-27
Retry "Connection reset by peer"

pjonsson:
Building the GDAL image can fail with:

Downloading https://cdn.proj.org/fr_ign_RAF18b.tif... (132 / 423)
Downloading https://cdn.proj.org/fr_ign_RAF20.tif... (133 / 423)
Downloading https://cdn.proj.org/fr_ign_RAGTBT2016.tif... (134 / 423)
Cannot open https://cdn.proj.org/fr_ign_RAGTBT2016.tif: OpenSSL SSL_read: Connection reset by peer, errno 104

so add the connection reset by peer
to the list of things that get a retry
time.

- [ ] Tests added

### DIFF
--- a/src/networkfilemanager.cpp
+++ b/src/networkfilemanager.cpp
@@ -1675,6 +1675,7 @@ static double GetNewRetryDelay(int response_code, double dfOldDelay,
         // S3 sends some client timeout errors as 400 Client Error
         (response_code == 400 && pszErrBuf &&
          strstr(pszErrBuf, "RequestTimeout")) ||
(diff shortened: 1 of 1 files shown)


pjonsson:
I'm not sure if my patch fixes the problem described, or even how to test it, but the line seems to be in good company with the "Connection timed out" comparison.

rouault:
Seems reasonable. As you also have interest in GDAL, would you mind issuing a similar patch to GDAL's CPLHTTPGetNewRetryDelay() method in port/cpl_http.cpp?

PR #4108, opened 2024-03-28
Add a WKTParser::grammarErrorList() method so that proj_create_from_wkt() can behave as documented

rouault:


rouault:
(removing backport label. As it moves reports on the C++ API from WKTParser::warningList() to WKTParser::grammarErrorList(), it is not fully backwards compatible)

Issue #4103, opened 2024-03-24
`"-` artifact in rendered cct man page


jidanni:
There is a weird `"-` artifact in

```
$ nroff -man Downloads/cct.1 | grep -A 3 concat 
troff:Downloads/cct.1:63: warning [p 1, 4.3i]: cannot adjust line
troff:Downloads/cct.1:194: warning: cannot select font 'C'
troff:Downloads/cct.1:207: warning: cannot select font 'C'
troff:Downloads/cct.1:219: warning: cannot select font 'C'
troff:Downloads/cct.1:233: warning: cannot select font 'C'
troff:Downloads/cct.1:247: warning: cannot select font 'C'
troff:Downloads/cct.1:261: warning: cannot select font 'C'
          • a  OGC  URN combining references for concatenated operations (e.g.
            "‐
            urn:ogc:def:coordinateOperation,coordinateOperation:EPSG::3895,co‐
            ordinateOperation:EPSG::1618")
troff:Downloads/cct.1:276: warning: cannot select font 'C'
troff:Downloads/cct.1:291: warning: cannot select font 'C'
troff:Downloads/cct.1:306: warning: cannot select font 'C'
troff:Downloads/cct.1:321: warning: cannot select font 'C'
```
But there is no smoking gun in cct.rst. Maybe the too long line triggered a bug.

Never mind the font C warnings. That is some other worldwide issue.

kbevers:
I can get a similar result using your commands but it doesn't show when viewed as an actual man page. Is this really a problem?

![Screenshot 2024-03-24 at 10 50 26](https://github.com/OSGeo/PROJ/assets/13132571/90fe831b-50fd-4631-a791-7251f43cb89e)

jidanni:
Seems like you're testing it on a super wide screen.

I read Man pages with M-X man in emacs, and stared at it for quite a while trying to figure out what it was  trying to say.

kbevers:
It doesn't make a difference if page is rendered wide or narrow:

![Screenshot 2024-03-24 at 14 24 39](https://github.com/OSGeo/PROJ/assets/13132571/68c3c5bc-dc7d-4f8e-84c3-b06165f76d3f)

jidanni:
I see what's going on.

It's trying to hyphen-
ate
a long word.

```
$ COLUMNS=80 man cct 2>&-|grep -B 1 EPSG::3895
            "‐
[...]

PR #4109, opened 2024-04-02
Update cct.rst to release pressure on man page formatter

jidanni:
For https://github.com/OSGeo/PROJ/issues/4103

Pssst, I didn't actually test it, in cct nor in man.

I suppose these newlines should be rendered as spaces and allow breaking the line.

### DIFF
--- a/docs/source/apps/cct.rst
+++ b/docs/source/apps/cct.rst
@@ -28,7 +28,9 @@ by :c:func:`proj_create`, provided it expresses a coordinate operation
     - an object name. e.g. "ITRF2014 to ETRF2014 (1)". In that case as
       uniqueness is not guaranteed, heuristics are applied to determine the appropriate best match.
     - a OGC URN combining references for concatenated operations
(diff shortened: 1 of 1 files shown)


kbevers:
> Pssst, I didn't actually test it, in cct nor in man.

Please do so. 

@rouault it is unclear to me if this is a list of several URN's or just one long and complicated URN. I suspect the latter, in which case the reformatting in this PR is problematic. Correct?

rouault:
> or just one long and complicated URN

yes, this is a single URN for a concatenated operation chaining 2 coordinate operations

kbevers:
I figured. In that case this change isn't going to cut it. @jidanni I think you'll just have to live with this quirk in the man page. Or get a wider monitor.

jidanni:
I didn't check, but maybe these URNs should allow spaces in them near commas in the first place. 

Sort of like Bourne shell sh -c arguments. You can even have  newlines in its single argument.

Or maybe they're like www urls where spaces have to be replaced by pluses...

PR #4128, opened 2024-04-26
Update proj-docs Docker image to Ubuntu 24.04

rouault:
- Fixes to make doc building work with Doxygen 1.9.8
- docbuild/Dockerfile: update to Ubuntu 24.04, and remove no longer pins of a few packages
- .github/workflows/docker.yml: (temporarily?) build proj-docs as proj-docs-2404

I believe the only user of proj-docs now is GDAL docs. I've succesfully tested the proposed update with the changes in https://github.com/OSGeo/gdal/pull/9784. I've also run PROJ doc building within the updated Docker image, although we only use readthedocs right now.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -25,12 +25,12 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        container: ["proj","proj-docs"]
+        container: ["proj","proj-docs-2404"]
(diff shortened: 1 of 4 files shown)


jjimenezshaw (approved):

rouault:
> .github/workflows/docker.yml: (temporarily?) build proj-docs as proj-docs-2404

I had to revert that part in https://github.com/OSGeo/PROJ/commit/6e2b5b5c8fd3999781aef8131a07aa524572f187 as it prevented GDAL to consume it

PR #4129, opened 2024-04-27
CI: switch from macos-latest to macos-13 to get x64 builds

rouault:
GHA has changed the macos-latest alias to macos-14 (arm64), hence to get Intel Mac one now needs to explicitly select macos-13 Cf https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          - platform: "macos-latest"
+          - platform: "macos-13"
             architecture: "x64"
           # macOS-14 is a M1 ARM64 MacOS runner: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
           - platform: "macOS-14"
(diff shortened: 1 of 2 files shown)


.github/workflows/mac.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          - platform: "macos-latest"
+          - platform: "macos-13"
             architecture: "x64"
           # macOS-14 is a M1 ARM64 MacOS runner: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
           - platform: "macOS-14"
```
jjimenezshaw, line 28: Should we change this to `latest`? otherwise it is probable that we do not update to the newer ones.

jjimenezshaw (approved):

.github/workflows/mac.yml
rouault: I don't know. The -latest is also a moving target that has switched between x64 and arm64, which is annoying. It is weird that they don't provide 2 separate keywords, one for the mac version and the other one for the architecture. But I believe Apple won't support x64 anymore, so perhaps at some point that will be arm64 only.

jjimenezshaw: Oh, yes. I think that the new models are all arm. The question is how long they will support the old ones.

PR #4131, opened 2024-05-01
Improve CLI testing

### LINKED ISSUE #4110: Proposed enhancement for command-line interface tests
This is an issue that I intend to resolve, but need to consult for feedback before starting.

The issue is that bash scripts like [test/cli/testproj](https://github.com/OSGeo/PROJ/blob/f4af2778ebd353a6a173da9689de02fc48a6c307/test/cli/testproj) are only run/tested for UNIX (i.e. [here](https://github.com/OSGeo/PROJ/blob/f4af2778ebd353a6a173da9689de02fc48a6c307/cmake/ProjTest.cmake#L18-L28)), which means these are not tested for Windows. Probably the most portable way to run these for all platforms is to make a Python script to do this, which shouldn't be an issue since Python is already used for building the data. More importantly, the tests would need a structured data format, which may contain arguments, input line(s) and expected output line(s).

One obvious candidate format is JSON:
```json
[
  {
    "args": "proj +ellps=WGS84 +proj=ob_tran +o_proj=latlon +o_lon_p=0.0 +o_lat_p=90.0 +lon_0=360.0 +to_meter=0.0174532925199433 +no_defs -E -f %.3f",
    "in": "2 49",
    "out": "2 49\t2.000\t49.000"
  },
  {
    "comment": "Test CRS option",
    "args": "proj EPSG:32620 -S",
    "in": "-63 0",
    "out": "500000.00\t0.00\t<0.9996 0.9996 0.9992 0 0.9996 0.9996>"
  },
  {
    "comment": "Test projection factors on projected CRS with non-Greenwhich prime meridian",
    "args": "proj EPSG:27571 -S",
    "in": "2.33722917 49.5",
    "out": "600000.00\t1200000.00\t<0.999877 0.999877 0.999755 0 0.999877 0.999877>"
  },
  {
    "comment": "Test projection factors on compound CRS with a projected CRS",
    "args": "proj EPSG:5972 -S",
    "in": "9 0",
    "out": "500000.00\t0.00\t<0.9996 0.9996 0.9992 0 0.9996 0.9996>"
  }
]
```
or possibly nicer as more compact and versatile YAML file, which can easily split tags like "exe" at the top:
```yaml
exe: proj
tests:
- args: +ellps=WGS84 +proj=ob_tran +o_proj=latlon +o_lon_p=0.0 +o_lat_p=90.0
    +lon_0=360.0 +to_meter=0.0174532925199433 +no_defs -E -f %.3f
  in: 2 49
  out: "2 49\t2.000\t49.000"
- comment: Test CRS option
  args: EPSG:32620 -S
  in: -63 0
  out: "500000.00\t0.00\t<0.9996 0.9996 0.9992 0 0.9996 0.9996>"
- comment: Test projection factors on projected CRS with non-Greenwhich prime meridian
  args: EPSG:27571 -S
  in: 2.33722917 49.5
  out: "600000.00\t1200000.00\t<0.999877 0.999877 0.999755 0 0.999877 0.999877>"
- comment: Test projection factors on compound CRS with a projected CRS
  args: EPSG:5972 -S
  in: 9 0
  out: "500000.00\t0.00\t<0.9996 0.9996 0.9992 0 0.9996 0.9996>"
```
versatility is more important with longer multi-line input/output from [test/cli/testvarious](https://github.com/OSGeo/PROJ/blob/f4af2778ebd353a6a173da9689de02fc48a6c307/test/cli/testvarious) and [test/cli/tv_out.dist](https://github.com/OSGeo/PROJ/blob/f4af2778ebd353a6a173da9689de02fc48a6c307/test/cli/tv_out.dist)

As for the specification of the input tests, there would be required and optional keys. I'd need to sort these specifications after looking at all of the test files.

After a Python script is built to run each to-be-converted file, it would then be called from CTest (replacing `proj_add_test_script_sh`). I may need to consider skipping these tests if a suitable YAML parser is not available for Python, since unfortunately there is no standard module for YAML (unlike JSON).

Should Python and YAML test data be used for a modernized CLI testing framework? Is there anything else to consider?

mwtoews:
This enhancement improves command-line interface tests across several platforms.

Until now CLI testing was done using a range of Unix shell scripts, comparing expected output and showing differences on failure. While this approach has worked fine for >20 years, it is limited to Unix shells (i.e. not Windows) and each test's inputs and expected outputs need to be cross-referenced between two files.

The more modern approach in this PR is to use a Python script to run tests described in YAML files. Each file describes a series of tests to run with an executable. Each test may have arguments (`args`), input (`in`), expected stdout, stderr, or combined stdout+stderr (`out`), as well as an expected `exitcode` (usually 0). There are a few optional components, such as use of a temporary directory (`tmpdir`) with files to copy; custom file content; the ability to skip tests based on a simple expression (`skipif`); and output post-processing commands (`grep`, `grep -v`, `sub`, `head`, `tail`, `sort`) used on command output streams. A full description of the format can be made later (after a review).

The Python script is currently located at `test/cli/run_cli_test.py` along with the YAML test files that have been mostly manually converted from the shell scripts. This script can be run directly, or via ctest. CMake does some checks before to ensure the requirements have been met, otherwise a warning is shown that the tests won't run. The requirements are: Python 3.7+, and one of either [PyYAML](https://pypi.org/project/PyYAML/) or [ruamel.yaml](https://pypi.org/project/ruamel.yaml/). Optionally pytest can be used to self-test the Python script.

One important maintainer feature of the script is `--update` to update the expected result (`stdout`, `stderr`, `out` and/or `exitcode`). This mode is enabled only with ruamel.yaml, which has round-trip read/write capabilities to preserve YAML styles and comments when re-writing the file. This mode might be needed when (e.g.) updating proj.db.

Some of the older shell script tests are kept, as they have tests that are better kept as shell scripts. The test names in ctest are the same as the input files, and they have been renamed to easier identify.

Closes #4110


test/cli/CMakeLists.txt
```diff
@@ -1,22 +1,112 @@
...
+set(PROJINFO_EXE "$<TARGET_FILE:projinfo>")
+set(PROJSYNC_EXE "$<TARGET_FILE:projsync>")
+
+if(UNIX)
+  # invproj is a symlink
+  # set(INVPROJ_EXE "$<TARGET_PROPERTY:invproj,DEPENDS>")  # TODO: find working genexpr
+  set(INVPROJ_EXE "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/invproj${CMAKE_EXECUTABLE_SUFFIX}")
+  proj_add_test_script_sh(test_cs2cs_locale.sh CS2CS_EXE)
+  proj_add_test_script_sh(test_projinfo.sh PROJINFO_EXE)
+  # if(BUILD_PROJSYNC_DATA)
```
rouault: shouldn't that if() be re-enabled ? (or just removed if useless)

test/cli/test_cct.yaml
```diff
@@ -0,0 +1,166 @@
+exe: cct
+tests:
+- args: -d 8 +proj=merc +R=1
+  in: 90 45 0
+  out: |2
```
rouault: what is the meaning of this "|2" ?

test/cli/CMakeLists.txt
mwtoews: Enabled for `BUILD_PROJSYNC` (note that this test previously never ran since `BUILD_PROJSYNC_DATA` does not exist; xref 87318283eedef680356ef272a6af39ac3140d333). Also add more of these for each of the [other BUILD_* options](https://github.com/OSGeo/PROJ/blob/8f29b8f48b1b78aed4cb54f19ce7ea5c6c6fb469/src/apps/CMakeLists.txt#L8-L28).

test/cli/test_cct.yaml
mwtoews: This is an indentation indicator, which is specified here since `cct` output has leading whitespace. A great resource on the topic that I've used for this PR is https://yaml-multiline.info/

.github/workflows/clang_linux/start.sh
```diff
@@ -4,22 +4,25 @@ set -e
 
 apt-get update -y
 DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
-    autoconf automake libtool make cmake ccache pkg-config python3-pip sqlite3 tar zip \
+    autoconf automake libtool make cmake ccache pkg-config python3-pip sqlite3 tar zip curl \
```
mwtoews, line 7: FYI, added `curl` here (and similar) so that networking is enabled for testing. Otherwise "Skipping network dependent tests" was shown.

.github/workflows/clang_linux/start.sh
```diff
@@ -4,22 +4,25 @@ set -e
...
     clang++-10 jq python3-clang-10 \
     libsqlite3-dev \
     libtiff-dev libwebp-dev libzstd-dev \
     libcurl4-openssl-dev libnghttp2-dev libidn2-dev librtmp-dev libssh-dev \
-      libpsl-dev libssl-dev libkrb5-dev comerr-dev libldap2-dev libbrotli-dev \
+    libpsl-dev libssl-dev libkrb5-dev comerr-dev libldap2-dev libbrotli-dev \
     nlohmann-json3-dev libgtest-dev
 
-python3 -m pip install --user jsonschema
 export PATH=$HOME/.local/bin:$PATH
```
mwtoews, line 15: re-order PATH export to avoid "WARNING: The script jsonschema is installed in '/root/.local/bin' which is not on PATH"

cmake/ProjTest.cmake
```diff
@@ -14,20 +14,28 @@ function(proj_test_set_properties TESTNAME)
     PROPERTY ENVIRONMENT ${_env})
 endfunction()
 
-function(proj_add_test_script_sh SH_NAME BIN_USE)
-  if(UNIX)
```
mwtoews, line 18: this `if(UNIX)` check was moved to where the function as used to make it more clear that these tests are skipped for Windows.

test/cli/CMakeLists.txt
```diff
@@ -1,22 +1,112 @@
...
+  proj_add_test_script_sh(test_projsync.sh PROJSYNC_EXE)
+  # endif()
+else()
+  set(INVPROJ_EXE "$<TARGET_FILE:invproj>")
+endif()
+
+macro(find_Python3_package PACKAGE VARIABLE)
```
mwtoews, line 39: This new macro could potentially be moved to `cmake/FindPython3Package.cmake`

test/cli/test_cs2cs_flaky.yaml
mwtoews, line 1: This test was revived from `test/cli/testflaky` which was previously not run. It could potentially be renamed or combined elsewhere.

test/cli/test_cs2cs_locale.sh
mwtoews, line 1: This is a subset from `test/cli/testvarious` to check that pj_init() is locale-safe. Is this sufficient? I'm not 100% sure what is being tested here, but it is a UNIX-only test.

test/cli/test_cs2cs_various.yaml
```diff
@@ -0,0 +1,1511 @@
...
+    +proj=topocentric +datum=WGS84 +X_0=-3982059.42 +Y_0=3331314.88 +Z_0=3692463.58 +no_defs +type=crs +to
+    EPSG:4978 -d 3 -E
+  in: 0 0 0
+  out: |
+    0 0 0	-3982059.420	3331314.880 3692463.580
+- comment: >
+    Test cs2cs EPSG:5488 (RGAF09) to EPSG:4559+5757 (RRAF 1991 / UTM zone 20N + Guadeloupe 1988 height)
+
+    Check that we use the horizontal transformation for Guadeloupe (RRAF 1991 to RGAF09 (2)), and not the one for Martinique
+  skipif: env.get("TIFF_ENABLED") == "NO"
```
mwtoews, line 1501: here's an example to skip a test if a TIFF_ENABLED environment variable is NO, otherwise it is not skipped (either set to a different value or undefined).

mwtoews (commented):
Note that all of the old tests should still run in some form, but no new tests have been added. New tests can be added later, e.g. for geod/invgeod. Also binary I/O tests can be added later (with `skipif: byteorder != "little"`), as I've enabled this support. When text I/O is defined in the YAML file it is otherwise utf-8 encoded on all platforms.

Inline comments added.

rouault (approved):

rouault:
LGTM. Excellent work!

mwtoews:
@rouault comment above to anything that I've raised if you have any strong opinions. Mostly I'm unsure how locale set-up and testing should work in `test/cli/test_cs2cs_locale.sh`.

Also, before merging, I could remove most of the repetition by removing `-E` echo flags. The expected stdout/out is easy to update with `--update`.

rouault:
> Mostly I'm unsure how locale set-up and testing should work in `test/cli/test_cs2cs_locale.sh`.

Your way is probably OK. Otherwise that logic (setting LC_ALL and PROJ_USE_ENV_LOCALE=1 in os.environ) could perhaps be ported to the Python launcher itself to run all tests with a non C-locale if available

> Also, before merging, I could remove most of the repetition by removing -E echo flags. The expected stdout/out is easy to update with --update.

Yes, the -E flag doesn't bring much with the new framework.

mwtoews:
Most tests have removed echo `-E` flag, except a few specific tests.

Some basic docs have been added (as part of the Python script, for now).

The run_cli_test.py with this PR is generic, and could potentially be adapted for other projects. It could be moved into a stand-alone Pure Python package, if there is interest.

Lastly, a note to packagers (e.g. https://github.com/conda-forge/proj.4-feedstock/). While there are no new build deps, there are new recommended test deps. A minimum of Python 3.7 and pyyaml should be good, otherwise most of the CLI tests won't run. A warning will be shown in the first cmake command.

jjimenezshaw:
> Otherwise that logic (setting LC_ALL and PROJ_USE_ENV_LOCALE=1 in os.environ) could perhaps be ported to the Python launcher itself to run all tests with a non C-locale if available

In PDAL, to test that it was working properly with a non classic locale, I did this in the github actions (only in Linux): https://github.com/PDAL/PDAL/blob/84d0a9d076d3ccfd9f48dbab609ec093e1121981/.github/workflows/linux.yml#L75
That locale is later enabled in the test suite, and all the tests run again.

mwtoews:
On locales, I was considering adding this effort to the Python scripting, but it mostly applies to POSIX systems, thus is easier to keep testing the .sh script (i.e., I copied excerpts from `testvarious` into `test_cs2cs_locale.sh`). (I still don't fully understand the interactions of various combinations of `PROJ_USE_ENV_LOCALE` with `LC_ALL`, as I never see the comma decimal separator).

rouault:
> I still don't fully understand the interactions of various combinations of `PROJ_USE_ENV_LOCALE` with `LC_ALL`, as I never see the comma decimal separator

PROJ_USE_ENV_LOCALE=1 commands the call to https://github.com/OSGeo/PROJ/blob/1d0354fd7b240387f81dcc9132085af9c426f200/src/apps/cs2cs.cpp#L799 , which enables the taking into account of the LOCALE environment variable by various libc functions

> as I never see the comma decimal separator

The test checks that we are able to parse inputs in a non-locale aware way, that is we don't use regular atof() for that (and more generally even internally, e.g when ingesting strings from the database to double), that would fail to parse numbers like "1.25" with a locale with a decimal separator that wouldn't be dot.

The shell script is probably good enough for now.

PR #4144, opened 2024-05-09
API: add proj_context_set_user_writable_directory()

### LINKED ISSUE #4142: User writable directory
Hi,

I would need to know if there is any API function to set up the user writable folder directly. On Android system, the default one, /tmp/proj, cannot be created and the files cannot be downloaded.

Thanks!

rouault:
Fixes #4142

### DIFF
--- a/src/proj_symbol_rename.h
+++ b/src/proj_symbol_rename.h
@@ -72,11 +72,14 @@
 #define proj_context_set_sqlite3_vfs_name                                      \
     internal_proj_context_set_sqlite3_vfs_name
 #define proj_context_set_url_endpoint internal_proj_context_set_url_endpoint
+#define proj_context_set_user_writable_directory                               \
+    internal_proj_context_set_user_writable_directory
 #define proj_context_use_proj4_init_rules                                      \
     internal_proj_context_use_proj4_init_rules
 #define proj_convert_conversion_to_other_method                                \
     internal_proj_convert_conversion_to_other_method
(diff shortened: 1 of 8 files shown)


src/proj_symbol_rename.h
```diff
@@ -72,11 +72,14 @@
...
     internal_proj_context_set_sqlite3_vfs_name
 #define proj_context_set_url_endpoint internal_proj_context_set_url_endpoint
+#define proj_context_set_user_writable_directory                               \
+    internal_proj_context_set_user_writable_directory
 #define proj_context_use_proj4_init_rules                                      \
     internal_proj_context_use_proj4_init_rules
 #define proj_convert_conversion_to_other_method                                \
     internal_proj_convert_conversion_to_other_method
 #define proj_coord internal_proj_coord
+#define proj_coordinate_metadata_create internal_proj_coordinate_metadata_create
```
jjimenezshaw, line 82: This was not created before because the script was not executed for some time, right? Was it risky?

jjimenezshaw (approved):

src/proj_symbol_rename.h
rouault: > This was not created before because the script was not executed for some time, right? 

yes

Was it risky?

moderately. This is only used on special builds with -DPROJ_RENAME_SYMBOLS as used in the GDAL Docker image to have a custom PROJ build used only by GDAL, while allowing other packaged libs to use the system libproj.

PR #4146, opened 2024-05-12
[Lint] WKTParser::Private: refactor code a bit in the hope to fix false-positive 'Untrusted value as argument' Coverity Scan warning

rouault:
Should hopefully fix CID 460619, 460620, 460621

Coverity Scan suddenly complains about propertyCount_ not being properly being validated (which is wrong). Let's take the opportunity to use a bit more C++ish style in the hope it will shut down that false positive.


QuLogic:
> Coverity Scan suddenly complains about numProperties_

Do you mean `propertyCount_` here?

rouault:
> Do you mean `propertyCount_` here?

yes, sorry. amended

jjimenezshaw (approved):

PR #4153, opened 2024-05-28
DOC: specify test requirements; add nlohmann/json to build reqs

mwtoews:
See discussion https://lists.osgeo.org/pipermail/proj/2024-May/011405.html

This PR splits the requirements into build and test requirements, and adds further details on nlohmann/json.

A slightly different version of this PR (with additional test deps) will be pushed to master later on.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -139,15 +139,27 @@ The following guides show how to compile and install the software using CMake.
     PROJ 9.0 and later releases only support builds using CMake.
 
 
-Build requirements
+Requirements
 --------------------------------------------------------------------------------
+
+Build requirements
+++++++++++++++++++
 
 - C99 compiler
 - C++11 compiler
 - CMake >= 3.16
-- SQLite3 >= 3.11: headers and library for target architecture, and sqlite3 executable for build architecture.
+- SQLite3 >= 3.11: headers and library for target architecture, and sqlite3 executable for build architecture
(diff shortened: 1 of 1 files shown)


kbevers:
This looks good. Please hold merging until 9.4.1 is released. Or a 9.4.1RC2 is needed.

docs/source/install.rst
```diff
@@ -139,15 +139,27 @@ The following guides show how to compile and install the software using CMake.
...
+++++++++++++++++++
+
 - C99 compiler
 - C++11 compiler
 - CMake >= 3.16
-- SQLite3 >= 3.11: headers and library for target architecture, and sqlite3 executable for build architecture.
+- SQLite3 >= 3.11: headers and library for target architecture, and sqlite3 executable for build architecture
 - libtiff >= 4.0 (optional but recommended)
 - curl >= 7.29.0 (optional but recommended)
+- JSON for Modern C++ (nlohmann/json) >= 3.7.0 (optional); if not found then version 3.9.1 from PROJ source tree is used
```
rouault: "optional" is perhaps a bit confusing here as it is different from the 2 above optional instances for libtiff and curl, and you can't disable use of nlohmann/json, so perhaps:
```suggestion
- JSON for Modern C++ (nlohmann/json) >= 3.7.0; if not found as an external dependency, then vendored version 3.9.1 from PROJ source tree is used
```

PR #4169, opened 2024-06-08
Add code comments in GeographicBoundingBox::Private::intersects()/intersections()

rouault:


rouault:
The Windows failures are mostly certainly the same issue I workaround in GDAL in https://github.com/OSGeo/gdal/pull/10161 which is an issue with the windows image of github actions. Let's home the github actions team revert their windows image or fix it soon

PR #4177, opened 2024-06-19
Support default context as nullptr in coordinateTransform

joukewitteveen:
In #3705, `coordinateTransformer` was added with documentation stating the
default context will be used when null is given as a context. This did
not work because the context was not sanitized.


src/iso19111/c_api.cpp
```diff
@@ -192,6 +192,7 @@ getDBcontextNoException(PJ_CONTEXT *ctx, const char *function) {
 // ---------------------------------------------------------------------------
 
 PJ *pj_obj_create(PJ_CONTEXT *ctx, const BaseObjectNNPtr &objIn) {
+    SANITIZE_CTX(ctx);
```
joukewitteveen: The definition of `SANITIZE_CTX` needs to be hoisted above this function. I'll do that later today.

PR #4184, opened 2024-06-28
Fix: not apply -Wextra-semi for gcc version<8

tomwillow:
I use vcpkg to get PROJ library, but error occurs due to this compiling argument "-Wextra-semi" is not avaliable at gcc which version is less than 8.

My environment:
Ubuntu 18.04
GCC 7.5.0

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,9 +63,13 @@ if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
   set(PROJ_C_WARN_FLAGS ${PROJ_common_WARN_FLAGS}
     -Wmissing-prototypes
   )
+  if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 8)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -63,9 +63,13 @@ if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
   set(PROJ_C_WARN_FLAGS ${PROJ_common_WARN_FLAGS}
     -Wmissing-prototypes
   )
+  if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 8)
+	  set(PROJ_CXX_WARN_FLAGS ${PROJ_common_WARN_FLAGS}
+		-Wextra-semi
+	  )
```
rouault: Removing tabulation characters:
```suggestion
        set(PROJ_CXX_WARN_FLAGS ${PROJ_common_WARN_FLAGS} -Wextra-semi)
```

PR #4199, opened 2024-07-15
Connect + and - option documentation

jidanni:
Here we show users the vital bridges between + and - style options.
As cs2cs and cct don't have -m, and cct doesn't allow mixing + and - options.

### DIFF
--- a/docs/source/apps/cct.rst
+++ b/docs/source/apps/cct.rst
@@ -76,7 +76,7 @@ The following control parameters can appear in any order:
 
 .. option:: -I
 
-    Do the inverse transformation.
+    Do the inverse transformation. (``+`` style option: ``+inv``).
 
 .. option:: -o <output file name>, --output=<output file name>
 
--- a/docs/source/apps/proj.rst
+++ b/docs/source/apps/proj.rst
@@ -127,7 +127,8 @@ The following control parameters can appear in any order
     in a forward projection mode the cartesian output values are multiplied by
     *mult* otherwise the input cartesian values are divided by *mult* before inverse
     projection. If the first two characters of *mult* are 1/ or 1: then the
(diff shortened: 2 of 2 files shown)


kbevers:
I don't like mixing application parameters (the ones with the minuses) with coordinate operation parameters (the plusses). This PR mainly displays misunderstanding of those two concepts. It's clear that this confuses many people but this isn't the way to fix that. I'm more likely to entertain a PR that explains the differences in general terms instead of spreading a patchwork of exceptions and quirks throughout the docs.

In addition, the use of old PROJ.4-style parameters such as `+to_meter` are discouraged in modern PROJ and are mainly kept for backwards compatability. We don't want to promote those to new users.

> cct doesn't allow mixing + and - options.

This isn't true:

```sh
echo 12 56 0 0 | cct -d 5 +proj=utm +zone=32 
 687071.43911  6210141.32675       0.00000        0.0000
```

jidanni:
All I know is
```
$ echo 12 56 0 0 | cct -d 5 +proj=utm +zone=32 | cct -d 5 +proj=utm +zone=32 -I
+ and - style options must not be mixed
$ cct --version
cct: Rel. 9.4.1, June 1st, 2024
```

jjimenezshaw:
Setting `-I` earlier works:
```
$ echo 12 56 0 0 | cct -d 5 +proj=utm +zone=32 | cct -d 5 -I +proj=utm +zone=32 
      12.00000        56.00000       0.00000        0.0000
```

kbevers:
> All I know is

It's stated clearly in the docs that minus parameters comes first, the plus parameters. See the synopsis at the top of https://proj.org/en/9.4/apps/cct.html

jidanni:
> \+ and - style options must not be mixed

Fixed in #4201.

PR #4207, opened 2024-07-24
CMake: add option EMBED_PROJ_DATA_PATH

WangWeiLin-MV:
Add an option `EMBED_PROJ_DATA_PATH` to allow `PROJ_DATA_PATH` not to be hardcoded. Defaults to `ON`, does not change existing behavior.

There are many ways to get the path, and the existing code allows macros not to be defined, for example [src/filemanager.cpp#L1390-L1394](https://github.com/OSGeo/PROJ/blob/9.4.1/src/filemanager.cpp#L1390-L1394).

There are many situations where you have to concern about hard coding path which might be fragile and breaks easily, or just want file path independent in build time. So, the option is added.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -414,6 +414,12 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
     Path to an existing directory used to cache :file:`proj.db` to speed-up
     subsequent builds without modifications to source SQL files.
 
+.. option:: EMBED_PROJ_DATA_PATH
+
+    .. versionadded:: 9.4
+
+    Embed ``PROJ_DATA_PATH`` data files location, default ON.
+
 
 Building on Windows with vcpkg and Visual Studio 2017 or 2019
 --------------------------------------------------------------------------------
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -418,9 +418,7 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
 
     .. versionadded:: 9.5
 
-    Embed ``PROJ_DATA_PATH`` data files location as an alternative paths for
-    external libraries. Disable to avoid setting this non-relocatable hardcoded
-    paths. Default ON.
+    Embed ``PROJ_DATA`` hard-coded alternative path for data files location. Disable to avoid setting this non-relocatable hard-coded path. Default ON.
 
 
 Building on Windows with vcpkg and Visual Studio 2017 or 2019
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -416,9 +416,11 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
 
 .. option:: EMBED_PROJ_DATA_PATH
 
-    .. versionadded:: 9.4
+    .. versionadded:: 9.5
 
-    Embed ``PROJ_DATA_PATH`` data files location, default ON.
+    Embed ``PROJ_DATA_PATH`` data files location as an alternative paths for
+    external libraries. Disable to avoid setting this non-relocatable hardcoded
+    paths. Default ON.
 
 
 Building on Windows with vcpkg and Visual Studio 2017 or 2019
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,6 +267,8 @@ endif()
 
 ################################################################################
 
+option(EMBED_PROJ_DATA_PATH "Whether the PROJ_DATA_PATH should be embedded" ON)
+
 if(DEFINED PROJ_LIB_ENV_VAR_TRIED_LAST)
   set(PROJ_DATA_ENV_VAR_TRIED_LAST ${PROJ_LIB_ENV_VAR_TRIED_LAST})
   message(WARNING "PROJ_LIB_ENV_VAR_TRIED_LAST option has been renamed to PROJ_DATA_ENV_VAR_TRIED_LAST. PROJ_LIB_ENV_VAR_TRIED_LAST is still working for now, but may be completely replaced by PROJ_DATA_ENV_VAR_TRIED_LAST in a future release")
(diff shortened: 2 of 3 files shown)


kbevers:
Could you please explain why this is desirable to you?

rouault:
> Could you please explain why this is desirable to you?

Besides original reporter's own needs, I see this could be of value for example for conda-forge builds, where PROJ is relocated in a arbitrary location on the user machine, and thus hardcoding PROJ_DATA makes little sense.

kbevers:
I'm sure there's a benefit, I just think it is good form to explain why we should consider a pull request.

This needs adding to the docs as well, by the way. Under CMake configure options in https://github.com/OSGeo/PROJ/blob/master/docs/source/install.rst

jjimenezshaw:
> > Could you please explain why this is desirable to you?
> 
> Besides original reporter's own needs, I see this could be of value for example for conda-forge builds, where PROJ is relocated in a arbitrary location on the user machine, and thus hardcoding PROJ_DATA makes little sense.

That would be useful for me too. When you do not know where your proj is going to be located, it makes no sense. Even worse, it could be that there is an older version installed there, and just takes the wrong proj.db

docs/source/install.rst
```diff
@@ -414,6 +414,12 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
     Path to an existing directory used to cache :file:`proj.db` to speed-up
     subsequent builds without modifications to source SQL files.
 
+.. option:: EMBED_PROJ_DATA_PATH
+
+    .. versionadded:: 9.4
```
jjimenezshaw: 9.4.1 is already released. Should it be 9.5, the next release?

kbevers: Yes

kbevers: Also, I think the description below this needs a bit more work. When I'm reading this it only clear to my that this hardcodes the `PROJ_DATA_PATH` into the binaries because I've read the discussion in this PR. A short explanation of why you might want to turn this off would be very helpful

jjimenezshaw:
> > > Could you please explain why this is desirable to you?
> > 
> > 
> > Besides original reporter's own needs, I see this could be of value for example for conda-forge builds, where PROJ is relocated in a arbitrary location on the user machine, and thus hardcoding PROJ_DATA makes little sense.
> 
> That would be useful for me too. When you do not know where your proj is going to be located, it makes no sense. Even worse, it could be that there is an older version installed there, and just takes the wrong proj.db

Another use case that happened to me (just in case it is happening to anybody else): "it works in my (developer's) computer"

I am using PROJ in an application that can be located anywhere in the filesystem. PROJ_DATA had to be set accordingly at the beginning (or right before) of the process. However it was working (in my computer) without setting it... because magically it was using the proj.db from my compilation folder, not the distribution folder. Even if the CI computer, because it was first compiling PROJ and then the application. We noticed it later when we tested manually somewhere else that never compiled PROJ. We changed the CI workflow to build and run in different instances.

rouault:
Actually, I stand corrected regarding conda-forge. Conda has magic to patch strings !

Hard-coded path (with placeholder padding) in the "raw" package downloaded from the forge:
```
$ strings /home/even/miniconda3/pkgs/proj-9.4.1-hb784bbd_0/lib/libproj.so.25.9.4.1 | grep share/proj
../share/proj
/home/conda/feedstock_root/build_artifacts/proj_1718272424126/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/proj
```

vs substituted paths in the package installed in the conda env:

```
$ strings /home/even/miniconda3/envs/gdal_3.9.1/lib/libproj.so.25.9.4.1 | grep share/proj
../share/proj
/home/even/miniconda3/envs/gdal_3.9.1/share/proj
```

jjimenezshaw:
> Conda has magic to patch strings !

I am a bit scared. How is it doing that?

rouault:
> I am a bit scared. How is it doing that?

https://docs.conda.io/projects/conda-build/en/stable/resources/make-relocatable.html

docs/source/install.rst
WangWeiLin-MV: Hi, I have made the changes, please review this PR and if you have any questions, please let me know.

docs/source/install.rst
```diff
@@ -414,6 +414,14 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
     Path to an existing directory used to cache :file:`proj.db` to speed-up
     subsequent builds without modifications to source SQL files.
 
+.. option:: EMBED_PROJ_DATA_PATH
+
+    .. versionadded:: 9.5
+
+    Embed ``PROJ_DATA_PATH`` data files location as an alternative paths for
```
jjimenezshaw: Sorry for asking. But I do not understand the meaning of. "data files location as an alternative paths for external libraries".

I was not aware until now of a variable called `PROJ_DATA_PATH`, but only about `PROJ_DATA`.

If I understand correctly, currently the variable `PROJ_DATA` has a hardcoded default value. You want to give an option (off by default) to not set that hardcoded value, leaving the responsibility to the final user.
`PROJ_DATA` is about configuration and resource files like `proj.db` and grid files, not about external libraries.

(I am in favor of this setting. I will use it! The question is about the explanation)

WangWeiLin-MV: Sorry I meant this definition is optional. I'm not very familiar with this project, I don't know what kind of comments should be added. I'd be grateful if you could give me some advice.

P.S. Actually, I think this doesn't need reason to "not hardcode variables", instead embed it requires a sufficient reason.

docs/source/install.rst
```diff
@@ -414,6 +414,14 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
...
+    external libraries. Disable to avoid setting this non-relocatable hardcoded
+    paths. Default ON.
```
jjimenezshaw:
```suggestion
    Embed ``PROJ_DATA`` hard-coded alternative path for data files location. Disable to avoid setting this non-relocatable hard-coded path. Default ON.
```

PR #4211, opened 2024-07-31
ISEA Inverse projection (#3047)

### LINKED ISSUE #3047: ISEA: inverse projection method is not implemented
As [expressed here](https://gis.stackexchange.com/q/423014/7505), even with QGIS is impossible to "see ISEA". 
A projection is useless if we cannot visualize it... Without visualization you lose confidence and a multitude of applications.

**At PROJ Readmes and/or documentation**, ideal is to citate a "how to visualize" tutorial, or explain *"Danger: you [can calculate distances and areas](https://gis.stackexchange.com/q/418623/7505), but can't visualize because there are implementation limitations"*.

jerstlouis:
- Integrating code ported from Java to eC, then eC to C++ implementing the inverse ISEA projection
- Originally from Franz-Benjamin Mocnik's ISEA implementation found at https://github.com/mocnik-science/geogrid/blob/master/src/main/java/org/giscience/utils/geogrid/projections/ISEAProjection.java (MIT License)
- NOTE: This only supports the default planar options (either `+orient=pole` or `+orient=isea`, but `+orient=pole` was not tested)
- [x] Closes #3047


jerstlouis:
@rouault @MichaelJendryke @ppKrauss @ldesousa would you be able to help test / validate? Thanks!

jerstlouis:
This inverse implementation is now working (successful latlon==>ISEA==>latlon roundtrip) now assuming the parameters:

`+proj=isea +R=6371007.18091875`.

For `+orient=pole`, it would just be a matter of using the `polarISEA` object instead of `standardISEA` (I am not sure how to detect that parameter).

For other values of `+R` / ellipsoid etc., I don't know if it's just a matter of multiplying `xy.x` and `xy.y` by a member of the `PJ *P` argument instead of by `earthAuthalicRadius`?

For the `+mode` other than `plane`, is it possible to simply not support these for the inverse?

Will different values of `+x_0`, `+y_0`, `+lat_0`, `+lon_0` be handled automatically by code outside _isea.cpp_?

Not sure about supporting the other parameters.

I've clarified that our (provisionary) [OGC:1534 CRS](https://maps.gnosis.earth/ogcapi/collections/gebco/map?crs=OGC:1534) is equivalent to:

```
+proj=pipeline
   +step +proj=geoc
   +step +proj=isea +R=6371007.18091875 +x_0=19186144.8709401525557041 +y_0=-3323137.7717845365405083 
```

which implies a geodetic->geocentric conversion that does not get performed automatically by `+isea`.
From past experiments, this conversion was required to align with data registered to an ISEA3H DGGS.

**Not sure whether this should automatically be done by `+proj=isea`?**

For the [ISEA 5x6 space](https://maps.gnosis.earth/ogcapi/collections/gebco/map?crs=OGC:153456) (provisionally called OGC:153456) used to define the [ISEA9R 2D Tile Matrix Set](https://maps.gnosis.earth/ogcapi/tileMatrixSets/ISEA9R) (where 10/30 level 0 tiles are valid -- could also define an ISEA4R quad-tree 2DTMS), which is also a convenient space for doing ISEA3H calculations, we can simply add an affine transform:

```
+proj=pipeline
   +step +proj=geoc
   +step +proj=isea +R=6371007.18091875 +x_0=19186144.8709401525557041 +y_0=-3323137.7717845365405083
   +step +proj=affine +inv
      +s11=3837228.9741880306974053 +s12=3837228.9741880302317441
      +s21=6646275.5435690730810165 +s22=-6646275.543569073081016
```

In the 5x6 space, the twelve icosahedron vertices are very conveniently/intuitively at:
- 1,0 (same as 2,1; 3,2; 4,3; 5,4) "top" of unfolded icosahedron -- corresponding to applied orientation: 58.454376167463 N, 11.25 E for standard `+orient=isea`
- 0,2 (same as 1,3; 2,4; 3,5; 4,6) "bottom" of unfolded icosahedron -- 58.454376167463 S, 168.75 W for `+orient=isea`
- 0,0 (same as 5,5) -- left vertex
- 1,1 -- first top dent
- 2,2 -- second top dent
- 3,3 -- third top dent
- 4,4 -- fourth top dent
- 0, 1 (same as 5, 6) -- right vertex
- 1,2 -- first bottom dent
- 2,3 -- second bottom dent
- 3,4 -- third bottom dent
- 4,5 -- fourth bottom dent

and these points are all on the equator:
- 0, 1 (same as 5, 6) -- right vertex
- 1, 1.5
- 2, 2 -- second top dent vertex
- 3, 3 -- third top dent vertex
- 3.5, 4
- 4, 5 -- fourth bottom dent vertex

cc. @allixender @perrypeterson @mpadillaruiz

jerstlouis:
For the appveyor MSVC failure, maybe we need to add `#include <algorithm>`:

https://stackoverflow.com/questions/19439670/min-max-not-a-member-of-std-errors-when-building-opencv-2-4-6-on-windows-8

jerstlouis:
The need for the pipeline to do the correct geodetic->geocentric conversion is problematic with e.g., `gdalwarp`:

```
gdalwarp  ~/Desktop/gebco.tiff test.tiff -s_srs "EPSG:4326" -t_srs "+proj=pipeline +step +proj=geoc +step +proj=isea +R=6371007.18091875 +x_0=19186144.8709401525557041 +y_0=-3323137.7717845365405083"
ERROR 1: PROJ: proj_crs_get_coordinate_system: Object is not a SingleCRS
ERROR 1: PROJ: proj_crs_get_coordinate_system: Object is not a SingleCRS
Using band 4 of destination image as alpha.
Using band 4 of source image as alpha.
Warning 1: The definition of geographic CRS EPSG:4326 got from GeoTIFF keys is not the same as the one from the EPSG registry, which may cause issues during reprojection operations. Set GTIFF_SRS_SOURCE configuration option to EPSG to use official parameters (overriding the ones from GeoTIFF keys), or to GEOKEYS to use custom values from GeoTIFF keys and drop the EPSG code.
ERROR 1: PROJ: proj_crs_get_coordinate_system: Object is not a SingleCRS
ERROR 1: PROJ: proj_crs_get_coordinate_system: Object is not a SingleCRS
ERROR 1: PROJ: proj_crs_get_datum: Object is not a SingleCRS
ERROR 1: PROJ: proj_crs_get_datum_ensemble: Object is not a SingleCRS
ERROR 1: PROJ: proj_crs_get_coordinate_system: Object is not a SingleCRS
ERROR 1: PROJ: proj_create_operations: target_crs is not a CRS or a CoordinateMetadata
ERROR 6: Cannot find coordinate operations from `EPSG:4326' to `{
  "$schema": "https://proj.org/schemas/v0.7/projjson.schema.json",
  "type": "Conversion",
  "name": "PROJ-based coordinate operation",
  "method": {
    "name": "PROJ-based operation method: +proj=pipeline +step +proj=geoc +step +proj=isea +R=6371007.18091875 +x_0=19186144.8709401525557041 +y_0=-3323137.7717845365405083 +type=crs"
  },
  "parameters": []
}'
```

whereas this kindof works now (complained about missing inverse before):

```
gdalwarp  ~/Desktop/gebco.tiff test.tiff -s_srs "EPSG:4326" -t_srs "+proj=isea +R=6371007.18091875 +x_0=19186144.8709401525557041 +y_0=-3323137.7717845365405083"
Using band 4 of destination image as alpha.
Using band 4 of source image as alpha.
Warning 1: The definition of geographic CRS EPSG:4326 got from GeoTIFF keys is not the same as the one from the EPSG registry, which may cause issues during reprojection operations. Set GTIFF_SRS_SOURCE configuration option to EPSG to use official parameters (overriding the ones from GeoTIFF keys), or to GEOKEYS to use custom values from GeoTIFF keys and drop the EPSG code.
Processing /home/jerome/Desktop/gebco.tiff [1/1] : 0...10...20...30...40...50...60...70...80...90...100 - done.
```

except for some missing bits at the extremities:

gdalWarpedToISEA

jerstlouis:
@rouault This inverse seems to be working quite well. This is the sample [GeoJSON file](https://portal.ogc.org/files/?artifact_id=108798) I mentioned in [#3047](https://github.com/OSGeo/PROJ/issues/3047#issuecomment-2209510495) inverse-projected to EPSG:4326 in QGIS by doing a _Set Layer CRS_ to:

`+proj=isea +R=6371007.18091875 +x_0=19186144.8709401525557041 +y_0=-3323137.7717845365405083`

qgisInverseISEA

(together with a Natural Earth countries layer in EPSG:4326).

Or de-projecting countries of the whole world in ISEA projection:

https://maps.gnosis.earth/ogcapi/collections/NaturalEarth:cultural:ne_10m_admin_0_countries/items.geojson?limit=1000&crs=OGC:1534

countriesISEA

Here's how a full ISEA GEBCO raster (https://maps.gnosis.earth/ogcapi/collections/gebco/map.tiff?crs=OGC:1534) deprojects in QGIS (just a few minor glitches around the interruptions):

GEBOUnprojected


Should I file a separate issue about whether `+proj=isea` should actually be doing the `+proj=geoc` step automatically?

I believe that the correct ISEA projection actually implies that step, and just as mentioned in https://github.com/OSGeo/gdal/issues/10537, it will still not be possible to accurately visualize things since the `+proj=pipeline` cannot be used with _Set Layer CRS_ etc.

cc. @ppKrauss @mpadillaruiz @ldesousa @perrypeterson @allixender

rouault:
> Should I file a separate issue about whether `+proj=isea` should actually be doing the `+proj=geoc` step automatically?

no idea. What does the literature about this projection says about that? But the fact that you define a spherical ellipsoid with +R=6371007.18091875 and you mention the authalic radius makes me think that perhaps you would need to use the +R_A switch instead (https://proj.org/en/9.4/usage/ellipsoids.html#ellipsoid-spherification-parameters instead), which is sometimes used for a number of projection methods

jerstlouis:
Thanks for triggering the builds @rouault .

The compilation warning as errors settings are a bit over the top... Any chance for some help addressing or bypassing those issues, or is there otherwise an easy way to test this locally? cmake / make / make test all work perfectly fine on my Linux system (_g++ (GCC) 14.1.1 20240720_), but I don't get those warnings as errors on what seems to be the compilers' opinionated idea of proper C++ style (not being able to trigger the CI builds myself doesn't help either).

Is there a particular set of GCC flags that will likely results in all builds passing if I manage to compile locally with them?

> What does the literature about this projection says about that?

As far as I know the official ISEA paper is the [Snyder1992 paper](https://ia800704.us.archive.org/view_archive.php?archive=/24/items/wikipedia-scholarly-sources-corpus/10.3138.zip&file=10.3138%252F27h7-8k88-4882-1752.pdf).

But this paper may actually be describing multiple projections as @ldesousa mentioned, and it definitely mentions multiple polyhedrons such as the truncated icosahedron for example, whereas ISEA is really about the icosahedron base solid (_Figure 12_, though that does not show the standard ISEA orientation but the `+orient=pole` one).
I could never really understand that paper, despite several attempts at reading it, because it's all over the place with these different solids, and how it talks about hexagonal faces where the ISEA projection itself has nothing to do with hexagons, and then later on page 16 _Application to the Platonic Polyhedral Globes_ describes this can be simplified ignoring hexagons for the platonic solids.

The term geodetic or geocentric do not appear, and the word _latitude_ appears on page 14 and 20 only, and the letter ϕ. and _lat_ is used and "geographic laititude" is used which would make you think it it is talking about geodetic latitude.
However, conceptually it is a spherical Earth with the surface area of the ellipsoid which is mapped to the icosahedron, so converting to geocentric first would make sense. 

The main reason I've started to think this geodetic ==> geocentric step was necessary is because it was needed to correctly integrate data served by PYXIS / uCalgary using their [PYX0 format](https://docs.ogc.org/per/23-010.html#_the_pyx0_format), for things to align properly in our visualization client. We mention this in the [FMSDI Phase 3 ER](https://docs.ogc.org/per/23-010.html#_challenges_and_lessons_learned_7).

I am hoping that experts on the topic (@sahrk @franz-benjamin @shatzi @perrypeterson @mpadillaruiz @ldesousa  @allixender) could chime in, so that we could clarify whether the geocentric conversion is implied or not. Essentially we need an agreement on what the ISEA projection really is. This is why a new issue might be a better place for this discussion than this pull request or the original missing inverse issue...

> But the fact that you define a spherical ellipsoid with +R=6371007.18091875 and you mention the authalic radius makes me think that perhaps you would need to use the +R_A switch instead (https://proj.org/en/9.4/usage/ellipsoids.html#ellipsoid-spherification-parameters instead), which is sometimes used for a number of projection methods

I did try the `+R_A` switch, but I was repeatedly/continuously mixing it up with `+R_a`.

The `6371007.18091875` is calculated as:

```eC
#define wgs84InvFlattening 298.257223563
#define wgs84Major         6378137.0
#define wgs84Minor         (wgs84Major - (wgs84Major / wgs84InvFlattening)) // 6356752.3142451792955399
#define a2                 (wgs84Major * wgs84Major)
#define c2                 (wgs84Minor * wgs84Minor)
#define eccentricity       0.0818191908426 // sqrt(1.0 - (c2/a2));
#define log1pe_1me         0.1640050079196   // log((1+e)/(1-e)))
#define S                  (Pi * (2 * a2 + c2/eccentricity * log1pe_1me))
#define earthAuthalicRadius 6371007.18091875 //sqrt(S / (4 * Pi)); // R -- authalic sphere radius for WGS84 [6371007.1809184747 m]
```

This also matches the [numbers here](https://www.jpz.se/Html_filer/wgs_84.html) -- √(510 065 621.724 km2 / (4 * π)) =
6 371 007.1809 m

The [documentation](https://proj.org/en/9.4/usage/ellipsoids.html#ellipsoid-spherification-parameters) is a bit confusing, compounded by the fact that both `R_a` and `R_A` exist and are completely different things.

> `+R_A`: A sphere with the same surface area as the ellipsoid.

I imagine this ellipsoid defaults to GRS80? So if I want to use an authalic sphere with the same surface area as WGS84, would I say `+ellps=WGS84 +R_A`?
Then that should be exactly the same as `+R=6371007.18091875`? The results are the same (when not confusing case-sensitive `+R_a` with `+R_A` as I must have done several times):

```
echo -168.7500000     58.2825256 | proj +proj=isea +R=6371007.18091875
-19186144.87	3323137.77

echo -168.7500000     58.2825256 | proj +proj=isea +ellps=WGS84 +R_A
-19186144.87	3323137.77
```

Omitting the `+ellps=WGS84` yields the same results up to three decimals, likely because WGS84 and GRS80 differ so slightly...

However, I still get a small difference between the `+R=6371007.18091875` vs. `+ellps=WGS84 +R_A`:

```
echo -168.7500000     58.2825256 | proj +proj=isea +ellps=WGS84 +R_A -f "%.10f"
-19186144.8722200952	3323137.7725913548

echo -168.7500000     58.2825256 | proj +proj=isea +R=6371007.18091875 -f "%.10f"
-19186144.8717271797	3323137.7725059795
```

Possibly related discussion by [other confused users](https://gis.stackexchange.com/questions/441036/proj-r-a-discrepancy-with-computed-authalic-radius).

Does any of this have any relation to the extra `+proj=geoc` step?

Thanks!

sahrk:
> As far as I know the official ISEA paper is the [Snyder1992 paper](https://ia800704.us.archive.org/view_archive.php?archive=/24/items/wikipedia-scholarly-sources-corpus/10.3138.zip&file=10.3138%252F27h7-8k88-4882-1752.pdf).
> 
> But this paper may actually be describing multiple projections as @ldesousa mentioned,

Snyder's paper gives projections for multiple polyhedra using the same approach. For convenience we named his icosahedral projection "ISEA". 

> The term geodetic or geocentric do not appear, and the word _latitude_ appears on page 14 and 20 only, and the letter ϕ. and _lat_ is used and "geographic laititude" is used which would make you think it it is talking about geodetic latitude. However, conceptually it is a spherical Earth with the surface area of the ellipsoid which is mapped to the icosahedron, so converting to geocentric first would make sense.

The ISEA projection is defined on the sphere. On the sphere there is no distinction between geodetic and geocentric latitudes; they are always equal. If you're integrating data defined on an ellipsoid then some sort of conversion will be necessary. 

Note that we chose the authalic radius for the ISEA3H DGGS so that authalic latitude could be used to define equal area cells on the ellipsoid. I believe any other conversion approach results in cells on the ellipsoid which are not equal area.

> ISEA is really about the icosahedron base solid (Figure 12, though that does not show the standard ISEA orientation but the +orient=pole one).

I don't think Snyder specified a "standard" orientation for his projection. Like the acronym ISEA, the orientation we chose for ISEA-based DGGS was something we came up with later.

> it talks about hexagonal faces where the ISEA projection itself has nothing to do with hexagons

He's referring to the hexagonal faces of a truncated icosahedron, which is one of the polyhedra he defines his projection on. This is unrelated to the hexagonal cells of a DGGS.

ldesousa:
@jerstlouis This is a commendable effort, glad that you took upon it. I can't test it right now, but will keep an eye on this PR.

The projection we call "Snyder Equal-area" is actually defined for a right triangle on the spherical surface. It can be used to project a sphere into any polyhedron whose faces can be decomposed into right triangles. Snyder himself was primarily concerned with the Dodecahedron (his favourite), the Icosahedron and the truncated Icosahedron, but the same formulation should apply beyond those.

Neither John Snyder, nor those who followed immediately on his steps towards DGGS (White, Kimmerling, etc), defined a "standard" or "default" orientation of the polyhedra. Concerning the Icosahedron, those articles from the early 90s debate to some extent the "inconvenience" of an icosahedron positioned with vertices on the poles always having at least one other vertex on land.

In conclusion: the orientation of the polyhedron should be a parameter to any implementation of Snyder's Equal-area projection. That could be the position of a vertex plus an azimuth to a proximate (adjacent?) vertex. That's what @sahrk did in dggrid.

jerstlouis:
@sahrk Thank you very much for the clarifications!

> The ISEA projection is defined on the sphere. On the sphere there is no distinction between geodetic and geocentric latitudes; they are always equal.

I guess this means that `+proj=isea` should not do this `+proj=geoc` step automatically, but it should also generally not be used without a perfectly spherical ellipsoid.

> Note that we chose the authalic radius for the ISEA3H DGGS so that authalic latitude could be used to define equal area cells on the ellipsoid. I believe any other conversion approach results in cells on the ellipsoid which are not equal area.

Assuming that this [authalic latitude](https://en.wikipedia.org/wiki/Latitude#Authalic_latitude) conversion is not already done by the ISEA projection in [PROJ](https://github.com/OSGeo/PROJ/blob/master/src/projections/isea.cpp#L302) and [geogrid](https://github.com/mocnik-science/geogrid/blob/main/src/main/java/org/giscience/utils/geogrid/projections/ISEAProjection.java#L395), then I suspect that our implementation of the ISEAH grid as well the PYXIS implementation we tested with in the FMSDI might not be the true ISEA3H DGGS, since we seemed to be using the geocentric latitude rather than the authalic latitude. @perrypeterson @shatzi Any insights / evidence of the contrary about that would be very useful.

The zones are clearly not perfectly equal area as computed on the WGS84 ellipsoid (though within the 1% margin allowed by Topic 21 v2 "Equal Area") [as can be seen here](https://maps.gnosis.earth/ogcapi/collections/gebco/dggs/ISEA3H/zones).

I am curious about whether [geogrid](https://github.com/mocnik-science/geogrid/) is actually using the geodetic latitude for its DGGS, or converting from/to geocentric or authalic elsewhere in the code? I could not find anywhere in the code where this is done. @franz-benjamin Any insight on that would be very useful.

So ideally this OGC:1534 CRS which we want to define for the purpose of facilitating the implementation of ISEA DGGs should probably imply this geodetic ==> authalic latitude conversion before applying the ISEA projection (rather than geodetic ==> geocentric).

That currently seems even more difficult to do with PROJ, since that is not one of the [conversions](https://proj.org/en/9.4/operations/conversions/index.html#) already supported by the pipelines.

@sahrk Does this sound like a reasonable assessment?

In order to achieve interoperability between multiple implementations of ISEA DGGs, we need clarity on how to bring in and take out georeferenced information correctly, including how geodetic latitude is mapped to the ISEA sphere prior to mapping to the icosahedron. Right now it seems like different implementations of ISEA3H are using three different types of latitudes: geodetic (geogrid), geocentric (PYXIS & GNOSIS), and authalic (DGGRID).

NOTE: [GDGGS03](https://webpages.sou.edu/~sahrk/sqspc/pubs/gdggs03.pdf) does not mention authalic latitudes.

@ldesousa 

> In conclusion: the orientation of the polyhedron should be a parameter to any implementation of Snyder's Equal-area projection. That could be the position of a vertex plus an azimuth to a proximate (adjacent?) vertex

Currently, PROJ supports `+orient=pole` and `+orient=isea` (the [standard ISEA orientation for DGGs](https://webpages.sou.edu/~sahrk/dgg/orientation/stdorient.html)), as well as an arbitrary azimuth parameter.

This implementation of the inverse supports both of these orientations (and could easily support additional ones), but does not yet support the azimuth parameter.

I'm also just realizing that we may need an additional parameter reflecting this azimuth for describing the orientation of polyhedron-based DGGs in the [draft DGGRS definition](https://docs.ogc.org/DRAFTS/21-038.html#_json_schema_for_defining_a_dggrs) -- e.g., to support the [R. Buckminster Fuller's Dymaxion Orientation](https://webpages.sou.edu/~sahrk/dgg/orientation/dymorient.html).

Thanks all!

sahrk:
> @sahrk Thank you very much for the clarifications!

@jerstlouis you're welcome! 

> > The ISEA projection is defined on the sphere. On the sphere there is no distinction between geodetic and geocentric latitudes; they are always equal.
> 
> I guess this means that `+proj=isea` should not do this `+proj=geoc` step automatically, but it should also generally not be used without a perfectly spherical ellipsoid.

The ISEA projection as defined by Snyder and implemented in DGGRID is a spherical projection only. So that sounds logical to me.

> > Note that we chose the authalic radius for the ISEA3H DGGS so that authalic latitude could be used to define equal area cells on the ellipsoid. I believe any other conversion approach results in cells on the ellipsoid which are not equal area.
> 
> Assuming that this [authalic latitude](https://en.wikipedia.org/wiki/Latitude#Authalic_latitude) conversion is not already done by the ISEA projection in [PROJ](https://github.com/OSGeo/PROJ/blob/master/src/projections/isea.cpp#L302) and [geogrid](https://github.com/mocnik-science/geogrid/blob/main/src/main/java/org/giscience/utils/geogrid/projections/ISEAProjection.java#L395), then I suspect that our implementation of the ISEAH grid as well the PYXIS implementation we tested with in the FMSDI might not be the true ISEA3H DGGS, since we seemed to be using the geocentric latitude rather than the authalic latitude. @perrypeterson @shatzi Any insights / evidence of the contrary about that would be very useful.
> 
> The zones are clearly not perfectly equal area as computed on the WGS84 ellipsoid (though within the 1% margin allowed by Topic 21 v2 "Equal Area") [as can be seen here](https://maps.gnosis.earth/ogcapi/collections/gebco/dggs/ISEA3H/zones).

I got a 503 error from your data server, but one caveat on cell area variance: if you’re calculating the area on the earth’s surface by treating the edges as great circle arcs then you’re not going to get the exact correct area for an ISEA cell, even on the sphere.

> So ideally this OGC:1534 CRS which we want to define for the purpose of facilitating the implementation of ISEA DGGs should probably imply this geodetic ==> authalic latitude conversion before applying the ISEA projection (rather than geodetic ==> geocentric).
> That currently seems even more difficult to do with PROJ, since that is not one of the [conversions](https://proj.org/en/9.4/operations/conversions/index.html#) already supported by the pipelines.
> 
> @sahrk Does this sound like a reasonable assessment?

If you want an ellipsoidal datum and you want equal area cells then I think that’s your only choice.

btw I’m not a fan of “zones” for cells. It minimizes the important (if so far unrealized) use cases where cell indexes indicate points, rather than areal zones.

> In order to achieve interoperability between multiple implementations of ISEA DGGs, we need clarity on how to bring in and take out georeferenced information correctly, including how geodetic latitude is mapped to the ISEA sphere prior to mapping to the icosahedron. Right now it seems like different implementations of ISEA3H are using three different types of latitudes: geodetic (geogrid), geocentric (PYXIS & GNOSIS), and authalic (DGGRID).

I’ll give you a fourth. Probably the most widely disseminated and used dataset on an ISEA grid is the ESA’s SMOS soil moisture products. They wanted 15km cell node spacing, so chose an ISEA4H resolution 9. They also needed it on the WGS84 ellipsoid. So they just took the spherical latitudes produced by DGGRID and treated them as WGS84 ellipsoid latitudes. At a resolution of 15km the error is too small to worry about; it may even be below the representational precision. I’ve referred to that faux ellipsoidal projection as “ISEL”.

> NOTE: [GDGGS03](https://webpages.sou.edu/~sahrk/sqspc/pubs/gdggs03.pdf) does not mention authalic latitudes.

I don’t think it mentions ellipsoids/spheres at all; that was beyond the scope of the paper. But DGGRID, which was essentially companion software for that paper, has always used the authalic radius.

jerstlouis:
@sahrk 

> I got a 503 error from your data server

Sorry, the server seems to be prone to these since this week's last update, and we're currently struggling to figure out the cause. I restarted it so hopefully you can try it again before it goes down. Trying to fix these ASAP.

> but one caveat on cell area variance: if you’re calculating the area on the earth’s surface by treating the edges as great circle arcs then you’re not going to get the exact correct area for an ISEA cell, even on the sphere.

We are not doing that -- we're using several points along the edges and de-projecting each of these to calculate this accurately. Here's a sample image of what that looks like for level 0 zones and the same info as a table below (in summary, compared to the mean zone area (Earth surface area / 12 zones), the 4 polar (Earth geographic poles) level 0 zones have 0.2% smaller areas, whereas the 4 equatorial level 0 zones have 0.17% larger areas, while the remaining 4 zones have 0.03% larger areas):

ZoneArea

**Level 0 Zones**
(variance relative to 42,505,468.48 km² reference mean zone surface area)

|Zone ID |Area|Variance|Min Lat|Min Lon|Max Lat|Max Lon|
|---|---|---|---|---|---|---|
|A0-0-A|42,420,592.22 km²|-0.2 %|21.0337617856831|101.2500000030306|90|-78.7500000030306|
|A0-0-B|42,420,594.2 km²|-0.2 %|21.0337617833224|-135.8114780636077|90|158.558247808623|
|A1-0-A|42,577,990.62 km²|+0.17 %|-35.4460114254118|-168.7500000000045|35.446011427561|-99.6551574473396|
|A2-0-A|42,517,825.61 km²|+0.03 %|0.0000000009586|-123.7499999994206|69.2228046463321|-33.7500000009366|
|A3-0-A|42,517,826.83 km²|+0.03 %|-69.2228046463321|-123.7499999990634|0.0000000010893|-33.7499999994906|
|A4-0-A|42,577,985.62 km²|+0.17 %|-35.446011425382|-57.8448425514859|35.4460114254118|11.2500000000041|
|A5-0-A|42,420,594.2 km²|-0.2 %|-90|-78.7499999969695|-21.0337617833224|101.2499999969694|
|A6-0-A|42,577,985.62 km²|+0.17 %|-35.446011425382|11.2499999999959|35.4460114254118|80.3448425514859|
|A7-0-A|42,517,827.45 km²|+0.03 %|-69.2228046463321|56.2499999994905|0.0000000010894|146.2499999990634|
|A8-0-A|42,517,825.61 km²|+0.03 %|0.0000000009586|56.2500000009365|69.2228046463321|146.2499999994207|
|A9-0-A|42,577,988.65 km²|+0.17 %|-35.4460114254118|122.1551574473396|35.446011427561|-168.7499999999959|
|A9-0-C|42,420,588.11 km²|-0.2 %|-90|43.9418186867605|-21.0337617833225|-21.6866565645132|

Total surface area: **510,065,624.74 km²**

> btw I’m not a fan of “zones” for cells. It minimizes the important (if so far unrealized) use cases where cell indexes indicate points, rather than areal zones.

I am really happy you bring this up! We had some recent discussion for [DGGS-JSON](https://docs.ogc.org/DRAFTS/21-038.html#rc_data-json) where I was advocating for a `representedValue` member that could indicate whether the sampled values best represents the _zone area_, the _centroid_ or a _vertex_ (a particular vertex to be defined -- not an easy task for ISEA3H). There was a lot of pushback to allow anything except _zone area_, but I believe there are some important, if less popular, use cases for points (which I think is what you're pointing out), including for example storing/transporting data registered to an ISEA3H even level zone using dual ISEA9R grids, where an area of ISEA3H corresponds to a vertex in ISEA9R (as we were doing in experiments with PYXIS for the FMSDI pilot project). I would really love to hear more about other use cases for representing points (but we're probably already getting quite off-topic).

As for the term "zone", itself, the two terms are defined as such in [OGC Topic 21](https://docs.ogc.org/as/20-040r3/20-040r3.html):

> 4.2. **cell** <DGGS> spatial, spatio-temporal or temporal unit of geometry with dimension greater than 0, associated with a zone ([Clause 4.52](https://docs.ogc.org/as/20-040r3/20-040r3.html#zone))
> 4.52. **zone** <DGGS> particular region of space-time

So in both cases, they seem to imply this areal meaning.
Another reason we're avoiding _cell_ in [OGC API - DGGS](https://docs.ogc.org/DRAFTS/21-038.html) is because of the potential confusion with a gridded coverage cell, and the possibility to retrieve coverage data for a particular zone, where the gridded coverage cells may not necessarily match DGGS zones at all.

> So they just took the spherical latitudes produced by DGGRID and treated them as WGS84 ellipsoid latitudes. 

Wouldn't that be the same as the first geodetic latitude case I mention that I believe geogrid may be using?
i.e., importing data registered to a WGS84 ellipsoid by taking the WGS84 latitudes and using that directly as the spherical latitude mapped to the icosahedron.

> But DGGRID, which was essentially companion software for that paper, has always used the authalic radius.

Using a sphere with a radius of 6 371 007.1809 m corresonding to √(510 065 621.724 km2 / (4 * π)) (the surface area of the WGS84 ellipsoid) is one thing, which we're doing in both GNOSIS and PYXIS, and which can be specified to PROJ with `+R=6371007.18091875`.

But using a preliminary conversion of geodetic ==> authalic latitude is a completely different thing, isn't it? i.e., when integrating data referenced to the WGS84 ellipsoid, applying this step to convert geodetic latitude to latitude on the sphere mapped to the icosahedron:

authlat1
where
authlat2
and
authlat3

as described at https://en.wikipedia.org/wiki/Latitude#Authalic_latitude .

This is how I understand what you mean by:

> so that authalic latitude could be used to define equal area cells on the ellipsoid. I believe any other conversion approach results in cells on the ellipsoid which are not equal area.

> If you want an ellipsoidal datum and you want equal area cells then I think that’s your only choice.

Which would result in the only area discrepency (as computed on the WGS84 ellipsoid) being the pentagonal zones having 5/6th the area of the hexagonal zones. 
Is that correct, and what DGGRID is doing?

Thanks again so much, your input is invaluable to getting this all right and achieving our goals of DGGS interoperability.

sahrk:
> @sahrk
> 
> We are not doing that -- we're using several points along the edges and de-projecting each of these to calculate this accurately. 

@jerstlouis That’s all I was getting at. As long as you’ve densified the edges in ISEA space and verified the area is stable you should be fine.

> > btw I’m not a fan of “zones” for cells. It minimizes the important (if so far unrealized) use cases where cell indexes indicate points, rather than areal zones.
> 
> I am really happy you bring this up! We had some recent discussion for [DGGS-JSON](https://docs.ogc.org/DRAFTS/21-038.html#rc_data-json)… would really love to hear more about other use cases for representing points (but we're probably already getting quite off-topic).

This is a deep topic beyond our scope here. I suggest reading the discussion on using DGGS to represent DGGS cell geometries (including vertices) in [Sahr 2019](https://utpjournals.press/doi/abs/10.3138/cart.54.1.2018-0022?journalCode=cart), in particular the section entitled “EXACT REPRESENTATION AND SUBSTRATE GRIDS”.

Using cells to represent “vector” points can be applied to any vector use case, not just the representation of cell vertices. The last time I looked at the OGC standard it appeared to me that you were claiming to solve the problem of unifying raster/vector datasets by rasterizing the vector datasets, which is not a solution to the problem at all. But we are indeed off topic…

> As for the term "zone", itself, the two terms are defined as such in [OGC Topic 21](https://docs.ogc.org/as/20-040r3/20-040r3.html):
> 
> > 4.2. **cell**  spatial, spatio-temporal or temporal unit of geometry with dimension greater than 0, associated with a zone ([Clause 4.52](https://docs.ogc.org/as/20-040r3/20-040r3.html#zone))
> > 4.52. **zone**  particular region of space-time

That looks to me like you’re doubling down on DGGS cell/zone indicating an areal unit only.

> So in both cases, they seem to imply this areal meaning. Another reason we're avoiding _cell_ in [OGC API - DGGS](https://docs.ogc.org/DRAFTS/21-038.html) is because of the potential confusion with a gridded coverage cell, and the possibility to retrieve coverage data for a particular zone, where the gridded coverage cells may not necessarily match DGGS zones at all.
 
I don’t know enough about OGC terminology to know what your definition of “gridded coverage cell” is. Just taking the words as english I would say that DGGS units are “gridded coverage cells”.

> > So they just took the spherical latitudes produced by DGGRID and treated them as WGS84 ellipsoid latitudes.
> 
> Wouldn't that be the same as the first geodetic latitude case I mention that I believe geogrid may be using? i.e., importing data registered to a WGS84 ellipsoid by taking the WGS84 latitudes and using that directly as the spherical latitude mapped to the icosahedron.

Could be. I don’t think geodetic latitude is defined the same on a sphere and an ellipse, but I’m not a geodesist.

> > But DGGRID, which was essentially companion software for that paper, has always used the authalic radius.
…
> But using a preliminary conversion of geodetic ==> authalic latitude is a completely different thing, isn't it? 

I’m not sure what point you’re trying to make. Why would we use the authalic radius otherwise? DGGRID does not currently support an ellipsoid model, but implementing the authalic conversion has been on my todo list since the beginning. 

But we’re getting off topic here….

> This is how I understand what you mean by:
> 
> > so that authalic latitude could be used to define equal area cells on the ellipsoid. I believe any other conversion approach results in cells on the ellipsoid which are not equal area.
> 
> > If you want an ellipsoidal datum and you want equal area cells then I think that’s your only choice.
> 
> Which would result in the only area discrepency being the pentagonal zones having 5/6th the area of the hexagonal zones. Is that correct?

I believe so, yes.

jerstlouis:
Thanks again @sahrk .

>  The last time I looked at the OGC standard it appeared to me that you were claiming to solve the problem of unifying raster/vector datasets by rasterizing the vector datasets, which is not a solution to the problem at all.

That is one approach (which is what the DGGS-JSON representation does for example), which I would have thought offers some partial solution to the problem, if not an ideal / perfect solution. There are other approaches, possibly better suited for vector data, such as clipping vector geometry to a particular areal zone (and generalizing the geometry with a minimum distance treshold between points based on that zone's resolution), or expressing the geometry vertices as zone IDs representing their centroids. I'll have to read your 2019 paper -- thanks!

> That looks to me like you’re doubling down on DGGS cell/zone indicating an areal unit only.

That's how the terms are specified in Topic 21 definitions, so I'm not really "doubling down", but at least for OGC standards we have to try to align with that terminology. I think addressing the use cases of representing / addressing "points" is a different topic, regardless of the terminology used.

> Just taking the words as english I would say that DGGS units are “gridded coverage cells”.

I believe that OGC "gridded coverage" as defined by [Topic 6](https://portal.ogc.org/files/?artifact_id=19820) (e.g., see 5.3.2 Grid on page 11-12) implies a **rectilinear** grid, which DGGS units are not necessarily. So for gridded coverage cells, although the cells themselves are not necessarily shaped in a particular way, they are at least organized in a rectilinear grid. e.g., the rectilinear pixels/gridded coverage cells in this PNG zone data response:

https://maps.gnosis.earth/ogcapi/collections/gebco/dggs/ISEA3H/zones/A6-0-E/data.png
zoneData

do not correspond to finer zones/cells of ISEA3H (this output is in EPSG:4326, but it would also apply to an ISEA projection output CRS). I can see however how the definition of "Grid" in Topic 6 could be extended to be more flexible so that most DGGs could potentially qualify.

Whereas for a [DGGS-JSON representation of the same zone data](https://maps.gnosis.earth/ogcapi/collections/gebco/dggs/ISEA3H/zones/A6-0-E/data.json), individual values for sub-zones correspond to finer hexagonal zones (as [visualized above](https://github.com/OSGeo/PROJ/pull/4211#issuecomment-2270520348)).

> I’m not sure what point you’re trying to make. Why would we use the authalic radius otherwise?

I actually always wondered about that!!! :) Since it seems that we've been doing data integration from WGS84-referenced data in GNOSIS and PYXIS **without** using authalic latitudes. Now I understand the purpose of that authalic radius because of that _authalic latitude_ concept, and I'll have a go at implementings this WGS84 geodetic latitude => authalic latitude conversion.

sahrk:
@jerstlouis I fear we've hijacked this pull request but just wanted to add one more thing that might interest you...

> There are other approaches, possibly better suited for vector data, such as ,,, expressing the geometry vertices as zone IDs representing their centroids.

This is the basic way to encode vector locations. This approach encodes the location at a single precision, and then (possibly) indexes that encoding using a hierarchical indexing. That is the best that is possible using systems like H3 or A3HT. But some hierarchical indexing systems allow for multi-precision quantization. See [Sahr 2013](https://discreteglobal.wpengine.com/wp-content/uploads/2016/01/GSGCBS-0.pdf) or the [corresponding slide deck](https://discreteglobal.wpengine.com/wp-content/uploads/2016/01/xuzhouTalk13.pdf).

jerstlouis:
@rouault 

Sorry for diverging into the above long parenthesis in this PR, but the discussions were highly insightful and did address the questions I had. To recap and summarize and try to move this forward:

1. Is there a set of compiler flags I can pass GCC to test compilation locally that will help me address the CI build failures (to turn on all these warnings / warnings as errors)?
2. The ISEA projection is only defined on the sphere, therefore it does not make sense to use it with a non-spherical ellipsoid. I don't know if some kind of warning would be appropriate when trying to use it with a non-spherical ellipsoid?
3. The correct preliminary step for going from WGS84 to ISEA would be to use geodetic ==> [authalic latitude](https://en.wikipedia.org/wiki/Latitude#Authalic_latitude) -- not _geocentric_ as we have been doing in GNOSIS and I believe PYXIS as well. This is the purpose of using an authalic radius for ISEA discrete global grids, and would result in truly equal area zones for zones of the same shape (e.g., ISEA3H hexagons). Does PROJ provide this geodetic ==> authalic latitude conversion capability?
4. I'm curious about the small difference between `+R=6371007.18091875` vs. `+ellps=WGS84 +R_A`, which I would have understood to be fully equivalent
```
echo -168.7500000     58.2825256 | proj +proj=isea +ellps=WGS84 +R_A -f "%.10f"
-19186144.8722200952	3323137.7725913548
echo -168.7500000     58.2825256 | proj +proj=isea +R=6371007.18091875 -f "%.10f"
-19186144.8717271797	3323137.7725059795
```

Thanks!

rouault:
> 1. Is there a set of compiler flags I can pass GCC to test compilation locally that will help me address the CI build failures (to turn on all these warnings / warnings as errors)?

With GCC: ``-Werror -Wall -Wdate-time -Werror=format-security -Werror=vla -Wextra -Wformat -Wmissing-declarations -Wshadow -Wswitch -Wunused-parameter -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Weffc++ -Woverloaded-virtual -Wzero-as-null-pointer-constant``. But we test other compilers in the CI which have their own set of warnings, so you have to fine tune your initial set of corrections, and look at the logs and fix remaining ones (usually it is straightforward)

Also make sure to use formatting by setting up pre-commit.
To do once:
```
pip install pre-commit
pre-commit install
```
Initial fixup (then this will be run automatically by git commit):
```
pre-commit run --files src/projections/isea.cpp
```

> 2. The ISEA projection is only defined on the sphere, therefore it does not make sense to use it with a non-spherical ellipsoid. I don't know if some kind of warning would be appropriate when trying to use it with a non-spherical ellipsoid?

https://proj.org/en/9.4/operations/projections/isea.html already mentions spherical. We don't warn for other projections that are meant for spherical only and used with non-spherical, so I'd say nothing to do here

> 3. Does PROJ provide this geodetic ==> authalic latitude conversion capability?

Not totally sure, but perhaps using the pj_qsfn() function as used by the laea, aea, healpix or cea projection

> 4. I'm curious about the small difference between +R=6371007.18091875 vs. +ellps=WGS84 +R_A, which I would have understood to be fully equivalent

I don't know. It looks like an approximation error somewhere. I see ell_set.cpp uses : ``P->a *= 1. - P->es * (SIXTH + P->es * (RA4 + P->es * RA6));`` for the R_A spherification. Perhaps it misses some terms?

jerstlouis:
@rouault 

Thanks. With these settings, I was only getting warnings/errors from included files which have nothing to do with this PR:

```
   ../include/proj/util.hpp:406:7: warning: ‘osgeo::proj::util::BoxedValue’ declared with greater visibility than the type of its field ‘osgeo::proj::util::BoxedValue::d’ [-Wattributes]
   ../include/proj/util.hpp:465:7: warning: ‘osgeo::proj::util::ArrayOfBaseObject’ declared with greater visibility than the type of its field ‘osgeo::proj::util::ArrayOfBaseObject::d’ [-Wattributes]
   ../include/proj/util.hpp:496:7: warning: ‘osgeo::proj::util::PropertyMap’ declared with greater visibility than the type of its field ‘osgeo::proj::util::PropertyMap::d’ [-Wattributes]
   ../include/proj/util.hpp:578:7: warning: ‘osgeo::proj::util::GenericName’ declared with greater visibility than the type of its field ‘osgeo::proj::util::GenericName::d’ [-Wattributes]
   ../include/proj/util.hpp:613:7: warning: ‘osgeo::proj::util::NameSpace’ declared with greater visibility than the type of its field ‘osgeo::proj::util::NameSpace::d’ [-Wattributes]
   ../include/proj/util.hpp:650:7: warning: ‘osgeo::proj::util::LocalName’ declared with greater visibility than the type of its field ‘osgeo::proj::util::LocalName::d’ [-Wattributes]
   ../include/proj/common.hpp:149:7: warning: ‘osgeo::proj::common::Measure’ declared with greater visibility than the type of its field ‘osgeo::proj::common::Measure::d’ [-Wattributes]
   ../include/proj/common.hpp:245:7: warning: ‘osgeo::proj::common::DateTime’ declared with greater visibility than the type of its field ‘osgeo::proj::common::DateTime::d’ [-Wattributes]
   ../include/proj/common.hpp:272:7: warning: ‘osgeo::proj::common::DataEpoch’ declared with greater visibility than the type of its field ‘osgeo::proj::common::DataEpoch::d’ [-Wattributes]
```

I did look through the Clang build and fixed a few things which hopefully will help, including adding `#undef` for macros which seemed to be causing issue because everything gets included in this `Unity/unity_8_cxx.cxx` file. I also did the pre-commit formatting thing.

> Not totally sure, but perhaps using the `pj_qsfn()` function as used by the laea, aea, healpix or cea projection

That does look like the _q(phi)_ used for the authalic latitude, where the authalic latitude is `ξ(ϕ) = asin(q(ϕ) / q(π/2))`.
However, what I meant is whether there is a `+proj=auth` equivalent to `+proj=geoc` which we could use as part of a `+proj=pipeline`. I am also wondering what the reasoning is for ISEA (which is a modification of LAEA) not using this automatically, while LAEA does?

> I don't know. It looks like an approximation error somewhere. I see ell_set.cpp uses : P->a *= 1. - P->es * (SIXTH + P->es * (RA4 + P->es * RA6)); for the R_A spherification. Perhaps it misses some terms?

I would have to try to figure out the math how this corresponds to the formulas I'm using:

eccentricity: e = sqrt(1.0 - (major^2 / minor^2))
surface of the ellipsoid: S = (Pi * (2 * major^2 + minor^2/e * log((1+e)/(1-e)))))
authalic radius: r = sqrt(S / (4 * Pi))

rouault:
> However, what I meant is whether there is a `+proj=auth` equivalent to `+proj=geoc` which we could use as part of a `+proj=pipeline`.

Doesn't exist as far as I know

>  I am also wondering what the reasoning is for ISEA (which is a modification of LAEA) not using this automatically, while LAEA does?

Perhaps it should do that transformation and the original implementor missed that detail? (and no, I don't know how to reach him). But if isea is meant to be used only on sphere, authalic latitude = geodetic latitude. A number of PROJ projections have started with spherical formulations only and have been improved for the non-spherical case later (using authalic latitude tricks, or more complex formulas, depending on the cases)

jerstlouis:
> But if isea is meant to be used only on sphere, authalic latitude = geodetic latitude. 

This is the subtlety that I am trying to understand / clarify...

Even though ISEA is only defined for a sphere, when integrating data georeferenced to an ellipsoid (which is the common case of WGS84 latitude, longitude), ideally one needs to apply this mapping of geodetic latitude on the source ellipsoid to the authalic altitude on that same ellipsoid before mapping it to the sphere with the authalic radius.

In other words, there is the ellipsoid of the input coordinates, and there is the spheroid used to apply the projection. I think the difference between the two is quite clear with the concept of source and target SRS in GDAL, but there is not really a concept of "input" SRS in PROJ.

> A number of PROJ projections have started with spherical formulations only and have been improved for the non-spherical case later (using authalic latitude tricks, or more complex formulas, depending on the cases)

Essentially I'm trying to understand whether doing this automatic geodetic ==> authalic conversion would be considered the ISEA projection using a non-spherical ellipsoid (_"improved for the non-spherical case"_) -- in which case we should do it automatically when detecting that the selected ellipsoid is not a sphere e.g. for `+proj=isea`, would simply be support for non-spherical input coordinates (in which case also whether it should be done by implementing a `+proj=auth` that could be used in `+proj=pipeline` instead), or whether there's really a conceptual difference at all between these two things?

Hopefully I've addressed most of the remaining issues with this last amended commit.

jerstlouis:
@rouault Seems like I've fixed the warnings.

Regarding `P->a *= 1. - P->es * (SIXTH + P->es * (RA4 + P->es * RA6))`, as you were suggesting, it seems like that polynomial approximation is not good enough for millimeter precision and needs the next power term.

This power series appears in this [book from 1880 on page 108](https://books.google.ca/books?id=PuhOAQAAMAAJ&pg=PA108), without additional term (also linked from geographiclib [here](https://geographiclib.sourceforge.io/geodesic-papers/helmert80-en.html) including full [english translation](https://zenodo.org/records/32050/files/helmert80-en.pdf?download=1)). I could not find any other reference outside the PROJ source code. Not sure if anyone figured out the next term(s) or if we would have to compute it from the [precise formulas I quoted above](https://github.com/OSGeo/PROJ/pull/4211#issuecomment-2284512882)?

If this is only for initialization rather than doing the actual projection computations, the exact values could be used to avoid sacrificing any precision, since it should not have a major performance impact? It's not a particularly expensive calculation anyways (one `log()` and one `sqrt()`).

rouault:
> If this is only for initialization rather than doing the actual projection computations, the exact values could be used to avoid sacrificing any precision, since it should not have a major performance impact?

I guess this would make sense. This should likely be done in a separate commit or pull request (might require adjustments to expected results)

src/projections/isea.cpp
```diff
@@ -1105,5 +1108,456 @@ PJ *PJ_PROJECTION(isea) {
...
+ * Snyder optimizations. In: Marina L. Gavrilova, and C. J. Kenneth Tan (Eds):
+ * Transactions on Computational Science XVI. Heidelberg, Springer, 2012. pp.
+ * 134–148. doi:10.1007/978-3-642-32663-9_8
+ */
+
+struct GeoPoint {
+    double lat, lon;
+}; // In radians
+
+#define Pi 3.1415926535897932384626433832795028841971
```
rouault: we can use the M_PI constant instead

src/projections/isea.cpp
```diff
@@ -1105,5 +1108,456 @@ PJ *PJ_PROJECTION(isea) {
...
+
+#define Degrees(x) ((x)*Pi / 180)
+
+#define wgs84InvFlattening 298.257223563
```
rouault: All of those WGS84 related parameters should likely be replaced by using P->a, P->b, P->es parameters

src/projections/isea.cpp
```diff
@@ -1105,5 +1108,456 @@ PJ *PJ_PROJECTION(isea) {
...
+ * 134–148. doi:10.1007/978-3-642-32663-9_8
+ */
+
+struct GeoPoint {
+    double lat, lon;
+}; // In radians
+
+#define Pi 3.1415926535897932384626433832795028841971
+
+#define Degrees(x) ((x)*Pi / 180)
```
rouault: We have the DEG_TO_RAD multiplicative constant to do that (in proj_internal.h)

src/projections/isea.cpp
```diff
@@ -1105,5 +1108,456 @@ PJ *PJ_PROJECTION(isea) {
...
+
+#define inf std::numeric_limits<double>::infinity()
+
+// distortion
+// static double maximumAngularDistortion = 17.27;
+// static double maximumScaleVariation = 1.163;
+// static double minimumScaleVariation = .860;
+
+// Vertices of dodecahedron centered in icosahedron faces
+#define E                                                                      \
```
rouault: the E_RAD constant could be reused

src/projections/isea.cpp
```diff
@@ -1105,5 +1108,456 @@ PJ *PJ_PROJECTION(isea) {
...
+    Degrees(52.6226318593487) //(Degrees)atan((3 + sqrt(5)) / 4); // Latitude of
+                              // center of top icosahedron faces
+#define F                                                                      \
```
rouault: the F_RAD constant could be reused

test/gie/builtins.gie
```diff
@@ -2776,6 +2776,35 @@ operation +proj=isea   +mode=hex +resolution=31
 accept  0 0
 expect  failure
 
+-------------------------------------------------------------------------------
+operation +proj=isea +R=6371007.18091875
+-------------------------------------------------------------------------------
+tolerance 2 mm
+direction inverse
```
rouault: it would be more convincing if we had the corresponding test for the forward test, to make it obvious that round tripping works.

The way to do it is to remove "direction inverse", switch the expect and accept parameters, and add a "roundtrip 1" line after each expect. So something like

accept -168.75 58.282525588539
expect -19186144.872455853968859 3323137.771243878174573
roundtrip 1

accept 11.25 58.282525588539
expect -15348915.898387057706714 9969413.314565381035209
roundtrip 1

...

jerstlouis: We can do that. The `expect` values come directly from copy-pasting results of the forward `+proj=isea` so I am convinced myself that it does work ;)

src/projections/isea.cpp
jerstlouis: The original `E_RAD` constant is not precise enough. Can I fix it and the original tests?

```diff
-/* 52.62263186 */
-#define E_RAD 0.91843818702186776133
+// Latitude of center of top icosahedron faces
+// atan((3 + sqrt(5)) / 4) = 52.6226318593487 degrees
+#define E_RAD 0.91843818701052843323

 accept  2 -1
-expect  -1097074.948264795 3233611.728585708
+expect  -1097074.948260187404 3233611.728216679767
 accept  -2 1
 expect  -1575486.353641554 3442168.342028188
 accept  -2 -1
-expect  -1575486.353880283 3234352.695594706
+expect  -1575486.353875697823 3234352.695233523380
```

src/projections/isea.cpp
jerstlouis: The original `F_RAD` constant is not precise enough. Can I fix it and the original tests?

```diff
-/* 10.81231696 */
-#define F_RAD 0.18871053072122403508
+// Latitude of center of faces mirroring top icosahedron face
+// atan((3 - sqrt(5)) / 4) = 10.8123169635739 degrees
+#define F_RAD 0.18871053078356206978
```

src/projections/isea.cpp
jerstlouis: That is turning out into quite an overhaul, but it will handle arbitrary `+R=` values... Working on it.

jerstlouis: I did that, but that turned out into trying to improve round-trip precision (including for the previous test cases), and as a result I had to merge some of the constants with the forward projection.

Now should I try to implement this automatic geodetic ==> authalic when a non-spherical ellipsoid is in use?
Do we keep that for a separate commit / PR?

Regarding authalic ==> geodetic conversion, there's no good reason that `pj_authset()` is doing a dynamic memory allocation for 3 doubles -- those `double *apa;` should just be `double apa[3]` in _cea.cpp_, _eqearth.cpp_, _healpix.cpp_ and _laea.cpp_ and passed as a return pointer to `pj_authset()`. Can I fix that? 

`cea_e_forward()` / `cea_e_inverse()` may be the simplest example of this automatic geodetic / authalic conversion.

src/projections/isea.cpp
jerstlouis: Changes applied.

src/projections/isea.cpp
jerstlouis: Changes applied.

src/projections/isea.cpp
jerstlouis: Changes applied.

test/gie/builtins.gie
jerstlouis: I've turned on roundtrip for all test cases, including the previous test cases using `+a=6400000`.
They now all pass (after sharing more of the constants between the forward and inverse, which I hope generaly improves the accuracy) but I had to update values and bump up tolerance to 1mm for everything.

src/projections/isea.cpp
jerstlouis: Changes applied.

src/projections/isea.cpp
rouault: > Now should I try to implement this automatic geodetic ==> authalic when a non-spherical ellipsoid is in use?
Do we keep that for a separate commit / PR?

Yes perhaps keep that for a follow-up PR

> Regarding authalic ==> geodetic conversion, there's no good reason that `pj_authset()` is doing a dynamic memory allocation for 3 doubles -- those `double *apa;` should just be `double apa[3]` in _cea.cpp_, _eqearth.cpp_, _healpix.cpp_ and _laea.cpp_ and passed as a return pointer to `pj_authset()`. Can I fix that?

That's more a micro-optimization that will likely not measurable. Should be kept for a later PR if you want to tackle that

src/projections/isea.cpp
```diff
@@ -1101,9 +1121,426 @@ PJ *PJ_PROJECTION(isea) {
...
+        if (Q->dgg.o_lat == ISEA_STD_LAT && Q->dgg.o_lon == ISEA_STD_LONG)
+            p = &standardISEA;
+        else if (Q->dgg.o_lat == M_PI / 2.0 && Q->dgg.o_lon == 0)
+            p = &polarISEA;
+        else
+            p = nullptr;
+    }
+
+    if (p != nullptr) {
+        if (P->e) {
```
rouault:
```suggestion
        if (P->e > 0) {
```

src/projections/isea.cpp
```diff
@@ -1101,9 +1121,426 @@ PJ *PJ_PROJECTION(isea) {
...
+    bool cartesianToGeo(const PJ_XY &inPosition,
+                        const isea_inverse_params &params, GeoPoint &result) {
+        bool r = false;
+        static const double epsilon =
+            2E-8; // NOTE: 1E-11 seems too small for forward projection
+                  // precision at boundaries
+        int face = 0;
+        PJ_XY position = inPosition;
+// sin(-60)
+#define sr -0.86602540378443864676372317075293618347140262690519
```
jjimenezshaw: This is more a general question, not a request for this PR.

Nowadays compilers are much better optimizing constants at compile time.
See https://godbolt.org/z/3KeGe6Psj (not using the last versions nor the aggressive optimizations in purpose)
What about not using a define, but a const variable?
```suggestion
const double sr = std::sin(-60 * DEG_TO_RAD);
```

Is there any drawback on that?

rouault: > Is there any drawback on that?

As far as I can see in https://en.cppreference.com/w/cpp/numeric/math/sin, it is guaranteed to be constexpr only in C++26, so even if most popular compilers have anticipated that, we don't have a guarantee of compile time computation of the constant with our current C++11 baseline.

src/projections/isea.cpp
jerstlouis: Thanks, applied.

jerstlouis:
@rouault There's a ~90 mm deviation failure in some of the builds that's proving quite annoying.

I've fought with similar ones yesterday, but now locally even with 32-bit GCC I can't seem to reproduce it.

It's that second new test again:

```
accept 11.25 58.282525588539
expect -15348915.896747918799520 9969413.314624363556504
roundtrip 1
```

It's right at a top icosahedron vertex so that might be related to the difficulties.
(it actually is projected to all 5 different top triangle vertices in the planar projection as I mention in https://github.com/OSGeo/PROJ/pull/4211#issuecomment-2262172061).

Do you get the failure locally? Thanks for all the help.

rouault:
> Do you get the failure locally?

I haven't tried and probably won't. It indeed happens that on i386 there are significant output differences near the validity border of some projections. The easiest in that situation is to relax the tolerance on the failing tests, with a comment to explain why.

jerstlouis:
Thanks @rouault .

> I haven't tried and probably won't. It indeed happens that on i386 there are significant output differences near the validity border of some projections.

That was a good hint. I was able to reproduce the problem locally with `-march=i386 -O2`, so I won't have to boot up an old machine from my 90's museum ;)

> The easiest in that situation is to relax the tolerance on the failing tests, with a comment to explain why.

I'm way too obstinate for that -- I'll try to fix it.

jerstlouis:
With a tiny bit of luck, this last amended commit will fix the i386 precision issue, still passing roundtrips at 1mm tolerance.

jerstlouis:
@rouault Maybe we need to change the header that says:

> This code was entirely written by Nathan Wagner and is in the public domain.

rouault:
> Maybe we need to change the header that says:
> 
> > This code was entirely written by Nathan Wagner and is in the public domain.

maybe something like

"The code for the forward direction was initially written by Nathan Wagner" ?

The "Available forms" line at top of docs/source/operations/projections/isea.rst also needs to be changed to "Forward and inverse, spherical", and perhaps with a longer explanation later in the page about the restrictions on which forms of isea are supported in the inverse direction.

jerstlouis:
> with a longer explanation later in the page about the restrictions on which forms of isea are supported in the inverse direction.

As far as I understand, what is not currently supported is the `+azi=<value>`, `+aperture=<value>`, `+resolution=<value>` and `+mode=<string>`.

However, there are no tests for any of these even for the forward projection, which at least in TDD practice would imply they do not work :)

I question whether `+aperture`, `+resolution` and `+mode` should be included at all in this projection. I have no idea what `di` or `dd` mean, and I have no idea what `hex` implies (and the documentation does not say anything about it), and I have been working heavily with ISEA for the last 3 years. At first glance, these parameters seem to be mixing up the concept of discrete global grids with the concept of the underlying projection.

Actually, there is one test expecting a failure:

```
operation +proj=isea   +mode=hex +resolution=31
accept  0 0
expect  failure
```

I imagine it's testing a resolution too high?

rouault:
> I question whether +aperture, +resolution and +mode should be included at all in this projection. I have no idea what di or dd mean, and I have no idea what hex implies

me neither. So perhaps just mentions that the inverse direction is only available when none of those options is specified?

> I imagine it's testing a resolution too high?

yes, this just tests we don't run into undefined behavior found by a fuzzer which generates "random" PROJ strings, and was fixed per 4c8a5cb8c7

jerstlouis:
> Snyder's Icosahedral Equal Area map projection on polyhedral globes for the dodecahedron and truncated icosahedron

This is actually confusing and probably wrong. The ISEA projection is strictly on the icosahedron.

As discussed above, the paper confusingly talks about several different projections on different solids, but ISEA is specifically about the icosahedron. The original code had some hints that the author wanted to implement different solids, but only the icosahedron is implemented.

> and for certain instructional tools and databases, the projections are useful for world maps.

This does not read well at all to me, and the description says nothing about the discrete global grids that this projection is widely used for.

I'm proposing to change the description to:

> Snyder's Icosahedral Equal Area map projections on an icosahedron polyhedral globe
offers relatively low scale and angular distortion. The equations involved are
relatively straight-forward. The interruptions remain a disadvantage, as with
any low-error projection system applied to the entire globe :cite:`Snyder1992`.
This projection is used as a basis for defining discrete global grid hierarchies.

and add this note regarding the use of ellipsoids:

> .. note:: As the projection is only defined on a sphere, it should only be used
with a spherical ellipsoid e.g., `+R=6371007.18091875` for a sphere with the
authalic radius of the WGS84 ellipsoid. For mapping coordinates on the WGS84
ellipsoid to the authalic sphere, the input latitude should be converted
from geodetic latitude to authalic latitude. A future version may automatically
perform this conversion when using a non-spherical ellipsoid.

jerstlouis:
What is this `trescale`? Does this correspond to an undocumented `+rescale=` parameter?

```c++
    if (pj_param(P->ctx, P->params, "trescale").i) {
        Q->radius = ISEA_SCALE;
    }
```

How do these `s`, `t`, `r` parameter prefixes work e.g., `sorient`, `tazi`, `razi`?

rouault:
> How do these `s`, `t`, `r` parameter prefixes work e.g., `sorient`, `tazi`, `razi`?

cf https://github.com/OSGeo/PROJ/blob/master/src/param.cpp#L112

jerstlouis:
The `isea_disn()` I removed in last commit was not reachable (like several other enum values in `isea_address_form`, nothing would set `output` to `ISEA_SEQNUM`).

That `serial` seems very much like an 64-bit integer zone index, similar to the ones we define for the [ISEA9R DGGRS](file:///home/jerome/specs/ogcapi-discrete-global-grid-systems/core/21-038.html#isea9r-dggrs).

Those `dd`, `di` and `hex` modes, and those related _aperture_, _resolution_ parameters, _isea_ptdd()_, _isea_dddi_ap3odd()_, _isea_dddi()_, _isea_ptdi()_, _isea_hex()_ really all seem related to particular discrete global grids / zone identifier reference systems that can be built on top of the ISEA planar projection.

The _isea_ptdd()_ is eerily familiar, looking very much like the 60 deg rotated / 30 deg horizontally sheared [5x6 ISEA space](https://github.com/OSGeo/PROJ/pull/4211#issuecomment-2262172061) I was talking about above:

_convert projected triangle coords to quad xy coords, return quad number_
```c++
    isea_rotate(pt, downtri ? 240.0 : 60.0);
    if (downtri) {
        pt->x += 0.5;
        pt->y += cos30;
    }
```

which can be achieved with an affine transform:
```
+proj=pipeline
   +step +proj=isea +R=6371007.18091875 +x_0=19186144.8709 +y_0=-3323137.7718
   +step +proj=affine +inv +s11=3837228.97419 +s12=3837228.97419 +s21=6646275.54357 +s22=-6646275.54357
```

Does all this really belong in PROJ? It seems like all that code belongs in discrete global grid reference system libraries like geogrid and DGGRID, not in a projection library.

With the exception of what `isea_tri_plane()` is doing, all of those extra modes perform additional operations on top of what the default planar mode does.

This `isea_tri_plane()` is doing a simple translation plus a 180 degrees rotation, which should really be optimized because it makes no sense to make 4 trigonometric calls to rotate a point by 180 degrees.

None of this is tested, and none of it is documented properly. Would there be any way to know if anyone is using these things, and if so, could we suggest to move this code to a separate module that can build on top of the default `plane` mode?

(Does PROJ have a deprecation mechanism?)

rouault:
> Would there be any way to know if anyone is using these things

Ask first on https://lists.osgeo.org/mailman/listinfo/proj , which I assume will be met with silence :-), and then remove them and wait for a few months to see if people complain when the new PROJ version hits them. In any case, that should be done in a separate pull request

> (Does PROJ have a deprecation mechanism?)

We don't really have a runtime warning level. If we wanted to do a stage deprecation, perhaps the isea.rst page should be updated to mention that parameters X, Y, Z will be removed in PROJ 9.6

Besides that, is there anything left for this PR? It looks good to me as it

jerstlouis:
> In any case, that should be done in a separate pull request

Are you OK with the undocumented `rescale` parameter removed in the last commit?
I believe that's the only reachable functionality that is removed.

I did separate commits in case you wanted to pick and choose what to merge. Feel free to squash them if you prefer.

> If we wanted to do a stage deprecation, perhaps the isea.rst page should be updated to mention that parameters X, Y, Z will be removed in PROJ 9.6

That would be a good idea. Would you want to do this separately and/or after asking on the mailing list?

> Besides that, is there anything left for this PR? It looks good to me as it

Just let me fix this `isea_rotate(pt, 180.0);` right now -- other than that it should be all good.

rouault:
> I did separate commits in case you wanted to pick and choose what to merge

separate (clean) commits is fine

> That would be a good idea. Would you want to do this separately and/or after asking on the mailing list?

I would (let you) asking on the mailing list first, and in the absence of negative feedback, proceed to update the doc

jerstlouis:
> I would (let you) asking on the mailing list first, and in the absence of negative feedback, proceed to update the doc

Thanks, will do.

I just tried adding tests for `+orient=pole` and one test case is failing (it might be an issue with the forward projection). Looking into it.

rouault:
> I just tried adding tests for `+orient=pole` and one test case is failing (it might be an issue with the forward projection). Looking into it.

I would appreciate if we could merge quickly this PR. Currently I need to click on a button in github UI each time you update the PR, so that github actions are run, because, as you don't have any commit merged yet into this repository, github doesn't trust you yet...

jerstlouis:
> I would appreciate if we could merge quickly this PR.

Sorry to be spamming PR updates.
If you prefer to merge this one now, I could open a separate PR for that `+orient=pole` fix.

rouault:
I've just merged this PR.

PR #4218, opened 2024-08-11
projinfo: 2 args equivalent to usage of -s and -t

jjimenezshaw:
From
https://lists.osgeo.org/pipermail/proj/2024-May/011388.html

Some times when I am using cs2cs I want to do also a projinfo, to see the
transformation(s) that is may apply.

It is always bothering me that I have to add "-s" and "-t".

Does it make sense that if there are two args without option identifier, it
assumes that the second is "-t"? Am I missing some corner-cases? (if you
like the idea, I am happy to do the PR)
I do not mean to remove "-t" completely, but to make the arg parser more
flexible.

- [x] Tests added

### DIFF
--- a/src/apps/projinfo.cpp
+++ b/src/apps/projinfo.cpp
@@ -122,8 +122,9 @@ struct OutputOptions {
         << "                {object_definition} | {object_reference} |"
         << std::endl
         << "                (-s {srs_def} [--s_epoch {epoch}] "
-           "-t {srs_def} [--t_epoch {epoch}])"
-        << std::endl;
+           "-t {srs_def} [--t_epoch {epoch}]) |"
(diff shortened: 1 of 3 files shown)


src/apps/projinfo.cpp
```diff
@@ -1042,8 +1043,7 @@ int main(int argc, char **argv) {
         usage();
     }
 
-    std::string user_string;
-    bool user_string_specified = false;
+    std::vector<std::string> user_strings;
```
rouault: perhaps ``positional_args`` would be clearer ?

jjimenezshaw: done

rouault (approved):

PR #4220, opened 2024-08-12
CMake: add TESTING_USE_NETWORK configure option

### LINKED ISSUE #4219: Change network dependence on configure / build time
This item is to address some points raised by @gdt starting in [this mail list thread](https://lists.osgeo.org/pipermail/proj/2024-May/011405.html), around "it's a bug to try to use the network at build time".

I have some work-in-progress to address this item (thus I'm assigned), which I hope to put out very soon, as I'm aware there could be a RC in the coming week.

My proposed solution is to have a CMake option `USE_NETWORK` with default `OFF`. It will show warnings if test dependencies cannot be met, and run a subset of tests that can still run.

mwtoews:
<s>This changes how CMake configures a build, where by default `USE_NETWORK=OFF` assumes use of the network is not permitted, which disables fetching GTest (if needed) and running network-dependant tests. The user (or packager) would need to opt-in with `USE_NETWORK=ON` to enable these features, which were the default for previous CMake-based PROJ builds.</s>

_Update:_ This PR now adds CMake `TESTING_USE_NETWORK` configure option to permit or disable network use, with a default ON to assume the network can be used to fetch GTest dependency (if needed) and run network-dependant tests.

This also finally cherry-picks #4153 to document test dependencies, which incorporate new requirements from #4131.

Closes #4219

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -139,15 +139,27 @@ The following guides show how to compile and install the software using CMake.
     PROJ 9.0 and later releases only support builds using CMake.
 
 
-Build requirements
+Requirements
 --------------------------------------------------------------------------------
+
+Build requirements
+++++++++++++++++++
 
 - C99 compiler
 - C++11 compiler
 - CMake >= 3.16
-- SQLite3 >= 3.11: headers and library for target architecture, and sqlite3 executable for build architecture.
+- SQLite3 >= 3.11: headers and library for target architecture, and sqlite3 executable for build architecture
 - libtiff >= 4.0 (optional but recommended)
 - curl >= 7.29.0 (optional but recommended)
+- JSON for Modern C++ (nlohmann/json) >= 3.7.0; if not found as an external dependency then vendored version 3.9.1 from PROJ source tree is used
+
+Test requirements
++++++++++++++++++
+
+This is only required if testing is built (see :option:`BUILD_TESTING`, default ON)
+
+- GoogleTest (GTest) >= 1.8.1; if not found then version 1.12.1 is fetched from GitHub and installed internally
+
 
 Build steps
 --------------------------------------------------------------------------------
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -153,12 +153,17 @@ Build requirements
 - curl >= 7.29.0 (optional but recommended)
 - JSON for Modern C++ (nlohmann/json) >= 3.7.0; if not found as an external dependency then vendored version 3.9.1 from PROJ source tree is used
 
+.. _test_requirements:
+
 Test requirements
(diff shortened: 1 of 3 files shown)


rouault:
> where by default `USE_NETWORK=OFF`

I would rather use ON by default, as packagers' constraints are not necessarily the ones of "regular" PROJ users/developers. If we keep the default to OFF, we should change our CI configurations (at least some of them) to set it to ON to avoid reducing our test coverage.

And the option should perhaps be renamed to something like TESTING_USE_NETWORK  to limit the risk of confusion with PROJ_NETWORK

mwtoews:
As for default ON/OFF, I'm open for opinions. I'd agree there are likely more that would prefer it to be ON, as PROJ is a web-enabled tool built on CURL.

I'll tune the CI options next (tomorrow), just testing to see what happens with changing the defaults. The result of a default OFF is a few CMake configure warnings, and fewer tests, but no failures. For example:
```
...
-- Looking for GTest
CMake Warning at test/unit/CMakeLists.txt:92 (message):
-- Looking for GTest - not found
  Tests that require GTest will be skipped; either install GTest dependency,
  or if possible, set USE_NETWORK=ON to fetch content from GitHub
...
```
A suggested rename to `TESTING_USE_NETWORK` seems ok, as it applies only to testing-related aspects. But if I'd go with that if we decide to keep default ON, otherwise it's a long option that a user would need to type to enable ON if that were not the default.

mwtoews:
I see the 👍 votes above to keep the default behavior that CMake can use the network during configure and test. Let me know if there are any objections, otherwise I'll re-write this PR using a default `TESTING_USE_NETWORK=ON`

(i.e. only users/packagers that do _not_ permit network access would need to specify `TESTING_USE_NETWORK=OFF` during CMake configure)

mwtoews:
Some other notes on this PR:

- Generally tests (via `ctest`) should always be run, but the number of tests may vary between 44 to 68 depending if the network is available, if GTest is available, if Python with YAML is available. There is one bonus test add if pytest is available, which self-tests the CLI tester.
- Warnings are shown with cmake configure in a few scenarios, e.g. is no network is detected but GTest is needed 
- Option `RUN_NETWORK_DEPENDENT_TESTS` has existed for a while, but this is moved outside an if-block.
- If 'curl' or 'ping' are not available then assume there is no network (Debian-like users can get these with `apt install -y iputils-ping curl`)
- If no network is detected (missing curl/ping), but it does work, then `RUN_NETWORK_DEPENDENT_TESTS=ON` can be specified to run `nkg` (the only network-dependant test)
- Many different combinations of configurations were tested locally, but I don't feel these test scenarios need to be replicated in CI.

PR #4225, opened 2024-08-14
ISEA: Fix bug introduced in 825682a2ea70fb923652fb921e3f9122adc0e22f

jerstlouis:
- Also added tests for `+orient=pole`
- step 1: double z = fabs(cosZ) < 1E-15 ? 0 : acos(cosZ); was wrong
- acos(0) is Pi/2, not 0

- [X] Tests added

Actually the bug was introduced in the first commit, I got the hash ID wrong:

https://github.com/OSGeo/PROJ/commit/825682a2ea70fb923652fb921e3f9122adc0e22f#diff-8c9143316955de6c44a9245f247245ba2c0b6cb91266afc522e484050a2a5202R406


jerstlouis:
These `safeArcSin/Cos()` functions make the difference between > 90 mm and < 0.2 mm roundtrip precision with the `-m32` / `-march=i386` flags that were causing the test failures for test cases near the interruptions yesterday. 

The bug introduced with the earlier attempt to address that was causing the build failure for the new `+orient=pole` tests (though it is not specific to that orientation -- it would also have broken specific coordinates for `+orient=isea` as well).

PR #4227, opened 2024-08-15
Fix very minor issues or workaround latest cppcheck false positives

rouault:


src/transformations/gridshift.cpp
```diff
@@ -372,6 +372,7 @@ PJ_XYZ gridshiftData::grid_interpolate(PJ_CONTEXT *ctx, const std::string &type,
         gridInfo.idxSampleXYZ[2] = idxSampleZ;
         iterCache = m_cacheGridInfo.emplace(grid, std::move(gridInfo)).first;
     }
+    // cppcheck-suppress derefInvalidIteratorRedundantCheck
```
jjimenezshaw, line 375: How long is the effect of this suppression? the rest of the file?

jjimenezshaw (approved):

src/transformations/gridshift.cpp
rouault: the line immediately below

PR #4229, opened 2024-08-16
Fix parameter mappings of Local Orthographic, and re-import ESRI database with the mapping for this method

rouault:
CC @jjimenezshaw


src/iso19111/operation/parammappings.cpp
```diff
@@ -403,6 +403,11 @@ static const ParamMapping paramScaleFactorPseudoStdParallel = {
     WKT1_SCALE_FACTOR, common::UnitOfMeasure::Type::SCALE,
     k}; /* ignored by PROJ currently */
 
+static const ParamMapping paramLongCentreLongCenter = {
```
jjimenezshaw, line 406: is `paramLongCentreLongCenter` name "duplicated" in purpose? Using "Centre" and "Center"? Or the editor tricked you?

jjimenezshaw (commented):
Thanks @rouault

src/iso19111/operation/parammappings.cpp
rouault: paraLongCentreLongCenter is indeed a very badly named variable. I believe it intends to capture the (WKT2, WKT1) names, but they are actually here ("Longitude of origin", "longitude_of_center"). I remember to have struggled establishing those mappings because the same WKT2 parameter name doesn't always map to the same WKT1 parameter name, depending on which projection method it is used in. Partly because WKT1 parameter names were invented before, and/or EPSG had renamed them after they were hardcoded in GDAL 20 years ago.

PR #4231, opened 2024-08-20
Fix race condition surrounding DB reopening during fork

lucasz93:

Hi all!

I was consistently running into this race condition in my project. One thread locked a copy of SQLiteHandleCache::sMutex_ while another thread was forking. Then the atfork callback tried to lock sMutex_. This caused a deadlock because the thread that owned the mutex isn't duplicated in a fork.

Kind regards,
-Lucas

### DIFF
--- a/src/iso19111/factory.cpp
+++ b/src/iso19111/factory.cpp
@@ -627,8 +627,23 @@ SQLiteHandleCache::getHandle(const std::string &path, PJ_CONTEXT *ctx) {
 #ifdef REOPEN_SQLITE_DB_AFTER_FORK
     if (firstTime_) {
         firstTime_ = false;
-        pthread_atfork(nullptr, nullptr,
-                       []() { SQLiteHandleCache::get().invalidateHandles(); });
+        pthread_atfork(
+            []() {
+                // This mutex needs to be acquired by 'invalidateHandles()'.
+                // The forking thread needs to own this mutex during the fork.
+                // Otherwise there's an opporunity for another thread to own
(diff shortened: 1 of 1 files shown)


src/iso19111/factory.cpp
```diff
@@ -627,8 +627,23 @@ SQLiteHandleCache::getHandle(const std::string &path, PJ_CONTEXT *ctx) {
...
-        pthread_atfork(nullptr, nullptr,
-                       []() { SQLiteHandleCache::get().invalidateHandles(); });
+        pthread_atfork(
+            []() {
+                // This mutex needs to be acquired by 'invalidateHandles()'.
+                // The forking thread needs to own this mutex during the fork.
+                // Otherwise there's an opporunity for another thread to own
+                // the mutex during the fork, leaving the child process unable
+                // to acquire the mutex in invalidateHandles().
+                SQLiteHandleCache::get().sMutex_.lock();
```
jjimenezshaw, line 637: Isn't this lock conflicting with the lock at the beginning of the function?

lucasz93: Not to my knowledge. These callbacks run during a call to fork() whereas the lock at the beginning of the function is protecting the installation of the callbacks.

rouault:
@lucasz93 Are you sure that playing with std::mutex around fork() is safe ? Particularly unlocking in a child process a std::mutex locked by its parent. I'm afraid we are in unspecified territory when mixing C++ mutex and fork()

rouault:
ok, let's merge this. We'll see if that cause issues

PR #4242, opened 2024-09-02
Output with roman numerals in cs2cs for lat-lon

jjimenezshaw:
After this post
https://mapstodon.space/@jjimenezshaw/113017850151676527
we could have this "Easter egg". It is totally fine if it is not added.

It is only intended to work in output (not parsing input) for lat-lon.

I have not documented it in `docs/source/cs2cs.rst`. If you consider it should be in the documentation, let me know.

- [x] Tests added

### DIFF
--- a/src/apps/cs2cs.cpp
+++ b/src/apps/cs2cs.cpp
@@ -67,6 +67,8 @@ static bool destIsLongLat = false;
 static double destToRadians = 0.0;
 static bool destIsLatLong = false;
 
+static bool romanNumerals = false;
+
 static int reversein = 0, /* != 0 reverse input arguments */
     reverseout = 0,       /* != 0 reverse output arguments */
     echoin = 0,           /* echo input data to output line */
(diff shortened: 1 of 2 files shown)


kbevers:
While this is fun I am -1 on this. It's just extra clutter in the code without any real benefit.

hobu:
I'm +1 to this. How else is someone going to be able to translate ancient roman coordinates? 😛 

I would be -1 on roman coordinate locale support however 😉

src/apps/cs2cs.cpp
```diff
@@ -94,6 +96,54 @@ using namespace NS_PROJ::metadata;
...
+                return five;
+            case (6):
+                return five + one;
+            case (7):
+                return five + one + one;
+            case (8):
+                return five + one + one + one;
+            case (9):
+                return one + ten;
+            default:
```
rouault, line 136: There's a flaw in the implementation due to the lack of zero. We should either fallback to one in that case, or throw an exception "Sorry, Romans could not express this coordinate"

jjimenezshaw, line 136: Uhmmm, I do not now if it is a flaw, or a feature. It was implemented in a way that `0` uses `no-symbol`. It would be strange something like `d9'12"M` , o even `d'"S d'"OC` for null island. But at the same time I like it even more, because it shows how roman numerals work (or not work).

rouault: Forget my comment. It seems I forgot how to use roman numbers... Sorry for the noise

skandalfo:
What have the Romans ever done for us?

jjimenezshaw:
> What have the Romans ever done for us?

Apart from the sanitation, the medicine, education, wine, public order, irrigation, roads, the freshwater system, and public health… what have the Romans ever done for us?

cffk:
I'm -1 on this too (sorry!).  It's awkwardly jimmying an obsolete counting system (no zero, no negative numbers, no fractions) into a modern application.  I fear it would also invite other silly PRs.

While I'm on the subject, does PROJ support centesimal degrees?  This was (still is?) used for surveying in some parts of Europe.

m-kuhn:
Hoc nobis opus est ut Carthaginem invadamus; quid possumus facere ut hoc consilium progrediatur?

jjimenezshaw:
In case you need it in your clock:
https://f-droid.org/en/packages/net.diffengine.romandigitalclock/

PR #4251, opened 2024-09-15
Doc: fix typos and add a CI check to spot them

rouault:


kbevers (approved):
This is good. I see that a lot of the spelling issues are found in news.rst. I guess that's mainly my fault. My workflow is usually to copy the titles of the relevant PRs and leave it at that (unless its very difficult to understand). Unfortunately, for the last maybe two years I've had to disable the spellchecks of the Sphinx setup, as it doesn't work on the newer generations of Macs (M1 processors and newer).

The ideal solution to this would be to automate more of the release process, similar to what we have for PROJ-data.

PR #4255, opened 2024-09-18
projinfo: display whether an operation is time-dependent

rouault:


rouault:
any feedback on this: good idea or not?  The downside I can see is that our "summary" of operations becomes a monstrous long line (especially due to the lengthy descriptions of the areas of use)

jjimenezshaw:
The idea sounds good to me. For some reason time dependent operations are getting more common. With the new US systems it is going to be even more, right?

rouault:
> With the new US systems it is going to be even more, right?

Maybe... ? (looks like we'll have reach retirement age at the time they are ready)  More seriously, I've that recollection that the dynamic stuff will be for advanced users and that for mortals who can only handle static CRS they will create a "snapshot" of the datum every 5 years of so.

kbevers:
I think this is a good solution

PR #4259, opened 2024-09-25
isTimeDependent(): do not mark 'static' Coordinate Frame helmert transformations as time-dependent, ...

rouault:
only time-dependent ones...


jjimenezshaw:
Thanks @rouault 
Has this bug any impact on the transformations used if there is no time provided?
(in other words, should I patch my PROJ 9.5.0 with this bugfix?)

rouault:
> Has this bug any impact on the transformations used if there is no time provided?

no, I don't think so. It is just that if you use the new isTimeDependent() API, it will lie to you without this fix. I noticed that in conjunction with https://github.com/OSGeo/PROJ/pull/4255 when the patched projinfo told me that a transformation was time dependent, whereas it was obviously not.

PR #4301, opened 2024-10-31
Add grid alternative for de_tlbg_thueringen_NTv2gridTH.tif

jjimenezshaw:
related: https://github.com/OSGeo/PROJ-data/pull/128

### DIFF
--- a/data/CMakeLists.txt
+++ b/data/CMakeLists.txt
@@ -45,7 +45,7 @@ set(ALL_SQL_IN "${CMAKE_CURRENT_BINARY_DIR}/all.sql.in")
 set(PROJ_DB "${CMAKE_CURRENT_BINARY_DIR}/proj.db")
 include(sql_filelist.cmake)
 
-set(PROJ_DB_SQL_EXPECTED_MD5 "4e70fd45d0ba64a6a80481f7e6e2ee4a")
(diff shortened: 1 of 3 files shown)


rouault (approved):

rouault:
@kbevers Will you be in a mood of release a PROJ-data 1.20 together with PROJ 9.5.1 with that referenced grid, and so this should be backported, or is it 9.6.0 material ?  (and if we do backport, do we need to update the PROJ.DATA.VERSION item in data/sql/metadata.sql while we are it, or will you do per the release procedure, as documented ?)

kbevers:
> Will you be in a mood of release a PROJ-data 1.20 together with PROJ 9.5.1 with that referenced grid, and so this should be backported, or is it 9.6.0 material ?

No worries, there's often a PROJ-data release bundled with patch releases of PROJ. 

> do we need to update the PROJ.DATA.VERSION item in data/sql/metadata.sql while we are it, or will you do per the release procedure, as documented ?

I don't mind bumping PROJ.DATA.VERSION now, it's easy to gloss over it during the release procedure. I usually check it and am most times delighted to see that's already been done by someone thinking ahead :)

rouault:
> I don't mind bumping PROJ.DATA.VERSION now

done

PR #4302, opened 2024-10-31
Fail consistently on network error and grid transformations

### LINKED ISSUE #705: PROJ_NETWORK: First transform fails, succeeds afterwards
https://gis.stackexchange.com/questions/373550/first-call-to-transform-fails-with-inf-all-subsequent-calls-are-ok-what-cou

Worth looking into later.

rouault:
Currently when we need to use a remote grid that can't be opened, we return HUGE_VAL coordinates values and a proj_errno = PROJ_ERR_OTHER_NETWORK_ERROR, (I guess) as expected... But if we do following proj_trans() calls on the same transformation object, the grid transformation is ignored and we fallback to other methods (Helmert, ballpark, etc.). Fix that by consistently returning the same error values as the initial failed call.

Fixes https://github.com/pyproj4/pyproj/issues/705

@snowman2 @jjimenezshaw @kbevers Thoughts?  I guess this wins on consistency. The potential downside is that if you transform a batch of one million points using the same grid and the network timeout is 1second, you'll spend 1 million seconds, as the current implementation (as modified by this PR) retries to open the grid each time a point is transformed  if it didn't manage to open it previously

### DIFF
--- a/src/iso19111/c_api.cpp
+++ b/src/iso19111/c_api.cpp
@@ -210,11 +210,13 @@ PJ *pj_obj_create(PJ_CONTEXT *ctx, const BaseObjectNNPtr &objIn) {
                     PROJStringFormatter::Convention::PROJ_5,
                     std::move(dbContext));
                 auto projString = coordop->exportToPROJString(formatter.get());
-                if (proj_context_is_network_enabled(ctx)) {
+                const bool defer_grid_opening_backup = ctx->defer_grid_opening;
+                if (!defer_grid_opening_backup &&
+                    proj_context_is_network_enabled(ctx)) {
                     ctx->defer_grid_opening = true;
                 }
                 auto pj = pj_create_internal(ctx, projString.c_str());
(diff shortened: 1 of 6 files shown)


rouault:
Or maybe we should remember, for a given PJ* object, that we previously failed to open the grid, and return the same error as the initial failed attempt, without actually retrying ?

kbevers:
> Or maybe we should remember, for a given PJ* object, that we previously failed to open the grid, and return the same error as the initial failed attempt, without actually retrying ?

I think this is the better option. If one or more coordinates in a dataset fails [0] to transform the dataset is corrupted, so we may as well do our best to make sure that it is clear to the user. You're more likely to notice that something is wrong when all or most of your coordinates are `HUGE_VAL`'s compared to only a few.

[0] Due to network issues, assuming it would transform ok otherwise

rouault:
> I think this is the better option.

implemented

snowman2 (approved):
Thanks @rouault :+1:

PR #4303, opened 2024-11-01
 Add Airocean projection (formerly Dymaxion)

### LINKED ISSUE #232: Add Dymaxion/Fuller Projection
**Reported by wildintellect on 12 Apr 2014 18:41 UTC**
Someone's already done the hard part, we just need to convert it to C
I believe this was written to work with Proj4js
http://mbostock.github.io/protovis/ex/dymax.js

Migrated-From: https://trac.osgeo.org/proj/ticket/232

plouvart:
- [X] Closes #232
- [X] Tests added


docs/source/operations/projections/airocean.rst
```diff
@@ -0,0 +1,57 @@
+.. _airocean:
+
+********************************************************************************
+Airocean
+********************************************************************************
+
+The Airocean map projection is based on the Dymaxion Fuller projection, which is a icosahedron polyhedral projection that intends
```
rouault: we should probably remove the Dymaxion word here...

src/projections/airocean.cpp
```diff
@@ -0,0 +1,224 @@
...
+        // Point lies outside icosahedron net faces
+        lp.lam = HUGE_VAL;
+        lp.phi = HUGE_VAL;
+        return lp;
+    }
+
+
+    PJ_XYZ sphereCoords = dym_to_ico(&xy, face_id);
+
+    double norm = sqrtl((sphereCoords.x * sphereCoords.x) + (sphereCoords.y * sphereCoords.y) + (sphereCoords.z * sphereCoords.z));
```
rouault:
```suggestion
    double norm = sqrt((sphereCoords.x * sphereCoords.x) + (sphereCoords.y * sphereCoords.y) + (sphereCoords.z * sphereCoords.z));
```

src/projections/airocean.cpp
```diff
@@ -0,0 +1,224 @@
...
...4014718}}, {{0.51545596, 0.3817169, -0.76720099}, {0.41468223, -0.65596241, -0.63067581}, {0.99500944, -0.0913478, 0.04014718}}, {{-0.42015243, -0.07814525, -0.90408255}, {-0.3557814, 0.84358, -0.40223423}, {-0.99500944, 0.0913478, -0.04014718}}, {{-0.42015243, -0.07814525, -0.90408255}, {-0.99500944, 0.0913478, -0.04014718}, {-0.51883673, -0.83542038, -0.18133184}}, {{-0.42015243, -0.07814525, -0.90408255}, {-0.51883673, -0.83542038, -0.18133184}, {0.41468223, -0.65596241, -0.63067581}}, {{-0.42015243, -0.07814525, -0.90408255}, {0.41468223, -0.65596241, -0.63067581}, {0.51545596, 0.3817169, -0.76720099}}, {{-0.3557814, 0.84358, -0.40223423}, {-0.38796691, 0.38271738, -0.65315839}, {0.51545596, 0.3817169, -0.76720099}}, {{-0.42015243, -0.07814525, -0.90408255}, {0.51545596, 0.3817169, -0.76720099}, {-0.38796691, 0.38271738, -0.65315839}}, {{-0.99500944, 0.0913478, -0.04014718}, {-0.3557814, 0.84358, -0.40223423}, {-0.58849102, 0.53029673, 0.0627648}}, {{-0.3557814, 0.84358, -0.40223423}, {-0.41468223, 0.65596241, 0.63067581}, {-0.58849102, 0.53029673, 0.0627648}}, {{-0.99500944, 0.0913478, -0.04014718}, {-0.58849102, 0.53029673, 0.0627648}, {-0.41468223, 0.65596241, 0.63067581}}};
```
rouault: Code comment explaining where those constants come from highly welcome

docs/source/operations/projections/airocean.rst
plouvart: Seems fair.

rouault: hum I just saw an issue with the appveyor target that complained about the sqrtl() usage, that should be replace with sqrt() (without l)
```
..\src\projections\airocean.cpp(187): error C2220: warning treated as error - no 'object' file generated
..\src\projections\airocean.cpp(187): warning C4244: 'initializing': conversion from 'long double' to 'double', 
```

rouault: hum there's a unrelated issue with -Wshadow that I've just fixed in https://github.com/OSGeo/PROJ/pull/4304. Please rebase your pull request on top of master, once PR #4304 has been merged into master

rouault: PR #4304 has been merged into master, so you can rebase on top of latest HEAD

plouvart: I'm on it

plouvart: There. It should be synced up with master now.

I altered the doc as you wanted.

src/projections/airocean.cpp
plouvart: Changed `sqrtl` to `sqrt` as suggested.

src/projections/airocean.cpp
plouvart: Added docstring to document the constants.
Removed unused constants.
Rebuild the lib and reran the test to check that it is still working. Tests pass.

src/projections/airocean.cpp
```diff
@@ -0,0 +1,249 @@
...
+        air_ico_trans[face_id][2][1] * p->y +
+        air_ico_trans[face_id][2][2],
+    };
+}
+
+inline PJ_XYZ cartesian_to_ico(const PJ_XYZ *p, unsigned char face_id) {
+    const PJ_XYZ * center = &ico_centers[face_id];
+    const PJ_XYZ * normal = &ico_normals[face_id];
+
+    double a = 1.0 - (
```
rouault, line 204:
```suggestion
    // cppcheck-suppress unreadVariable
    double a = 1.0 - (
```

plouvart:
Seems like cppcheck is complaining about some unused variable, despite the fact that it is used next line.
Edit: just saw your fix above :)

rouault:
A few typos in the doc reported in  https://github.com/OSGeo/PROJ/actions/runs/11632419978/job/32395481110?pr=4303 . For words which are legitimate and must be allowed, you can add them in docs/source/spelling_wordlist.txt

src/projections/airocean.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+
+static PJ_LP airocean_inverse(PJ_XY xy, PJ *P) {
+    PJ_LP lp = {0.0, 0.0};
+
+    unsigned char face_id = get_dym_face_index(&xy);
+
+    if (face_id == 23) {
+        // Point lies outside icosahedron net faces
+        lp.lam = HUGE_VAL;
```
rouault, line 282:
```suggestion
        proj_errno_set(P, PROJ_ERR_COORD_TRANSFM_OUTSIDE_PROJECTION_DOMAIN);
        lp.lam = HUGE_VAL;
```

plouvart:
Thanks for the fix for the out of projection domain error

docs/source/operations/projections/airocean.rst
```diff
@@ -0,0 +1,57 @@
+.. _airocean:
+
+********************************************************************************
+Airocean
+********************************************************************************
+
+The Airocean map projection is a icosahedron polyhedral projection that intends
```
rouault, line 9:
```suggestion
.. versionadded:: 9.6.0

The Airocean map projection is a icosahedron polyhedral projection that intends
```

rouault (approved):

test/gie/builtins.gie
```diff
@@ -1162,6 +1162,112 @@ accept  -2 -1
...
+accept 4800000 32900000
+expect 98.90848938837188 -48.50784207235999
+accept 7700000 1100000
+expect 114.00231130949376 -71.99425265168686
+accept 3200000 27900000
+expect 143.20066829467564 -8.847739058325072
+accept 5200000 26800000
+expect 123.33794956763194 6.861286881197438
+accept 2600000 22300000
+expect 147.63769509437498 15.818879769693432
```
rouault: @plouvart could you add a test point that triggers the out-of-projection-domain error ?

plouvart: Yes. I'll do that in a few minutes

docs/source/operations/projections/airocean.rst
```diff
@@ -0,0 +1,59 @@
...
+   :width: 500 px
+   :align: center
+   :alt:   Airocean
+
+   proj-string: ``+proj=airocean``
+
+.. note::
+   The airocean map of the whole globe has an extent of approximately
+   :math:`[0, 0, 17404000, 36844000]` in meters. Although the airocean map has no
+   true up or down, the choice was made to lay it out vertically, from Antarctica to Australia.
```
mwtoews, line 45: It would be neat to have a `+orient=<string>` parameter to control this aspect, where "vertical" would be the current layout and "horizontal" would be the classic layout with Australia on the left and Antarctica on the right.

src/projections/airocean.cpp
```diff
@@ -0,0 +1,251 @@
...
+#endif
+
+#include <cmath>
+#include <cstdint>
+#include <errno.h>
+
+#include "proj.h"
+#include "proj_internal.h"
+
+PROJ_HEAD(airocean, "Airocean") "\n\tMisc, Sph&Ell";
```
mwtoews, line 15: The docs in this PR specify that only spherical form is supported, so probably remove `&Ell` unless it is mistaken.

docs/source/operations/projections/airocean.rst
```diff
@@ -0,0 +1,59 @@
...
+In comparison to the Snyder's Icosahedral Equal Area, this projection relies on a specific way 
+to orient and cut the icosahedron so that most interruptions lie within the ocean. The projection
+was first invented by Buckminster Fuller in 1943 using a cuboctahedron as basis.
+The idea to use an icosahedron came later in 1954, and it wasn't until 1978 that the projection 
+was formally mathematically defined :cite:`Gray1995`.
+
++---------------------+----------------------------------------------------------+
+| **Classification**  | Polyhedral, equal area                                   |
++---------------------+----------------------------------------------------------+
+| **Available forms** | Forward and inverse, spherical                           |
```
mwtoews: The `PROJ_HEAD` has "Sph&Ell" and it appears that ellipsoidal forms are supported, so should the docs reflect this?
```suggestion
| **Available forms** | Forward and inverse, spherical and ellipsoidal           |
```

docs/source/references.bib
```diff
@@ -73,6 +73,18 @@ @TechReport{Deakin2004
   Url                      = {http://www.mygeodesy.id.au/documents/Molodensky%20V2.pdf}
 }
 
+@Article{Gray1995,
+  Title                    = {Exact Transformation Equations for Fuller's World Map},
```
mwtoews: Preserve title case for BibTeX via:
```suggestion
  Title                    = {Exact Transformation Equations for {Fuller's World Map}},
```

src/projections/airocean.cpp
```diff
@@ -0,0 +1,251 @@
...
+    };
+} // anonymous namespace
+
+
+/*
+    The vertices of the faces of the icosahedron are inspired by those used by Robert W. Gray.
+
+    Original Reference:
+    Robert W. Gray (1995) Exact Transformation Equations for
+        Fuller's World Map, Vol. 32. Autumn, 1995, pp. 17-25.
```
mwtoews, line 32:
```suggestion
        Fuller's World Map. Cartographica, 32(3), 17-25.
        https://doi.org/10.3138/1677-3273-Q862-1885
```

mwtoews (commented):
A few suggestions and comments below.

plouvart:
@mwtoews I saw your comments. Those fixes seem fair. I'll get on those as soon as a I have some free time

docs/source/operations/projections/airocean.rst
plouvart: I modified the code to be able to alter matrices from icosahedral to airocean space and back using custom transformations.
Then I added the +orient argument with choice for either vertical (default) or horizontal layout.

rouault:
a few extra tests to check the behavior of +orient would be nice

plouvart:
> a few extra tests to check the behavior of +orient would be nice

Tests seem to fail in builtins, but I see no error about airocean specifically. What could be the reason?

test/gie/builtins.gie
```diff
@@ -1162,6 +1162,215 @@ accept  -2 -1
...
+accept 143 -9
+expect 8926966.62914534 3215425.0474725077
+accept 123 7
+expect 10051797.28039541 5233302.981464034
+accept 148 16
+expect 14580207.201673953 2606103.035139274
+
+direction inverse
+accept 13400000 13500000
+except 23.38101068889326 28.613417194415643
```
rouault: typo: except --> expect (here and below)

src/projections/airocean.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+
+// By default the resulting orientation of the projection is vertical
+// the following transforms are used to alter the projection data
+// so that the resulting orientation is horizontal instead
+constexpr double orient_horizontal_trans[4][4] = {{0.0, -1.0, 0.0, 36843762.068421006}, {1.0, 0.0, 0.0, 0.0}, {0.0, 0.0, 1.0, 0.0}, {0.0, 0.0, 0.0, 1.0}};
```
rouault: Isn't the value 36843762.068421006 specific to the use of the GRS80 ellipsoid ?
And I suspect that P->left and P->right shouldn't be overridden, and P->a not used directly, to let the generic code in src/fwd.cpp and src/inv.cpp do the scaling from the unit ellipsoid to the target one.

rouault: @plouvart gentle ping w.r.t my above question

plouvart: Hello @rouault 
Sorry for the lack of response over the last few week's. Lots of personal and work related obligations.
I will have time for a closer look at the issue this week end.

plouvart: @rouault 
I remember that I got the inspiration for the P->right, P->left and P->a lines from the [s2 projection file](https://github.com/plouvart/PROJ/blob/18f9d9eafa6693a32ae0391501b174b79e70345b/src/projections/s2.cpp#L429)
I didn't know that src/fwd and src/inv does that for us automatically. I can remove those lines.

About the 36843762.068421006 value, it is related to the final scale of the airocean space whose dimensions stays constant whatever ellipsoid model we use (since it all gets _normalized_ to the unit sphere in the end anyway).
Unless I'm mistaken?

rouault: > About the 36843762.068421006 value, it is related to the final scale of the airocean space whose dimensions stays constant whatever ellipsoid model we use (since it all gets _normalized_ to the unit sphere in the end anyway).

sounds surprising to me that the extent of the projection is constant and not dependent on the ellipsoid semi-major axis, but I don't know anything about that projection. I guess a could way to check that would be to add round-trip tests with the unit ellipsoid

plouvart: > I guess a could way to check that would be to add round-trip tests with the unit ellipsoid

That's a good idea. I'll try that

plouvart: Round trip tests work when the ellipsoid/no ellipsoid is the same for fwd and inv successively.
Obviously choosing a different ellipsoid model for the fwd and inv model result in a difference.

The dimensions of the Airocean space are determined by the length of the edges of the triangle faces of the icosahedron, which depends on the arbitrary choice of the radius of the sphere it is circumscribed within.
The choice was made to have that scaling factor equals the arithmetic mean radius of the earth, ie. 6,371,008.8 meters.

So it does not mater what ellipsoid model was chosen in the end, because whatever the model chosen, it always gets normalized to a unit sphere as an intermediary step before being scaled to that arbitrary dimension.

plouvart: image
Here's a small sketch I drew to illustrate my point.
blue process is a fwd/inv with +ellps=sphere and red process is fwd/inv with +ellps=GRS80 for instance.

rouault: I still believe there's something wrong related to taking into account ellipsoid characteristics. The value of scale factors compared to other projections seems to be totally off:
```
$ echo 2 49 | bin/proj +proj=merc -S 
222638.98	6242596.00	<1.52134 1.52134 2.31449 0 1.52134 1.52134>

$ echo 2 49 | bin/proj +proj=tmerc -S 
146339.48	5431555.61	<1.00026 1.00026 1.00053 0 1.00026 1.00026>

 echo 2 49 | bin/proj +proj=s2 -S 
0.53	1.05	<0.823654 1.0871 0.895218 15.8914 1.08739 0.823271>

$ echo 2 49 | bin/proj +proj=airocean -S 
13269980.92	20976603.56	<5.70685e+06 6.08587e+06 3.46329e+13 5.67366 6.18381e+06 5.60058e+06>
```

plouvart: Sorry for the reply delay.
It is really weird to me that those factors would be so out of place considering the end result of projecting vectors and raster work so well.

I will be investigating this issue today. I'm open to any idea/tip about what may be causing this discrepancy as I am unfamiliar with parallel and meridional scale factors computations

plouvart: You were right @rouault . I rebuild the transform matrices and airocean unfolded faces so they do not make an assumption on the semi-major earth axis.
I also recomputed values for the tests and made sure they pass for both vertical and horizontal orientations.

rouault: Things look ok now:
```
$ echo 2 49 | bin/proj +proj=airocean -S +ellps=GRS80
13284846.38	21000102.22	<0.895755 0.955245 0.853244 5.67366 0.970618 0.879073>
$ echo 2 49 | bin/proj +proj=airocean -S +a=1 +rf=298.257
2.08	3.29	<0.895755 0.955245 0.853244 5.67366 0.970618 0.879073>
```

> I rebuild the transform matrices

So I understand you have a scripts to generate those base_ico_faces[], etc. arrays? Would it be possible to commit it into the repository too (and add a comment in the .cpp file to point to it), otherwise next maintainers (PROJ has a 40+ years history...) will wonder how those magic values have been generated.

plouvart: Sure. It is in the form of a python notebook file (.ipynb). Where in the repo should I put it?

rouault: > Where in the repo should I put it?

I'd say in ``scripts/``

plouvart: I added the notebook (with markdown comments), as well as a mention in the projection source file.

kbevers: Would it be possible to convert the notebook to a pure python script? While notebooks are fairly common these days it's definitely not something everyone is familiar with (me included) and just browsing the file without opening it in the intended way is quite overwhelming. And again, this might be useful again in a couple of decades when notebooks will have been replaced by whatever is in at the time.
The image that is encoded within the notebook is likely very helpful in general and could be placed in the user documentation.

plouvart: @kbevers 
I see your point. I will convert it to plain python when I have time tonight.

plouvart: I removed the notebook and added a script that takes no arguments and generates the 2 plot images and prints the C code on the standard output.
The only required import for printing the C code is the numpy library.

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,558 @@
...
+translation(h, 0) @ rotation(np.pi / 2 * 1) @ upper_right
+orient_horizontal = translation(h, 0) @ rotation(np.pi / 2 * 1)
+inv_orient_horizontal = np.linalg.inv(orient_horizontal)
+
+
+"""
+    Definition of icosahedron vertices, faces and centers.
+"""
+
+ico_faces = np.array(
```
rouault, line 252: Where do the below numeric values come from ?

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,558 @@
...
+            [-0.58849102, 0.53029673, 0.0627648, 1.0],
+        ],
+        [
+            [-0.99500944, 0.0913478, -0.04014718, 1.0],
+            [-0.58849102, 0.53029673, 0.0627648, 1.0],
+            [-0.41468223, 0.65596241, 0.63067581, 1.0],
+        ],
+    ]
+)
+ico_20_centers = np.array(
```
rouault, line 717: Where do the below numeric values come from ?

scripts/build_airocean_parameters.py
plouvart: Those are values for the vertices of the icosahedron used by Gray Fuller.
Those can be found here http://www.rwgrayprojects.com/rbfnotes/maps/docs/graypr.c
Do I need to put a mention/link to the original source?

rouault: > Do I need to put a mention/link to the original source?

yes, please! That will help immensely anyone that would have to dig into that with no knowledge of the context

rouault:
Looks good to me. At others, any remaining comment before merging?

scripts/build_airocean_parameters.py
mwtoews: It would be nice to have the full precision as the source too, i.e. `0.42015243`->`0.420152426708710003`

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,558 @@
...
+    ]
+)
+# Centers corresponding to the split faces around Japan
+ico_3_centers = np.array(
+    [
+        [-0.64642729, 0.48840818, -0.12653887],
+        [-0.45298488, 0.67661305, 0.09706879],
+        [-0.66606089, 0.42586898, 0.21776448],
+    ]
+)
```
mwtoews: These are defined, but unused. So remove?

scripts/build_airocean_parameters.py
mwtoews: This appears to be unused, so could be removed.

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,558 @@
...
+
+"""
+    Definition of the vertices of the unfolded icosahedron net:
+"""
+
+
+# Icosahedron properties
+circumscribed_radius = 1
+inscribed_radius = 3**0.5 / 12 * (3 + 5**0.5)
+triangle_edge = circumscribed_radius / np.sin(2 * np.pi / 5)
```
mwtoews: keep a native Python float (rather than mixing np.float64) by a cast:
```suggestion
triangle_edge = float(circumscribed_radius / np.sin(2 * np.pi / 5))
```

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,558 @@
...
+triangle_height = 3**0.5 / 2 * triangle_edge
+
+# Icosahedron properties scaled to a specifi radius. Left to 1.0
```
mwtoews:
```suggestion
# Icosahedron properties scaled to a specific radius. Left to 1.0
```

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,560 @@
...
+
+def rotation(angle):
+    res = np.eye(4)
+    res[:2, :2] = np.round(
+        [
+            [np.cos(angle), -np.sin(angle)],
+            [np.sin(angle), np.cos(angle)],
+        ],
+        10,
+    )
```
mwtoews: keep full precision, but remove the near zeros from `np.cos(np.pi / 2)`
```suggestion
    res[:2, :2] = zap_zero_or_one(
        np.array(
            [
                [np.cos(angle), -np.sin(angle)],
                [np.sin(angle), np.cos(angle)],
            ]
        )
    )
```

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,560 @@
...
+w, h = airocean_vertices[:, 0].max(), airocean_vertices[:, 1].max()
+lower_left = [0, 0, 0, 1]
+lower_right = [w, 0, 0, 1]
+upper_left = [0, h, 0, 1]
+upper_right = [w, h, 0, 1]
+
+translation(h, 0) @ rotation(np.pi / 2 * 1) @ lower_left
+translation(h, 0) @ rotation(np.pi / 2 * 1) @ lower_right
+translation(h, 0) @ rotation(np.pi / 2 * 1) @ upper_left
+translation(h, 0) @ rotation(np.pi / 2 * 1) @ upper_right
```
mwtoews: these are not used, so either remove or comment-out

QuLogic: These lines don't do anything?

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,560 @@
...
+            crs=None,
+        )
+        ax = gdf.plot()
+        gdf.apply(
+            lambda x: ax.annotate(
+                text=x["id"], xy=x.geometry.centroid.coords[0], ha="center"
+            ),
+            axis=1,
+        )
+        ax.get_figure().savefig(str(outfile))
```
QuLogic: `savefig` accepts `Path`s
```suggestion
        ax.get_figure().savefig(outfile)
```

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,560 @@
...
```
QuLogic:
```suggestion
        ax.get_figure().savefig(outfile)
```

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,560 @@
...
+                "constexpr PJ_XYZ base_ico_normals[23]": ico_normals,
+                "constexpr pj_face base_airocean_faces[23]": airocean_23_face_vertices[
+                    :, :, [0, 1, 3]
+                ],
+                "constexpr double base_ico_air_trans[23][4][4]": ico_air_trans[
+                    :, :, :
+                ],
+                "constexpr double base_air_ico_trans[23][4][4]": air_ico_trans[
+                    :, :, :
+                ],
```
QuLogic: No need for copies?
```suggestion
                "constexpr double base_ico_air_trans[23][4][4]": ico_air_trans,
                "constexpr double base_air_ico_trans[23][4][4]": air_ico_trans,
```

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,560 @@
...
+        v2 /= v2[3]
+        assert np.isclose(v, v2, rtol=1e-7).all()
+
+"""
+    End of definitions.
+"""
+
+
+def generate_20_faces_airocean_net(
+    outfile: Path = Path("./20_faces_airocean_net.png"),
```
QuLogic: There's only one caller; why bother with a default?

mwtoews (commented):
Great work! I have a few other suggestions...

Some of the parameters very close to zero or one could do with a wee bit of rounding. Here is a def that could be put near the top and used in a few places:
```python
def zap_zero_or_one(vals, eps=1e-12, verbose=False):
    """Return copy of array with values very close to zero and one set exactly."""
    vals = vals.copy()
    sel = (np.abs(vals) < eps) & (vals != 0.0)
    if sel.any():
        vals[sel] = 0.0
        if verbose:
            print(f"adjusted {sel.sum()} zeros")
    ones = np.abs(np.abs(vals) - 1.0)
    sel = (ones < eps) & (ones != 0.0)
    if sel.any():
        vals[sel] = np.sign(vals[sel])
        if verbose:
            print(f"adjusted {sel.sum()} ones")
    return vals
```

scripts/build_airocean_parameters.py
```diff
@@ -0,0 +1,558 @@
...
+        a.T @ np.linalg.inv(b.T)
+        for i, (a, b) in enumerate(zip(ico_basis, airocean_basis))
+    ]
+)
+ico_air_trans = np.array(
+    [
+        b.T @ np.linalg.inv(a.T)
+        for i, (a, b) in enumerate(zip(ico_basis, airocean_basis))
+    ]
+)
```
mwtoews: (missed this one from my previous review). Using the `zap_zero_or_one` function that I mentioned earlier, many near zeros and ones can be set exactly zero and one. These were introduced from the iterative matrix inversion algorithms. I suggest to amend this:
```python
air_ico_trans = zap_zero_or_one(air_ico_trans)
ico_air_trans = zap_zero_or_one(ico_air_trans)
```
with `verbose=True` the messages show many tiny adjustments:
```
adjusted 27 zeros
adjusted 8 ones
adjusted 36 zeros
adjusted 6 ones
```

scripts/build_airocean_parameters.py
plouvart: Was useful during testing but indeed not used anymore. Will remove.

scripts/build_airocean_parameters.py
plouvart: Used during verification but not needed anymore indeed.

scripts/build_airocean_parameters.py
plouvart: I thought it might depend on matplotlib version but you are correct

scripts/build_airocean_parameters.py
plouvart: Indeed

scripts/build_airocean_parameters.py
plouvart: In case someone wants to generate files somewhere else. I agree this has limited usefulness, but that'd be counterproductive to remove it.

scripts/build_airocean_parameters.py
plouvart: @mwtoews 
Indeed. I now added the full precision of the source

scripts/build_airocean_parameters.py
QuLogic: If they were generating files somewhere else, they'd specify a path, so I don't understand why that means there should be an unused default.

plouvart: I removed the defaults.

PR #4318, opened 2024-11-09
createObjectsFromName(): be more tolerant about N/S vs North/South, absence of zone or height

rouault:
We want the following matches to be possible:

```
user entry            official name
------------------    ----------------
EGM96                 EGM96 height
WGS84 UTM31 north     WGS 84 / UTM zone 31N
```

### DIFF
--- a/src/iso19111/factory.cpp
+++ b/src/iso19111/factory.cpp
@@ -9171,7 +9171,8 @@ AuthorityFactory::createObjectsFromNameEx(
         auto sqlRes = d->run(sql, params);
         bool isFirst = true;
         bool firstIsDeprecated = false;
-        bool foundExactMatch = false;
+        size_t countExactMatch = 0;
+        size_t countExactMatchOnAlias = 0;
         std::size_t hashCodeFirstMatch = 0;
         for (const auto &row : sqlRes) {
             const auto &name = row[3];
@@ -9262,9 +9263,12 @@ AuthorityFactory::createObjectsFromNameEx(
                 throw std::runtime_error("Unsupported table_name");
             };
             const auto obj = getObject(table_name, code);
-            if (metadata::Identifier::canonicalizeName(obj->nameStr()) ==
-                canonicalizedSearchedName) {
-                foundExactMatch = true;
+            if (metadata::Identifier::isEquivalentName(
+                    obj->nameStr().c_str(), searchedName.c_str(), false)) {
+                countExactMatch++;
+            } else if (metadata::Identifier::isEquivalentName(
+                           name.c_str(), searchedName.c_str(), false)) {
+                countExactMatchOnAlias++;
             }
 
             const auto objPtr = obj.get();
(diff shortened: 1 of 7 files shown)


src/iso19111/metadata.cpp
```diff
@@ -1249,11 +1257,34 @@ static const struct utf8_to_lower *get_ascii_replacement(const char *c_str) {
...
-std::string Identifier::canonicalizeName(const std::string &str) {
+
+static bool matchesLowerCase(const char *c_str, const char *needle) {
+    size_t i = 0;
+    for (; c_str[i] && needle[i]; ++i) {
+        if (lower(c_str[i]) != lower(needle[i])) {
+            return false;
+        }
+    }
+    return needle[i] == 0;
```
jjimenezshaw, line 1272: `matchesLowerCase("Java", "javaScript");` returns `false`, but `matchesLowerCase("JavaScript", "java");` returns `true`. Is that intentional?
Otherwise
```suggestion
    return c_str[i] == 0 && needle[i] == 0;
```

src/iso19111/metadata.cpp
```diff
@@ -1311,18 +1386,84 @@ bool Identifier::isEquivalentName(const char *a, const char *b) noexcept {
             ++j;
             continue;
         }
-        if (aCh == '1' && !(lastValidA >= '0' && lastValidA <= '9') &&
-            a[i + 1] == '9' && a[i + 2] >= '0' && a[i + 2] <= '9') {
+        if (aCh == '1' && !isdigit(lastValidA) && a[i + 1] == '9' &&
```
jjimenezshaw, line 1404: What is the point of checking if the string has "19"+digit? Is it about years in the XX century? (I know this PR is not changing the logic, I am just curious).

src/iso19111/metadata.cpp
```diff
@@ -1311,18 +1386,84 @@ bool Identifier::isEquivalentName(const char *a, const char *b) noexcept {
...
+            isdigit(b[j + 2])) {
             j += 2;
             lastValidB = '9';
             continue;
         }
+
+        if (biggerDifferencesAllowed) {
+            bool skip = false;
+            // Skip "zone" if preceding character is a space
+            if (aCh == 'z' && i > 0 && isIgnoredChar(a[i - 1]) &&
```
jjimenezshaw: What about a lambda to do these repetitive tasks?

src/iso19111/metadata.cpp
rouault: > Is that intentional?

yes, it checks if it finds needle (substring) within the string. Code comment added

src/iso19111/metadata.cpp
rouault: yes, "WGS1984" is seen as equivalent to "WGS84". Code comment added

src/iso19111/metadata.cpp
rouault: excellent idea. Done

jjimenezshaw (approved):

src/iso19111/factory.cpp
```diff
@@ -9171,7 +9171,8 @@ AuthorityFactory::createObjectsFromNameEx(
         auto sqlRes = d->run(sql, params);
         bool isFirst = true;
         bool firstIsDeprecated = false;
-        bool foundExactMatch = false;
+        size_t countExactMatch = 0;
+        size_t countExactMatchOnAlias = 0;
```
jjimenezshaw, line 9175: Is this "alias" the alias from EPSG?

rouault: > Is this "alias" the alias from EPSG?

yes, or ESRI. We request not only on the "name" field of objects, but also join the "alias_name" table

PR #4322, opened 2024-11-14
Removed leftover setting of CMAKE_REQUIRED_FLAGS

seanm:
Back in 88d91f709c86fad6c0cc8b1f15706216cba81ffe I removed a `check_c_source_compiles` to detect some pthread stuff and converted to std::mutex instead.

This stuff I believe I should have removed back then too.

The -Werror caused a problem in VTK, which still has the `check_c_source_compiles`, because a totally unrelated warning made the try-compile fail when it should have passed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -289,14 +289,6 @@ if(Threads_FOUND AND CMAKE_USE_PTHREADS_INIT)
     "${CMAKE_REQUIRED_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT}")
 endif()
 
-include(CheckCSourceCompiles)
-if(MSVC)
-  set(CMAKE_REQUIRED_FLAGS "${CMAKE_C_FLAGS} /WX /W4")
-else()
-  set(CMAKE_REQUIRED_LIBRARIES m)
-  set(CMAKE_REQUIRED_FLAGS "${CMAKE_C_FLAGS} -Werror -Wall")
-endif()
-
 # Set a default build type for single-configuration cmake generators if
 # no build type is set.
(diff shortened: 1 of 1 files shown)


seanm:
This I think I should have done back in https://github.com/OSGeo/PROJ/pull/3425.

The Werror removed in this PR is causing me grief in VTK: https://gitlab.kitware.com/vtk/vtk/-/issues/19532

CMakeLists.txt
```diff
@@ -287,14 +287,6 @@ if(Threads_FOUND AND CMAKE_USE_PTHREADS_INIT)
     "${CMAKE_REQUIRED_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT}")
 endif()
 
-include(CheckCSourceCompiles)
-if(MSVC)
-  set(CMAKE_REQUIRED_FLAGS "${CMAKE_C_FLAGS} /WX /W4")
-else()
-  set(CMAKE_REQUIRED_LIBRARIES m)
```
rouault, line 296: we should probably keep ``set(CMAKE_REQUIRED_LIBRARIES m)``

seanm: So the whole block isn't there just for the now-removed check_c_source_compiles?

rouault:
your pull request is based on a not so recent master. Can you please rebase it on top of latest master HEAD ?

CMakeLists.txt
rouault: hum, I see there were past math function detection in 2622d38515, but they have gone to now. So, probably disregard my comment and this is indeed no longer needed

seanm:
> your pull request is based on a not so recent master. Can you please rebase it on top of latest master HEAD ?

Weird, I swear I did, but indeed not. Should be fixed now.

PR #4336, opened 2024-12-04
Fix typo Tranverse->Transverse and many other fixes

mwtoews:
This fixes a number of typos found and corrected with a modern version of [codespell](https://github.com/codespell-project/codespell). Here are some notes and observations:

- The most visible correction is Tranverse->Transverse
- Some of the corrections were not necessarily wrong before, but are dated forms, e.g., [advertize](https://en.wiktionary.org/wiki/advertize) or [synthetize](https://en.wiktionary.org/wiki/synthetize)
- Others are French, e.g. [instanciation](https://en.wiktionary.org/wiki/instanciation)
- There is [`scripts/fix_typos.sh`](https://github.com/OSGeo/PROJ/blob/9.5.1/scripts/fix_typos.sh) in the repo (based on [an older fork of codespell](https://github.com/rouault/codespell/tree/gdal_improvements)), but either it hasn't been run for a while or it doesn't have the dictionaries to catch the ones in this PR? I could consider a subsequent PR to this one to run modern codespell in CI, which runs in a few seconds.
- There are a number of typos from EPSG's data that are not fixed here, but I might report them to IOGP.


CODE_OF_CONDUCT.md
```diff
@@ -6,7 +6,7 @@ In the interest of fostering an open and welcoming environment, we as
 contributors and maintainers pledge to make participation in our project and
 our community a harassment-free experience for everyone, regardless of age, body
 size, disability, ethnicity, sex characteristics, gender identity and expression,
-level of experience, education, socio-economic status, nationality, personal
+level of experience, education, socioeconomic status, nationality, personal
```
kbevers: I guess this is a bit like changing the wording in a LICENSE file. The code of conduct is versioned and verbatim copy of what's been put out by the contributor-covenant, so I think this should be kept as-is. They repeat the same spellign in newer versions, it seems.

docs/source/community/code_of_conduct.rst
```diff
@@ -17,7 +17,7 @@ In the interest of fostering an open and welcoming environment, we as
 contributors and maintainers pledge to make participation in our project and
 our community a harassment-free experience for everyone, regardless of age, body
 size, disability, ethnicity, sex characteristics, gender identity and expression,
-level of experience, education, socio-economic status, nationality, personal
+level of experience, education, socioeconomic status, nationality, personal
```
kbevers: Same comment here as above.

kbevers (approved):
Hard to argue against improving spelling. Good work. 

I must be the culprit for a lot of these. Both because English isn't my first language, but more likely because I have to turn off the spellchecker when building the docs locally before releases (I haven't found a way to make the sphinx spellchecker work on my ARM-based Mac).

mwtoews:
I've certainly made my fair share of typos, so no need to take any blame. I'll look into embedding spellchecking into the CI process to minimise these going forward, which could also replace "sphinxcontrib.spelling" too.

kbevers:
> I'll look into embedding spellchecking into the CI process to minimise these going forward, which could also replace "sphinxcontrib.spelling" too.

We already have that:
https://github.com/OSGeo/PROJ/blob/master/.github/workflows/doc_checks.yml 

It has a long list of ignored words in [this file](https://github.com/OSGeo/PROJ/blob/master/docs/source/spelling_wordlist.txt). One of the words is "Tranverse" :)

mwtoews:
"Tranverse" was added to the ignore list recently; see 693620fe71c17f689332109de36162d8bbd931ec

Issue #4327, opened 2024-11-22
WebAssembly build


birkskyum:
Hi, I'm interested in a WebAssembly build of PROJ.

The intention is to mitigate differences across PROJ and proj4js, by simply running PROJ in the browser.

I'm curious if there are known blockers to this.

I've previously managed to make [large C++ codebases](https://github.com/maplibre/maplibre-native) run online [as WebAssembly](https://birkskyum.github.io/maplibre-native-wasm/qt-for-webassembly/webgl1-from-opengl2-legacy-renderer.html).

It's sometimes a bit slow still, but there're are all the time small performance gains coming from improved browser support for wasm, and [improvements in emscripten](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md) to leverage it.

jjimenezshaw:
Hi @birkskyum 
I want to do it as well... but I do not know anything about WebAssembly. It would be great to have it (and include it in https://spatialreference.org ;)

There is no "hard" blocker. There is an implementation of GDAL in WebAssembly: https://github.com/bugra9/gdal3.js
So it is doable. 

This gdal implementation is disabling curl, so it cannot use grid files.
About the curl part, somebody got this https://github.com/emscripten-core/emscripten/issues/3270#issuecomment-817345275 , but we can leave that for a later step.

In the release of PROJ 9.6.0 (planned for March) there is an option to include the proj.db file embedded in the library. That would make it even easier. https://github.com/OSGeo/PROJ/pull/4265

If you want to do it, please contact me.

dg0yt:
I don't know if it is usable at runtime, but I can say that the default build of proj:wasm32-emscripten succeeds in vcpkg. This is the list of dependencies (host: x64-linux):
~~~
$ ./vcpkg depend-info proj:wasm32-emscripten
vcpkg-cmake: 
vcpkg-cmake-config: 
vcpkg-cmake-get-vars: vcpkg-cmake
openssl:wasm32-emscripten: vcpkg-cmake, vcpkg-cmake-config, vcpkg-cmake-get-vars
zlib:wasm32-emscripten: vcpkg-cmake
libjpeg-turbo:wasm32-emscripten: vcpkg-cmake, vcpkg-cmake-config
liblzma:wasm32-emscripten: vcpkg-cmake, vcpkg-cmake-config
[...]

PR #4349, opened 2024-12-13
Add a CMake EMBED_GRIDS_DIRECTORY option to embed .tif/.json files in to libproj

rouault:
and also generalize the existing EMBED_RESOURCE_FILES option to embed the ITRFxxxx resource files.

```
.. option:: EMBED_GRIDS_DIRECTORY=<directory>

    .. versionadded:: 9.6

    Embed files from <directory> ending with .tif or .json in the PROJ library itself.

    The pointed directory can potentially be the full PROJ-data package (uncompressed).
    In that case, about 6 GB of free disk and 16 GB of RAM are required to build PROJ.

    When using this parameter, EMBED_RESOURCE_FILES must be set to ON.

    If the content of the directory changes, you need to run CMake again to
    update the list of files.
```


hobu:
All-or-nothing for now, but would you expect to allow configuration of which grids should be embedded?

rouault:
> All-or-nothing for now, but would you expect to allow configuration of which grids should be embedded?

why not just creating a directory containing all grids you are interested in ? Pointing at individual files with CMake options could be tedious, couldn't it?

hobu:
> why not just creating a directory containing all grids you are interested in ? Pointing at individual files with CMake options could be tedious, couldn't it?

Fair point, although I was thinking about standard distribution situations like a widely distributed WASM PROJ with embedded resources on cdn.proj.org. Maybe that's always every grid file...

willcohen:
My work-in-progress WASM proj could be a good candidate for this (though, to be clear, it's still squarely in javascript+wasm land because of needing to handle the malloc and freeing of objects via the C api -- https://github.com/OSGeo/PROJ/issues/4327). It'll prob be a little while before I stabilize enough to throw all the grid files in there, but I could imagine there being, effectively, two different distributions of WASM PROJ, one ~10-20MB one where it's the library + DB (size dependent on whether it's base64'd into a JS file, as is the case right now), or a mega one for people who are going for full-intensity projections.

PR #4353, opened 2024-12-16
Take into account new EPSG methods 'Cartesian Grid Offsets by TIN Interpolation (JSON)' and 'Vertical Offset by TIN Interpolation (JSON)'

rouault:
and remove custom transformations for Finland TIN base transformations that were previously under the PROJ authority

PR #4354, opened 2024-12-16
ConcatenatedOperation::fixStepsDirection(): add even more hacks to deal with EPSG:10675 'BES2020 to Saba height'

rouault:
On top of PR #4353


rouault:
Closing. This is not the right approach to deal with this transformation. It happens the Helmert transformation must be done in 3D. I've suggested EPSG uses an explicit 3D Helmert operation as the first step to remove any ambiguity

PR #4356, opened 2024-12-19
Added an option native_ca to proj.ini and an environment variable PROJ_NATIVE_CA to be able to configure curl to use the operating system CA store.

### LINKED ISSUE #4338: Possibility to set CURLSSLOPT_NATIVE_CA through proj.ini
I would like to suggest a feature to make it possible to use the Windows Certificate Store to get trusted CA certificates. I think a nice way to do this would be to add a boolean option to proj.ini to allow the user to set CURLSSLOPT_NATIVE_CA, maybe called `native_ca`. CURLSSLOPT_NATIVE_CA is part of : [CURL_SSL_OPTIONS](https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html)

The reason we want to do this is because we have a customer that uses Zscaler. Zscaler is network security software that does deep packet inspection by doing a MITM on all ssl handshakes. For applications to be able to work with this setup, they need to trust the Zscaler CA certificate because that is the one they will actually see. In PROJ it is currently only possible to use a file that contains CA certs (curl-ca-bundle.crt), it would be very inconventient for our customer to have to manually add a certificate here on all machines. On Windows it is also most common for applications to use the Windows Certificate Store, so it would be very nice if PROJ supported this.

I am aware that you can configure the SSL_OPTIONS (which includes the CURLSSLOPT_NATIVE_CA flag) in PROJ (#3936) but we would prefer not to make and maintain our own build of PROJ.

Thank you.

Robrecht-VS:
- [x] Closes #4338


src/networkfilemanager.cpp
```diff
@@ -1622,7 +1627,15 @@ CurlFileHandle::CurlFileHandle(PJ_CONTEXT *ctx, const char *url, CURL *handle)
 #if defined(SSL_OPTIONS)
     // https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html
     auto ssl_options = static_cast<long>(SSL_OPTIONS);
+    if (pj_context_get_native_ca(ctx)) {
+        ssl_options = ssl_options | CURLSSLOPT_NATIVE_CA;
```
rouault, line 1634: From CI failures, it seems that CURLSSLOPT_NATIVE_CA is only available in "recent" Curl versions, so this will need to be conditioned by a #ifdef on the Curl version

Robrecht-VS: Thank you for your clarification @rouault , I have committed a fix.

data/proj.ini
```diff
@@ -42,6 +42,13 @@ only_best_default = off
 ; (added in PROJ 9.0)
 ; ca_bundle_path = /path/to/cabundle.pem
 
+; When this is set to on, the operating systems native CA store will be used for certificate verification
+; If you set this option to on and also set ca_bundle_path then during verification those certificates are
+; searched in addition to the native CA store.
+; Valid values = on, off
```
rouault, line 49:
```suggestion
; (added in PROJ 9.6)
; Valid values = on, off
```

docs/source/usage/environmentvars.rst
```diff
@@ -106,3 +106,9 @@ done by setting the variable with no content::
     Define a custom path to the CA Bundle file. This can be useful if `curl`
     and :envvar:`PROJ_NETWORK` are enabled. Alternatively, the 
     :c:func:`proj_curl_set_ca_bundle_path` function can be used.
+
+.. envvar:: PROJ_NATIVE_CA
+
+    When this is set to ON, the operating systems native CA store will be used for certificate verification
```
rouault, line 114:
```suggestion
    .. versionadded:: 9.6.0
    
    When this is set to ON, the operating systems native CA store will be used for certificate verification
```

src/networkfilemanager.cpp
```diff
@@ -1622,7 +1629,19 @@ CurlFileHandle::CurlFileHandle(PJ_CONTEXT *ctx, const char *url, CURL *handle)
...
     // https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html
     auto ssl_options = static_cast<long>(SSL_OPTIONS);
+#if CURL_AT_LEAST_VERSION(7, 71, 0)
+    if (pj_context_get_native_ca(ctx)) {
+        ssl_options = ssl_options | CURLSSLOPT_NATIVE_CA;
+    }
+#endif
     CHECK_RET(ctx, curl_easy_setopt(handle, CURLOPT_SSL_OPTIONS, ssl_options));
+#else
+#if CURL_AT_LEAST_VERSION(7, 71, 0)
```
rouault, line 1639: It looks to me that this branch ( SSL_OPTIONS not defined but curl >= 7.71) cannot be taken as SSL_OPTIONS is anterior to 7.71

Robrecht-VS: As far as I understand, SSL_OPTIONS is a preprocessor definition that is specific to PROJ, it is not defined by curl. It was introduced in this pull request: #3936

rouault: ah indeed, I missed that. Looks ok then

PR #4366, opened 2025-01-02
Build: Update to C++17 version

rouault:
The immediate driver for this change is related when fetching GTest. Currently we are stuck with GTest-1.12.1 because is the latest version supporting C++11. Later versions require C++14.
However, GTest-1.12.1 has a CMakeLists.txt with
``cmake_minimum_required(VERSION 3.5)`` which causes the following warning with modern CMake versions:
```
CMake Deprecation Warning at build/_deps/googletest-src/CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.
```
Updating to >= C++14 allows us to use GTest-1.15.2 which has a ``cmake_minimum_required(VERSION 3.13)``


src/iso19111/factory.cpp
```diff
@@ -3114,9 +3114,9 @@ std::string
 DatabaseContext::suggestsCodeFor(const common::IdentifiedObjectNNPtr &object,
                                  const std::string &authName,
                                  bool numericCode) {
-    const char *tableName = "";
+    const char *tableName = "prime_meridian";
```
jjimenezshaw, line 3117: Modifying a const char * ?
(I see it is not changing the chars, but the pointee)
What was the checker complain to change it now?
Would it make more sense to use an std::string?

rouault: the checker complained about the initial value never being read and always modified. This is tiny more bit performant (and contribute to binary size reduction) to use a const char* here given that it is a fixed set of strings, although that shouldn't matter in practice.

PR #4375, opened 2025-01-09
Use std::make_unique<> instead of internal::make_unique<>

rouault:


src/filemanager.cpp
```diff
@@ -2059,11 +2058,10 @@ void pj_load_ini(PJ_CONTEXT *ctx) {
...
                     ci_equal(value, "TRUE");
-            }
-            else if (native_ca == nullptr && key == "native_ca"){
-                ctx->native_ca =
-                    ci_equal(value, "ON") || ci_equal(value, "YES") ||
-                    ci_equal(value, "TRUE");
+            } else if (native_ca == nullptr && key == "native_ca") {
+                ctx->native_ca = ci_equal(value, "ON") ||
+                                 ci_equal(value, "YES") ||
+                                 ci_equal(value, "TRUE");
```
jjimenezshaw, line 2064: This seems to be a change by clang-format. Why now? Didn't we run it over the whole repository?

rouault: > This seems to be a change by clang-format

yes I've isolated the reformats in a separate commit

> Why now? Didn't we run it over the whole repository?

Come from recent contributions from occasional contributors who haven't set up pre-commit, and we don't have yet a CI job to detect formatting violations.

PR #4379, opened 2025-01-14
createOperations(): use more appropriate operation when using a 'PROJ {grid_name}' geoid model, based on matching the vertical datum

rouault:
Helps when transforming from 'NAD83(2011) / Puerto Rico and Virgin Is. + "VIVD09 height + PROJ us_noaa_g2012bp0.tif' to NAD83(2011). Without that fix, PROJ would use the EPSG record for the NAD83(2011) <--> PRVD02 height (Puerto Rico) to get the extent of the transformation, instead of the one for NAD83(2011) <--> VIVD09 height (Virgin Islands), causing the resulting global transformation to not be usable in practice.


kbevers:
I am not sure I understand this. Is this specific to CRS's described in WKT or a more general thing?

rouault:
> I am not sure I understand this. Is this specific to CRS's described in WKT or a more general thing?

The fix is general to geoid grids that are referenced multiple times in the grid_transformation table with different extents. This is just the use case that triggered the need for this fix.

PR #4384, opened 2025-01-16
Défouloir: arrêtons de faire de la pub pour cette organisation

rouault:


busstoptaktik:
I suppose a higher degree of clarity would be obtained by changing "standardization organism" to either "standards body" or "standardization organization" (en-US) / "standardisation organisation" (en-GB).

For the submitter's declared purpose of letting out steam, however, it probably doesn't matter much which exact term is used... steam is steam, after all, and it exits willingly through any sufficiently sized rupture 😊

kbevers:
I understand the frustrations with OGC. But could we please - for the sake of our projects credibility - take the high road and not add more fuel to the fire?

PR #4388, opened 2025-01-24
Use C++17 function ellint_1 in adams.cpp

cffk:
Replace approximate series for F(x, 1/sqrt(2)) by a call to ellint_1.

The approximate series was only accurate to "better than 1e-7", which meant that the errors in the projection were typically about 1 m.  The adoption of ellint_1 (with its considerably better accuracy) then leads to a torrent of errors in the tests.  These have been fixed by changing the tolerances from 1 mm to 1 m.

The values in the tests should be replaced by more accurate ones, but I'm reluctant merely to update the values with the ones the code now produces given that the code depends on a parameter TOL = 1e-9.  Also, I don't understand why K(1/sqrt(2)) = std::comp_ellint_1(RSQRT2) = 1.8540746773013719 is sometimes given to only 5 decimal places as 1.85407.

- [x] Tests updated
- [x] Fully documented in commit message

### DIFF
--- a/src/projections/adams.cpp
+++ b/src/projections/adams.cpp
@@ -78,34 +78,7 @@ struct pj_adams_data {
 } // anonymous namespace
 
 #define TOL 1e-9
-#define RSQRT2 0.7071067811865475244008443620
-
-static double ell_int_5(double phi) {
-    /* Procedure to compute elliptic integral of the first kind
-     * where k^2=0.5.  Precision good to better than 1e-7
-     * The approximation is performed with an even Chebyshev
-     * series, thus the coefficients below are the even values
-     * and where series evaluation  must be multiplied by the argument. */
-    constexpr double C0 = 2.19174570831038;
-    static const double C[] = {
-        -8.58691003636495e-07, 2.02692115653689e-07, 3.12960480765314e-05,
-        5.30394739921063e-05,  -0.0012804644680613,  -0.00575574836830288,
-        0.0914203033408211,
-    };
-
-    double y = phi * M_2_PI;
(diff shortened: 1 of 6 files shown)


rouault:
it seems clang 18.1.8 both on MacOS with Conda and OSSFuzz Linux lacks support for std::ellint_1() which feels weird   (the failure on AppVeyor is likely due to it using MSVC17. we should probably retire that configuration)

cffk:
This is non-essential PR.  I was just seeing how C++17 provides.  I close this for now and we can revisit this in a year or so.

rouault:
> I was just seeing how C++17 provides

that's a bit strange as my reading of https://en.cppreference.com/w/cpp/numeric/special_functions/ellint_1 is that it should be a mandatory feature of a C++17 compliant environment (there are some parts sometimes of the C++ standards that are optional). Wondering if there aren't mixup of compilers and header of standard library from older compiler, or something like that

jjimenezshaw:
To document this behavior for a future development:
In the recent PRs #4401 and #4402 we got sample points for ESRI:54098 (WGS_1984_Adams_Square_II) ESRI:54099 (WGS_1984_Spilhaus_Ocean_Map_in_Square). (https://github.com/kgjenkins/test-coordinates)
We are much more near to the values from ESRI using `std::ellint_1` (from millimeters using `std::ellint_1` to 20 or 50 cm using the current formula. Near the borders of the projection, with big distortion, the differences between PROJ and ESRI are bigger).

Said that, I don't know if it is really impacting any user. Those projections are mainly used to display the whole world.

PR #4393, opened 2025-02-04
Dispatch content of 4D_api.cpp into multiple .cpp files

rouault:
No functional change, just moving code around.

4D_api.cpp has become quite large and a set of 'random' things, that didn't correlate well with the '4D' naming. Some re-organization was needed.

Create following files:
- src/area.cpp: PJ_AREA related code
- src/coordinates.cpp: functions related to coordinate manipulation
- src/coord_operation.cpp: PJCoordOperation methods
- src/create.cpp: pj_create_internal() related stuff
- src/crs_to_crs.cpp: proj_create_crs_to_crs() and the like
- src/dist.cpp: distance related functions
- src/info.cpp: proj_info() and proj_pj_info()
- src/trans.cpp: proj_trans(), proj_trans_array(), proj_trans_generic(), proj_roundtrip()
- src/trans_bounds.cpp: proj_trans_bounds()

Some other functions also moved to more relevant existing source files.


kbevers:
This is a much needed restructuring. It's a lot simpler to navigate the code this way. 

It seems all the functions in `misc.cpp` is related to coordinates, one way or another. Would it make sense to simply call it `coordinates.cpp`?

rouault:
> Would it make sense to simply call it `coordinates.cpp`?

I don't know. One of my initial thoughts was that it was mostly about units... But as that wasn't completely the case, I preferred misc...  Not opposed to "coordinates", but I've about the same mixed feeling as "units"

kbevers:
I'm not a massive fan of having "misc" or "utils" modules in a code base. It's never really obvious what's to be expected in them. Sometimes you really need a place to put random functions but in this case I think there is actually a common thread and the filename may as well reflect that. "coordinates" was what immediately came to my mind. There may be a better name for it.

rouault:
> Would it make sense to simply call it `coordinates.cpp`?

ok, I've settled on that

PR #4394, opened 2025-02-04
Add proj_trans_bounds_3D()

### LINKED ISSUE #4392: proj_trans_bounds for 3d bounding boxes
The existing proj_trans_bounds function is handy for transforming 2d bounding boxes, but cannot be used with 3d dimensional transforms (such as those involving geocentric CRS). I'd love to see an equivalent method exposed for transformation of 3d bounding boxes too.

rouault:
(on top of PR https://github.com/OSGeo/PROJ/pull/4393; only last commit is specific to this PR)

Fixes #4392

```c++
/** \brief Transform boundary, taking into account 3D coordinates.
 *
 * Transform boundary densifying the edges to account for nonlinear
 * transformations along these edges and extracting the outermost bounds.
 *
 * Note that the current implementation is not "perfect" when the source CRS is
 * geocentric, the target CRS is geographic, and the input bounding box
 * includes the center of the Earth, a pole or the antimeridian. In those
 * circumstances, exact values of the latitude of longitude of discontinuity
 * will not be returned.
 *
 * If one of the source or target CRS of the transformation is not 3D, the
 * values of *out_zmin / *out_zmax may not be significant.
 *
 * For 2D or "2.5D" transformation (that is planar component is
 * geographic/coordinates and 3D axis is elevation), the documentation of
 * proj_trans_bounds() applies.
 *
 * @param context The PJ_CONTEXT object.
 * @param P The PJ object representing the transformation.
 * @param direction The direction of the transformation.
 * @param xmin Minimum bounding coordinate of the first axis in source CRS
 *             (target CRS if direction is inverse).
 * @param ymin Minimum bounding coordinate of the second axis in source CRS.
 *             (target CRS if direction is inverse).
 * @param zmin Minimum bounding coordinate of the third axis in source CRS.
 *             (target CRS if direction is inverse).
 * @param xmax Maximum bounding coordinate of the first axis in source CRS.
 *             (target CRS if direction is inverse).
 * @param ymax Maximum bounding coordinate of the second axis in source CRS.
 *             (target CRS if direction is inverse).
 * @param zmax Maximum bounding coordinate of the third axis in source CRS.
 *             (target CRS if direction is inverse).
 * @param out_xmin Minimum bounding coordinate of the first axis in target CRS
 *             (source CRS if direction is inverse).
 * @param out_ymin Minimum bounding coordinate of the second axis in target CRS.
 *             (source CRS if direction is inverse).
 * @param out_zmin Minimum bounding coordinate of the third axis in target CRS.
 *             (source CRS if direction is inverse).
 * @param out_xmax Maximum bounding coordinate of the first axis in target CRS.
 *             (source CRS if direction is inverse).
 * @param out_ymax Maximum bounding coordinate of the second axis in target CRS.
 *             (source CRS if direction is inverse).
 * @param out_zmax Maximum bounding coordinate of the third axis in target CRS.
 *             (source CRS if direction is inverse).
 * @param densify_pts Recommended to use 21. This is the number of points
 *     to use to densify the bounding polygon in the transformation.
 * @return an integer. 1 if successful. 0 if failures encountered.
 * @since 9.6
 */
```


src/trans_bounds.cpp
```diff
@@ -0,0 +1,894 @@
...
+                y_boundary_array[iii] = ymax;
+                x_boundary_array[iii] = xmax - i * delta_x;
+                z_boundary_array[iii] = z;
+                ++iii;
+            }
+        }
+
+        const double delta_z = (zmax - zmin) / side_pts;
+
+        // (xmin, ymin) corner
```
nyalldawson: Should this instead be:
```suggestion
        // (xmin, ymin) edge
```
?

nyalldawson:
Thanks @rouault !

src/trans_bounds.cpp
rouault: yes, fixed

snowman2:
Would there be any benefit to adding the time component?

One idea I have was for this to behave like proj_trans and make z and t optional. Then, this would become the successor of proj_trans_bounds.

rouault:
> Would there be any benefit to adding the time component?

I considered that for about 1000 ms, but failed to see the practical interest, so I preferred to keep things simpler.

PR #4395, opened 2025-02-05
Database: update to EPSG v11.031

rouault:


jjimenezshaw (approved):
It is interesting that there are "low distortion projections" coming into EPSG. USA will be the next, with a few thousands. Let's see if they run out of numbers in the catalog.

rouault:
> they run out of numbers in the catalog.

we are going to get into serious trouble when CRS codes are going to be above 32767 since this is the maximum allowed by GeoTIFF

PR #4401, opened 2025-02-16
Add Spilhaus projection

### LINKED ISSUE #1851: Spilhaus projection
Example of Spilhaus projection
Also see [A story about the "Spilhaus projection" – a map projection that went viral in fall 2018 and will be supported in the next release of ArcGIS](https://storymaps.arcgis.com/stories/756bcae18d304a1eac140f19f4d5cb3d)

The article finds that the spilhaus projection is a `Adams World in a Square II` projection with these params:

```
    Geographic Coordinate System: GCS WGS 1984
    Projection: Adams Square II
    False Easting: 0 m
    False Northing: 0 m
    Scale Factor: 1
    Azimuth: 40.17823482°
    Longitude of Center: 66.94970198°E
    Latitude of Center: 49.56371678°S
    XY Plane Rotation: 45°
```

It would be fun to have this projection in proj. [This document](https://github.com/Esri/projection-engine-db-doc/blob/master/other/proj4-esri.txt) suggests that the "Adams World in a Square II" projection was implemented in the `libproject` distribution of proj. So maybe it is even possible to find working code somewhere.

jjimenezshaw:
Implementation of the Spilhaus projection, following most of the recommendations done in  #1851.
spilhaus

It is working for both the sphere and the ellipsoid. To support the ellipsoid it uses conformal latitudes.
An additional correction factor for the conformal latitude is included, that was not present in other implementations.

In ticket #1851 there was a discussion about the implementation in ESRI products, and the scale factor (apparently `sqrt(2)`) between both implementations. In case we want to do something else, please add a comment here. There is a CRS that uses it: [ESRI:54099](https://spatialreference.org/ref/esri/54099/)

- [x] Closes #1851
- [x] Tests added


jjimenezshaw:
Before merging this PR I have to squash the commits (and probably address the comments).

src/projections/spilhaus.cpp
```diff
@@ -0,0 +1,179 @@
...
+    double sinalpha;
+    double beta;
+    double lambda_0;
+    double conformal_distortion;
+    double cosrot;
+    double sinrot;
+    PJ *adams_ws2;
+};
+
+double conformal_lat(double lat, double e) {
```
rouault: could we move that do a dedicated file in src/  ? (if it is not already there, but can't spot it)
I can see that mod_ster_setup() does this computation too and could be another user

src/projections/spilhaus.cpp
```diff
@@ -0,0 +1,179 @@
...
+    lp.phi = aasin(P->ctx,
+                   Q->sinalpha * sinphi_s + Q->cosalpha * cosphi_s * coslam_s);
+
+    lp.lam = Q->lambda_0 +
+             aatan2(cosphi_s * sinlam_s,
+                    Q->sinalpha * cosphi_s * coslam_s - Q->cosalpha * sinphi_s);
+
+    // latitude (iterative formula from
+    // https://mathworld.wolfram.com/ConformalLatitude.html)
+    const double taninit = tan(M_PI / 4 + lp.phi / 2);
```
rouault: could be worth moving in the same file as conformat_lat(), and find other users of it in the code base

test/gie/spilhaus.gie
```diff
@@ -0,0 +1,3436 @@
+<gie-strict>
+
```
rouault, line 2: That's a lot of test points. Do we need so much ? Especially given there is hardly any branch in the implementation. Perhaps keep a few ones close to the projection center and ones far away that are challenging for numerical precision ?

test/gie/spilhaus.gie
```diff
@@ -0,0 +1,3436 @@
...
+------------------------------------------------------------
+accept   130.4   -16.2
+expect   3720561.6630   309609.603620
+------------------------------------------------------------
+operation  +proj=spilhaus +k_0=0.9
+tolerance  1 mm
+------------------------------------------------------------
+accept   130.4   -16.2
+expect   3360069.0106   -8388.7716
+
```
rouault, line 421: It would be good having a test for a sphere, to check there's no divide by zero or other similar issue when e = 0.

rouault (commented):
Great work !

src/projections/spilhaus.cpp
jjimenezshaw: Ok. Is there already a file with that kind of things, or better to a dedicated one?

test/gie/spilhaus.gie
jjimenezshaw: During the development I spent a lot of time until I realized that longitudes should be normalized before calling Adams_ws2. It failed only aprox in one eighth of the projection. That made me a bit paranoid about sampling more points.
The first case is every 30 deg. That is not that much... or maybe yes?
The second group is getting every 100th element of the file from ersi (and removing 2 outlayers).
Runtime is 0.02 s (in debug).

src/projections/spilhaus.cpp
rouault: Probably a dedicated one. We have some files phi2.cpp, mlfn.cpp, msfn.cpp, qsfn.cpp, tsfn.cpp with "similar" things but whose naming is apparently meant to be obscure enough that I never know what they do...
Or maybe have a "latitudes.cpp" file where we would group together code to transform between geographic latitude and all those "other latitudes ? But that may be work for another pull request

test/gie/spilhaus.gie
rouault: I don't have a too strong opinion about this. Probably more a personal taste where I tend to keep testing at the strict minimum to reduce their potential future maintenance.

src/projections/spilhaus.cpp
jjimenezshaw: I created `latitudes.cpp`, and move the functions there.

src/projections/spilhaus.cpp
jjimenezshaw: done

test/gie/spilhaus.gie
```diff
@@ -0,0 +1,3456 @@
...
+------------------------------------------------------------
+# Sphere
+------------------------------------------------------------
+operation  +proj=spilhaus +R=6378137
+tolerance  1 mm
+------------------------------------------------------------
+accept   130.4   -16.2
+expect   3737644.5177   -7049.7883
```
rouault, line 429: to test the reverse operation as well;
```suggestion
expect   3737644.5177   -7049.7883
roundtrip 1
```

src/projections/spilhaus.cpp
```diff
@@ -0,0 +1,162 @@
...
+    double lambda_0;
+    double conformal_distortion;
+    double cosrot;
+    double sinrot;
+    PJ *adams_ws2;
+};
+
+} // anonymous namespace
+
+static PJ_XY spilhaus_s_forward(PJ_LP lp, PJ *P) {
```
rouault: the s_ means spherical, so:

```suggestion
static PJ_XY spilhaus_forward(PJ_LP lp, PJ *P) {
```

src/projections/spilhaus.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    PJ_XY xyadams = Q->adams_ws2->fwd(lpadams, Q->adams_ws2);
+    const double factor = Q->conformal_distortion * P->k0;
+    xy.x = -(xyadams.x * Q->cosrot + xyadams.y * Q->sinrot) * factor;
+    xy.y = -(xyadams.x * -Q->sinrot + xyadams.y * Q->cosrot) * factor;
+
+    return xy;
+}
+
+static PJ_LP spilhaus_s_inverse(PJ_XY xy, PJ *P) {
```
rouault:
```suggestion
static PJ_LP spilhaus_inverse(PJ_XY xy, PJ *P) {
```

src/projections/spilhaus.cpp
```diff
@@ -0,0 +1,162 @@
...
+    Q->cosalpha = sqrt(1 - Q->sinalpha * Q->sinalpha);
+    Q->lambda_0 = lon_center + atan2(tan(azimuth), -sin(conformal_lat_center));
+    Q->beta = M_PI + atan2(-sin(azimuth), -tan(conformal_lat_center));
+
+    Q->conformal_distortion =
+        cos(lat_center) / sqrt(1 - P->es * sin(lat_center) * sin(lat_center)) /
+        cos(conformal_lat_center);
+
+    P->fwd = spilhaus_s_forward;
+    P->inv = spilhaus_s_inverse;
```
rouault:
```suggestion
    P->fwd = spilhaus_forward;
    P->inv = spilhaus_inverse;
```

rouault (commented):
-

test/gie/spilhaus.gie
jjimenezshaw: I reduced significantly the sample points. In the generated points, the spacing is bigger. For the samples from ESRI, I picked manually (with the help of a geojson) some points around the projection.

rouault (approved):

PR #4402, opened 2025-02-17
Export ESRI:54099 WGS_1984_Spilhaus_Ocean_Map_in_Square as PROJ string

rouault:
and  fix ESRI:54098 mapping to use +proj=spilhaus with +k_0=sqrt(2)

On top of PR #4401

With it:
~/proj/proj/build_cmake$ echo 41.34837298 -117.28057438 | bin/cs2cs EPSG:4326 ESRI:54099
12421874.60	-14140625.11 0.00

Which matches ESRI values of https://github.com/OSGeo/PROJ/issues/1851#issuecomment-2638222300


src/iso19111/operation/conversion.cpp
```diff
@@ -4126,17 +4126,29 @@ void Conversion::_exportToPROJString(
             "Latitude_Of_Center", common::UnitOfMeasure::DEGREE);
         const double XYPlaneRotation = parameterValueNumeric(
             "XY_Plane_Rotation", common::UnitOfMeasure::DEGREE);
-        if (scaleFactor != 1.0 || azimuth != 0.0 || latitudeOfCenter != 0.0 ||
-            XYPlaneRotation != 0.0) {
+        if (scaleFactor == 1.0 && std::fabs(azimuth - 40.17823482) <= 1e-8 &&
+            std::fabs(longitudeOfCenter - 66.94970198) <= 1e-8 &&
+            std::fabs(latitudeOfCenter - -49.56371678) <= 1e-8 &&
+            std::fabs(XYPlaneRotation - 45) <= 1e-8) {
```
jjimenezshaw: If you prefer, you can set the parameters `+lat_c`, `+lon_c`, `rot` and `azi` to use the spilhaus projection.
I added `rot` to make it more generic, and usable with ESRI Adams_ws2

src/iso19111/operation/conversion.cpp
```diff
@@ -4126,17 +4126,29 @@ void Conversion::_exportToPROJString(
...
+            // ESRI:54099 WGS_1984_Spilhaus_Ocean_Map_in_Square
+            formatter->addStep("spilhaus");
+            // ESRI implementation differs from PROJ one by a sqrt(2) factor
+            // Cf
+            // https://github.com/OSGeo/PROJ/issues/1851#issuecomment-2638222300
+            formatter->addParam("k_0", M_SQRT2);
```
jjimenezshaw: This is probably more generic.
`            formatter->addParam("k_0", M_SQRT2 * scaleFactor);`

jjimenezshaw (commented):
I tried to make `spilhaus` projections flexible enough to match ESRI adams ws2.
The only difference is the scale factor affected by sqrt(2).
But I only had the points reported by @mdsumner in the spilhaus issue #1851
If you set `spilhaus` settings `+lat_c`, `+lon_c`, `rot` and `azi` all to 0, you should have an Adams_ws2

rouault:
> If you set `spilhaus` settings `+lat_c`, `+lon_c`, `rot` and `azi` all to 0, you should have an Adams_ws2

yes I tested that.

We could indeed map ESRI Adams_WS2 as PROJ spillhaus with Scale_factor * sqrt(2) if we get confirmation that there's also a sqrt(2) factor between ESRI:54098 WGS_1984_Adams_Square_II and PROJ's +proj=adams_ws2, hence my request to the community for test points for ESRI:54098 in https://github.com/OSGeo/PROJ/issues/1851#issuecomment-2663842781 . Or maybe @melitakennedy can provide us with some?

rouault:
@kbevers I believe this PR / #4401 should be merged before 9.6.0 is cut, even if they are not "perfect", if we don't get inputs before that. For the record, I won't be available next week.

rouault:
PR updated to be more general and also replicate (almost) ESRI results for ESRI:54098

Thanks to @kgjenkins for https://github.com/kgjenkins/test-coordinates/blob/main/coordinates.csv test points:

$  echo 80 -170 | bin/cs2cs EPSG:4326 ESRI:54098
-9457033.96	13296625.29 0.00

vs ESRI: -9457034.1001	13296625.6969

$ echo 80 -170 | bin/cs2cs EPSG:4326 ESRI:54099
-6624823.23	14253430.38 0.00

vs ESRI: -6624823.16092	14253430.1778

src/iso19111/operation/conversion.cpp
rouault: done

src/iso19111/operation/conversion.cpp
rouault: done

kbevers:
> I believe this PR / https://github.com/OSGeo/PROJ/pull/4401 should be merged before 9.6.0 is cut, even if they are not "perfect",

I agree, but now that the problem has been dealt with we are in an even better position. I expect to put out the release candidate at the beginning of next week.

PR #4406, opened 2025-02-20
Insert a 'wisdom of the day' in our version string

rouault:
In the same vain as https://github.com/OSGeo/gdal/pull/11865 and https://github.com/MapServer/MapServer/pull/7233

### DIFF
--- a/src/release.cpp
+++ b/src/release.cpp
@@ -6,8 +6,10 @@
 #define STR_HELPER(x) #x
 #define STR(x) STR_HELPER(x)
 
-char const pj_release[] = "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(
-    PROJ_VERSION_MINOR) "." STR(PROJ_VERSION_PATCH) ", "
-                                                    "March 1st, 2025";
+char const pj_release[] =
+    "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(PROJ_VERSION_MINOR) "." STR(
+        PROJ_VERSION_PATCH) ", "
+                            "March 1st, 2025. Our mission: \"Diversity, Equity "
+                            "and Inclusion (DEI) for all CRS of the world.\" ";
 
 const char *pj_get_release() { return pj_release; }
--- a/HOWTO-RELEASE.md
+++ b/HOWTO-RELEASE.md
@@ -213,6 +213,7 @@ version numbers in the files.
 - src/proj.h:           Update PROJ_VERSION_MAJOR, PROJ_VERSION_MINOR and
                         PROJ_VERSION_PATCH.
 - src/release.cpp:      Update date to the scheduled release date.
+                        Possibly update the wisdom of the day sentence.
 - CMakeLists.txt:       Update proj_version()
 - docs/source/conf.py:  Update version number
 


src/release.cpp
```diff
@@ -6,8 +6,10 @@
...
 #define STR(x) STR_HELPER(x)
 
-char const pj_release[] = "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(
-    PROJ_VERSION_MINOR) "." STR(PROJ_VERSION_PATCH) ", "
-                                                    "March 1st, 2025";
+char const pj_release[] =
+    "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(PROJ_VERSION_MINOR) "." STR(
+        PROJ_VERSION_PATCH) ", "
+                            "March 1st, 2025. Our mission: \"Diversity, Equity "
+                            "and Inclusion (DEI) of all CRS of the world.\" ";
```
nyalldawson:
```suggestion
                            "and Inclusion (DEI) for all CRS of the world.\" ";
```

rouault: thanks

kbevers:
While I appreciate the sentiment on a personal level, I really do, this is a can of worms I am very hesitant to open. My work on PROJ is done in the capacity as a civil servant in a government agency in a country that has received military threats from the receiver of this protest. According to the code for Danish civil servants I am not to be political in my work - that should be left to the politicians. Of course they are working behind the scenes to resolve this weird and scary situation we find us in now. Waving red flags, however tiny they may be, I don't think is fruitful.

rouault:
> My work on PROJ is done in the capacity as a civil servant in a government agency in a country that has received military threats from the receiver of this protest. According to the code for Danish civil servants I am not to be political in my work - that should be left to the politicians

Fully understood. You don't need to +1 this if you can't.

snowman2:
I agree with @kbevers that we should leave politics out of the code as it tends to be divisive.

rouault:
> I agree with @kbevers that we should leave politics out of the code as it tends to be divisive.

I'm happy with people not happy with that to leave the project as I likely don't share their values.
As developers, we should use the tools we have to express our protest against what is going on, and exercise our little power.

src/release.cpp
```diff
@@ -6,8 +6,10 @@
...
 #define STR(x) STR_HELPER(x)
 
-char const pj_release[] = "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(
-    PROJ_VERSION_MINOR) "." STR(PROJ_VERSION_PATCH) ", "
-                                                    "March 1st, 2025";
+char const pj_release[] =
+    "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(PROJ_VERSION_MINOR) "." STR(
+        PROJ_VERSION_PATCH) ", "
+                            "March 1st, 2025. Our mission: \"Diversity, Equity "
+                            "and Inclusion (DEI) for all CRS of the world.\" ";
```
jayenashar, line 13: is GCJ-02 technically a CRS?  (i don't know if it's supported by PROJ but many projects don't include it)

mdsumner:
I think we need to make sure the list of wisdoms is entirely thoughtful and relevant, overseen by a diverse community, make sure it's not just a time-stamped or single-person's reflections. Totally onboard, it's an excellent addition to daily details about what software just presented at your service.

jayenashar:
> While I appreciate the sentiment on a personal level, I really do

Are you allowed to say that publicly?

rouault:
closing

PR #4412, opened 2025-02-28
Adding Latvia 2000 height transformation

yurish:
Added vertical transformation between EPSG:4949 and EPSG:7700. Grid was added to proj-data repository as well.


rouault:
@yurish I now realize that I've merged this without seeing this was a custom transformation record and something not registered in EPSG. Would you mind contacting EPSG at https://epsg.org/dataset-change-requests.html so they add a record for this transformation. This will minimize long term maintenance on the PROJ side if there's an official record for that transformation. Thanks!

yurish:
@rouault yes, I will contact them shortly

PR #4414, opened 2025-03-03
proj_download_file(): invalidate in-memory caches related to downloaded file (for the current context)

### LINKED ISSUE #4397: Downloading grids during runtime & updating available grids
From: https://github.com/pyproj4/pyproj/issues/1474

Each context has a cache of available grids it generates once and doesn't look for grids after that for performance. However, some programs download grids during runtime. Having a way to update the cache or clear it would help with this use case.

Ideas:
- Add `proj_reset_found_grids` to invalidate cache or for a specific grid.
- Have`proj_download_file` invalidate the cache or for a specific grid.

rouault:
Fixes #4397


snowman2:
Thanks 👍

PR #4416, opened 2025-03-08
Remove hashbang from bash completion.

sebastic:
Completions are meant to be sourced, not executed.

See also: https://lintian.debian.org/tags/bash-completion-with-hashbang.html

### DIFF
--- a/scripts/projinfo-bash-completion.sh
+++ b/scripts/projinfo-bash-completion.sh
@@ -1,3 +1,5 @@
+# Hashbang deliberately missing because this file should be sourced, not executed
+
 function_exists() {
     declare -f -F "$1" > /dev/null
     return $?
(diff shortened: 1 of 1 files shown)


mwtoews:
The first line should be a comment that it needs to be sourced, othwise it would be tempting to re-add the hashbang for a `.sh` file in the scripts dir.

sebastic:
The gdal completion on which this seems to be based doesn't do that either:

 https://github.com/OSGeo/gdal/blob/master/scripts/gdal-bash-completion.sh

kbevers:
@rouault I assume you are okay with this change?

> The first line should be a comment that it needs to be sourced, othwise it would be tempting to re-add the hashbang for a .sh file in the scripts dir.

@mwtoews do you mind adding such a comment?

rouault:
> @rouault I assume you are okay with this change?

yes

rouault:
> do you mind adding such a comment?

I've just done it

PR #4417, opened 2025-03-08
Fix spelling errors.

sebastic:
The lintian QA tool reported spelling errors for the Debian package build:

 * infomation -> information
 * positon    -> position


kbevers:
These are errors in the EPSG dataset ([an example](https://epsg.org/concatenated-operation_8633/Yoff-to-WGS-84-1.html?sessionkey=4l6v4597q2)) and I would prefer they be fixed upstream.

sebastic:
The Debian package already carries these changes as a patch.

The EPSG template for data submission is far too overwhelming to figure out how to translate my patch to that.

I've sent an email to [feedback@epsg.org](mailto:feedback@epsg.org) linking to this issue, that's the best I can do.

kbevers:
> I've sent an email to [feedback@epsg.org](mailto:feedback@epsg.org) linking to this issue, that's the best I can do.

I think that's a very reasonable approach and in my experience they are happy to handle this through the feedback email. Thanks for reporting the errors.

I'll close this on and we'll get the corrections once they are fixed upstream.

sebastic:
FTR:
> In v12.005 of the EPSG Dataset we have made corrections to the records which you identified, except for one record that had been deprecated [EPSG:8646]. This previously-deprecated record was not changed. We have now closed this issue.

PR #4419, opened 2025-03-10
Retry "SSL connection timeout"

pjonsson:
Building the GDAL image can fail with:

Downloading https://cdn.proj.org/de_lgvl_saarland_README.txt... (86 / 439)
Downloading https://cdn.proj.org/de_lgvl_saarland_SeTa2016.tif... (87 / 439)
Cannot open https://cdn.proj.org/de_lgvl_saarland_SeTa2016.tif: SSL connection timeout
Cannot download https://cdn.proj.org/de_lgvl_saarland_SeTa2016.tif

so add SSL connection timeout
to the list of things that get
a retry time.

Sibling of commit 72307126.

- [ ] Tests added

### DIFF
--- a/src/networkfilemanager.cpp
+++ b/src/networkfilemanager.cpp
@@ -1696,7 +1696,8 @@ static double GetNewRetryDelay(int response_code, double dfOldDelay,
         (response_code == 400 && pszErrBuf &&
          strstr(pszErrBuf, "RequestTimeout")) ||
         (pszCurlError && strstr(pszCurlError, "Connection reset by peer")) ||
-        (pszCurlError && strstr(pszCurlError, "Connection timed out"))) {
+        (pszCurlError && strstr(pszCurlError, "Connection timed out")) ||
+        (pszCurlError && strstr(pszCurlError, "SSL connection timeout"))) {
         // Use an exponential backoff factor of 2 plus some random jitter
         // We don't care about cryptographic quality randomness, hence:
         // coverity[dont_call]


hobu:
Is the SSL handshake actually timing out or do you have a MITM or proxy that's timing out? Have the TIFFs from the CDN ever been having SSL handshake time failures?

pjonsson:
I'm not sure what you are asking with the first question. If I state that I don't have a proxy or anything similar in the way, does that answer your question?

I'm also not sure if there has ever been a SSL handshake time failure for me since failures might have been silently retried without any notification. It's the first time I see this error in 1000+ builds of GDAL, but I have been busy with other things the last couple of months so it's possible I missed the error during that time.

hobu:
> I'm not sure what you are asking with the first question. 

I don't understand why this patch is needed, the ticket doesn't clearly describe the scenario in which it is needed, and this architecture has been structured this way for 5+ releases and there has been no need for it. 

It does not seem such a wise thing to be retrying SSL connections, especially when we're not given any root description of why it is needed.

pjonsson:
> the ticket doesn't clearly describe the scenario in which it is needed,

I'm building the full Ubuntu-based GDAL image in the same way as the GDAL CI:
https://github.com/OSGeo/gdal/blob/25fd4e023ee7bc1b2903cf226567818ec8284c20/.github/workflows/docker.yml#L67

> and this architecture has been structured this way for 5+ releases and there has been no need for it.
>
> It does not seem such a wise thing to be retrying SSL connections, especially when we're not given any root description of why it is needed.

I'm not sure why I got a SSL timeout after building GDAL 1000+ times. I also don't know why I intermittently got build failures caused by connection reset by peer a couple of times per month before #4106 landed.

I have to admit I'm a bit confused, are you arguing that all the retry-code in PROJ should be removed, or that SSL timeouts are special and should cause a build failure?

PR #4421, opened 2025-03-10
[Backport 9.6] Remove hashbang from bash completion.

rouault:
Backport #4416
 **Authored by:** @sebastic


kbevers:
I'll prepare RC2 tomorrow with this and the SSL patch included.

PR #4422, opened 2025-03-12
Build: make sure that BUILD_SHARED_LIBS is initialized before defining DEFAULT_EMBED_RESOURCE_FILES

rouault:
otherwise DEFAULT_EMBED_RESOURCE_FILES will be set to ON

@kbevers This is kind of a blocker, so a RC3 seems to be needed. Sorry for the inconvenience

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,6 +156,13 @@ include(ProjConfig)
 include(ProjMac)
 include(policies)
 
+##############################################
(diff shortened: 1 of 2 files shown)


kbevers:
> @kbevers This is kind of a blocker, so a RC3 seems to be needed. Sorry for the inconvenience

No worries. This is why we do the release candidates. I'll push it out tomorrow afternoon.

PR #4443, opened 2025-03-22
Add missing 'limits' and 'algorithm' includes

### LINKED ISSUE #4442: Missing `#include <limits>` in v9.6.0
Apologies for terseness but I think that this issue is so trivial that it requires a few words to explain.

### Problem

Proj 9.6.0 build fails due to missing `std::numeric_limits` template. 

### Solution

Include `limits` header file (https://en.cppreference.com/w/cpp/types/numeric_limits).

### Patch
```
--- a/src/crs_to_crs.cpp
+++ b/src/crs_to_crs.cpp
@@ -33,6 +33,7 @@
 #include <math.h>
 
 #include <algorithm>
+#include <limits>
 
 #include "proj/internal/internal.hpp"
 
--- a/src/trans.cpp
+++ b/src/trans.cpp
@@ -33,6 +33,8 @@
 #include "proj_internal.h"
 #include <math.h>
 
+#include <limits>
+
 #include "proj/internal/io_internal.hpp"
 
 inline bool pj_coord_has_nans(PJ_COORD coo) {
```

rouault:
and add CI test to check for that (borrowed from GDAL)

Fixes #4442

### DIFF
--- a/src/trans.cpp
+++ b/src/trans.cpp
@@ -33,6 +33,9 @@
 #include "proj_internal.h"
 #include <math.h>
 
+#include <cmath>
+#include <limits>
+
 #include "proj/internal/io_internal.hpp"
 
 inline bool pj_coord_has_nans(PJ_COORD coo) {
--- a/.github/workflows/code_checks.yml
+++ b/.github/workflows/code_checks.yml
@@ -65,6 +65,15 @@ jobs:
           # Do not fail the job. This is just used as a tool to monitor how we are regarding recent cppcheck
           ./scripts/cppcheck.sh || /bin/true
 
+  code_quality_checks:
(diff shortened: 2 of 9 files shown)


src/trans.cpp
```diff
@@ -33,6 +33,9 @@
 #include "proj_internal.h"
 #include <math.h>
 
+#include <cmath>
```
jjimenezshaw, line 36: [minor] There is an include <math.h> a few lines above. Is there any difference?

rouault: yes, the scripts check that if std::isnan / isinf / isfinie is used, cmath is included.  Including math.h itself will not bring those symbols. I suspect cmath is already included indirectly here.

PR #4446, opened 2025-03-23
Implement uniform conversions between auxiliary latitudes

cffk:
This follows on the discussion in #4441.  In particular...

latitudes.cpp now offers methods for converting between auxiliary laitudes using trigonometric series whose coefficients are power series in n.  The series are evaluated using Clenshaw summation and Horner's method.  For small flattening, these often offer better accuracy than the analytical expressions.

Now mlfn.cpp and tmerc.cpp use these conversions, simplifying the implementations in these files.  In particular tmerc's functions gatg and clens (which puzzlingly seemed to do exactly the same thing!) have been removed.  tmerc's clenS (with a capital "S") stays -- it does complex Clenshaw summation

Along the way, I've simplified pj_conformal_lat and pj_conform_lat_inverse in latitudes.cpp.  For the former, I've substituted a better conditioned formula; for the latter, I used the existing (and more efficent) approach implemented in pj_sinhpsi2tanphi.

The computation of authalic latitudes in latitudes.cpp now uses the series method if the flattening is small.  However some of the test cases of the healpix projection use large flattening (is this really necessary?); so I've had to retain the poorly conditioned analytical formulas as well.

If we really wanted to support large flattening in this case, it's necessary to evaluate the authalic latitude using Eqs. (52), (53), (54) of "On auxiliary latitudes" (https://arxiv.org/abs/2212.05818) and using Newton's method on the tangents of the angles for the inverse, using Eq. (57).  I don't see the need for these at present.

I've simplified the interface somewhat: passing P instead of P->e, P->onees, etc. separately; dropping "_coeffs" off the q function name and the tagging of the inverse function name with "_exact".  I had hoped to retire the external use of Snyder's q(sinphi).  However, some equal area projections, e.g., the cylindrical version, really do need this function instead of the authalic latitude directly.


cffk:
There is no hurry on this -- it doesn't fix any bugs.  However, it removes code duplication and makes some improvements in accuracy.

jjimenezshaw:
> However some of the test cases of the spillhaus projection use large flattening (is this really necessary?)

Are we changing the flattening in the spilhaus tests? I am not aware of that. We use either the normal ellipsoid or a sphere.

cffk:
> > However some of the test cases of the spillhaus projection use large flattening (is this really necessary?)
> 
> Are we changing the flattening in the spilhaus tests? I am not aware of that. We use either the normal ellipsoid or a sphere.

I misspoke.  It was healpix and not spillhaus that used large flattening in its tests.  In any case, I changed no tests, and they all still pass.

src/latitudes.cpp
```diff
@@ -10,72 +10,52 @@
...
-    const double esinphi = e * sin(phi);
-    const double chi =
-        2 * atan(tan(M_FORTPI + phi / 2) *
-                 std::pow(((1 - esinphi) / (1 + esinphi)), (e / 2))) -
-        M_HALFPI;
-    return chi;
+    // Instead of calling tan and sin, call sin and cos which the compiler
+    // optimizes to a single call to sincos.
+    double sphi = sin(phi), cphi = cos(phi);
+    return atan(sinh(asinh(sphi / cphi) - P->e * atanh(P->e * sphi)));
```
jjimenezshaw, line 32: Isn't if throwing for phi = 90 deg?
In Mercator projection it is not defined. But in spilhaus it is (near the center of the projection). Maybe se should add a test case in spilhaus for latitudes -90 and 90 if it is not already there.

cffk: There's no error.  cos(pi/2) evaluates in double precision is a small positive quantity, so the returned value will be very close to (or exactly) pi/2 (rounded to a double).  Even if cos(pi/2) is zero, we would have sinh(asinh(inf)) -> inf, and atan(inf) = pi/2.

The only issue is if the precision is changed.  With long double (64 bit precision), cos(pi/2) is negative and then we would get -pi/2 for the result.  I suppose we could guard against this with fabs(cphi); but so many other things would need to be tweaked in PROJ, I don't think this is warranted.  N.B. with quad precision (113 bits), cos(pi/2) is positve once again.

rouault (approved):

src/latitudes.cpp
```diff
@@ -204,3 +152,211 @@ double pj_authalic_lat_inverse_exact(double beta, const double *APA,
...
     return phi;
 }
+#undef PROJ_AUTHALIC_SERIES_VALID
+
+// Compute the coefficients of the trigonometric series for conversions between
+// auxiliary latitudes.  The coefficients are produced by the Maxima code
+// auxlat.mac bundled with GeographicLib.  Load this file into Maxima and run
+// writecppproj()$ -- this produces the 'static const' declarations below.
+// Only a subset of conversions is provided.  To add others, include them in
+// the list "required" in writecppproj().
```
kbevers: This comment confuses me. Specifically the "Load this file ... " part. What is "this" referring to? I get the feeling that this might be copied from somewhere else?

src/latitudes.cpp
```diff
@@ -204,3 +152,211 @@ double pj_authalic_lat_inverse_exact(double beta, const double *APA,
...
+        }
+    } else {
+        for (int l = 0; l < Lmax; ++l) {
+            int m = (Lmax - l - 1); // order of polynomial in n
+            c[l] = d * pj_polyval(n, coeffs + o, m);
+            o += m + 1;
+            d *= n;
+        }
+    }
+    // assert (o == ptrs[k+1]);
```
kbevers, line 370: Consider removing this

src/proj_internal.h
```diff
@@ -453,6 +453,27 @@ enum class TMercAlgo {
...
+  PHI = GEOGRAPHIC,
+  BETA = PARAMETRIC,
+  THETA = GEOCENTRIC,
+  MU = RECTIFYING,
+  CHI = CONFORMAL,
+  XI = AUTHALIC,
+  COMMON = GEOGRAPHIC,
+  GEODETIC = GEOGRAPHIC,
+  REDUCED = PARAMETRIC,
+  AUXORDER   = 6,
```
kbevers: Doesn't do a whole lot, but follows the pattern from lines above:
```suggestion
  AUXORDER = AUXNUMBER,
```

src/projections/tmerc.cpp
```diff
@@ -488,109 +446,33 @@ static PJ *setup_exact(PJ *P) {
...
+    static_assert( PROJ_ETMERC_ORDER == int(AuxLat::AUXORDER),
+                   "Inconsistent orders etmerc vs auxorder" );
     /* third flattening */
     const double n = P->n;
-    double np = n;
 
     /* COEF. OF TRIG SERIES GEO <-> GAUSS */
     /* cgb := Gaussian -> Geodetic, KW p190 - 191 (61) - (62) */
     /* cbg := Geodetic -> Gaussian, KW p186 - 187 (51) - (52) */
     /* PROJ_ETMERC_ORDER = 6th degree : Engsager and Poder: ICC2007 */
```
kbevers, line 462: Are these lines still relevant here? Seeing as the lines below are removed in favour of a centralised function I'm wondering if keeping the comment here will be confusing for someone down the line.

kbevers (commented):
The math is generally beyond me, so I've mostly commented on stuff that seemed odd to me. Overall I'm happy with this pull request.

src/latitudes.cpp
cffk: I've now provided detail instructions on how to regenerate the array of coefficients.

src/latitudes.cpp
cffk: You mean the assert?  This provided useful information for the next developer to understand how the ptrs array is used.  Let's keep.

src/proj_internal.h
cffk: I've clarified this.  (For the record AUXORDER and AUXNUMBER are distinct quantities only accidently are they equal.)

src/projections/tmerc.cpp
cffk: It's important to keep these comments since they correspond to the terms in E&P (2007).  I've inserted a key given the correspondence between these and the names of the auxiliary latitudes.

PR #4449, opened 2025-03-29
Bug correction in Wagner VI projection parameters

jgpallero:
Wagner VI projection parameters Cx and Cy must have both the value 1.0 according to the projection definition. See
Evenden (2005). libproj4: A Comprehensive Library of Cartographic Projection Functions (Preliminary Draft), p. 55, where the parameters for Wagner VI are in accordance with https://en.wikipedia.org/wiki/Wagner_VI_projection and https://www.boehmwanderkarten.de/kartographie/is_netze_wagner_123456789_inversions.html


rouault:
There is a bit of confusion in that area. Current coefficients Cx = Cy = 0.94745 come from the very initial PROJ.4 release (https://github.com/OSGeo/PROJ/blob/proj_4_4_3/src/PJ_eck3.c)  . They seem to come from https://download.osgeo.org/proj/proj.4.3.I2.pdf page 10 , except that in that pdf those coefficients are supposed to be the ones of Putnin.˘s P1, so the PDF is in error.
The coefficients Cx = Cy = 1 are the ones indicated at page 205 of Snyder's Two Thousand Years of Map Projections (https://books.google.fr/books?id=0UzjTJ4w9yEC&pg=PA205&redir_esc=y#v=onepage&q&f=false)
And more interestingly Gerald Evenden's libproj4 , which is a fork/continuation of PROJ.4, also use them : cf https://github.com/rouault/libproj4/blob/master/libproject-1.01/src/proj_eck3.c#L37
So I concur this is the right fix.
And those 0.94745 coefficients according to libproj4 are not the one of  Putnin.˘s P1 but the variant  Putnin.˘s P1'  (not the final apostroph), hence the confusion in the proj.4.3.I2.pdf

rouault:
@jgpallero why closing the pull request ?

jjimenezshaw:
> @jgpallero why closing the pull request ?

by accident I think.

rouault:
re-opening

rouault:
actually I can't as the commit has been removed...

PR #4495, opened 2025-05-15
Database: update to EPSG v12.009

rouault:


rouault:
FYI @jjimenezshaw  this adds 'ETRS89/DREF91/2016 to GNTRANS2016 height (1)'

PR #4499, opened 2025-05-21
Add grid alternative for lv_lgia_lks92to2020.tif

kbevers:
In combination with https://github.com/OSGeo/PROJ-data/pull/140


jjimenezshaw (approved):

rouault:
backported to 9.6 per dde234c0fb   (the automated backport  wouldn't work as there's a  difference in the content of the database between master and 9.6 branch, and thus the expected MD5 isn't the same between both branches)

PR #4500, opened 2025-05-21
Add an option to ship PDB files

ag-pix4d:
A new CMake option `EXPORT_PDB` is added, when enabled, on Windows, a PDB file will be generated. Default value for `EXPORT_PDB` is `OFF`.


rouault:
@ag-pix4d  can you set this option in .github/workflows/windows.yml ? (at line 78)

ag-pix4d:
> @ag-pix4d can you set this option in .github/workflows/windows.yml ? (at line 78)

@rouault sure, done

PR #4519, opened 2025-06-03
Database: change ESRI deprecated records back to their deprecated state

rouault:
Fixes https://github.com/OSGeo/PROJ-CRS-Explorer/pull/36#issuecomment-2935170928

During the ArcGIS Pro 3.5 database import, the import script wrongly considered the new value 'codechange' as meaning not deprecated, which resulted in deprecated CRS no longer being flagged as such.


jjimenezshaw (commented):
What does it actually mean `codechange`?

This change in the database is reason for a bugfix release? (just asking)

rouault:
> What does it actually mean `codechange`?

My intuition is that they deprecate an old ESRI code in favor of the regular EPSG one.

> This change in the database is reason for a bugfix release? 

I'm not sure if the impact is that strong

jjimenezshaw:
> > This change in the database is reason for a bugfix release?
> 
> I'm not sure if the impact is that strong

I can think on this use case: If any application is showing the list of (ESRI) available CRSs, removing the deprecated ones, with PROJ 9.6.1 they will have a "few more" (hundreds?) in the combo box.
I don't know how many people is doing that with ESRI.

@kbevers any opinion?

kbevers:
> @kbevers any opinion?

Well... I guess the list of CRS's in QGIS and similar apps will be longer and somewhat incorrect. It's not a problem for me personally and preparing a new release is not a priority at this time. I'm not against a quick 9.6.2 release if someone else takes the lead.

rouault:
> I'm not against a quick 9.6.2 release if someone else takes the lead.

I'll tackle a 9.6.2 release

melitakennedy:
> > What does it actually mean `codechange`?
> 
> My intuition is that they deprecate an old ESRI code in favor of the regular EPSG one.

A rarer case is when EPSG has done a deprecate/replace but Esri already had the correct definition, or the change is something that Esri doesn't support anyway. 

(sorry, made 2 edits to fix my original comment made when distracted)

PR #4539, opened 2025-07-03
Fix tolower()/toupper() implementation to not lead to undefined behavior

### LINKED ISSUE #4537: proj can invoke ctype(3) functions with invalid input, leading to UB
#### Problem description

As background, ctype(3) requires that arguments to `tolower` etc., which are of type `int`, have a value which is `EOF` or which can be represented as an unsigned char.  Otherwise, the call is UB.  It is common to use `char`, which is signed on many CPU types, including x86.  If the value, viewed as unsigned, is within 128..255, then it is negative, and UB results.

On NetBSD current (to be 11), the ctype functions check for conditions specified as UB and aborts.

Building proj 9.6.2 and running tests results in crashes in several tests.   The same code/tests pass on NetBSD 10.

Further, building qgis crashes in crssync.

#### Environment Information

- PROJ 9.6.2
- NetBSD-current amd64

#### Installation method

- pkgsrc

#### Fix

Audit the sources for all uses of ctype(3) functions, and ensure that the argument is of type unsigned char, or that there is some other guarantee that all values will be within the defined range.

The following changes do not entirely resolve the test issues, but do make qgis build.   It would be nice for someone who is more skilled at C++ to opine on the proper fix.

[proj-ctype-patch.txt](https://github.com/user-attachments/files/21043290/proj-ctype-patch.txt)

rouault:
Fixes #4537

### DIFF
--- a/src/iso19111/internal.cpp
+++ b/src/iso19111/internal.cpp
@@ -129,8 +129,9 @@ std::string tolower(const std::string &str)
 
 {
     std::string ret(str);
-    for (size_t i = 0; i < ret.size(); i++)
-        ret[i] = static_cast<char>(::tolower(ret[i]));
+    for (char &ch : ret)
+        ch =
+            (ch >= 'A' && ch <= 'Z') ? static_cast<char>(ch + ('a' - 'A')) : ch;
     return ret;
 }
 
@@ -144,8 +145,9 @@ std::string toupper(const std::string &str)
 
 {
     std::string ret(str);
-    for (size_t i = 0; i < ret.size(); i++)
-        ret[i] = static_cast<char>(::toupper(ret[i]));
+    for (char &ch : ret)
+        ch =
+            (ch >= 'a' && ch <= 'z') ? static_cast<char>(ch - ('a' - 'A')) : ch;
     return ret;
 }
 
--- a/src/dmstor.cpp
+++ b/src/dmstor.cpp
@@ -46,7 +46,9 @@ double dmstor_ctx(PJ_CONTEXT *ctx, const char *is, char **rs) {
      * It is possible that a really odd input (like lots of leading zeros)
      * could be truncated in copying into work.  But ...
      */
-    while ((isgraph(*p) || *p == DEG_SIGN1 || *p == DEG_SIGN2) && --n)
+    while ((isgraph(static_cast<unsigned char>(*p)) || *p == DEG_SIGN1 ||
+            *p == DEG_SIGN2) &&
+           --n)
         *s++ = *p++;
     *s = '\0';
     int sign = *(s = work);


src/iso19111/internal.cpp
```diff
@@ -129,8 +129,9 @@ std::string tolower(const std::string &str)
 
 {
     std::string ret(str);
-    for (size_t i = 0; i < ret.size(); i++)
-        ret[i] = static_cast<char>(::tolower(ret[i]));
+    for (char &ch : ret)
+        ch =
+            (ch >= 'A' && ch <= 'Z') ? static_cast<char>(ch + ('a' - 'A')) : ch;
```
gdt, line 134: Will likely fail with EBCDIC but probably that's ok :-)

But more seriously, I wonder why the cast to unsigned char is bad.  It allows ctype(3) implementations that use lookup tables and are thus fast, in general.   (I get it that this isn't used much and that its performance isn't important.)

I don't object, though, and will test this.

rouault: > that use lookup tables and are thus fast,

I'm quite skeptical that lookup based code would be faster than branch testing. Modern processors are much better at doing branch prediction than fetching bytes from slow memory. Anyway this implementation or the alternate can only deal with ASCII, and that's fine given its intended purpose.

gdt:
With this PR on top of 9.6.2, proj has one test failure (vs 2), and qgis builds.

The remaining test issue is
```
Command: "/tmp/work/geography/proj/work/proj-9.6.2/cmake-pkgsrc-build/bin/gie_self_tests"
Directory: /tmp/work/geography/proj/work/proj-9.6.2/cmake-pkgsrc-build/test/unit
"gie_self_tests" start time: Jul 03 18:08 EDT
Output:
----------------------------------------------------------
[==========] Running 18 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 15 tests from gie
[ RUN      ] gie.cart_selftest
utm: Invalid latitude
utm: Invalid latitude
utm: Invalid latitude
[       OK ] gie.cart_selftest (9 ms)
[ RUN      ] gie.info_functions
Unrecognized vertical grid format for filename 'proj.db'
Unrecognized horizontal grid format for filename 'proj.db'
ctype(3) isgraph: invalid input: -62
<end of output>
Test time =   0.02 sec
----------------------------------------------------------
Test Failed.
"gie_self_tests" end time: Jul 03 18:08 EDT
"gie_self_tests" time elapsed: 00:00:00
----------------------------------------------------------
```

which looks to be in `src/dmstor.cpp:dmstor_ctx`.

Adding a cast to unsigned char gets me a clean pass.
```
--- src/dmstor.cpp.orig 2025-04-01 21:34:48.000000000 +0000
+++ src/dmstor.cpp
@@ -46,7 +46,7 @@ double dmstor_ctx(PJ_CONTEXT *ctx, const
      * It is possible that a really odd input (like lots of leading zeros)
      * could be truncated in copying into work.  But ...
      */
-    while ((isgraph(*p) || *p == DEG_SIGN1 || *p == DEG_SIGN2) && --n)
+    while ((isgraph((unsigned char) *p) || *p == DEG_SIGN1 || *p == DEG_SIGN2) && --n)
         *s++ = *p++;
     *s = '\0';
     int sign = *(s = work);
```

jjimenezshaw:
Interesting comment in the "notes" in https://en.cppreference.com/w/cpp/string/byte/isgraph

rouault:
isgraph() use now fixed

jjimenezshaw (approved):

PR #4540, opened 2025-07-05
Database: update to EPSG 12.014

rouault:
The main change is the improved metadata accuracy declaration for "WGS 84 to EGM2008 height (1)" from 1 m to 0.113 m.

And also the mention to cdn.proj.org GTG grids for EGM96 and EGM2008 :-)


jjimenezshaw:
Do you know who suggested referencing cdn.proj.org?
btw, "The grid file is available in GTG format from https://cdn.proj.org/us_nga_egm08_1.tif/" ... that file does not exist, and it would be huge.
(in case anybody needs that data, there is a backup in https://github.com/jjimenezshaw/EGM2008_data and a link to read it with python)

rouault:
> Do you know who suggested referencing cdn.proj.org?

yeah I've some idea...  I've already reached to IOGP about that (and the more mundane issue that the URL terminating with a slash are not valid)

rouault:
extract of answer from IOGP: "For all 6 records, we will remove the remarks regarding availability from Proj. There are many grids available from Proj/GDAL and we mostly do not make reference to their availability."  .

In any case, that's just comments, so the only bonus of the 12.014 is the updated accuracy for EGM08

jjimenezshaw (approved):

PR #4544, opened 2025-07-11
Database: update to EPSG v12.015

rouault:
also update consistency checks and code to make sure we deal properly with new Helmert transformation EPSG:10905 "ETRS89/DREF91/2016 to Asse 2025 + Asse 2025 height (1)" whose source CRS is a geog3D CRS but target is a compound CRS...


jjimenezshaw (commented):
in data/sql/unit_of_measure.sql there are many units with 3 less decimal numbers. Is that from EPSG database, or a script change?

rouault:
> Is that from EPSG database, or a script change?

yes, from EPSG dataset. They still have 18 significant digits, which is one more than what is needed to guarantee preserving a 64-bit floating point value...

PR #4546, opened 2025-07-14
Use correct path alias syntax on windows

SunBlack:
When reading through the documentation, I saw that you have to adjust the path alias manually before you can use it.

### DIFF
--- a/docs/source/resource_files.rst
+++ b/docs/source/resource_files.rst
@@ -39,7 +39,7 @@ The following paths are checked in order:
 
   The PROJ user writable directory, which is :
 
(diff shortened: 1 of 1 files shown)


docs/source/resource_files.rst
```diff
@@ -39,7 +39,7 @@ The following paths are checked in order:
 
   The PROJ user writable directory, which is :
 
-    * on Windows, ``${LOCALAPPDATA}/proj``
+    * on Windows, ``%LOCALAPPDATA%/proj``
```
kbevers: May as well go all in then:

```suggestion
    * on Windows, ``%LOCALAPPDATA%\proj``
```

PR #4560, opened 2025-08-19
BUG-fix: geodesic inverse for meridional points on prolate ellipsoid

cffk:
This fixes a bug in the inverse geodesic routine for extreme prolate ellipsoids (f < -2 or b > 3*a).  The utility routine invgeod does not allow you to specify prolate ellipsoid (an overall gripe I have with PROJ).  However, the API does let you set f < 0.  Of course, the results you get are still approximate, since f is outside the recommended range f < |1/50|.

This syncs with version 2.2 of geographiclib-c.  This includes a couple of workarounds for compiler issues:

(1) clang -Wkeyword-macro triggered a warning with nullptr being defined.  Switch to NULLPTR and define this to be nullptr, NULL, or 0.

(2) "error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]".  Fix by changing the declaration "int f()" to "int f(void)".

Finally, update geod.rst to replace +ellps=<ellipse> by +ellps=<ellispoid>.  Presumably the man page will catch up shortly.

- [x] Tests added


cffk:
For the record, the bug is exhibited with ellipsoid `a = 1e6`, `f = -3` (i.e., `b = 4e6`) calculating the distance between `30N 0E` and `30E 180E`.  The old code gave
```
direction = 0 deg
distance  = 1077696 m
```
The new code gives
```
direction = 21.63068 deg
distance  = 1074660 m (i.e., shorter by 3 km)
```
This "right" answer (calculated with elliptic functions and not
supported by PROJ) is
```
direction = 22.36881 deg
distance  = 1074082 m (shorter by another 0.5 km)
```
so the new code is still in error.  But the direction is now plausible.

PR #4565, opened 2025-09-07
Build arm64 docker images with attestation

j9ac9k:
Update the CI so that the docker images use provenance attestations for the github container repository.  In addition, have the CI generate arm64 images of proj; make both images available through a manifest.

Happy to explain any of the changes.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,6 +4,7 @@ on:
   push:
       paths-ignore:
           - 'docs/**'
+      branches-ignore: ["dependabot/**"]
   pull_request:
       paths-ignore:
           - 'docs/**'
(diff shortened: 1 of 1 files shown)


kbevers:
> Happy to explain any of the changes.

Please. Start with the why.

j9ac9k:
> > Happy to explain any of the changes.
> 
> Please. Start with the why.

Because @hobu asked me to to start with :)

Having native arm64 is beneficial for a variety of reasons, and using manifests the "correct" platform image will be fetched when doing a `docker pull osgeo/proj:latest`.  

Regarding the attestation, that's a feature that github container registry offers that allows for auditing of the docker images, but it's mostly implemented here because it uses minimal extra complexity in the CI script to get that feature (which maybe some agencies may require in the future if they don't already).

rouault:
At first sight, this looks reasonable to me. But I see in your fork that there are some issues: https://github.com/j9ac9k/PROJ/actions/runs/17530060616/job/49785809104  . Ideally you would also slighly modify it to test the push parts in your fork

j9ac9k:
> At first sight, this looks reasonable to me. But I see in your fork that there are some issues: https://github.com/j9ac9k/PROJ/actions/runs/17530060616/job/49785809104 . Ideally you would also slighly modify it to test the push parts in your fork

My bad on that; looks like I'm missing some brackets.  I'll sort it out as much as I can on my fork, and then update this branch so I don't waste your CI cycles unnecessarily.

hobu:
Regarding the attestation stuff: PROJ and GDAL are low level core libraries that often have complex system built upon them. [Attestations](https://github.blog/news-insights/product-news/introducing-artifact-attestations-now-in-public-beta/) allow the project to authenticate the artifacts we're putting out there and let our users verify they are indeed coming from the project. While nothing significant *requires* attestation and certificate infrastructure at this time, it is likely that governments and larger institutions may start requiring artifact lineage information before software can be installed.

The GitHub attestation APIs and GitHub Actions make this easy enough to add. BTW, we should also add add attestation to our release tarballs in addition to the Docker images.

kbevers:
> BTW, we should also add add attestation to our release tarballs in addition to the Docker images.

I'll leave that for you to take care of :-)

PR #4589, opened 2025-10-18
Database: update to EPSG 12.033

rouault:
A few hacks have had to be made. Below email sent to epsg.org

```
PROJ consistency checks have detected the following issues with the 12.033 update (or changes done after 12.029):

- Transformation 11064 "ETRS89-BEL [BEREF2002] to ETRF2000 (1) " lacks the unit of measure for the 3 X/Y/Z-translation terms.

- Transformation 10996 "ETRS89-GBR [OSNet v2009] to Xrail84 (1)" uses EPSG:11008, a geographic 3D CRS, as source CRS, whereas its target CRS is EPSG:10991, a geocentric one. Given the transformation method is "Coordinate Frame rotation (geocentric domain)", the source should likely be EPSG:11007

- Similarly, transformation 11050 " ETRS89-ALB [KRGJSH 2010] to WGS 84 (1) " uses EPSG:11047, a geographic 2D CRS as source CRS, and EPSG:4326 as target CRS, but with "Geocentric translations (geocentric domain)Open". It should likely be either source_crs = EPSG:11045 and target_crs = EPSG:4978, or the operation method should be changed to "Geocentric translations (geog2D domain)".

- Transformation 11185 "ETRS89-BGR [BGS2005] to ETRF2000 (1)" is between 2 geocentric CRS, but the operation method is "Geocentric translations (geog2D domain)". I presume it should be "Geocentric translations (geocentric domain)" / 1031 instead

Independently of the above, I've some concerns regarding backwards compatibility of creating all those specialized CRS  ETRS89-xxx and re-assigning past transformations between national CRS and ETRS89 to be between that national CRS and ETRS89-xxx. Such changes can be seen as regressions from users used to use generic ETRS89.

For example for Romania, people doing EPSG:4179 (Pulkovo 1942(58)) to EPSG:4258 (ETRS89) previously got EPSG:15994 as the most precise result. But now EPSG:15994 is no longer between those 2 CRS but beween EPSG:4179 and EPSG:11119 ("ETRS89-ROU [ETRF2000]"), and now they get only EPSG:15993 that was superseded per EPSG:15994.  My understanding was that existing objects in the database could not be changed in significant ways. I see such change of target CRS to be a significant change. Wouldn't have it been better to copy EPSG:15994 to a new object with the new target CRS?

Furthermore there are no transformation records between generic ETRS89 and ETRS89-XXX (that should presumably be null transformations with some centimeric accuracy), which makes it difficult to "bridge" records.
```

Given above concerns wondering how much appropriate it is to backport those changes in the 9.7 branch


test/unit/gie_self_tests.cpp
```diff
@@ -1018,7 +1018,7 @@ TEST(gie, horner_only_fwd_selftest) {
 
 // ---------------------------------------------------------------------------
 
-TEST(gie, proj_create_crs_to_crs_PULKOVO42_ETRS89) {
+TEST(gie, proj_create_crs_to_crs_PULKOVO42_ETRS89_ROU) {
     auto P = proj_create_crs_to_crs(PJ_DEFAULT_CTX, "EPSG:4179", "EPSG:4258",
```
jjimenezshaw: ETRS89_ROU is still EPSG:4258 ?

rouault: > ETRS89_ROU is still EPSG:4258 ?

nope. Restoring the initial name of that function  (I initially changed the target CRS to be ETRS89_ROU, and then reverted to initial EPSG:4258, but forgot to revert the test name). Fixed

kbevers:
> Given above concerns wondering how much appropriate it is to backport those changes in the 9.7 branch

I agree. I suppose we might see further changes upstream that might make it more suitable for a backport.

I am curious to see the answer to your email. This topic has been discussed fairly extensively in the European geodetic community. The idea is to make life simpler in the long run by naming individual national realisations of ERTS89 and including them in the ETRS89 datum ensemble. 

I've raised similar concerns as you some time ago but was reassured that since the new ETRS89-xxx's would become part of the ETRS89 datum ensemble these problems shouldn't appear. For now that isn't the case, at least. I can't really tell from the SQL if they've been added to the ensemble or not. If they have, perhaps we need to tweak how we find connections between CRS's to also take members of datum ensembles into account? (no idea if this is already happening).

It would probably be good if we could get release candidates of the EPSG dataset...

rouault:
>  I can't really tell from the SQL if they've been added to the ensemble or not.

They have not. The ensemble only lists ETRF89 to 2020 as datum ensemble members.

 > If they have, perhaps we need to tweak how we find connections between CRS's to also take members of datum ensembles into account? (no idea if this is already happening).

We'll definitely need to tweak the DB by creating null transformations between each of the member and the datum ensemble itself

kbevers:
> They have not. The ensemble only lists ETRF89 to 2020 as datum ensemble members.

Okay. There's a chance that is coming in a later update, perhaps as a new separate ensemble that doesn't include the ETRF's. But we can't really work on that assumption.

> We'll definitely need to tweak the DB by creating null transformations between each of the member and the datum ensemble itself

Wouldn't it be better to deal with this upstream? If so, we might be able to convince Roger et al to do that. This dataset is becoming quite complex and I don't think the IOPG has the means to uncover all the pitfalls before releasing a new version, so if we can help guide them a bit that would be for everyones benefit.

rouault:
Feedback from R. Lott:
"""
We will correct transformations 11064, 10996, 11050 and 11185 as quickly as possible. [...]
 
Regarding your concerns about backward compatibility and the Romanian example, there is no right answer here. We cannot replace all earlier transformations to the ETRS89 ensemble (CRS 4258) with a new transformation to the modern nationalization because they often pre-date the current national realization and in several cases the national authority have told us that the transformation should not apply to the latest national realization and should instead be left as to the ETRS89 ensemble. We have therefore updated only the latest transformation to have the explicit national realization as the target CRS. In the particular case of Romania, 15993 probably does predate ETRS89-ROU [ETRF2000] and so, like 15994, could have been changed to have ETRS89-ROU [ETRF2000] rather than ETRS89 ensemble as the target CRS. But that will not be the case for all countries.
 
As Javier has noted, our plan is to include all of these national realizations into a very much enlarged ETRS89 datum ensemble (datum code 6258). Then, at the datum ensemble accuracy (for ETRS89, 0.1m), all of these national realizations and the ensemble itself can be considered to be insignificantly different. This is the bridge between ETRS89 and the national realizations. The enlarged ensemble also can act as a mechanism for merging at lower accuracy data referenced to different national realizations of ETRS89, in conjunction with transformation paths to ETRF2000 allowing more accurate relationships between national realizations. ETRF2000 rather than WGS 84 should be used as a hub for inter-European transformations. (The ~0.1m accuracy compared to ~1m for WGS 84 transformations should encourage that). Of course, where a direct transformation exists (e.g. between RGF93 v2 and RGF93 v2b in France), this should be prioritised over any indirect path through a hub. For historic systems not aligned with ETRS89, software should prioritise a transformation from an historic system such as Pulkovo 1942(58) to a specific national realization (e.g. ETRS89-ROU [ETRF2000]) over any transformation to the ensemble (ETRS89). We have a guidance note describing all this in preparation (and clearly need to accelerate its publication).

 

Because we cannot add members to the ensemble until their new (to the EPSG Dataset) national realizations are in the Dataset, it had been our intention to update the ensemble after we had added all of the CRSs and transformations, but perhaps we need to consider updating the ensemble as we publish the new realizations.

 We have already discussed the issue which Javier has noted that a large datum ensemble will make for unwieldy WKT. Unfortunately in the WKT specification ISO 19162, a datum ensemble requires that its members to be included. We do not want to deviate from that standard (and would hope that in the future it might be changed to make it optional).
 """"

rouault:
> > We'll definitely need to tweak the DB by creating null transformations between each of the member and the datum ensemble itself
> 
> Wouldn't it be better to deal with this upstream?

I'm just realizing that our import script actually already populates null transformations (under the PROJ authority) between the datum ensemble and its members, with the accuracy of the ensemble. This is done in create_datumensemble_transformations() of scripts/build_db.py

jjimenezshaw:
> I'm just realizing that our import script actually already populates null transformations (under the PROJ authority) between the datum ensemble and its members, with the accuracy of the ensemble. This is done in create_datumensemble_transformations() of scripts/build_db.py

Yes. The potential problem is that it is an extra step, and we have it limited to two steps. I don't know if it is a real problem. I remember I had it to go from ITRFxx to CH1903+. One way was ITRFxx ->  ETRFxx -> ETRS89 -> CH1903+
The step in the middle is the datum ensemble.
But maybe it is a corner case.

rouault:
> The potential problem is that it is an extra step, and we have it limited to two steps.

yes that's indeed a problem. But the solution (for your above case) to fix it at the DB level would be a bit ugly because that would mean that each time you have a transformation between some CRS and a member of a datum ensemble, you'd also need to create a copy between the CRS and the ensemble itself
Ultimately we'd need to implement a proper routing engine that navigates within the DB :-)

PR #4598, opened 2025-11-01
Database: update to EPSG v12.039

rouault:


rouault:
@jjimenezshaw This brings a new geoid model for Faroe Islands : fo_geoid2012a.gri  .  It is available there  https://www.us.fo/kort/geodesi/  (as well as in GeoTIFF) . No mention of license

PR #4602, opened 2025-11-07
Database: update to EPSG v12.041

rouault:


hobu:
This is happening so often we almost need to write a CI automation for it 😉

PR #4603, opened 2025-11-09
tinshift: support TIN GeoPackage files

### LINKED ISSUE #3732: Large JSON TIN support
We are running into trouble extending the PROJ structure where we have large JSON TIN files like we were doing in ticket #3616.  The limit seems to be set [here](https://github.com/OSGeo/PROJ/blob/c31146bee252d5b8bf4dc0b35168ab5e3ddf94b9/src/transformations/tinshift.cpp#L99), but this limit is labeled as arbitrary.

Our current use case has an 80 MB file for the Alaska region.  In the short term we would like to explore increasing this limit (perhaps as part of some local configuration?), but perhaps in the long run there could be support for some Cloud Optimized TIN format for the CDN?

rouault:
Such files are equivalent to already supported TIN JSON files, but scale better for arbitrarily large triangulations, in particular for network-based access.

The format is defined in [source/specifications/tin_gpkg.rst](https://osgeo-proj--4603.org.readthedocs.build/en/4603/specifications/tin_gpkg.html) and the https://github.com/rouault/PROJ-data/blob/tin_json_to_tin_gpkg/grid_tools/tin_json_to_tin_gpkg.py Python script may be used o convert an existing TIN JSON into a TIN GeoPackage.

$ echo 3432087 6995748 0 |  PROJ_NETWORK=ON PROJ_DATA=data bin/cct  +proj=tinshift +file=fi_nls_ykj_etrs35fin.gpkg
  431943.0905   6992816.7826        0.0000           inf
$ echo 3432087 6995748 0 |  PROJ_NETWORK=ON PROJ_DATA=data bin/cct  +proj=tinshift +file=fi_nls_ykj_etrs35fin.json
  431943.0905   6992816.7826        0.0000           inf
$ echo 3432087 6995748 0 |  PROJ_NETWORK=ON PROJ_DATA=data bin/cct  +proj=tinshift +file=fi_nls_n60_n2000.gpkg
 3432087.0000   6995748.0000        0.3161           inf
$ echo 3432087 6995748 0 |  PROJ_NETWORK=ON PROJ_DATA=data bin/cct  +proj=tinshift +file=fi_nls_n60_n2000.json
 3432087.0000   6995748.0000        0.3161           inf

Fixes #3732


src/transformations/tinshift_gpkg.cpp
```diff
@@ -0,0 +1,881 @@
...
+        lambda2 =
+            ((vY[2] - vY[0]) * (x - vX[2]) + (vX[0] - vX[2]) * (y - vY[2])) /
+            det_T;
+        if (!checkWithinTriangle) {
+            lambda3 = 1 - lambda1 - lambda2;
+            return true;
+        }
+        if (lambda1 >= -EPSILON && lambda1 <= 1 + EPSILON &&
+            lambda2 >= -EPSILON && lambda2 <= 1 + EPSILON) {
+            lambda3 = 1 - lambda1 - lambda2;
```
jjimenezshaw: Why writing lambda3 only in that case? It is strange to fill lambda1 and 2 always, but lambda3 only if return is true. I know that it can be computed (as it is there). The extra computation time is negligible.

jjimenezshaw (commented):
I didn't review the complete algorithm yet. But one usual optimization is to start searching on the last triangle successfully returned. If searched points an near, they will probably be in the same triangle. Is it something to include in the API somehow? Or just an implementation detail? (I want to think it is an implementation detail, but I am not sure. In that case we should store the last used triangle threadsafe somewhere)

src/transformations/tinshift_gpkg.cpp
rouault: fixed

rouault:
> But one usual optimization is to start searching on the last triangle successfully returned.

good idea. Implemented now

appveyor.yml
```diff
@@ -5,18 +5,12 @@ branches:
 environment:
   matrix:
 
-# VS 2015
+# VS 2017
   - platform: x86
     APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
```
jjimenezshaw: This is contradictory. The comment says 2017, while the text is "Visual Studio 2015"

src/transformations/tinshift.cpp
```diff
@@ -26,19 +26,29 @@
...
 
 #define PROJ_COMPILATION
+#define FROM_PROJ_CPP
 
-#include "tinshift.hpp"
 #include "filemanager.hpp"
+#include "proj/internal/internal.hpp"
 #include "proj_internal.h"
+#include "tinshift_gpkg.h"
+#include "tinshift_iface.h"
```
jjimenezshaw: Is there any reason to name them `.h` and not `.hpp`? Do we have any rule for that?

src/transformations/tinshift_gpkg.cpp
```diff
@@ -0,0 +1,945 @@
...
+}
+
+// ---------------------------------------------------------------------------
+
+std::unique_ptr<TINShiftGeopackageFile>
+TINShiftGeopackageFile::open(PJ_CONTEXT *ctx, const std::string &filename) {
+    std::string sqliteURI(filename);
+
+    auto poTinshift =
+        std::unique_ptr<TINShiftGeopackageFile>(new TINShiftGeopackageFile());
```
jjimenezshaw, line 48: Does it fit here a `make_unique`? It is used in other points of the PR

src/transformations/tinshift_gpkg.hpp
```diff
@@ -0,0 +1,135 @@
...
+    double max_shift_y_ = 0;
+
+    enum FallbackStrategy {
+        FALLBACK_NONE,
+        FALLBACK_NEAREST_SIDE,
+        FALLBACK_NEAREST_CENTROID,
+    };
+    FallbackStrategy fallbackStrategy_ = FALLBACK_NONE;
+
+    // Cached results
```
jjimenezshaw, line 94: I assume this class cannot be used from several threads, right? (it is not threadsafe)

test/gie/tinshift_gpkg_network.gie
```diff
@@ -0,0 +1,27 @@
+
+-------------------------------------------------------------------------------
+===============================================================================
+Test +proj=tinshift with GeoPackage files on network
```
jjimenezshaw, line 4: Is this using curl or sqlite3 directly? Is it getting the whole file, or is it "cloud optimized"?

appveyor.yml
rouault: comment fixed

src/transformations/tinshift.cpp
rouault: both header files renamed to .hpp

src/transformations/tinshift_gpkg.cpp
rouault: As the constructor is private (the intent is to use the static method instead), we unfortunately can't use std::make_unique (annoying C++ limitation)

src/transformations/tinshift_gpkg.hpp
rouault: yes, it is used by a single PJ object at a time, and PJ objects are not threadsae

test/gie/tinshift_gpkg_network.gie
rouault: We use our NetworkFile implementation that uses curl underneath and range request, and plug it into sqlite3 I/O layer. That way we can have piece wise access to the parts of the file we need, and benefit from the local caching in our ~/.local/share/proj/cache.db of the chunks we have already read.

From my test on a synthetic grid of 500x500 nodes, with each square divided into 2 triangles, that generates a GeoPackage file of ~ 64 MB. The first access to extract a triangle involves ~ 20 network requests of 16 KB each. Accesses to other parts (not close to each other) of the file afterwards take between 6 and 14 requests.

It seems that FlatGeoBuf spatial index (packed Hilbert RTree) would involve less requests, but FlatGeoBuf supports only one layer per file, and we need 2 layers. The FlatGeoBuf files would also be larger than the GeoPackage one (presumably because both the triangle geometry and the RTree need to be stored, whereas with GeoPackage we only store the RTree of the bbox of the triangles, but not the geometry of the triangles themselves, just the indices to the vertices), although file size is probably less of a concern. We could possibly have a .tar of 2 flatgeobuf files? I'm not sure if that's worth the complication.

rouault:
@GlenRice-NOAA Did you get a chance to test this feature?

GlenRice-NOAA:
No, I have not.  I will try to get to it in the next couple of weeks.

jjimenezshaw:
@GlenRice-NOAA any progress? Release 9.8.0 is coming soon

GlenRice-NOAA:
I'm planning on having it tested by the end of the week.  Thanks for checking in.

GlenRice-NOAA:
Can we bump this to the next release?

rouault:
> Can we bump this to the next release?

you mean 9.9 (likely to be released towards end of this year) ? Sure, no emergency from my side

GlenRice-NOAA:
Thank you.

rouault:
Keep open

rouault:
@GlenRice-NOAA Any ETA on when you will be able to give feedback on this?

GlenRice-NOAA:
We have someone new working on it (again).  We think we can provide feedback in the next month.

GlenRice-NOAA:
Thank you for your patence.

We validated the new PROJ tinshift gpkg format by pulling an operational Alaskan mesh dataset from [AWS S3](https://noaa-ocs-nationalbathymetry-pds.s3.amazonaws.com/index.html#Test-and-Evaluation/NOAA_NOS_Vertical_Datum_Register/regional_tins/), converting it using tin_json_to_tin_gpkg.py, and testing it across local and streaming channels. Cross-validation against the source .json and .2dm meshes confirmed absolute numerical parity with 0.00 mm of coordinate drift across all test points. 

Multi-tier speed benchmarking (100 to 1,000,000 points) demonstrated that the gpkg format successfully eliminates the baseline ~3.4-second JSON text-parsing penalty, delivering a 3.9x to 5.2x speedup at typical operational scales. At the scale of 1 million points, a performance crossover occurs where the fully parsed, in-memory JSON array outpaces the repeated sequential SQLite index queries of the gpkg.

JSON size: 49 MB, GPKG: 89MB, Time: second

The comparison table is shown below:
| Points | Channel | JSON (s) | GPKG (s) | Winner/X |
|   100   |   Local   |  3.4676   |  0.6606 |  GPKG (5.2x) |
|   100   |  Stream | 15.5844  | 12.0151 | GPKG (1.3x) |
|   10k   |   Local   |  3.5109   |  0.8917 |  GPKG (3.9x) |
|   10k   |  Stream |  15.5346 | 10.6877 | GPKG (1.5x) |
|   1M   |   Local   |  5.7736   |  24.4672 | JSON (4.2x) |
|   1M   |  Stream |  16.6272 |  25.8323 | JSON (1.6x) |

Thank you for your work on this functionality in PROJ.  We think it could be a vauable contribution to the library.

rouault:
rebased on top of latest master

PR #4606, opened 2025-11-14
WKT/PROJJSON datum ensemble import: allow no members

rouault:
With the future extension of the ETRS89 datum ensemble to all recently created ETRS89-XX national datums, the number of members in the ensemble will be huge and will make ugly WKT/PROJJSON, hence it is highly likely that listing the members will be made optional. Be tolerant of that on the parsing size, but still emit them on export for now.


jjimenezshaw:
Was it a problem not having a member of the ensemble? What happened if you had an "old" WKT from WGS84 before they added the last realization (not so long ago) ? Was it failing?
For instance https://raw.githubusercontent.com/OSGeo/PROJ-CRS-Explorer/8.2.1/wkt2/EPSG/4326.txt

rouault:
> Was it a problem not having a member of the ensemble?

lacking all members was rejected.  Lacking one member or having extra members was fine, at least for coordinate transformation purposes, since testing for equality stops at comparing the datum ensemble name, not its members

> What happened if you had an "old" WKT from WGS84 before they added the last realization (not so long ago) ? Was it failing?

no, that was fine

PR #4610, opened 2025-11-17
Build arm64 docker images with attestation

j9ac9k:
Update the CI so that the docker images use provenance attestations for the github container repository.  In addition, have the CI generate arm64 images of proj; make both images available through a manifest.

Since https://github.com/OSGeo/PROJ/pull/4565 got closed to being stale, I'm re-creating the PR.  

Since the previous PR was opened, similar changes have been merged into GDAL with attestation here: https://github.com/OSGeo/gdal/pull/13066 and multi-platform builds here: https://github.com/OSGeo/gdal/pull/13164

Happy to answer any questions or make requested changes.

PR #4612, opened 2025-11-18
Add relavent permissions for image attestation

j9ac9k:
Follow up to #4610 

Once merged, there was an error (see here: https://github.com/OSGeo/PROJ/actions/runs/19471499706/job/55719775695 ).  I suspect the error has to do with the respective permissions for attestation not being set.  As things are currently, the permissions are only set when creating the multi-platform manifest.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -33,6 +33,11 @@ jobs:
             dockerfile: "./docs/docbuild/Dockerfile"
             runner: "ubuntu-24.04"
     runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+      packages: write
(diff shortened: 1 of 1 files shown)


j9ac9k:
Since this sort of thing is difficult to test locally, if someone with push rights to this repo can test the diff on a separate branch, that would be greatly appreciated.

rouault (approved):

rouault:
> Since this sort of thing is difficult to test locally, if someone with push rights to this repo can test the diff on a separate branch, that would be greatly appreciated.

merging, hoping for the best :-)

j9ac9k:
I see the new error... I'm on it.

PR #4616, opened 2025-11-24
Fix test_fork intermittent failure on macOS by using safe EPSG code

chikle9090:
This PR fixes the intermittent failure in the test_fork test on macOS by preventing database re-initialization inside forked child processes. The change makes the test stable by using a safe EPSG code in child processes so no new registry is created. This resolves the occasional SQLite I/O error seen on macOS.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -149,8 +149,13 @@ Build requirements
 - C++17 compiler
 - CMake >= 3.16
 - SQLite3 >= 3.11: headers and library for target architecture, and sqlite3 executable for build architecture
+<<<<<<< HEAD
 - libtiff >= 4.0 (optional but recommended to read grid formats based on GeoTIFF. Only basic TIFF support is required; optional features like WebP or ZStd are not needed.)
 - curl >= 7.29.0 (optional but recommended to download required grids at runtime.)
+=======
+- libtiff >= 4.0 (optional but recommended to enable reading of grid formats based on GeoTIFF. Only basic TIFF support is required; optional features like WebP or ZStd are not needed.)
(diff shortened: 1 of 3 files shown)


rouault:
> The change makes the test stable by using a safe EPSG code in child processes so no new registry is created. This resolves the occasional SQLite I/O error seen on macOS.

This change actually hides the purpose of his test. But this test only works if the sqlite build supports pread(), which I believe Mac doesn't. So this means using proj with fork() on Mac is not safe. Probably that this test should just be enabled on Linux.

chikle9090:
Thanks for the explanation. I understand. Would the appropriate solution in this case be to disable this test on macOS instead of modifying the EPSG codes? If yes, could you please guide me on where/how the test should be conditionally disabled for macOS?

rouault:
> Thanks for the explanation. I understand. Would the appropriate solution in this case be to disable this test on macOS instead of modifying the EPSG codes? If yes, could you please guide me on where/how the test should be conditionally disabled for macOS?

Likely this patch:
```
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -255,7 +255,7 @@ add_test(NAME test_misc COMMAND test_misc)
 set_property(TEST test_misc
   PROPERTY ENVIRONMENT ${PROJ_TEST_ENVIRONMENT})
 
-if (Threads_FOUND AND CMAKE_USE_PTHREADS_INIT)
+if (Threads_FOUND AND CMAKE_USE_PTHREADS_INIT AND NOT APPLE)
 add_definitions(-DPROJ_HAS_PTHREADS)
 add_executable(test_fork
   test_fork.c)
```

rouault:
Please rebase on top of latest master, and revert the EPSG code changes

PR #4623, opened 2025-11-26
Support for MTM CGVD2013 epoch 1997, 2002, 2010 and UTM CGVD28 epoch 1997, 2002, 2010

jsmoreau:
Addition of support for MTM CGVD2013 epoch 1997, 2002, 2010 and UTM CGVD28 epoch 1997, 2002, 2010.


rouault:
You actually need to also commit the .sql file regenerated by running that file

rouault:
ah, and as this changes the database content, you also need to update data/CMakeLists.txt as indicated by the below error message
```
  Update 'set(PROJ_DB_SQL_EXPECTED_MD5 ...)' line in data/CMakeLists.txt with
  714c60bfd143234b311b6a934e028e10 value
```

PR #4627, opened 2025-12-01
Use Emscripten fetch in networkfilemanager

jjimenezshaw:
The main purpose of this PR is an implementation in the networkfilemanager that substitutes cURL when running in wasm. For that purpose emscripten_fetch is used, as suggested in https://github.com/emscripten-core/emscripten/issues/3270#issuecomment-2260232768

The added implementation in `networkfilemanager.cpp` is a copy of the cURL implementation, trying to mimic it as much as possible. Emscripten does convert it into whatever code in wasm + js, using the fetch in the browser.

To enable this feature, PROJ must be compiled with the option `ENABLE_EMSCRIPTEN_FETCH`, that by default is OFF. It works only in `emscripten`.

As a needed tool for my tests I implemented a Docker image and a build script to produce a wasm compilation, and test it. It has been very useful, as it showed up some problems due to the synchronous fetch used in PROJ. Already having the build infrastructure, I added it as a github action. Anybody can take that to produce a wasm of PROJ.

[emscripten documentation](https://emscripten.org/docs/api_reference/fetch.html#synchronous-fetches) says clearly that to get synchronous fetch the `-pthread` flag is needed. It also needs that the javascript calls are done in a Web Worker, not in the main thread. I was not able to make `--proxy-to-worker` work, as there is no `main` to use it. The build script is then using the `-pthread` flag for all the dependencies that have to be compiled: `zlib`, `libtiff` and `libsqlite3`.
If a developer doesn't need the fetch functionality, it is much easier not linking with libtiff (and zlib).

If anybody with a better empscripten knowledge has a better solution, please let me know.

A basic example is running in https://jjimenezshaw.github.io/wasm-proj/example.html

For the usage in the browser, an special configuration for Cross-origin isolation is needed. As that is not included in GitHub Pages, I am using https://github.com/gzuidhof/coi-serviceworker in the example. Probably you need it too.

- [ ] Closes #xxxx
- [ ] Tests added


jjimenezshaw:
fyi @willcohen

jjimenezshaw:
BTW, the github action is creating an artifact with the js and wasm files.

CMakeLists.txt
```diff
@@ -282,6 +282,16 @@ if(ENABLE_CURL)
...
 
+################################################################################
+# Check for Emscripten fetch
+################################################################################
+
+option(ENABLE_EMSCRIPTEN_FETCH "Enable Emscripten fetch support" OFF)
+set(EMSCRIPTEN_FETCH_ENABLED FALSE)
+if(ENABLE_EMSCRIPTEN_FETCH)
+  set(EMSCRIPTEN_FETCH_ENABLED TRUE)
+endif()
```
rouault, line 293: why not turning it ON by default, but only exposing it if (EMSCRIPTEN) ?
```suggestion
if (EMSCRIPTEN)
  option(ENABLE_EMSCRIPTEN_FETCH "Enable Emscripten fetch support" ON)
  set(EMSCRIPTEN_FETCH_ENABLED FALSE)
  if(ENABLE_EMSCRIPTEN_FETCH)
    set(EMSCRIPTEN_FETCH_ENABLED TRUE)
  endif()
endif()
```

src/networkfilemanager.cpp
```diff
@@ -1922,6 +1939,267 @@ static const char *pj_curl_get_header_value(PJ_CONTEXT *,
...
+            } else {
+                break;
+            }
+        } // end of while(true)
+
+        if (out_error_string && error_string_max_size) {
+            out_error_string[0] = '\0';
+        }
+
+        // cast 64b to 32b. Hopefully we dont download more than 2GB
```
rouault: is that comment accurate ? I believe 64-bit WASM builds are possible

src/networkfilemanager.cpp
```diff
@@ -1922,6 +1939,267 @@ static const char *pj_curl_get_header_value(PJ_CONTEXT *,
...
+
+            emscripten_fetch_get_response_headers(fetch, headerString,
+                                                  headersLengthBytes);
+            TRACE_FETCH("Got headers: " << headerString);
+
+            char **responseHeaders =
+                emscripten_fetch_unpack_response_headers(headerString);
+            assert(responseHeaders);
+
+            free(headerString);
```
rouault, line 2134:
```suggestion
            const size_t headersLengthBytes =
                emscripten_fetch_get_response_headers_length(fetch);
            TRACE_FETCH("headers length: " << headersLengthBytes);
            std::string headerString;
            headerString.resize(headersLengthBytes);

            emscripten_fetch_get_response_headers(fetch, headerString.data(),
                                                  headersLengthBytes + 1);
            TRACE_FETCH("Got headers: " << headerString);

            char **responseHeaders =
                emscripten_fetch_unpack_response_headers(headerString.c_str());
            assert(responseHeaders);
```

scripts/ci/emscripten/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    sqlite3 \
+    ninja-build && \
+    # Clean up apt cache to reduce image size
+    rm -rf /var/lib/apt/lists/*
+
+# --- Setup Build Directories ---
+# Define standard locations for building and installing
+ENV BUILD_DIR="/build"
+ENV INSTALL_DIR="/usr/local/wasm"
+ENV PROJ_VERSION="9.7.0"
```
rouault: unneeded in this repository as we use the git checkout
```suggestion
```

scripts/ci/emscripten/build_wasm.sh
```diff
@@ -0,0 +1,318 @@
+#!/bin/bash
+# Exit immediately if a command exits with a non-zero status.
+set -e
+
+# --- Configuration Variables ---
+# Use environment variables from Dockerfile or runtime.
+PROJ_VERSION=${PROJ_VERSION:-9.7.0}
+PROJ_REPO=${PROJ_REPO:-https://github.com/OSGeo/PROJ.git}
+PROJ_BRANCH=${PROJ_BRANCH:-${PROJ_VERSION}} # If PROJ_BRANCH is not set, use PROJ_VERSION
```
rouault:
```suggestion
```

scripts/ci/emscripten/build_wasm.sh
```diff
@@ -0,0 +1,318 @@
...
+    if [ -d "${PROJ_SRC_DIR}" ] && [ "$(ls -A ${PROJ_SRC_DIR})" ]; then
+        echo "Local PROJ source found at ${PROJ_SRC_DIR}. Skipping clone."
+    else
+        echo "Cloning PROJ repository: ${PROJ_REPO}"
+        git clone --depth 1 --branch ${PROJ_BRANCH} ${PROJ_REPO} ${PROJ_SRC_DIR}
+    fi
+else
+    echo "PROJ source directory already exists. Skipping clone."
+fi
+
```
rouault, line 211:
```suggestion
```

scripts/ci/emscripten/build_wasm.sh
```diff
@@ -0,0 +1,318 @@
...
+    mkdir -p ${PROJ_BUILD_WASM_DIR}
+    cd ${PROJ_BUILD_WASM_DIR}
+
+    # Configure PROJ
+    configure_cmake .. \
+        -D BUILD_TESTING=OFF \
+        -D BUILD_APPS=OFF \
+        -D ENABLE_TIFF=ON \
+        -D ENABLE_CURL=OFF \
```
rouault, line 239: could make sense to default ENABLE_CURL to OFF for emscripten builds

rouault (commented):
Very cool!

CMakeLists.txt
jjimenezshaw: I see your point. But it may have some side effects. At the be beginning of networkfilemanager.cpp I check it with the macro  `__EMSCRIPTEN_PTHREADS__` .
Some users (a majority?) will prefer to build without the fetch functionality. That produces a smaller wasm file, without libtiff, zlib, pthreads, and it works in the main thread. A Web Worker is not needed.
(well, compiling with pthreads also works in the main thread... until it does a fetch request ;)
In this case I think opt-in is easier and safer. Maybe should we say something else in the documentation? In https://proj.org/en/stable/usage/network.html ?

scripts/ci/emscripten/Dockerfile
jjimenezshaw: My idea was that the Dockerfile and build_wasm.sh could be used for the "normal users" when needed. In that case it would be useful just to set the release or branch. (I am thinking on making configurable the `EXPORTED_FUNCTIONS` parameter via a file, something already available in emcc. In a later PR probably)
I agree that having a default version is bad, I will remove it in any case.
Maybe having the scripts somewhere else in the code-tree makes it more accessible.

Said that, I don't have really a strong opinion. The "user" can always clone the code themselves. I can remove those options.

scripts/ci/emscripten/build_wasm.sh
jjimenezshaw: Is it worth it? (including to document it)

src/networkfilemanager.cpp
jjimenezshaw: `fetch->numBytes` is 64 bits, so it is possible. However `size_t` is 32 bits (surprise!)
I have the impression that the function calling with `size_to_read` was not expecting the 32 bits limit, does it?
What should I do?

src/networkfilemanager.cpp
jjimenezshaw: I had serious problems with the memory trying to implement it on my own. So finally I decided to copy literally what they have in the unit tests.
There was something fishy in the emscripten side.
If it is not wrong, I would prefer to leave it as it is.

CMakeLists.txt
rouault: > At the be beginning of networkfilemanager.cpp I check it with the macro `__EMSCRIPTEN_PTHREADS__` .
> Some users (a majority?) will prefer to build without the fetch functionality.

ok, I see your point. Let's forget that then

src/networkfilemanager.cpp
rouault: My point was that on a 64-bit WASM build size_t is going to be 64 bit, hence the comment about 64->32bit cast wouldn't be accurate. That's a minor point.
That cast isn't much of a concern. We shouldn't request more than 4 GB at once...

src/networkfilemanager.cpp
rouault: > If it is not wrong, I would prefer to leave it as it is.

ok, that's fine

scripts/ci/emscripten/Dockerfile
rouault: maybe add some comment in the files that the version pin isn't used by PROJ CI ?

scripts/ci/emscripten/build_wasm.sh
rouault: not necessarily. I see that in the doc of ENABLE_EMSCRIPTEN_FETCH, you've added "It is incompatible with cURL, so ``ENABLE_CURL`` must be OFF."

rouault (approved):

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,43 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+
+  emscripten:
+    runs-on: ubuntu-latest
+
+    if: "!contains(github.event.head_commit.message, '[ci __skip]') && !contains(github.event.head_commit.message, '[skip ci]')"
```
rouault:
```suggestion
    if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]')"
```

src/networkfilemanager.cpp
jjimenezshaw: Oh, pretty recent: https://webassembly.org/news/2025-09-17-wasm-3.0/

scripts/ci/emscripten/Dockerfile
jjimenezshaw: Removed from the Dockerfile. In the script PROJ_TAG has to be explicitly set, or populate the proj_src folder with your local copy (usually mounting a volume).

The code (or tag) to compile has to be explicitly provided.

.github/workflows/emscripten.yml
jjimenezshaw: done

src/networkfilemanager.cpp
jjimenezshaw: I just removed the comment.

docs/source/usage/network.rst
```diff
@@ -45,6 +45,12 @@ Authorizing network access can be done in multiple ways:
     with :cpp:func:`proj_context_set_network_callbacks`. Enabling network use
     must still be done with one of the above mentioned method.
 
+Emscripten
+^^^^^^^^^^
+
+To compile with emscripten (to produce a WASM file) with network capability,
+the CMake configure option `ENABLE_EMSCRIPTEN_FETCH` can be used.
```
jjimenezshaw: I would like to make it a link to the proper section for `ENABLE_EMSCRIPTEN_FETCH` in `install.rst`, but I don't know how.

willcohen:
I am positively thrilled to see this moving forward. I’m currently working on putting proj-wasm to work on some projects to reproject transpiled-to-WASM JTS objects and or allow them to work with CRSes beyond just EPSG in the browser and once I get that a little integrated I will be very excited to be able to seamlessly add grids in too. Long way of saying I may not be able to use this downstream for a few weeks, but I’m working up to it.

docs/source/usage/network.rst
rouault: insert a marker ``.. _install_emscripten_fetch:`` before the line that documents ENABLE_EMSCRIPTEN_FETCH in install.rst,   and link to it with `` :ref:`some text <install_emscripten_fetch>` ``

jjimenezshaw: Thanks!

rouault (approved):

PR #4632, opened 2025-12-04
fix normalizeForVisualization to skip extent checks for axis-swap ope…

### LINKED ISSUE #4631: Add option to skip extent validation in `proj_normalize_for_visualization()`
`proj_normalize_for_visualization()` currently fails for certain concatenated coordinate operations where the sub-operations have non-overlapping validity areas. This is because the function internally calls `ConcatenatedOperation::createComputeMetadata()` with `checkExtent=true`, which throws an exception when validity areas don't intersect.

I propose adding an option to allow callers to bypass this extent check when needed.

## Use Case

When using pyproj's `always_xy=True` parameter (which calls `proj_normalize_for_visualization()` internally), users expect coordinate transformations to work with (x, y) / (longitude, latitude) axis order. However, for certain valid concatenated operations like [EPSG:8047 (ED50 → ED87 → WGS84)](https://epsg.org/concatenated-operation_8047/ED50-to-WGS-84-15.html?sessionkey=w0balagpl1), the function returns NULL because:

1. [EPSG:1147 (ED50 → ED87) covers Norway N65 area](https://epsg.org/transformation_1147/ED50-to-ED87-2.html?sessionkey=qdvcdc3xyc)
2. [EPSG:1146 (ED87 → WGS84) covers North Sea area](https://epsg.org/transformation_1146/ED87-to-WGS-84-1.html?sessionkey=w8gxl5skml)
3. These areas don't have a perfect intersection when combined with the axis-swap wrapper operations

The transformation itself is mathematically valid and works correctly with native axis order—it's only the normalization wrapper that fails.

## Reproduction

```cpp
#include <proj.h>
#include <stdio.h>

int main() {
    PJ_CONTEXT *ctx = proj_context_create();
    
    // EPSG:8047 is a concatenated operation: ED50 -> ED87 -> WGS84
    PJ *P = proj_create(ctx, "urn:ogc:def:coordinateOperation:EPSG::8047");
    if (!P) {
        printf("Failed to create operation\n");
        return 1;
    }
    
    printf("Operation: %s\n", proj_get_name(P));
    
    // This returns NULL due to extent validation failure
    PJ *P_norm = proj_normalize_for_visualization(ctx, P);
    if (!P_norm) {
        int err = proj_context_errno(ctx);
        printf("proj_normalize_for_visualization failed: %s\n", proj_errno_string(err));
    } else {
        printf("Normalization succeeded\n");
        proj_destroy(P_norm);
    }
    
    proj_destroy(P);
    proj_context_destroy(ctx);
    return 0;
}
```

Output:
```
Operation: ED50 to WGS 84 (15)
proj_normalize_for_visualization failed: (error message about empty intersection)
```

### Option 1: Add `proj_normalize_for_visualization_ex()` with `checkExtent` parameter

```c
// Existing API (unchanged)
PJ *proj_normalize_for_visualization(PJ_CONTEXT *ctx, const PJ *obj);

// New extended API
PJ *proj_normalize_for_visualization_ex(PJ_CONTEXT *ctx, const PJ *obj, int check_extent);
```

Usage:
```c
// Skip extent validation
PJ *P_norm = proj_normalize_for_visualization_ex(ctx, P, false);
```

### Option 2: Skip extent check for axis-swap wrappers

Since the axis-swap operations added by `normalizeForVisualization()` have no spatial effect (they only reorder coordinates), the extent check could be skipped specifically for these wrapper operations.

In `src/iso19111/operation/singleoperation.cpp`:

```cpp
// Current code
return util::nn_static_pointer_cast<CoordinateOperation>(
    ConcatenatedOperation::createComputeMetadata(subOps, true));  // checkExtent=true

// Proposed change
return util::nn_static_pointer_cast<CoordinateOperation>(
    ConcatenatedOperation::createComputeMetadata(subOps, false)); // checkExtent=false for axis-swap wrappers
```

This approach recognizes that axis-swap operations don't change the geographic validity of the underlying operation.

### Option 3: Use union instead of intersection for axis-swap wrappers

When wrapping with axis-swap operations, compute the **union** of validity areas instead of the intersection, since the wrapper operations themselves are globally valid.

## Rationale

1. **Axis swaps have no spatial effect** — They only change coordinate order, not the geographic area where the transformation is valid

2. **The underlying operation is valid** — EPSG:8047 works correctly with native axis order; only the normalization wrapper fails

3. **Better downstream library support** — Libraries like pyproj can provide consistent `always_xy` behavior without workarounds

4. **Backward compatibility** — The existing `proj_normalize_for_visualization()` behavior can remain unchanged; only the new `_ex()` variant or internal logic change would be affected

## Current Workaround

Users must manually wrap the PROJ pipeline with axis swaps:

```c
const char *proj_str = proj_as_proj_string(ctx, P, PJ_PROJ_5, nullptr);

char pipeline[2048];
snprintf(pipeline, sizeof(pipeline),
    "+proj=pipeline "
    "+step +proj=axisswap +order=2,1 "
    "+step %s "
    "+step +proj=axisswap +order=2,1",
    proj_str + 15);

PJ *P_normalized = proj_create(ctx, pipeline);
```

This is error-prone and requires users to understand the internal structure of PROJ pipelines.

## Related

- This affects pyproj's `Transformer(..., always_xy=True)` functionality
- Related code: `CoordinateOperation::normalizeForVisualization()` in `src/iso19111/operation/singleoperation.cpp`
- Related code: `ConcatenatedOperation::createComputeMetadata()` in `src/iso19111/operation/concatenatedoperation.cpp`

phaarnes:
This pull request is adding the improvement required in issue #4631.

- ✅ Closes #4631
- ✅ Test added in ``test_operation.cpp``

``proj_normalize_for_visualization()`` currently fails for certain concatenated coordinate operations where the sub-operations have non-overlapping validity areas. This happens because the function internally calls ``ConcatenatedOperation::createComputeMetadata()`` with ``checkExtent = true``, which throws an exception when the validity areas do not intersect.

This PR disables the extent check for the axis-swap wrapper. It is not necessary because the axis-swap operations added by ``normalizeForVisualization()`` have no spatial effect (they only reorder coordinates), so the extent check can be safely skipped.

### DIFF
--- a/src/iso19111/operation/concatenatedoperation.cpp
+++ b/src/iso19111/operation/concatenatedoperation.cpp
@@ -176,10 +176,13 @@ ConcatenatedOperationNNPtr ConcatenatedOperation::create(
 
     crs::CRSPtr interpolationCRS;
     bool interpolationCRSValid = true;
+    bool hasBallparkTransformation = false;
     for (size_t i = 0; i < operationsIn.size(); i++) {
         auto l_sourceCRS = operationsIn[i]->sourceCRS();
         auto l_targetCRS = operationsIn[i]->targetCRS();
 
+        hasBallparkTransformation |=
+            operationsIn[i]->hasBallparkTransformation();
         if (interpolationCRSValid) {
             auto subOpInterpCRS = operationsIn[i]->interpolationCRS();
             if (interpolationCRS == nullptr)
@@ -249,6 +252,7 @@ ConcatenatedOperationNNPtr ConcatenatedOperation::create(
         operationsIn);
     op->assignSelf(op);
     op->setProperties(properties);
+    op->setHasBallparkTransformation(hasBallparkTransformation);
     op->setCRSs(l_sourceCRS, l_targetCRS, interpolationCRS);
     op->setAccuracies(accuracies);
 #ifdef DEBUG_CONCATENATED_OPERATION
--- a/src/iso19111/operation/singleoperation.cpp
+++ b/src/iso19111/operation/singleoperation.cpp
@@ -432,8 +432,13 @@ CoordinateOperation::normalizeForVisualization() const {
                     l_targetCRS->normalizeForVisualization(), nullptr);
         subOps.emplace_back(op);
     }
+    // Use checkExtent=false because axis-swap operations don't change the
+    // geographic validity of the underlying operation - they only reorder
+    // coordinates. This allows normalizeForVisualization() to work on
+    // concatenated operations where sub-operations may have non-overlapping
+    // validity areas (e.g., EPSG:8047).
     return util::nn_static_pointer_cast<CoordinateOperation>(
-        ConcatenatedOperation::createComputeMetadata(subOps, true));
+        ConcatenatedOperation::createComputeMetadata(subOps, false));
 }
 
 // ---------------------------------------------------------------------------
--- a/src/iso19111/operation/singleoperation.cpp
+++ b/src/iso19111/operation/singleoperation.cpp
@@ -37,6 +37,7 @@
 #include "proj/metadata.hpp"
 #include "proj/util.hpp"
 
+#include "proj/internal/crs_internal.hpp"
 #include "proj/internal/internal.hpp"
 #include "proj/internal/io_internal.hpp"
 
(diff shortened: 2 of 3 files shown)


phaarnes:
A simple test script to verify:

```cpp
/**
 * Example: proj_normalize_for_visualization
 *
 * Test whether "always_xy" (proj_normalize_for_visualization) works
 * for ED50 to WGS84 transformations.
 *
 * Tests both:
 * - EPSG:1612 (simple Helmert, non-concatenated)
 * - EPSG:8047 (concatenated: ED50->ED87->WGS84) (note: superseded)
 *
 */

#include <cstdio>
#include <proj.h>

void test_transformation(PJ_CONTEXT *ctx, const char *op_code, double lon, double lat) {
    char urn[100];
    snprintf(urn, sizeof(urn), "urn:ogc:def:coordinateOperation:EPSG::%s", op_code);

    PJ *P = proj_create(ctx, urn);
    if (!P) {
        printf("Failed to create EPSG:%s\n\n", op_code);
        return;
    }

    printf("=== EPSG:%s ===\n", op_code);
    printf("Name: %s\n", proj_get_name(P));
    printf("Accuracy: %.2f m\n", proj_coordoperation_get_accuracy(ctx, P));

    // Test WITHOUT normalization (native axis order: lat, lon)
    printf("\nWithout normalization (native lat/lon order):\n");
    PJ_COORD input_native = proj_coord(lat, lon, 0, 0);
    PJ_COORD output_native = proj_trans(P, PJ_FWD, input_native);
    printf("  Input:  lat=%.6f, lon=%.6f\n", lat, lon);
    printf("  Output: lat=%.6f, lon=%.6f\n", output_native.lp.lam, output_native.lp.phi);

    // Test WITH normalization (always_xy: lon, lat order)
    printf("\nWith proj_normalize_for_visualization (always_xy):\n");
    PJ *P_norm = proj_normalize_for_visualization(ctx, P);
    if (P_norm) {
        PJ_COORD input_xy = proj_coord(lon, lat, 0, 0);
        PJ_COORD output_xy = proj_trans(P_norm, PJ_FWD, input_xy);
        printf("  Input:  lon=%.6f, lat=%.6f\n", lon, lat);
        printf("  Output: lon=%.6f, lat=%.6f\n", output_xy.lp.lam, output_xy.lp.phi);
        printf("  SUCCESS: always_xy works!\n");
        proj_destroy(P_norm);
    } else {
        printf("  FAILED: proj_normalize_for_visualization returned NULL\n");

    }

    proj_destroy(P);
    printf("\n");
}

int main() {
    printf("=== Testing proj_normalize_for_visualization ===\n\n");

    PJ_CONTEXT *ctx = proj_context_create();

    // Test point in Norway (valid for both transformations)
    double lon = 10.0;
    double lat = 60.0;

    printf("Test point: lon=%.4f, lat=%.4f\n\n", lon, lat);

    // Test 1: Simple (non-concatenated) transformation
    test_transformation(ctx, "1612", lon, lat);  // ED50 to WGS 84 (23) - Helmert

    // Test 2: Concatenated transformation
    test_transformation(ctx, "8047", lon, lat);  // ED50 to WGS 84 (15) - via ED87

    proj_context_destroy(ctx);

    return 0;
}

```

rouault:
@phaarnes I've added a commit to your pull request to better address the issue by re-applying most of the metadata of the original operation to the normalized one (name, domain of validity, accuracy, remarks), that also avoids revalidating the domain

phaarnes:
@rouault
Great, thanks!

PR #4645, opened 2025-12-30
Edits to TRACE_FETCH debug messages

willcohen:
Small tweak to #4627, just updating formatting of the debug messages (noticed while using it to work on incorporating fetch into proj-wasm).


jjimenezshaw (approved):
I'm happy those traces were useful to somebody else.

willcohen:
Force pushed to address linter from CI.

PR #4646, opened 2025-12-30
projinfo as library function

jjimenezshaw:
The main purpose if this PR is to allow the usage of  `projinfo` as a library, not only as a CLI program.
That allows to call it, for instance, in the browser. See an example at https://jjimenezshaw.github.io/wasm-proj/projinfo.html

The code from the CLI was kept unmodified. To replicate the usage of std::cout and std::cerr, streams are used internally in the library, and the API is using a callback to report every call to those streams. See that some commands, like `--list-crs` can be very fragmented, with many calls to the callback.

For the specific usage of the callback in emscripten, a nicer helper in C++ is added to the `build_wasm.sh` script. That makes the usage from javascript much simpler and small.

I was not sure about the file structure to be used, and how to do it in CMake. Please, make any suggestion on that regard.

- [ ] ~AI/LLM (Copilot, ChatGPT, Claude or something similar) supported my development of this PR~
- [ ] ~Closes #xxxx~
- [x] Tests added
- [ ] Fully documented, including updating `docs/source/*.rst` for new API: Where and what exactly?


src/apps/projinfo_lib.cpp
```diff
@@ -0,0 +1,2238 @@
...
+                             "PROJ_NETWORK=ON not specified"
+                          << std::endl;
+            }
+#else
+            strm.cout << "Status: disabled" << std::endl;
+            strm.cout << "Reason: build without Curl support" << std::endl;
+#endif
+            return 0;
+        } else if (arg == "-?" || arg == "--help") {
+            return usage(strm);
```
jjimenezshaw: The code in master is considering `--help` an error, and returns 1 (probably because it was simpler)
I kept the same behaviour in the code, but it is easy to change with this.
```suggestion
            usage(strm);
            return 0;
```
Should I do it?

jjimenezshaw:
cc @willcohen

src/apps/projinfo_lib.cpp
jjimenezshaw: Well, it is a bit more code if we want to be consistent, but easy. I have it already done in my computer.

jjimenezshaw: I added the changes in a commit. Easy to remove if wanted.

willcohen:
I haven't had a chance to test this build first -- I'm still trying to line up the fetch functionality w wasm versus JNA -- but I am extremely supportive of any moves to migrate the CLI functionality exposed as functions. This will mean that all (most?) of the projinfo logic would then be available in WASM and also all native wrappers, and the callback logic seems totally appropriate. (Perhaps there's another pre-existing way for other downstream consumers to embed the tools into libraries, but I'm not aware of it.)

jjimenezshaw:
@hobu @rouault

rouault (approved):

PR #4657, opened 2026-01-15
Update proj_symbol_rename.h

pjonsson:
Make the script that generates `proj_symbol_rename.h` use the `libproj.so` built by the regular build system instead of compiling its own library.

Re-generate `proj_symbol_rename.h` with the updated script, and keep the long lines to make it easier to spot differences in the future. This adds 2 new symbols to the header.

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,2 @@
+# Auto-generated file.
+src/proj_symbol_rename.h
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,6 +17,7 @@ repos:
                 src/wkt2_generated_parser.h|
                 src/geodesic.c|
                 src/geodesic.h|
+                src/proj_symbol_rename.h|
                 src/tests/geodtest.c|
                 src/tests/geodsigntest.c
               )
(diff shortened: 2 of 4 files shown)


rouault:
Maybe add an exception to ``.pre-commit-config.yaml`` (at root of the repo) so that it doesn't try to reformat proj_symbol_rename.h

rouault (approved):

pjonsson:
> Maybe add an exception to `.pre-commit-config.yaml` (at root of the repo)

Ok, added a clang-format-ignore as well while I was at it. I'm afraid I clogged up the CI with my 2 consecutive pushes though, sorry about that!

rouault:
> I'm afraid I clogged up the CI with my 2 consecutive pushes though, sorry about that!

No, that should be fine. We have the following statement in the CI .yml that cancels previous tasks related to a PR:

```
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
  cancel-in-progress: true

```

pjonsson:
I'm not sure how the PROJ release schedule intertwines with the GDAL release schedule, will this fix be a part of GDAL 3.13.0?

rouault:
> will this fix be a part of GDAL 3.13.0?

more exactly PROJ 9.8 is planned to be released beginning of march, so before GDAL 3.13

PR #4661, opened 2026-01-30
Emphasize modern use patterns in docs

kbevers:
A very minimal attempt at removing bad usage patterns from the docs. In 2026 there's no need to advertise old practice - we've got something that's much better today.

Ideally the proj-string examples in the cs2cs docs should get a revamp as well. I might get to that later.

### DIFF
--- a/docs/source/usage/quickstart.rst
+++ b/docs/source/usage/quickstart.rst
@@ -48,22 +48,14 @@ transform from one coordinate reference system to another. Say we want to
 convert the above Mercator coordinates to UTM, we can do that with
 :program:`cs2cs`:
 
-::
-
-    $ echo 3399483.80 752085.60 | cs2cs +proj=merc +lat_ts=56.5 +ellps=GRS80 +to +proj=utm +zone=32
-    6103992.36      1924052.47 0.00
-
-Notice the ``+to`` parameter that separates the source and destination
-projection definitions.
-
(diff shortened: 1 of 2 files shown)


docs/source/usage/quickstart.rst
```diff
@@ -48,22 +48,14 @@ transform from one coordinate reference system to another. Say we want to
...
 
-   $ echo 56 12 | cs2cs +init=epsg:4326 +to +init=epsg:25832
-   231950.54      1920310.71 0.00
+   $ echo 56 12 | cs2cs EPSG:4326 EPSG:25832
+687071.44       6210141.33 0.00
 
 In the above example we transform geodetic coordinates in the WGS84 reference
 frame to UTM zone 32N coordinates in the ETRS89 reference frame.
-UTM coordinates 
+UTM coordinates.
```
rouault:
```suggestion
   $ echo 56 12 | cs2cs EPSG:4326 EPSG:25832
687071.44       6210141.33 0.00

In the above example we transform geodetic coordinates in latitude, longitude
order in the WGS84 reference frame to UTM zone 32N coordinates in the
ETRS89 reference frame.
UTM coordinates.
```

kbevers: Thanks, that's a helpful addition!

PR #4665, opened 2026-02-07
oss-fuzz: Copy proj.db and additional data files to output directory

am009:
When fuzzing, it frequently reports that proj.db is not found:

```
proj_create: Cannot find proj.db
```

The proj.db is generated under build/data, not data/.

### DIFF
--- a/test/fuzzers/build.sh
+++ b/test/fuzzers/build.sh
@@ -88,3 +88,4 @@ echo "[libfuzzer]" > $OUT/proj_crs_to_crs_fuzzer.options
 echo "max_len = 10000" >> $OUT/proj_crs_to_crs_fuzzer.options
 
 cp -r data/* $OUT
+cp -r build/data/* $OUT


rouault:
@am009 Your fix looks plausible, but can you explain how you detected the issue and how you verified your PR fixes it ? Looking at a run of ci-fuzz on another pull request (https://github.com/OSGeo/PROJ/actions/runs/20169922244/job/57902633665?pr=4603), I don't see a "Cannot find proj.db" trace (but I'm aware the run might not be long enough to trigger it)

am009:
This script was copied verbatim into [FuzzBench](https://github.com/google/fuzzbench), and I encountered this issue there. In FuzzBench, the build environment and the execution environment are separated, and only the $OUT directory is copied into the isolated runtime environment; as a result, proj.db cannot be found.

I am not very familiar with the OSS-Fuzz architecture, but it seems that its build and runtime environments may be the same. The preceding `make install` command in the script installs proj.db into the system, so proj.db can be located correctly.

sorry for opening this pull request without a thorough analysis. Based on the logs you provided, the issue indeed does not appear to exist. In that case, whether this fix is applied does not seem to matter much.

PR #4667, opened 2026-02-08
Derived projected CRS related improvements

rouault:
- proj_factors(): make it work with derived projected CRS
- Add proj_create_linear_affine_parametric_conversion(), proj_create_derived_projected_crs(), proj_crs_add_horizontal_derived_conversion()


test/unit/gie_self_tests.cpp
```diff
@@ -640,6 +640,254 @@ TEST(gie, info_functions) {
...
+            "            LENGTHUNIT[\"metre\",1],\n"
+            "            ID[\"EPSG\",8639]],\n"
+            "        PARAMETER[\"B1\",0,\n"
+            "            SCALEUNIT[\"coefficient\",1],\n"
+            "            ID[\"EPSG\",8640]],\n"
+            // 3.2811679790026242 = 1.0 / 0.3048 * 1.0001
+            "        PARAMETER[\"B2\",3.2811679790026242,\n"
+            "            SCALEUNIT[\"coefficient\",1],\n"
+            "            ID[\"EPSG\",8641]]],\n"
+            "    CS[Cartesian,2],\n"
```
jjimenezshaw, line 860: As we know, in an "Affine parametric transformation" the axis order in the WKT is considered in a strange way.
What happens with a northing-easting system? And when A2 or B1 have non-zero values?

src/proj.h
```diff
@@ -1798,6 +1816,16 @@ PJ PROJ_DLL *proj_create_projected_crs(PJ_CONTEXT *ctx, const char *crs_name,
                                        const PJ *conversion,
                                        const PJ *coordinate_system);
 
+PJ PROJ_DLL *proj_create_derived_projected_crs(PJ_CONTEXT *ctx,
```
jjimenezshaw, line 1819: How do these changes behave with `proj_normalize_for_visualization`?

test/unit/gie_self_tests.cpp
rouault: > What happens with a northing-easting system? 

That PR doesn't change this behaviour. The axis order of the coordinate system of the derived projected CRS doesn't influence the computation. The CS of the derived projected CRS is mostly informative

> And when A2 or B1 have non-zero values?

well just as documented:
- value of the first axis of the derived CRS = A0 +  value of first axis of base CRS * A1 + value of second axis of base CRS * A2
- value of the second axis of the derived CRS = B0 +  value of first axis of base CRS * B1 + value of second axis of base CRS * B2

src/proj.h
rouault: From what I can see, CRS::normalizeForVisualization() does act on a DerivedProjectedCRS, but I don't think that currently it has any effect on changing actual coordinate values.

PR #4679, opened 2026-02-23
CMake: use check_cxx_compiler_flag() for -Wdeprecated-copy-dtor

mwtoews:
Fixes gcc warning on NetBSD 9 by using CMake's [`check_cxx_compiler_flag()`](https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html) command for `-Wdeprecated-copy-dtor`.

See https://lists.osgeo.org/pipermail/proj/2026-February/011976.html


kbevers:
Thanks, Mike!

PR #4685, opened 2026-03-01
Make sure that epoch is set in more scenarios of time-dependent transformations

rouault:
and Doc: add a page specific to time dependent transformations


kbevers (approved):
The new doc page is a very good ressource. Gives a great overview of what's possible.

PR #4686, opened 2026-03-02
Doc: copyedits in about_core.rst

garnerrobert:
Removing redundant article and adding period at end of sentence


docs/source/about_core.rst
```diff
@@ -1,7 +1,7 @@
-PROJ is a generic coordinate transformation software that transforms geospatial
+PROJ is generic coordinate transformation software that transforms geospatial
```
rouault: 'a' seems fine, so keeping it
```suggestion
PROJ is a generic coordinate transformation software that transforms geospatial
```

PR #4699, opened 2026-03-06
fully respect crs_extent_use=none in pipeline composition

### LINKED ISSUE #4697: SourceTargetCRSExtentUse::NONE (--crs-extent-use) does not fully disable extent filtering when pipelines are composed/chained
When SourceTargetCRSExtentUse::NONE is set (e.g. --crs-extent-use none in projinfo, or crs_extent_use="none" in pyproj), PROJ still filters out valid intermediate coordinate operations whose extents don't overlap. This affects pipeline composition and intermediate-CRS-based operation chaining.

## Reproduce
These produce identical output — NONE has no effect

``projinfo -s EPSG:7406 -t EPSG:5500 --spatial-test intersects --hide-ballpark --summary``
``projinfo -s EPSG:7406 -t EPSG:5500 --crs-extent-use none --spatial-test intersects --hide-ballpark --summary``

 Both returns 3 candidates. Expected with NONE: 6

## Description

Three locations hard-code extent intersection checks, ignoring `SourceTargetCRSExtentUse::NONE`:

### 1. `createHorizVerticalHorizPROJBased` call sites
**File:** `src/iso19111/operation/coordinateoperationfactory.cpp`

The `disallowEmptyIntersection` parameter is hard-coded to `true`:
```cpp
createHorizVerticalHorizPROJBased(..., interpolationGeogCRS, true);
//                                                           ^^^^
```

### 2. `createFromCRSCodesWithIntermediates` SQL filter
**File:** `src/iso19111/factory.cpp` 

SQL query always includes `intersects_bbox()` between the two intermediate operation extents:
```sql
AND intersects_bbox(south_lat1, west_lon1, north_lat1, east_lon1,
                    south_lat2, west_lon2, north_lat2, east_lon2) = 1
```

### 3. `createBetweenGeodeticCRSWithDatumBasedIntermediates` C++ filter
**File:** `src/iso19111/factory.cpp` 

Hard-coded bounding box intersection check between intermediate operation pairs:
```cpp
if (!bbox1->intersects(bbox2))
    continue;
```

## Proposed Fix

1. The two `createHorizVerticalHorizPROJBased` call sites hard-code `disallowEmptyIntersection` to `true`. Changed to `context.disallowEmptyIntersection()` so it respects the user's extent-use setting.

2. `createFromCRSCodesWithIntermediates` always applies `intersects_bbox()` in its SQL query between the two intermediate operation extents. Added a `skipIntermediateExtentIntersection` parameter (default `false`) to make this conditional.

3. `createBetweenGeodeticCRSWithDatumBasedIntermediates` always checks `bbox1->intersects(bbox2)` in C++ between intermediate operation pairs. Added the same `skipIntermediateExtentIntersection` parameter to skip this check when appropriate.

Both new parameters default to `false` (no behavior change) and are set to `!context.disallowEmptyIntersection()` at call sites, which evaluates to `true` only when `SourceTargetCRSExtentUse::NONE`.

#### Environment Information

- PROJ version 9.8.0

phaarnes:
This PR includes a fix for an issue where PROJ did not fully respect the argument to ignore extents.

- ✅  Closes #4697 
- ✅  Tests added


rouault:
What is the actual use case for this change ? The new added test doesn't seem very compelling to me. Why would we want to get transformations where the extent of the horizontal one doesn't intersect  the extent of the vertical one

phaarnes:
> What is the actual use case for this change ? The new added test doesn't seem very compelling to me. Why would we want to get transformations where the extent of the horizontal one doesn't intersect the extent of the vertical one

That’s a fair point, Even. I totally agree that from a strictly geodetic standpoint, this behavior is indeed non-standard. However, this change is driven by real-world "Custom Authority" scenarios where users define transformations for specific sub-regions in a custom proj.db.


**The Core Argument:** 

API Consistency
The strongest motivation here is that SourceTargetCRSExtentUse::NONE currently behaves as a partial "no-op." While it disables filtering for the top-level source/target, the three code paths I modified still enforce "silent" intersection checks for intermediate sub-operations within compound CRS pipelines.

This creates a logical inconsistency: if a user explicitly requests NONE, the engine should respect that intent throughout the entire pipeline construction, rather than re-introducing extent-based filtering for internal steps.

The Use Case:

We see historical datasets using custom CRSs and/or custom transformations where registered extents for sub-operations have very narrow margins or small gaps. When a user sets crs_extent_use=NONE, they are signaling that they trust these operations for their specific workflow, regardless of whether the registered extents overlap perfectly.

**The Test Case**

I agree, its not the best. The Alaska/CONUS test (7406 → 5500) was chosen simply because it is reproducible using only the standard EPSG dataset. It demonstrates the bug clearly: currently, NONE produces the exact same result as the default filtering because of these internal checks.

I’m happy to adjust the approach and revise the test if you’d prefer a different approch or a more targeted example.

rouault:
ok thanks for the context . You just need to update the scripts/reference_exported_symbols.txt file to make CI happy

PR #4700, opened 2026-03-06
Database: update to EPSG 12.053

rouault:


jjimenezshaw (approved):
Nice. Another site localization implemented via a Local Orthographic.
Is it going to become popular? O_O

PR #4705, opened 2026-03-07
Check for NULL from proj_as_proj_string in proj_factors derived CRS handling

ClayWarren:
### Motivation

- A code path in `proj_factors()` for derived projected CRS constructed a `std::string` from the return value of `proj_as_proj_string()` without checking for `NULL`, which can occur when an object is not exportable or on export error and causes undefined behavior/crash. 
- The change prevents denial-of-service / crash scenarios when processing crafted or non-exportable CRS definitions while preserving existing functionality.

### Description

- Add a `NULL` check for the result of `proj_as_proj_string()` and return safely (after cleaning up) if export fails, avoiding construction of `std::string` from `NULL` in the derived projected CRS branch. (modified `src/factors.cpp`).
- When rebuilding an operation from the trimmed PROJ string, check the result of `proj_create()` and return safely on failure instead of relying only on `assert()` (modified `src/factors.cpp`).
- Add a regression unit test that constructs a derived projected CRS that cannot be exported to a PROJ string and asserts that `proj_factors()` returns an error condition rather than crashing (added test in `test/unit/gie_self_tests.cpp`).

### Testing

- Configured with CMake (`-DCMAKE_BUILD_TYPE=Release`) successfully in this environment. (succeeded)
- Built the project with `cmake --build` successfully and produced binaries. (succeeded)
- Exercised the produced `proj` binary with a simple projection invocation (`./build/bin/proj +proj=merc +ellps=WGS84`) which produced expected numeric output. (succeeded)

- [X] AI/LLM (Codex Security in research preview) supported my development of this PR
- [X] Tests added

### DIFF
--- a/src/factors.cpp
+++ b/src/factors.cpp
@@ -257,8 +257,13 @@ PJ_FACTORS proj_factors(PJ *P, PJ_COORD lp) {
             proj_destroy(cs);
 
             // Remove trailing unit conversion and axis swap
-            std::string osProjStr =
+            const char *pszProjStr =
                 proj_as_proj_string(ctx, newOp, PJ_PROJ_5, nullptr);
+            if (!pszProjStr) {
+                proj_destroy(newOp);
+                return factors;
+            }
+            std::string osProjStr = pszProjStr;
             bool resized = false;
             for (const char *suffix :
                  {" +step +proj=unitconvert", " +step +proj=axiswap"}) {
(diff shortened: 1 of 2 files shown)


rouault:
@ClayWarren Why did you delete this PR ? This looked good to me

ClayWarren:
I deleted my fork, I'm still learning.

PR #4708, opened 2026-03-10
Chain vertical CRS transformations through intermediate same-datum vertical CRS

### LINKED ISSUE #4707: Chain vertical CRS transformations through intermediate vertical CRS sharing the same datum
## Summary

When transforming between two vertical CRSs and no direct registered operation exists for the exact pair, PROJ should look for registered operations involving a different vertical CRS that shares the same datum as the source or target. The intermediate CRS may differ in axis direction (height vs depth), units (metres vs feet), or both. Currently PROJ falls back to a ballpark transformation in these cases, discarding available registered operations.

## Example

**EPSG:5705** (Baltic 1977 height) → **EPSG:5706** (Caspian depth)

The EPSG registry contains operation **EPSG:5438** which transforms 5705 → **5611** (Caspian *height*). CRS 5611 and 5706 share the same datum (Caspian Sea) but differ in axis direction (UP vs DOWN).

**Expected:** PROJ composes `5705 →(EPSG:5438)→ 5611 →(axis/unit conversion)→ 5706`
**Actual:** PROJ returns a ballpark vertical transformation, discarding the registered operation.

### Current Behaviour

> The registered height→height operation works correctly:
>
> ```python
> from pyproj.transformer import TransformerGroup
>
> # ✅ Works: Baltic 1977 height → Caspian height (EPSG:5438)
> tg = TransformerGroup("EPSG:5705", "EPSG:5611", always_xy=True)
> t = tg.transformers[0]
> print(t.to_proj4())                   # +proj=geogoffset +dh=28
> print(t.transform(50.0, 40.0, 100))   # (50.0, 40.0, 128.0) ← correct
> ```
>
> But height→depth on the same datum falls back to ballpark:
>
> ```python
> # ❌ Fails: Baltic 1977 height → Caspian depth (same datum, axis down)
> tg = TransformerGroup("EPSG:5705", "EPSG:5706", always_xy=True)
> t = tg.transformers[0]
> print(t.description)                  # "... (ballpark vertical transformation)"
> print(t.to_proj4())                   # +proj=affine +s33=-1 ← geogoffset lost
> print(t.transform(50.0, 40.0, 100))   # (50.0, 40.0, -100.0) ← wrong: should be -128.0
> ```

### Should Be

> PROJ chains the registered transformation with the axis conversion:
>
> ```python
> # ✅ 5705 → 5611 (height → height): uses registered EPSG:5438
> print(t.to_proj4())                   # +proj=geogoffset +dh=28
> print(t.transform(50.0, 40.0, 100))   # (50.0, 40.0, 128.0)
>
> # ✅ 5705 → 5706 (height → depth): chains EPSG:5438 + axis conversion
> print(t.description)                  # Baltic 1977 height to Caspian height (1)
>                                        #   + Conversion from Caspian height to Caspian depth
> print(t.to_proj4())                   # +proj=pipeline +step +proj=geogoffset +dh=28
>                                        #   +step +proj=axisswap +order=1,2,-3
> print(t.transform(50.0, 40.0, 100))   # (50.0, 40.0, -128.0)
> ```

## Affected Scenario

Any vertical-to-vertical transformation where:
1. No direct registered operation exists between the exact source/target CRS pair, **but**
2. A registered operation exists to/from a variant vertical CRS that shares the same datum as the target (or source), differing in axis direction, units, or both.

## Proposed Behaviour

When searching for vertical-to-vertical operations, if no direct result is found, PROJ should:

1. **Strategy 1 (pivot on target datum):** Find candidate vertical CRSs sharing the *target's* datum. If a registered operation exists from the source to any candidate, compose it with the candidate→target axis/unit conversion.
2. **Strategy 2 (pivot on source datum):** Find candidate vertical CRSs sharing the *source's* datum. If a registered operation exists from any candidate to the target, compose the source→candidate axis/unit conversion with that operation.

This mirrors the existing logic in `createOperationsGeogToVertWithIntermediateVert` (which handles geographic→vertical paths) but extends it to vertical→vertical paths.

phaarnes:
## Summary

When transforming between two vertical CRSs and no direct registered operation exists for the exact pair, PROJ now searches for registered operations involving an intermediate vertical CRS that shares the same datum as the source or target. The intermediate CRS may differ in axis direction (height vs depth), units (metres vs feet), or both. Previously PROJ fell back to a ballpark transformation, discarding available registered operations.

- ✅ Closes #4707
- ✅ Tests added
- [x] Developed with AI-assistance (Claude Opus 4.6)

## Example

**EPSG:5705** (Baltic 1977 height) → **EPSG:5706** (Caspian depth)

EPSG:5438 transforms 5705 → 5611 (Caspian *height*). CRS 5611 and 5706 share the Caspian datum but differ in axis direction. PROJ now composes:

`5705 →(EPSG:5438)→ 5611 →(height-to-depth axisswap)→ 5706`

### Before (ballpark fallback)

> ```
> +proj=affine +s33=-1
> (50.0, 40.0, 100) → (50.0, 40.0, -100.0)  ← wrong
> ```

### After (registered operation + axis conversion)

> ```
> +proj=pipeline +step +proj=geogoffset +dh=28 +step +proj=axisswap +order=1,2,-3
> (50.0, 40.0, 100) → (50.0, 40.0, -128.0)  ← correct
> ```

## Implementation

New method `createOperationsVertToVertWithIntermediateVert` in `coordinateoperationfactory.cpp`, using two strategies:

1. **Pivot on target datum**: Find candidate vertical CRSs sharing the target's datum. If a registered operation exists from the source to any candidate, compose it with the candidate→target axis/unit conversion.
2. **Pivot on source datum**: Find candidate vertical CRSs sharing the source's datum. If a registered operation exists from any candidate to the target, compose the source→candidate conversion with that operation.

This mirrors the existing `createOperationsGeogToVertWithIntermediateVert` (geographic→vertical paths) but extends it to vertical→vertical paths. Uses the existing `findCandidateVertCRSForDatum` infrastructure.

## Changes

- **`src/iso19111/operation/coordinateoperationfactory.cpp`**: Added `createOperationsVertToVertWithIntermediateVert` method with anti-recursion guard; invoked from `createOperationsFromDatabaseWithVertCRS` when no direct result is found.
- **`test/unit/test_operationfactory.cpp`**: Two new tests — Strategy 1 (5705→5706, height→depth) and Strategy 2 (5706→5705, depth→height).
- **`NEWS.md`**: Added entry under 9.8.0 Updates.
- **`docs/source/operations/operations_computation.rst`**: Updated "Vertical CRS to a Vertical CRS" section documenting the new intermediate CRS pivot logic.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,6 +45,10 @@
 * `createOperationsCompoundToGeog()`: discard 2D-only transformations from interpolation
   3D CRS to target 3D CRS (#4677)
 
+* `createOperations()`: chain vertical CRS transformations through intermediate
+  same-datum vertical CRS when no direct operation exists, avoiding ballpark fallback
+  for CRS pairs differing in axis direction, units, or both
+
 
 ### Bug Fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,11 +45,6 @@
 * `createOperationsCompoundToGeog()`: discard 2D-only transformations from interpolation
   3D CRS to target 3D CRS (#4677)
 
-* `createOperations()`: chain vertical CRS transformations through intermediate
-  same-datum vertical CRS when no direct operation exists, avoiding ballpark fallback
-  for CRS pairs differing in axis direction, units, or both
-
-
 ### Bug Fixes
 
 * CRS identification: fix issue when identifying WKT from older EPSG releases with newer
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,6 +45,7 @@
 * `createOperationsCompoundToGeog()`: discard 2D-only transformations from interpolation
   3D CRS to target 3D CRS (#4677)
 
+
 ### Bug Fixes
 
 * CRS identification: fix issue when identifying WKT from older EPSG releases with newer
--- a/src/iso19111/operation/coordinateoperationfactory.cpp
+++ b/src/iso19111/operation/coordinateoperationfactory.cpp
@@ -4605,21 +4605,15 @@ std::vector<CoordinateOperationNNPtr> CoordinateOperationFactory::Private::
                 // first one, which is likely/hopefully the only one.
                 for (const auto &opFirst : opsFirst) {
                     if (hasIdentifiers(opFirst)) {
-                        if (candidateVert->_isEquivalentTo(
-                                targetCRS.get(),
-                                util::IComparable::Criterion::EQUIVALENT)) {
-                            res.emplace_back(opFirst);
-                        } else {
-                            try {
-                                res.emplace_back(
-                                    ConcatenatedOperation::
-                                        createComputeMetadata(
-                                            {opFirst, opsSecond.front()},
-                                            context
-                                                .disallowEmptyIntersection()));
-                            } catch (
-                                const InvalidOperationEmptyIntersection &) {
-                            }
+                        try {
+                            res.emplace_back(
+                                ConcatenatedOperation::
+                                    createComputeMetadata(
+                                        {opFirst, opsSecond.front()},
+                                        context
+                                            .disallowEmptyIntersection()));
+                        } catch (
+                            const InvalidOperationEmptyIntersection &) {
                         }
                     }
                 }
@@ -4648,21 +4642,15 @@ std::vector<CoordinateOperationNNPtr> CoordinateOperationFactory::Private::
             if (!opsFirst.empty()) {
                 for (const auto &opSecond : opsSecond) {
                     if (hasIdentifiers(opSecond)) {
-                        if (candidateVert->_isEquivalentTo(
-                                sourceCRS.get(),
-                                util::IComparable::Criterion::EQUIVALENT)) {
-                            res.emplace_back(opSecond);
-                        } else {
-                            try {
-                                res.emplace_back(
-                                    ConcatenatedOperation::
-                                        createComputeMetadata(
-                                            {opsFirst.front(), opSecond},
-                                            context
-                                                .disallowEmptyIntersection()));
-                            } catch (
-                                const InvalidOperationEmptyIntersection &) {
-                            }
+                        try {
+                            res.emplace_back(
+                                ConcatenatedOperation::
+                                    createComputeMetadata(
+                                        {opsFirst.front(), opSecond},
+                                        context
+                                            .disallowEmptyIntersection()));
+                        } catch (
+                            const InvalidOperationEmptyIntersection &) {
                         }
                     }
                 }
--- a/src/iso19111/operation/coordinateoperationfactory.cpp
+++ b/src/iso19111/operation/coordinateoperationfactory.cpp
@@ -4632,7 +4632,7 @@ std::vector<CoordinateOperationNNPtr> CoordinateOperationFactory::Private::
     // Strategy 2: pivot through candidates sharing the SOURCE's datum.
     // Find vertical CRSs on the same datum as the source, then look for
     // operations from each candidate to the target.
-    {
+    if (res.empty()) {
         auto candidatesSrc = findCandidateVertCRSForDatum(
             authFactory, vertSrc->datumNonNull(dbContext).get());
         for (const auto &candidateVert : candidatesSrc) {
--- a/src/iso19111/operation/coordinateoperationfactory.cpp
+++ b/src/iso19111/operation/coordinateoperationfactory.cpp
@@ -560,6 +560,7 @@ struct CoordinateOperationFactory::Private {
         bool inCreateOperationsWithDatumPivotAntiRecursion = false;
         bool inCreateOperationsGeogToVertWithAlternativeGeog = false;
         bool inCreateOperationsGeogToVertWithIntermediateVert = false;
+        bool inCreateOperationsVertToVertWithIntermediateVert = false;
         bool skipHorizontalTransformation = false;
         int nRecLevelCreateOperations = 0;
         std::map<std::pair<io::AuthorityFactory::ObjectType, std::string>,
@@ -647,6 +648,15 @@ struct CoordinateOperationFactory::Private {
     static std::vector<CoordinateOperationNNPtr>
     createOperationsGeogToVertWithAlternativeGeog(
         const crs::CRSNNPtr &sourceCRS, const crs::CRSNNPtr &targetCRS,
+        Context &context);
+
+    static std::vector<CoordinateOperationNNPtr>
+    createOperationsVertToVertWithIntermediateVert(
+        const crs::CRSNNPtr &sourceCRS,
+        const util::optional<common::DataEpoch> &sourceEpoch,
+        const crs::CRSNNPtr &targetCRS,
+        const util::optional<common::DataEpoch> &targetEpoch,
+        const crs::VerticalCRS *vertSrc, const crs::VerticalCRS *vertDst,
         Context &context);
 
     static void createOperationsFromDatabaseWithVertCRS(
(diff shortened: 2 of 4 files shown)


NEWS.md
```diff
@@ -45,6 +45,10 @@
 * `createOperationsCompoundToGeog()`: discard 2D-only transformations from interpolation
   3D CRS to target 3D CRS (#4677)
 
+* `createOperations()`: chain vertical CRS transformations through intermediate
+  same-datum vertical CRS when no direct operation exists, avoiding ballpark fallback
+  for CRS pairs differing in axis direction, units, or both
+
```
rouault: 9.8.0 has been released, so its content can't be changed:

```suggestion
```

src/iso19111/operation/coordinateoperationfactory.cpp
```diff
@@ -4530,6 +4540,123 @@ std::vector<CoordinateOperationNNPtr> CoordinateOperationFactory::Private::
...
+                targetCRS.get(), util::IComparable::Criterion::EQUIVALENT)) {
+            continue; // Skip the target itself — already tried by the caller.
+        }
+        auto opsFirst = createOperations(sourceCRS, sourceEpoch, candidateVert,
+                                         sourceEpoch, context);
+        if (!opsFirst.empty()) {
+            const auto opsSecond = createOperations(
+                candidateVert, sourceEpoch, targetCRS, targetEpoch, context);
+            if (!opsSecond.empty()) {
+                // The transformation from candidateVert to targetCRS should
```
rouault: I see this comment is the same as in createOperationsGeogToVertWithIntermediateVert(), but I actually don't understand it (although I'm its author. ahem...)  Either the comment is wrong, and we might do several iterations because opsFirst.size() > 1. Or it is right, and then we should add a break after the emplace_back()

test/unit/test_operationfactory.cpp
```diff
@@ -7275,6 +7275,62 @@ TEST(operation, vertCRS_to_vertCRS_New_Zealand_context) {
...
+    auto list = CoordinateOperationFactory::create()->createOperations(
+        // Baltic 1977 height
+        authFactory->createCoordinateReferenceSystem("5705"),
+        // Caspian depth (same datum as Caspian height 5611, but axis: down)
+        authFactory->createCoordinateReferenceSystem("5706"), ctxt);
+    ASSERT_GE(list.size(), 1U);
+    // Must NOT be a ballpark transformation
+    EXPECT_FALSE(list[0]->hasBallparkTransformation());
+    // The pipeline should contain geogoffset +dh=28 and axisswap
+    auto projStr =
```
rouault: would be more convincing to test the whole pipeline string

test/unit/test_operationfactory.cpp
```diff
@@ -7275,6 +7275,62 @@ TEST(operation, vertCRS_to_vertCRS_New_Zealand_context) {
...
+        authFactory->createCoordinateReferenceSystem("5706"),
+        // Baltic 1977 height
+        authFactory->createCoordinateReferenceSystem("5705"), ctxt);
+    ASSERT_GE(list.size(), 1U);
+    // Must NOT be a ballpark transformation
+    EXPECT_FALSE(list[0]->hasBallparkTransformation());
+    // The pipeline should contain axisswap (depth→height) and geogoffset
+    auto projStr =
+        list[0]->exportToPROJString(PROJStringFormatter::create().get());
+    EXPECT_TRUE(projStr.find("geogoffset") != std::string::npos) << projStr;
```
rouault: would be more convincing to test the whole pipeline string

rouault (commented):
is this generated with LLM ? Asking because the use of non ASCII characters like ↔ is quite typical of them. We have a checkbox in the pull request template about AI/LLM usage. It must be ticked when appropriate

phaarnes:
> is this generated with LLM ? Asking because the use of non ASCII characters like ↔ is quite typical of them. We have a checkbox in the pull request template about AI/LLM usage. It must be ticked when appropriate

Yes, this code was developed with assistance from an LLM, and I forgot to check the box. I will remove the non-ASCII characters and update the PR description. Thanks for pointing that out

test/unit/test_operationfactory.cpp
phaarnes: @rouault 
Good point, thanks. Full pipeline is now added.

test/unit/test_operationfactory.cpp
phaarnes: @rouault 
Good point, thanks. Full pipeline is now added.

src/iso19111/operation/coordinateoperationfactory.cpp
phaarnes: @rouault 
Yes, I copied that comment from `createOperationsGeogToVertWithIntermediateVert`, and I agree it a bit unclear.. 

The phrase "take only the first one" is meant to explain why we use `opsSecond.front()`. Since the conversion from the candidate vertical CRS to the target CRS is expected to be a trivial change (like switching units or axis direction), we only need the first result from that lookup.

The loop over `opsFirst`, on the other hand, is intentional and should not be broken out of early. Its purpose is to collect every registered operation from the source CRS to the candidate, since those operations may differ in accuracy or area of use, and we want all of them available for the later ranking/filtering step. In short, no `break` is needed after `emplace_back()` — the current logic is correct, but the comment should be rephrased a bit to make this clearer.

I suggest changing the comment to something like:

```cpp
        // Collect all registered source-to-candidate operations, which
        // may differ in accuracy or area of use, for later ranking.
        auto opsFirst = createOperations(sourceCRS, sourceEpoch, candidateVert,
                                         sourceEpoch, context);
        if (!opsFirst.empty()) {
            // The candidate-to-target conversion is expected to be a
            // trivial unit or axis change, so we only use the first
            // result (opsSecond.front()).
            const auto opsSecond = createOperations(
                candidateVert, sourceEpoch, targetCRS, targetEpoch, context);
```

Thoughts?

rouault: > but the comment should be rephrased a bit to make this clearer. I suggest changing the comment to something like:

yes much better

phaarnes: > > but the comment should be rephrased a bit to make this clearer. I suggest changing the comment to something like:
> 
> yes much better

I have change it now. Also change the comments in ``createOperationsGeogToVertWithIntermediateVert``. 
[3498a5d](https://github.com/OSGeo/PROJ/commit/3498a5d8b277d01489bd1639607f6c42eae8cc25)

NEWS.md
phaarnes: Fixed

test/unit/test_operationfactory.cpp
```diff
@@ -7275,6 +7275,65 @@ TEST(operation, vertCRS_to_vertCRS_New_Zealand_context) {
...
+        // Baltic 1977 height
+        authFactory->createCoordinateReferenceSystem("5705"), ctxt);
+    ASSERT_GE(list.size(), 1U);
+    // Must NOT be a ballpark transformation
+    EXPECT_FALSE(list[0]->hasBallparkTransformation());
+    // The exported pipeline is normalized to the same PROJ steps as the
+    // forward case.
+    auto projStr =
+        list[0]->exportToPROJString(PROJStringFormatter::create().get());
+    EXPECT_EQ(projStr, "+proj=pipeline +step +proj=geogoffset +dh=28 +step "
```
rouault: Either this pipeline is wrong , or the one at line 7301 is wrong. But they can't certainly be both correct, since one should be the opposite of the other, and here they are identical

NEWS.md
```diff
@@ -45,7 +45,6 @@
 * `createOperationsCompoundToGeog()`: discard 2D-only transformations from interpolation
   3D CRS to target 3D CRS (#4677)
 
-
```
rouault, line 48: please re-add the newline, so that there is no change at all in that file

phaarnes: Done

test/unit/test_operationfactory.cpp
phaarnes: The old code always tried Strategy 1 first and returned immediately if it found any result.
Forward (5705 height → 5706 depth): Strategy 1 found Caspian height (5611) on the target's datum. Result: geogoffset +dh=28 | axisswap -3.

Inverse (5706 depth → 5705 height): Strategy 1 again ran first, finding Baltic depth (5612) on the target's (Baltic) datum. It picked up a different registered transformation — inverse of EPSG:5440 — which also had dh=+28, and also appended axisswap. Result: geogoffset +dh=28 | axisswap -3 — identical to forward. Strategy 2 never got a chance to run because of the early return.

The fix implemented: 

Both strategies now run unconditionally and collect all candidates into the same result vector. No early returns within the function. The caller's existing filterAndSort ranks them by accuracy, area of use, etc., and picks the best. This matches the pattern used by createOperationsGeogToVertWithAlternativeGeog.

For the inverse case, Strategy 2 now contributes axisswap -3 | geogoffset +dh=-28, which filterAndSort correctly ranks above Strategy 1's identical-to-forward result.

A side benefit: for Kuwait PWD (5788↔5789), the collect-all approach finds the direct EPSG:7981 single-step path (dh=-4.25) instead of the 2-step chain via KOC CD height (dh=0.49 + dh=-4.74). Same numeric result, shorter pipeline.

Tests cover 6 CRS pairs (Caspian, AIOC95, KOC, Kuwait PWD, KOC ft, Black Sea) in both forward and inverse directions (12 checks total).

rouault: I was actually wrong (sorry but I'm super paranoid with anything LLM generated).  "+proj=pipeline +step +proj=geogoffset +dh=OFFSET +step +proj=axisswap +order=1,2,-3"   means new_Z = - (old_Z + OFFSET )   ,  and thus old_Z = - new_Z - OFFSET = - (new_Z + OFFSET). So such pipeline is its own inverse.  I've added a commit partially reverting your latest changes, ie using strategy 2 only when 1 didn't return anything, and cutting down added tests to the strict minimum demonstrating the new code

rouault (approved):

test/unit/test_operationfactory.cpp
phaarnes: No worries. I fully understand!

Issue #4716, opened 2026-03-18
Add proper way to list proj.db content


jidanni:
There is `proj`
```
       -l<[=id]>
              List projection identifiers that can be selected with +proj. proj
              -l=id  gives  expanded  description  of  projection id, e.g. proj
              -l=merc.
```
OK, but how about a way to list the things in
/usr/share/proj/proj.db ?

Currently users must use e.g.,
```
$ sqlite3 /usr/share/proj/proj.db .dump|grep TWD67.*121
INSERT INTO projected_crs VALUES('EPSG',3828,'TWD67 / TM2 zone 121'...
```

rouault:
This request isn't actionable. You have to be way more specific about what you want to list. Most top level information can be queried with projinfo. And if you want to know more yes using sqlite3 or ogrinfo or whatever user interface on top of sqlite3 please you is the way to go.

jidanni:
Oops, forgot about projinfo.

jidanni:
```
$ projinfo --area taiwan
Missing user string ...
```
"User string" is confusing. `--user $USER`?

(Anyway I found what I was looking for with
`projinfo --list-crs | grep TWD` )

mwtoews:
Correct usage of [`--area`](https://proj.org/en/stable/apps/projinfo.html#cmdoption-projinfo-area) is:
```
$ projinfo --list-crs --area taiwan
EPSG:3821 "TWD67"
EPSG:3822 "TWD97"
EPSG:3823 "TWD97"
EPSG:3824 "TWD97"
EPSG:3825 "TWD97 / TM2 zone 119"
EPSG:3826 "TWD97 / TM2 zone 121"
EPSG:3827 "TWD67 / TM2 zone 119"
EPSG:3828 "TWD67 / TM2 zone 121"
EPSG:3829 "Hu Tzu Shan 1950 / UTM zone 51N"
EPSG:4236 "Hu Tzu Shan 1950"
EPSG:8904 "TWVD 2001 height"
ESRI:115800 "TWD_1997"
```
[...]

PR #4719, opened 2026-03-19
Clarify usage of --area option in projinfo documentation

jidanni:
Adding a genuine working example.
See also https://github.com/OSGeo/PROJ/issues/4716#issuecomment-4085865276


rouault:
this would be better suited as an example in the section at the bottom of the page

PR #4722, opened 2026-03-20
CMake should only add cpack related options when PROJ is the top level project.

dbs4261:
Updated CMakeLists.txt to only set CPACK variables and uninstall target when proj is the top level project. This will help when proj is consumed using something like CMake's FetchContent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -501,51 +501,52 @@ install(FILES
 # "make dist" workalike
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


rouault (approved):

dbs4261:
Note: once CMake is updated to be [3.21 or greater](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html), the if statement should switch to `if(PROJECT_IS_TOP_LEVEL)` instead.

PR #4725, opened 2026-03-23
libproj: Add proper rpath for linux and macOS

mwestphal:
This add proper rpath in order to be able to use libsqlite not from system.

- [ ] Closes #xxxx
- [ ] Tests added

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -425,6 +425,8 @@ if (USE_ONLY_EMBEDDED_RESOURCE_FILES AND NOT EMBED_RESOURCE_FILES)
   message(WARNING "USE_ONLY_EMBEDDED_RESOURCE_FILES=ON set but EMBED_RESOURCE_FILES=OFF: some drivers will lack required resource files")
 endif()
 
+option(PROJ_SET_INSTALL_RELATIVE_RPATH "Whether to set rpath on installed libraries" OFF)
+
 ################################################################################
 # Build configured components
 ################################################################################
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -433,9 +433,9 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
     Path to an existing directory used to cache :file:`proj.db` to speed-up
     subsequent builds without modifications to source SQL files.
 
-.. option:: PROJ_SET_INSTALL_RELATIVE_RPATH
+.. option:: PROJ_SET_INSTALL_RELATIVE_RPATH=OFF
 
-    Whether to set rpath on installed libraries
+    Whether to set relative rpath on installed libraries.
 
 .. option:: TESTING_USE_NETWORK=ON
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -433,6 +433,10 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
     Path to an existing directory used to cache :file:`proj.db` to speed-up
     subsequent builds without modifications to source SQL files.
 
+.. option:: PROJ_SET_INSTALL_RELATIVE_RPATH
+
+    Whether to set rpath on installed libraries
+
 .. option:: TESTING_USE_NETWORK=ON
 
     .. versionadded:: 9.5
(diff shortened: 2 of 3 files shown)


src/lib_proj.cmake
```diff
@@ -388,6 +388,12 @@ add_library(proj
 )
 add_library(PROJ::proj ALIAS proj)
 
+if(APPLE)
```
rouault: I don't think setting INSTALL_RPATH unconditionnaly is the way to go (otherwise that would be a CMake default).

Would you mind replicating the logic we use for GDAL ? See https://github.com/OSGeo/gdal/blob/master/gdal.cmake lines 298 to 317

CMakeLists.txt
```diff
@@ -425,6 +425,8 @@ if (USE_ONLY_EMBEDDED_RESOURCE_FILES AND NOT EMBED_RESOURCE_FILES)
   message(WARNING "USE_ONLY_EMBEDDED_RESOURCE_FILES=ON set but EMBED_RESOURCE_FILES=OFF: some drivers will lack required resource files")
 endif()
 
+option(PROJ_SET_INSTALL_RELATIVE_RPATH "Whether the rpath of installed libraries should be written as a relative path to the library" OFF)
```
rouault:
```suggestion
option(PROJ_SET_INSTALL_RELATIVE_RPATH "Whether to set rpath on installed libraries" OFF)
```

mwtoews:
Have you checked if setting built-in [`-D CMAKE_BUILD_RPATH_USE_ORIGIN=ON`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_RPATH_USE_ORIGIN.html)? More docs on [BUILD_RPATH_USE_ORIGIN](https://cmake.org/cmake/help/latest/prop_tgt/BUILD_RPATH_USE_ORIGIN.html#prop_tgt:BUILD_RPATH_USE_ORIGIN).

I've tried it locally with Ubuntu, but I don't see anything with `objdump -x /tmp/inst/lib/libproj.so | grep 'R.*PATH'`, so perhaps I'm missing something. <s>I've also checked this PR too, also nothing found in libproj.so with objdump.</s> This PR has expected:
```
$ objdump -x /tmp/pr4725/lib/libproj.so | grep 'R.*PATH'
  RUNPATH              $ORIGIN
```

mwtoews:
I was unsuccessfully checking to see if built-in CMAKE variables could be used for this feature, but I can't see any. This PR does enable the feature. Could it be documented [here](https://github.com/OSGeo/PROJ/blob/master/docs/source/install.rst#cmake-configure-options)?

jjimenezshaw:
> I've tried it locally with Ubuntu, but I don't see anything with `objdump -x /tmp/inst/lib/libproj.so | grep 'R.*PATH'`,

Have you tried `readelf -d /tmp/inst/lib/libproj.so` ?

mwtoews:
@jjimenezshaw yes, tried that too. I believe that `readelf -d` and `objdump -x` are equivalent for inspecting RPATH/RUNPATH.

mwestphal:
BUILD_RPATH_USE_ORIGIN is for build rpath, not install rpath.

mwestphal:
> I was unsuccessfully checking to see if built-in CMAKE variables could be used for this feature, but I can't see any. This PR does enable the feature. Could it be documented [here](https://github.com/OSGeo/PROJ/blob/master/docs/source/install.rst#cmake-configure-options)?

done.

docs/source/install.rst
```diff
@@ -433,6 +433,10 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
     Path to an existing directory used to cache :file:`proj.db` to speed-up
     subsequent builds without modifications to source SQL files.
 
+.. option:: PROJ_SET_INSTALL_RELATIVE_RPATH
+
+    Whether to set rpath on installed libraries
```
mwtoews:
```suggestion
.. option:: PROJ_SET_INSTALL_RELATIVE_RPATH=OFF

    Whether to set relative rpath on installed libraries.
```

mwestphal:
I've been talking with CMake devs and this approach is wrong. `-DCMAKE_INSTALL_RPATH` should be enough to fix that locally and should not be handled by the project like that. Sorry for the noise.

PR #4736, opened 2026-04-02
DRAFT: Yet another MEGA (Make ETRS89 Great Again) fix!

rouault:
- Database: import aliases for coordinate operations involving ETRS89

- Yet another MEGA (Make ETRS89 Great Again) fix!
    
    I've been made aware that ETRS89 to MGI was semi broken since the EPSG
    update that added a ETRS89-AUT [2002] datum and migrated the 'MGI to
    ETRS89 (8)' transformation using the last released grid AT_GIS_GRID_2021_09_28
    to be 'MGI to ETRS89-AUT [2002]' and thus leaving user with the older
    transformation 'MGI to ETRS89 (5)'.
    This is just one manifestation of a problem that can be seen with ETRS89
    in the Netherlands, Romania, etc.
    Until EPSG perhaps figures out a plan to restore cleanly backward
    compatibility (they have been made aware of those issues), I've devised
    the following s/hack/strategy that seems to work (TM):
    
    When looking for ETRS89 (EPSG:4258 the datum ensemble) to XXXX
    transformations, also query the ``alias_name`` table for old names
    of transformations that start with 'ETRS89 to XXXX' or
    'XXXX to ETRS89', and are aliases to 'ETRS89-YYY to XXXX[something]'
    or 'XXXX to ETRS89-YYY[something]' newly named transformation. And then
    consider those 'migrated' transformations to be also used for plain
    ETRS89, because they used to be used for it!
    
    No new test cases, but actually reverting to old reference values before
    commit 047e5dbc74ef872828c06f24c6f67bb7b55432cf (update to EPSG 12.033)
    that changed them.

PR #4739, opened 2026-04-03
[DRAFT] [Backport 9.8] Yet another MEGA (Make ETRS89 Great Again) fix!

rouault:
Backport of https://github.com/OSGeo/PROJ/pull/4736


rouault:
will need rework as steps on the toes of https://github.com/OSGeo/PROJ/pull/4741. Actually if we pursue this PR, the best way forward would likely be to revert PR 4741 which is a (complex) revert

PR #4743, opened 2026-04-08
[9.8] Prepare for 9.8.1 release

rouault:
@kbevers if you can have a look and check I haven't messed up things (we must have slightly different sphinx or whatever versions, but that's likely fine)


NEWS.md
```diff
@@ -1,5 +1,48 @@
...
+
+* pj_obj_create: use database context if already open for grid name resolution (#4703)
+
+* Chain vertical CRS transformations through intermediate same-datum vertical CRS (#4711)
+  Helps for example for **EPSG:5705** (Baltic 1977 height) to **EPSG:5706** (Caspian depth)
+  by using intermediate operation from Baltic 1977 height to Caspian *height*
+
+### Documentation
+
+* Add a page specific to time dependent transformations (#4688)
```
kbevers: I generally leave documentation updates out of the release notes. In this case I guess this new page already has been online for a while

kbevers (commented):
Looks okay to me. I note that the release date is set for today. I would put that a bit in the future, leaving room for a release candidate.

rouault:
> I note that the release date is set for today. I would put that a bit in the future, leaving room for a release candidate.

I'm never sure tomorrow will happen, so I tend to put it at the date where the code is frozen :-)

NEWS.md
rouault: ok, remove

kbevers:
> I'm never sure tomorrow will happen, so I tend to put it at the date where the code is frozen :-)

These days you never really know... In the end it doesn't matter much, but it's different to what I usually do. No harm in mixing it up a bit :)

PR #4750, opened 2026-04-13
Separate the checksum value of proj.db.sql into its own file

dbs4261:
Separate the checksum value of proj.db.sql into its own file so that changes to the CMake and changes to the database are tracked separately.


rouault (approved):

data/sql/proj.db.sql.expected.md5
```diff
@@ -0,0 +1 @@
+497ccd15dc1072dce88a5bdc2856d307
```
jjimenezshaw, line 1: Is it a problem if there is an empty line afterwards? Some editors add it automatically (and PR kind of complains with "No newline at end of file")

dbs4261: Adding a new line is totally fine. The file is read with `file(STRINGS <FILENAME> <VARIABLE> LIMIT_COUNT 1)` so, based on the [CMake Documentation](https://cmake.org/cmake/help/v3.16/command/file.html#strings), only one line will be extracted from the file.

PR #4693, opened 2026-03-04
CMake: handle deprecated SQLite::SQLite3 target

jmckenna:
- CMake >= 4.3 deprecates the `SQLite::SQLite3` target (see [FindSQLite3.html](https://cmake.org/cmake/help/latest/module/FindSQLite3.html))
- using today's CMake 4.3.0-rc2, with Visual Studio, and PROJ master, gives the following warnings:
```
CMake Warning (dev) at src/lib_proj.cmake:596 (target_link_libraries):
  The library that is being linked to, SQLite::SQLite3, is marked as being
  deprecated by the owner.  The message provided by the developer is:

  The target name SQLite::SQLite3 is deprecated.  Please use SQLite3::SQLite3
  instead.

Call Stack (most recent call first):
  src/CMakeLists.txt:9 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at src/lib_proj.cmake:596 (target_link_libraries):
  The library that is being linked to, SQLite::SQLite3, is marked as being
  deprecated by the owner.  The message provided by the developer is:

  The target name SQLite::SQLite3 is deprecated.  Please use SQLite3::SQLite3
  instead.

Call Stack (most recent call first):
  src/CMakeLists.txt:9 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at src/lib_proj.cmake:596 (target_link_libraries):
  The library that is being linked to, SQLite::SQLite3, is marked as being
  deprecated by the owner.  The message provided by the developer is:

  The target name SQLite::SQLite3 is deprecated.  Please use SQLite3::SQLite3
  instead.

Call Stack (most recent call first):
  src/CMakeLists.txt:9 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at src/lib_proj.cmake:596 (target_link_libraries):
  The library that is being linked to, SQLite::SQLite3, is marked as being
  deprecated by the owner.  The message provided by the developer is:

  The target name SQLite::SQLite3 is deprecated.  Please use SQLite3::SQLite3
  instead.

Call Stack (most recent call first):
  src/CMakeLists.txt:9 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at src/lib_proj.cmake:596 (target_link_libraries):
  The library that is being linked to, SQLite::SQLite3, is marked as being
  deprecated by the owner.  The message provided by the developer is:

[...]

PR #4753, opened 2026-04-15
Fix SQLite3 version check so it is done by find_package

dbs4261:
Right now there is a separate version check from the find_package call for SQLite3. This can cause issues with some dependency providers, particularly FetchContent, that bypass the find_package version check because they do their own, and do not provide the package version variables that the CMake script currently checks.


CMakeLists.txt
```diff
@@ -238,19 +238,12 @@ if(DEFINED SQLITE3_LIBRARY)
   set(SQLite3_LIBRARY ${SQLITE3_LIBRARY})
 endif()
 
-find_package(SQLite3 REQUIRED)
+find_package(SQLite3 3.11 REQUIRED)
 
 if(NOT TARGET SQLite3::SQLite3) # CMake < 4.3
-  set_target_properties(SQLite::SQLite3 PROPERTIES IMPORTED_GLOBAL TRUE)
```
rouault, line 244: why do you remove that line ? There was some rationale for it, although I've already forgotten it

dbs4261: Good spotting, I should have mentioned that in the PR.

The `IMPORTED_GLOBAL` property changes the scope of a target. Based on [the documentation](https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_GLOBAL.html#prop_tgt:IMPORTED_GLOBAL), this allows a find_package call in a subdirectory to supply that dependency to the entire project. I'm not sure if, historically, the dependencies were found in a different CMake script, but they are currently in the top level script for the project, so the targets created will always be accessible.

Now somebody might be using PROJ as a dependency with FetchContent or git submodules, and they would no longer be able to use `SQLite::SQLite3` or `SQLite3::SQLite3` without their own `find_package` statement (Or by setting `CMAKE_FIND_PACKAGE_TARGETS_GLOBAL=ON`).

rouault: I dug up where the need for that line was discussed: https://github.com/OSGeo/PROJ/pull/4693#discussion_r2886660452

dbs4261: Phoey, I think that implies I'm missing a CMake policy or something. I'll need to poke around some more.

dbs4261: Ugh, its not even a CMake policy, its a [change in the C++ part of cmake from verson 3.18](https://cmake.org/cmake/help/latest/release/3.18.html). 

So, the obvious answer is to bump the CMake required version to 3.18 from mid July 2020. Thats almost 6 years ago, should be easy, right? No. CI runs on Ubuntu 20.04LTS, which provides v3.16.2 from December 2019. Should we update it? Yes, 20.04 has been dead for a year. But, I haven't worked with CI before, so I think I need to draw the edge of my scope here for now.

Alternate plan: check if SQLite::SQLite3 is imported. If so, check the CMake version, if less than 3.18 then make it global and produce a dev warning. I pushed this approach.

rouault: @dbs4261 Thanks for having dug into those gory details!

mwtoews: I can look into a follow-up to bump the minimum CMake to 3.18 for the PROJ 9.9 series. There are several other tasks related to resetting the minimum. But keep this PR assuming that the minimum version is unchanged.

jjimenezshaw: > So, the obvious answer is to bump the CMake required version to 3.18 from mid July 2020. Thats almost 6 years ago, should be easy, right? No. CI runs on Ubuntu 20.04LTS, which provides v3.16.2 from December 2019. Should we update it? Yes, 20.04 has been dead for a year. But, I haven't worked with CI before, so I think I need to draw the edge of my scope here for now.

Is PROJ CI still on Ubuntu 20.04? Really?

rouault: > Is PROJ CI still on Ubuntu 20.04? Really?

might be some remains. But PROJ claims to be C++17, so this is one of the oldest environment that has C++17 capabilities

dbs4261: If you are going to bump up the CMake minimum version, lets consider going to 3.22 (Nov 2021), the version Ubuntu 22.04LTS uses.  I'm curious to see what changes are necessary, so I would appreciate it if you could at me in that PR.

As an aside, I'd like to bump to 3.31 (Nov 2024) for a number of reasons, mainly the standardization of the codegen target. But that is modern enough that even Ubuntu 24.04LTS would need to specifically install a more modern version. So that work might just exist in a fork for a while..

mwtoews: @dbs4261 see #4756 for a discussion, and describe some use-cases of necessitating modern versions of CMake; thanks!

mwtoews (approved):
Thanks! The CMake<3.18 logic is nice to have for now, but we'd probably like to increase the minimum CMake version as already discussed.

PR #4764, opened 2026-04-21
Database: update to EPSG v12.055

rouault:
- does *not* address ETRS89 issues
- rename JGD2011 to JGD2024... sigh ... and creates a distinct JGD2024 height from existing JGD2011 height
- adds gllat_2025 and glmsl_2025 (CC @kbevers)

Besides that, tune import script to not import ISO geodetic registry codes as name aliases.


rouault:
Regarding Japan, for JGD2024 (geographic) to "JGD2024 (vertical) height"  (what used to be JGD2011 to JGD2024 vertical), the recommended way (new concatenated operation 11441) is to apply :
- the geoid model jp_gsi_gsigeo2011.tif to get JGD2011 height (which we have),
- followed by EPSG:11440 "JGD2011 (vertical) height to JGD2024 (vertical) height" using vertical offset file hyokorevBM_jgd2024_h.par, which we don't have

rather than applying jp_gsi_jpgeo2024.tif which we have...

EPSG:11441 "This concatenated operation is the path recommended by GSI for deriving JGD2024 (vertical) height from GNSS-derived ellipsoidal height in accordance with Survey Act requirements, as using the JPGEO2024 geoid separation model directly (JGD2024 to JGD2024 (vertical) height (1), CT code 11334) is not aligned with Survey Act requirements for coordinates to be at epoch 2011.395."

Japanese grid (& license) hunters, the floor is yours... (CC @tohka @hernando @jjimenezshaw )

jjimenezshaw:
> * using vertical offset file hyokorevBM_jgd2024_h.par, which we don't have

Apparently it is in https://www.gsi.go.jp/sokuchikijun/sokuchikijun41012.html
However, I don't understand the PatchJGD format (using `par` extension). Opening with a text editor it seems to be  just a grid of corrections, but there is some binary code (or non UTF8 encoding) at the beginning and no idea how to decode the MeshCode.
About the license, as always, it doesn't say anything.

We would need somebody with more knowledge about the PatchJGD format and the license of the file.

There is a program to open PatchJGD files. I cannot open it, as it is only Windows. And I will not understand the Japanese menus.

jjimenezshaw:
> * rename JGD2011 to JGD2024... sigh ... and creates a distinct JGD2024 height from existing JGD2011 height

:open_mouth: :facepalm: 

"On adoption of JGD2024 (vertical) with effect from 1st April 2025, the name of this geodetic CRS was changed from JGD2011 to JGD2024; there is no change of definition or of coordinate values"

Issue #3822, opened 2023-07-12
Transverse Mercator Zoned Grid System (CRS EPSG:32600/32700) not implemented


olsen232:
In a service I help maintain, users are able to reproject geometries from EPSG:4326 into the CRS of their choice for export. One relatively common cause of error is that users select either EPSG:32600 or EPSG:32700 as the output CRS.

#### Example of problem
The reprojection EPSG:4326 -> EPSG:32600 (Northern hemisphere) doesn't seem to work at all:

```
$ cs2cs 'EPSG:4326' 'EPSG:32600'
0 0
*       * inf
10 10
*       * inf
-10 -10
*       * inf
```
The `*       *` is apparently the default way that cs2cs indicates that an error has occurred.

Similarly for EPSG:32700 (southern hemisphere)

```
$ cs2cs 'EPSG:4326' 'EPSG:32700'
0 0
*       * inf
10 10
*       * inf
-10 -10
*       * inf
```

I don't know enough to know if these transforms are supposed to work but broken / misconfigured somehow, or if they are not supposed to work.

Based on this projinfo output, it looks like such a transformation should be possible: 
```
$ projinfo -s EPSG:4326 -t EPSG:32600
Candidate operations found: 1
-------------------------------------
Operation No. 1:

EPSG:16000, UTM grid system (northern hemisphere), 0 m, Northern hemisphere between equator and 84°N, onshore and offshore.

PROJ string:
Error when exporting to PROJ string: Unsupported conversion method: Transverse Mercator Zoned Grid System

WKT2:2019 string:
CONVERSION["UTM grid system (northern hemisphere)",
    METHOD["Transverse Mercator Zoned Grid System",
        ID["EPSG",9824]],
    PARAMETER["Latitude of natural origin",0,
        ANGLEUNIT["degree",0.0174532925199433],
        ID["EPSG",8801]],
    PARAMETER["Initial longitude",-180,
        ANGLEUNIT["degree",0.0174532925199433],
        ID["EPSG",8830]],
    PARAMETER["Zone width",6,
        ANGLEUNIT["degree",0.0174532925199433],
        ID["EPSG",8831]],
    PARAMETER["Scale factor at natural origin",0.9996,
        SCALEUNIT["unity",1],
        ID["EPSG",8805]],
    PARAMETER["False easting",500000,
        LENGTHUNIT["metre",1],
        ID["EPSG",8806]],
    PARAMETER["False northing",0,
        LENGTHUNIT["metre",1],
[...]

PR #4770, opened 2026-05-02
Add interrupted projection for "sinu", "tmerc", "cass" and "poly"

### LINKED ISSUE #4533: Interrupted projection proposal.
Hi

I have been developing an interrupted projection. We could consider it a "metaprojection", because it just call the proper base projection adjusted to create properly the gores (apparently that is the term).

Before I continue with it and do the documentation and tests (that is a lot of work), I would like to know if it makes sense.

It so far supports the projections "sinu", "tmerc", "cass" and "poly". Adding more cylindrical base projections should be easy.

Let me show you some outputs done with QGIS.

`+proj=interrupted +base=sinu +gores=4`
Image

`+proj=interrupted +base=tmerc +gores=4`
Image

`+proj=interrupted +base=cass +gores=4`
Image

`+proj=interrupted +base=poly +gores=4`
Image

`+proj=interrupted +base=tmerc +gores=20`
Image

You can also specify non uniform gores, and change the central meridian.
`+proj=interrupted +base=sinu +gores=2,3,2 +lon_0=45`
Image

Daan Strebe suggested me these papers:
https://cartographicperspectives.org/index.php/journal/article/view/cp42-strebe
https://research.manchester.ac.uk/en/publications/evaluating-map-projections-for-globe-making

The main use case could be (I think) in visual software like QGIS, to directly display the gores as you need them. I do not know how many people need gores. I personally like it, maybe because it makes explicit the connection to a real globe (watch https://www.youtube.com/watch?v=4RWcWSN4HhI at 1:10)

If you find it useful I can continue with it. If not, I can just leave [the branch](https://github.com/jjimenezshaw/PROJ/tree/interrupted) as it is now, and anyone interested can take the file interrupted.cpp and compile it.

jjimenezshaw:
Creates a "meta" projection to make an interrupted projection with gores. It can be used with `sinu`, `tmerc`, `cass`, and `poly` projections, using their algorithms. It could be easily extended for other pseudo-cylindrical projections.

The definition of the gores can be just the number, like in `+gores=3` or with a comma separated list, like in `+gores=1,3,1`. For the latter, the values are relative sizes, that will be translated into the 360 degrees.

There are several graphic examples in #4533

See that a single point in the frontier of two gores can only be projected in one side or the other. But not in both. That makes that meridians right in the division of gores are plotted only in one side. As a hack you can define another meridian with a tiny offset. Due to rounding problems it is not defined if the meridian is plotted on the right or left.

[Evaluating Map Projections for Globemaking](https://www.tandfonline.com/doi/full/10.1080/00087041.2024.2436324)

- [x] Closes #4533
- [x] Tests added


src/projections/interrupted.cpp
```diff
@@ -0,0 +1,203 @@
...
+    Q->base_pj = get_fun_ptr()(nullptr);
+    if (Q->base_pj == nullptr)
+        return interrupted_destructor(P, PROJ_ERR_OTHER /*ENOMEM*/);
+    Q->base_pj->ctx = P->ctx;
+    Q->base_pj->es = P->es;
+    Q->base_pj->one_es = P->one_es;
+    Q->base_pj->k0 = P->k0;
+    Q->base_pj->f = P->f;
+    Q->base_pj->f2 = P->f2;
+    Q->base_pj->n = P->n;
```
jjimenezshaw: Those parameters are the ones I found needed for the ellipsoidal projections. Do I need more? Is there already any function that copies them?

jjimenezshaw:
I can squash it in one commit before merging

busstoptaktik:
@jjimenezshaw I think I understand the usefulness of interrupting a cylindrical projection along a set of longitudes outside the domain of interest (i.e. ocean vs. continent), in order to limit the poleward deformation of wide area coverage maps.

I just don't see why this should be useful for transverse mercator, which does not deform significantly poleward (but rather as one approaches +/-90 degreees from the central meridian.

Wouldn't it make more sense supporting `proj=merc`, rather than `proj=tmerc`?

jjimenezshaw:
@busstoptaktik The usefulness of this meta-projection is for projections that do not have "vertical" (parallel to the Y axis) meridians. For Mercator it does not bring any value: the final output would be the same.

Daan Strebe (who suggested me the paper mentioned above https://www.tandfonline.com/doi/full/10.1080/00087041.2024.2436324 ) mentioned that transverse Mercator, being conformal, is better than sinusoidal if you want to later put the gores in a physical sphere: the parallels do not have an angle in the union of the gores.

busstoptaktik:
@jjimenezshaw Thanks! (now still confused, but at a higher level)

wrenoud:
In the process of exploring an implementation for #3822, which I have working on the branch https://github.com/OSGeo/PROJ/compare/master...wrenoud:PROJ:zoned-grid-system, I realized the zoned grid system implementation looks basically like this PR, but with a specified gore spacing and underlaying projection.

Here the gore spacing is based on the width at the equator of the gore. The EPSG method 9824 (used in [EPSG:32600](https://spatialreference.org/ref/epsg/32600/) and [EPSG:32700](https://spatialreference.org/ref/epsg/32700/)) specify the zone spacing as 1'000'000 units. Adding a parameter that defaults to "auto" but can be overridden would solve the case for method 9824.

So I'm wondering if I should rebase on this and add an "alias" projection, kind of like how "+proj=utm" maps to "+proj=tmerc" but specifies the defaults?

jjimenezshaw:
> In the process of exploring an implementation for #3822, which I have working on the branch [master...wrenoud:PROJ:zoned-grid-system](https://github.com/OSGeo/PROJ/compare/master...wrenoud:PROJ:zoned-grid-system), I realized the zoned grid system implementation looks basically like this PR, but with a specified gore spacing and underlaying projection.
> 
> Here the gore spacing is based on the width at the equator of the gore. The EPSG method 9824 (used in [EPSG:32600](https://spatialreference.org/ref/epsg/32600/) and [EPSG:32700](https://spatialreference.org/ref/epsg/32700/)) specify the zone spacing as 1'000'000 units. Adding a parameter that defaults to "auto" but can be overridden would solve the case for method 9824.
> 
> So I'm wondering if I should rebase on this and add an "alias" projection, kind of like how "+proj=utm" maps to "+proj=tmerc" but specifies the defaults?

Thanks @wrenoud . You are right, what both "projections" are doing is very similar: use a current cylindrical projection (in your case only transverse Mercator) and repeat it along the world. However I think it is going to be more clear to keep them separated. The "noise" that this PR would need does not compensate the few code that you would remove from your PR.

Also I think the "target audience" have different needs. This PR is more oriented to display the whole world divided in gores (with the corresponding settings), ready to be printed. On the other hand your PR is more oriented to implement EPSG:32600 and EPSG:32700, used in maps at a much bigger scale.

PR #4771, opened 2026-05-03
Add get_build_git_revision in emscripten compilation

jjimenezshaw:
To be sure what is run in emscripten outputs, this change adds a function `get_build_git_revision` that returns the git revision.
It is oriented to be used in github actions, but not only. In case the git revision is not readable (due to git configuration) or directly not available (not a git clone), the returned value is an empty string.

Would this functionality be useful in any compilation?

 - [x] AI (Copilot or something similar) supported my development of this PR. See our [policy about AI tool use](https://proj.org/community/ai_tool_policy.html). Use of AI tools *must* be indicated. -> I asked Gemini the best way to ask git the revision, knowing that there are corner cases. In github actions apparently it is better to use `${{ github.sha }}` than asking git directly.
- [ ] Closes #xxxx
- [ ] Tests added

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -35,6 +35,7 @@ jobs:
       - name: Run docker build
         run: |
           docker run --user $(id -u):$(id -g) --rm \
+            -e GITHUB_SHA=${{ github.sha }} \
             -v "$PWD":/build/proj_src \
             -v "$PWD"/wasm_out:/build/install \
             proj-emscripten-builder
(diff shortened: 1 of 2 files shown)


kbevers:
> Would this functionality be useful in any compilation?

Yes! I think it would be fantastic if `pj_get_release` could spit out the git revision as well.

rouault:
GDAL's cmake/helpers/GdalVersion.cmake and cmake/helpers/generate_gdal_version_h.cmake could potentially be of help to replicate GDAL's gdalinfo --version working (might be overkill though)

PR #4774, opened 2026-05-10
Add git hash in release string

jjimenezshaw:
- [x] AI (Copilot or something similar) supported my development of this PR. See our [policy about AI tool use](https://proj.org/community/ai_tool_policy.html). Use of AI tools *must* be indicated. -> gemini helped to call git in cmake.
- [ ] Closes #xxxx
- [ ] Tests added

Continuing #4771 , this PR adds the git hash to the release string
For instance:
```
./bin/proj
Rel. 9.9.0 (a51f7b60), September 15th, 2026
usage: proj [-bdeEfiIlmorsStTvVwW [args]] [+opt[=arg] ...] [file ...]
```

In case the git tree is "dirty" (there are modified files), an asterisk `*` is added after the hash.
If this feature is not needed, it can be removed.
```
./bin/proj
Rel. 9.9.0 (a51f7b60*), September 15th, 2026
usage: proj [-bdeEfiIlmorsStTvVwW [args]] [+opt[=arg] ...] [file ...]
```

To help in the docker images, that may not have proper access to git, the environment variable `PROJ_GITHUB_SHA` is used before asking git for the revision. (`.git` is included in `.dockerignore`, so it is not found)

### DIFF
--- a/cmake/ProjConfig.cmake
+++ b/cmake/ProjConfig.cmake
@@ -51,4 +51,54 @@ if(EXISTS ${AUTOCONF_PROJ_CONFIG_H})
     "before CMake's build.")
 endif()
 
+# Get the git hash to write it in the release message.
+set(PROJ_GIT_HASH "")
+if(DEFINED ENV{PROJ_GITHUB_SHA})
+    string(SUBSTRING "$ENV{PROJ_GITHUB_SHA}" 0 7 PROJ_GIT_HASH)
+else()
+    find_package(Git QUIET)
+    if(Git_FOUND AND EXISTS "${PROJ_SOURCE_DIR}/.git")
+        # Get the short hash
+        execute_process(
+            COMMAND ${GIT_EXECUTABLE} rev-parse --short HEAD
+            WORKING_DIRECTORY "${PROJ_SOURCE_DIR}"
+            RESULT_VARIABLE GIT_REV_RESULT
+            OUTPUT_VARIABLE PROJ_GIT_HASH
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+            ERROR_QUIET
+        )
+
+        if(GIT_REV_RESULT EQUAL 0)
+            # Check if the working directory is dirty
+            execute_process(
+                COMMAND ${GIT_EXECUTABLE} status --porcelain --untracked-files=no
+                WORKING_DIRECTORY "${PROJ_SOURCE_DIR}"
+                RESULT_VARIABLE GIT_STATUS_RESULT
+                OUTPUT_VARIABLE GIT_STATUS_OUTPUT
+                OUTPUT_STRIP_TRAILING_WHITESPACE
+                ERROR_QUIET
+            )
+
+            # If the command succeeds and outputs anything, the tree is dirty
+            if(GIT_STATUS_RESULT EQUAL 0 AND NOT GIT_STATUS_OUTPUT STREQUAL "")
+                string(APPEND PROJ_GIT_HASH "*")
+            endif()
+        else()
+            set(PROJ_GIT_HASH "")
+        endif()
+    endif()
+endif()
+if(PROJ_GIT_HASH)
+    add_compile_definitions(PROJ_GIT_HASH=\"${PROJ_GIT_HASH}\")
+endif()
+message(PROJ_GIT_HASH=\"${PROJ_GIT_HASH}\")
+
+# Tell CMake to re-configure if the Git HEAD or index changes
+if(Git_FOUND AND EXISTS "${PROJ_SOURCE_DIR}/.git")
+    set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS
+        "${PROJ_SOURCE_DIR}/.git/HEAD" # for the hash
+        "${PROJ_SOURCE_DIR}/.git/index" # for the *
+    )
+endif()
+
 configure_file(cmake/proj_config.cmake.in src/proj_config.h)
--- a/src/release.cpp
+++ b/src/release.cpp
@@ -6,8 +6,15 @@
 #define STR_HELPER(x) #x
 #define STR(x) STR_HELPER(x)
 
-char const pj_release[] = "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(
-    PROJ_VERSION_MINOR) "." STR(PROJ_VERSION_PATCH) ", "
-                                                    "September 15th, 2026";
+#ifdef PROJ_GIT_HASH
+#define PROJ_GIT_REV_STR " (" PROJ_GIT_HASH ")"
+#else
+#define PROJ_GIT_REV_STR ""
+#endif
+
+char const pj_release[] =
+    "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(PROJ_VERSION_MINOR) "." STR(
+        PROJ_VERSION_PATCH) PROJ_GIT_REV_STR ", "
+                                             "September 15th, 2026";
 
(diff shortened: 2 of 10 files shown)


cmake/ProjConfig.cmake
```diff
@@ -51,4 +51,54 @@ if(EXISTS ${AUTOCONF_PROJ_CONFIG_H})
...
+                RESULT_VARIABLE GIT_STATUS_RESULT
+                OUTPUT_VARIABLE GIT_STATUS_OUTPUT
+                OUTPUT_STRIP_TRAILING_WHITESPACE
+                ERROR_QUIET
+            )
+
+            # If the command succeeds and outputs anything, the tree is dirty
+            if(GIT_STATUS_RESULT EQUAL 0 AND NOT GIT_STATUS_OUTPUT STREQUAL "")
+                string(APPEND PROJ_GIT_HASH "*")
+            endif()
```
jjimenezshaw, line 85: This piece of code is to detect a dirty tree (modified files). This is more helpful in the developer's environment, where things can be modified. It makes clear that the code used was not exactly what is in the git hash.

I think it is not that helpful in the CI environment, where source code should not be modified.

cmake/ProjConfig.cmake
```diff
@@ -51,4 +51,54 @@ if(EXISTS ${AUTOCONF_PROJ_CONFIG_H})
...
+endif()
+message(PROJ_GIT_HASH=\"${PROJ_GIT_HASH}\")
+
+# Tell CMake to re-configure if the Git HEAD or index changes
+if(GIT_FOUND AND EXISTS "${PROJ_SOURCE_DIR}/.git")
+    set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS
+        "${PROJ_SOURCE_DIR}/.git/HEAD" # for the hash
+        "${PROJ_SOURCE_DIR}/.git/index" # for the *
+    )
+endif()
```
jjimenezshaw, line 102: In case something is modified in git, proj is reconfigured. This is helpful in the developer's environment. I don't know if it is too much overhead.

src/release.cpp
```diff
@@ -6,8 +6,15 @@
...
+#ifdef PROJ_GIT_HASH
+#define PROJ_GIT_REV_STR " (" PROJ_GIT_HASH ")"
+#else
+#define PROJ_GIT_REV_STR ""
+#endif
+
+char const pj_release[] =
+    "Rel. " STR(PROJ_VERSION_MAJOR) "." STR(PROJ_VERSION_MINOR) "." STR(
+        PROJ_VERSION_PATCH) PROJ_GIT_REV_STR ", "
+                                             "September 15th, 2026";
```
jjimenezshaw, line 18: This string is the simplest modification to the previous one: it just adds the hash in parenthesis . I am open to other options.

jjimenezshaw:
cc @kbevers

cmake/ProjConfig.cmake
rouault: that should be fine

cmake/ProjConfig.cmake
```diff
@@ -51,4 +51,54 @@ if(EXISTS ${AUTOCONF_PROJ_CONFIG_H})
     "before CMake's build.")
 endif()
 
+# Get the git hash to write it in the release message.
+set(PROJ_GIT_HASH "")
+if(DEFINED ENV{PROJ_GITHUB_SHA})
+    string(SUBSTRING "$ENV{PROJ_GITHUB_SHA}" 0 7 PROJ_GIT_HASH)
+else()
+    find_package(Git QUIET)
```
mwtoews, line 59: This should be put above the `if` block, since `GIT_FOUND` is a variable used on L97.

Furthermore, looking at [CMake docs for FindGit](https://cmake.org/cmake/help/latest/module/FindGit.html), I see that `GIT_FOUND` is deprecated since version 4.2, so these instances should be set to `Git_FOUND` (which is valid since [older versions](https://cmake.org/cmake/help/v3.16/module/FindGit.html)).

jjimenezshaw: Thanks @mwtoews . It is now Git_FOUND.

About the location of the `find_package`, in case `PROJ_GITHUB_SHA` is defined, the second part later in the file is not needed. It can be used to disable the automatic check of the status of the repo.
Do you think it is enough that way? Would you prefer something else?

kbevers:
This is great!

> In case the git tree is "dirty" (there are modified files), an asterisk * is added after the hash.
If this feature is not needed, it can be removed.

Keep it. It can be helpful in debugging situations.

jjimenezshaw:
Is there a way to get the docker image from github actions (before merging into master) to check it behaves as expected?

rouault:
> Is there a way to get the docker image from github actions (before merging into master) to check it behaves as expected?

no, not in the current state of the action

cmake/ProjConfig.cmake
mwtoews: Perhaps the automatic check of the status of the repo (L98-L101) could instead be moved up to this block (L60-L86) instead. (The original point I was making is that the same condition appears twice, except the second check at L97 may not always have `Git_FOUND` defined, which is actually fine with CMake, since undefined vars evaluate false).

rouault (approved):

cmake/ProjConfig.cmake
jjimenezshaw: Is it worth it? I separated in two because it is doing two different things (but based on the same if statement, I know).

jjimenezshaw:
Once merged, the docker image (from master) is doing this:

```
# proj
Rel. 9.9.0 (ad36bbd), September 15th, 2026
usage: proj [-bdeEfiIlmorsStTvVwW [args]] [+opt[=arg] ...] [file ...]
```

cmake/ProjConfig.cmake
gdt: This feels like papering over bugs.    dependency rules should work, regardless of where the sources are stored.  They are in git this week, but anyone could be storing them in mercurial.    Not a big deal, but it seems off.

gdt:
This doesn't change the build instructions, where the semantics could be specified.

In particular, packaging uses releases, which doesn't have .git (and really shouldn't).  I think from reading the diff that was merged that nothing will change.

gdt:
The usual approach for a version from git is to use "git describe" to get something that's a tag, number of commits past the tag, and a short hash.    NUT (ups driver software) does this, for example.   But, I think this is just for people who are hacking on proj, to keep them from being confuse about the N versions on thieir system, and the details are not important.

For any successful software, which certainly includes proj, the overwhelming majority of uses are from being built and distributed in some kind of packaging system, so packaging systems are the most important target.   At the same time, almost all contributions are from people who are building from sources.   The trick is to satisfy both situations at once - I think things are ok, but I do see from time to time projects that are only aligned with source builds and "you can run our CI artifacts".

gdt:
I also wonder if this is going to overall be make cache busting.   In other contexts I am increasingly seeing every file get rebuilt, with cache misses in ccache, on changes that seem like they should not have that kind of scope of change.    For example, if someone changes a single .c file, and commits, and then types make, really only that file should be built, and relinked.   It's ok to write version.h from git and rebuild version.c, but only version.c should include the .h so the the rest of the files don't need recompiling.   I haven't tried this but I didn't see it mentioned in the discussion.

jjimenezshaw:
hi @gdt this "feature" is just a helper to clearly identify the git revision... if that information is available.
If it is not the case, the behaviour is like before. No problem.

As you say, if your source of the files to compile and package is not git... well, the information is not there.

PR #4780, opened 2026-05-22
Clarify installation instructions: building from source, and guidance for packagers

gdt:
This PR contains multiple commits improving the installation instructions.  The big parts are rewording the start to say that it covers building from source (I almost stopped and went looking for that), and cautioning packagers about downloading and vendoring that is likely surprising and contrary to typical packaging system policy.

- [ x] Added clear title that can be used to generate release notes

(To be clear: there was no AI. I left it, unchecked, to state affirmatively that there was none.)

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -144,12 +144,8 @@ Alternatively, one can also check out the master branch or any other revision fr
 Note that some sections below presume an unpacked release tarball, and
 some say to use a git checkout.
 
-The following guides show how to compile and install the software using CMake.
-
-  .. note::
-
-    Support for Autotools was maintained until PROJ 8.2 (see :ref:`RFC7`).
-    PROJ 9.0 and later releases only support builds using CMake.
+As of PROJ 9.0, the only build system is CMake.  (The last release
+with Autotools was 8.2.)
 
 
 Requirements
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -4,10 +4,19 @@
 Installation
 ================================================================================
 
-These pages describe how to install PROJ on your computer without compiling it
-yourself. Below are guides for installing on Windows, Linux and Mac OS X. This
-is a good place to get started if this is your first time using PROJ. More
-advanced users may want to compile the software themselves.
+These pages describe how to install PROJ on your computer, covering a
+variety of operating systems.
+
+There is a section for using package management systems.  If PROJ is
+available for a package management system you already use, and you
+wish to run the release of PROJ offered by that system (usually that's
+fine), this is is the easiest path.
+
+There is a second section for building from source.  This section can
+be followed if there is no binary package available for your
+operating and CPU combination or if you wish to install a different
+version, including the tip of master in git, or pending fixes.  It is
+of course also useful to those creating packages.
 
 Installation from package management systems
 ################################################################################
@@ -130,6 +139,10 @@ Compilation and installation from source code
 
 The classic way of installing PROJ is via the source code distribution. The
 most recent version is available from the :ref:`download page<current_release>`.
+Alternatively, one can also check out the master branch or any other revision from GitHub.
+
+Note that some sections below presume an unpacked release tarball, and
+some say to use a git checkout.
 
 The following guides show how to compile and install the software using CMake.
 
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -130,29 +139,39 @@ Compilation and installation from source code
...
+- If nlohman/json is not found, a vendored version (that may not be
+  current) will be used (details below).
+- If tests are enabled and GoogleTest is not found, a version will be
+  downloaded from the network, compiled and run (details below).
+
 Build requirements
 ++++++++++++++++++
 
-- C99 compiler
+- C99 compiler with POSIX threads
```
rouault: PROJ does build on native Win32 API, so POSIX theads aren't  required

```suggestion
- C99 compiler
```

gdt: sorry,  read CMakeLists to check that against deps list as I know it's easy for them to get out of sync and assumed too much.

I see the `pthread_atfork()` bit that's included conditionally now, and nothing else.

But, `src/test/multistresstest.cpp` includes pthread.h on any system that isn't WIN32.

So is it:

   mulitstresstest.cpp is in the sources but not part of the build (even with tests) and thus we really don't need threads, or

   we aren't supposed to need threads, but multistresstest is built and on !WIN32 it will fail without threads.

Withdrawn, guessing multistresstest is a red herring.

rouault: > mulitstresstest.cpp is in the sources but not part of the build

that

docs/source/install.rst
```diff
@@ -130,18 +139,27 @@ Compilation and installation from source code
...
 --------------------------------------------------------------------------------
 
+Tests default to on, and thus one should either ensure that all of the
+test requirements are met or disable tests.
+
+Packagers and others with policy constraints should note two things
+that may be surprising:
+- If nlohmann/json is not found, a vendored version (that may not be current) will be used (details below).
+- If tests are enabled and GoogleTest is not found, a version will be downloaded from the network, compiled and run (details below).
+
```
rouault: This whole paragraph is a bit oddly placed. Maybe extract some bit of information and dispatch it in the 2 below subparagraphs "Build requirements" and "Test requirements" ?

gdt: I have moved it and shortened it, trying to point out where the details are and not say much more, and otherwise keeping the edits minimal.

rouault (approved):

mwtoews (approved):

docs/source/install.rst
```diff
@@ -130,14 +139,18 @@ Compilation and installation from source code
...
-The following guides show how to compile and install the software using CMake.
-
-  .. note::
+Note that some sections below presume an unpacked release tarball, and
+some say to use a git checkout.
 
-    Support for Autotools was maintained until PROJ 8.2 (see :ref:`RFC7`).
-    PROJ 9.0 and later releases only support builds using CMake.
+As of PROJ 9.0, the only build system is CMake.  (The last release
+with Autotools was 8.2.)
```
kbevers, line 148: Why not just discard this entirely? Autotools is not relevant in the context of current PROJ versions. This just sounds like we're making excuses. In reality the original wording is probably a remnant of a time when we didn't offer versioned docs. If you somehow find yourself with the source code of an old PROJ you can go back to a previous version of the docs.

docs/source/install.rst
```diff
@@ -130,14 +139,18 @@ Compilation and installation from source code
 
 The classic way of installing PROJ is via the source code distribution. The
 most recent version is available from the :ref:`download page<current_release>`.
+Alternatively, one can also check out the master branch or any tag (or
+really any other version) from GitHub.
```
kbevers: Avoids the parentheses:

```suggestion
Alternatively, one can also check out the master branch or any other revision from GitHub.
```

docs/source/install.rst
gdt: Doesn't really matter to be, but one point is that I was trying to edit somewhat lightly.  I don't see it as making excuses so much as being clear about withdrawal timing.

The second point is that people may be coming to a new release from various old releases, perhaps with build scripts, and the build system switch is a big deal -- but it can be described in one brief sentence.  README points to unversioned docs, which change, so for any given release, the online version might be from the future.   I see the benefit of unconfusing someone as larger than the harm for someone who doesn't care to skip quickly over a sentence.

Happy to drop the remark about autoconf/8.2, if the three of you have a combined opinion.

PR #4782, opened 2026-05-22
Fix use of  vendored json when external is chosen

### LINKED ISSUE #4781: vendored nlohmann/json is used even if external was chosen
# AI

No AI!

# Example of problem

With a proj source tree, I removed `include/proj/internal/vendor/nlohmann/json.hpp` and configured the build to use external nlohmann/json by having it present and passing
`-DNLOHMANN_JSON_ORIGIN=EXTERNAL`.

Building, all is well until:
```
[ 82%] Building CXX object src/apps/CMakeFiles/projsync.dir/projsync.cpp.o
In file included from /n0/gdt/SOFTWARE/GEO/PROJ/proj/src/apps/projsync.cpp:43:
/n0/gdt/SOFTWARE/GEO/PROJ/proj/include/proj/internal/include_nlohmann_json.hpp:46:10: fatal error: vendor/nlohmann/json.hpp: No such file or directory
   46 | #include "vendor/nlohmann/json.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
*** Error code 1
```

# Problem description

The build of projsync did not pass the `EXTERNAL_NLOHMANN_JSON` define to the projsync build.  Reading CMakeLists, it is only passed to builds in lib.

After adding cmake code to pass that define for projsync, observe that there is then a test program with the same problem.

# Environment Information

- proj master
- NetBSD 10 amd64 with deps from pkgsrc

# Installation method

- source build

# fix

The cleanest fix is to remove the vendored code, the selection logic for choosing it, and the conditionals, and just expect those building proj to have installed nlohmann/json, just as they have to install cmake, sqlite3, curl, tiff and so on.

Otherwise, apply a patch like

[patch-src-apps-bin_projsync.cmake.txt](https://github.com/user-attachments/files/28164802/patch-src-apps-bin_projsync.cmake.txt)

and repeat for other build failures.

It would be nice to have a CI test that builds with external json and removes the vendored copy, to validate that the build doesn't reference it.   That's just an rm before the build - I found this by adding such an rm to the pkgsrc build recipe.

gdt:
- [x]  zero AI use!
- [x] Closes #4781
- [ ] Tests added (there is no new code to test)
- [x] No new APis; this is a bug fix

nlohmann/json can either be a normal dependency ("external", the standard approach) or a vendored copy (for those who don't have it installed).   There is a header that switches on `EXTERNAL_NLOHMANN_JSON` to use the system version instead of the vendored version.  The build system fails to pass it to some compilations that use json.

This PR adds those defines.

An alternative change in the same spirit would be to add the define to all builds, which would be more robust than adding it to builds that use json today.   A further change that would improve robustness would be to flip the define to VENDORED_NLOHMANN_JSON and thus if there are new targets that use json with passing it, builds in environments lacking system nlohmann/json will fail, instead of silently using the old code.


gdt:
On my system, with normal-system-dependency nlohmann/json, this branch builds and passes tests as well as master does (one issue in test_network already reported on the list), providing confidence that now all compilations using the json implementation switcher have the flag passed.

gdt:
Hey stale-bot, this PR is still active.

rouault (approved):

rouault:
Looks good to me

jjimenezshaw (approved):

gdt:
Thanks for reviewing and merging!

PR #4792, opened 2026-06-21
add proj_crs_is_dynamic(), and cs2cs: warn when source/destination CRS is dynamic, but no epoch is provided

rouault:
Relates to https://lists.osgeo.org/pipermail/proj/2026-June/thread.html#12125 thread

### DIFF
--- a/src/iso19111/c_api.cpp
+++ b/src/iso19111/c_api.cpp
@@ -2138,6 +2138,59 @@ int proj_crs_is_derived(PJ_CONTEXT *ctx, const PJ *crs) {
 
 // ---------------------------------------------------------------------------
 
+/** \brief Returns whether (at least one component of a) CRS has a dynamic
+ * reference frame
+ *
+ * @param ctx PROJ context, or NULL for default context
+ * @param crs Object of type CRS (must not be NULL)
+ * @return TRUE if the CRS is a dynamic CRS.
+ * @since 9.9
+ */
+int proj_crs_is_dynamic(PJ_CONTEXT *ctx, const PJ *crs) {
+    if (!crs) {
+        proj_context_errno_set(ctx, PROJ_ERR_OTHER_API_MISUSE);
+        proj_log_error(ctx, __FUNCTION__, "missing required input");
+        return false;
+    }
+    auto l_crs = dynamic_cast<const CRS *>(crs->iso_obj.get());
+    if (!l_crs) {
+        proj_log_error(ctx, __FUNCTION__, "Object is not a CRS");
+        return false;
+    }
+
+    const auto singleCRSIsDynamic = [](const CRS *singleCRS) {
+        if (auto geodCRS = singleCRS->extractGeodeticCRSRaw()) {
+            const auto &datum = geodCRS->datum();
+            if (dynamic_cast<const DynamicGeodeticReferenceFrame *>(
+                    datum.get()))
+                return true;
+        } else if (auto vertCRS =
+                       dynamic_cast<const VerticalCRS *>(singleCRS)) {
+            const auto &datum = vertCRS->datum();
+            if (dynamic_cast<const DynamicVerticalReferenceFrame *>(
+                    datum.get()))
+                return true;
+        }
+
+        return false;
+    };
+
+    if (dynamic_cast<const SingleCRS *>(l_crs)) {
+        return singleCRSIsDynamic(l_crs);
+    } else if (auto compoundCRS = dynamic_cast<const CompoundCRS *>(l_crs)) {
+        const auto components = compoundCRS->componentReferenceSystems();
+        for (const auto &comp : components) {
+            if (singleCRSIsDynamic(comp.get())) {
+                return true;
+            }
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


kbevers (approved):

src/iso19111/c_api.cpp
```diff
@@ -2138,6 +2138,59 @@ int proj_crs_is_derived(PJ_CONTEXT *ctx, const PJ *crs) {
...
+    auto l_crs = dynamic_cast<const CRS *>(crs->iso_obj.get());
+    if (!l_crs) {
+        proj_log_error(ctx, __FUNCTION__, "Object is not a CRS");
+        return false;
+    }
+
+    const auto singleCRSIsDynamic = [](const CRS *singleCRS) {
+        if (auto geodCRS = singleCRS->extractGeodeticCRSRaw()) {
+            const auto &datum = geodCRS->datum();
+            if (dynamic_cast<const DynamicGeodeticReferenceFrame *>(
```
jjimenezshaw, line 2164: I  remember I asked about dynamic systems some months ago, and the code was checking two things: PJ_TYPE_DYNAMIC_GEODETIC_REFERENCE_FRAME and something else, that I do not remember.

rouault: > and something else

might have been the presence of a velocity grid. I don't remember the exact context. I'd probably stick with the "official" definition of a dynamic... with that recurring impression the concept is somewhat fuzzy...

jjimenezshaw: I think the code is this:

``` C++
CoordinateMetadataNNPtr
CoordinateMetadata::create(const crs::CRSNNPtr &crsIn, double coordinateEpochIn,
                           const io::DatabaseContextPtr &dbContext) {

    if (!crsIn->isDynamic(/*considerWGS84AsDynamic=*/true)) {
        bool ok = false;
        if (dbContext) {
            auto geodCrs = crsIn->extractGeodeticCRS();
            if (geodCrs) {
                auto factory = io::AuthorityFactory::create(
                    NN_NO_CHECK(dbContext), std::string());
                ok = !factory
                          ->getPointMotionOperationsFor(NN_NO_CHECK(geodCrs),
                                                        false)
                          .empty();
            }
        }
        if (!ok) {
            throw util::Exception(
                "Coordinate epoch should not be provided for a static CRS");
        }
    }
```

kbevers: > I'd probably stick with the "official" definition of a dynamic... with that recurring impression the concept is somewhat fuzzy...

I don't know if this is helpful or not, but I tend to avoid using the static/dynamic descriptors and instead talk about plate-fixed and time-dependant reference frames. I think it conveys better what is actually going on. Not sure if this goes against the terminology of ISO19111 though...

test/cli/test_cs2cs_various.yaml
```diff
@@ -1088,6 +1088,7 @@ tests:
     59.4967 -117.61748 329.396
     59.4967 -117.61748 329.396 1988
   out: |
+    Input coordinates lack a coordinate epoch, whereas the source CRS is dynamic. Results might be inaccurate.
```
mwtoews: Maybe split the expected output streams? (e.g. see L1388-1392 in this file)
```suggestion
 stderr: |
    Input coordinates lack a coordinate epoch, whereas the source CRS is dynamic. Results might be inaccurate.
 stdout: |
```
same suggestion for the other two below.

src/iso19111/c_api.cpp
jjimenezshaw: I understand that `plate-fixed` and `time-dependant` are independent flags, right? Looking at the changes in cs2cs, I assume that what Even is trying to find is `time-dependant` systems to check if the epoch is provided.

test/cli/test_cs2cs_various.yaml
rouault: thanks! adjusted

src/iso19111/c_api.cpp
kbevers: > I understand that plate-fixed and time-dependant are independent flags, right? 

Generally yes, but then you can have a little think about ETRFyyyy which pin the Eurasian plate to it's location in 1989 but is still somehow time-dependant (dynamic) due to intra-plate deformation etc... The terminology around all this is FUBAR, I'm afraid 🙃

Probably the smart thing to do is consider if a CRS is time-dependant or not. From that we can infer if a transformation requires temporal coordinates. Which is what this PR adresses, so that's all good.

jjimenezshaw: > but is still somehow time-dependant (dynamic) due to intra-plate deformation

There you are using `dynamic` as equivalent to `time-dependant`, right?
I don't know if EPSG is doing exactly that as well.

I read somewhere that a dynamic system is one with velocities... that is like saying that it depends on the time.

kbevers: Yes, I am. I *think* EPSG has aligned their terminology with ISO19111 so I would assume they use dynamic as well.

PR #4803, opened 2026-07-23
Database: update to EPSG v13.001

rouault:
Just integrating the DB, but not updating unit tests. Most of them should be just updating some names, but at least 2 of them are broken. See below email sent to EPSG:

```
I've just tried to integrate EPSG 13.001 into PROJ and it does break a few of our regression tests related to ETRS89 geoid transformations.

One of those test cases is transforming between BD72 + Ostend height (epsg:4313+5710) and  Amersfoort + NAP height (epsg:4289+5709). This is a test case I made up myself a few years ago to exercise some of PROJ capabilities, so I'm not completely sure if that makes sense, although I imagine that for parts at the Belgium/Netherlands border, that could possibly be relevant. Prior to 13.001, the transformation went through ETRS89, chaining BD72 to ETRS89 (3) + Inverse of ETRS89 to Ostend height (1) + ETRS89 to NAP height (2) + Inverse of Amersfoort to ETRS89 (9)  with following PROJ pipeline:

+proj=pipeline
  +step +proj=hgridshift +grids=be_ign_bd72lb72_etrs89lb08.tif
  +step +proj=vgridshift +grids=be_ign_hBG18.tif +multiplier=1
  +step +inv +proj=vgridshift +grids=nl_nsgi_nlgeo2018.tif +multiplier=1
  +step +inv +proj=hgridshift +grids=nl_nsgi_rdtrans2018.tif

Now that "ETRS89 to Ostend height (1)" and "ETRS89 to NAP height (2)" have been deprecated,  PROJ no longer manages to find the ETRS89 pivot and fallbacks to a lower accuracy transformation going through EVRF2000 height: Ostend height to EVRF2000 height (1) + Inverse of NAP height to EVRF2000 height (1) + BD72 to ETRS89 (3) + Inverse of Amersfoort to ETRS89 (9)

+proj=pipeline
  +step +proj=vertoffset +lat_0=50.7166666666667 +lon_0=4.76666666666667
        +dh=-2.311 +slope_lat=-0.016 +slope_lon=0
  +step +proj=geogoffset +dh=0.005
  +step +proj=hgridshift +grids=be_ign_bd72lb72_etrs89lb08.tif
  +step +inv +proj=hgridshift +grids=nl_nsgi_rdtrans2018.tif


Another test case affected is epsg:8360 "ETRS89 + Baltic 1957 height" to epsg:7423 "ETRS89 + EVRF2007 height"  . With EPSG 12.059a, PROJ offers the use of "Baltic 1957 height to EVRF2007 height (1)" for Czechia, and   'ETRS89 to ETRS89 + Baltic 1957 height (1)' + "ETRS89 to ETRS89 + EVRF2007 height (1)"  for Slovakia  using  Slovakia_ETRS89h_to_Baltic1957 and Slovakia_ETRS89h_to_EVRF2007.tif grids.   But now with 13.001 that those 2 latter ones have been deprecated, it no longer proposes a transformation for Slovakia.


I'm not sure to understand why deprecating those records, rather than just superseding them, and perhaps increase their (in)accuracy to reflect they are not quite registered to an ensemble but rather a particular realization, was needed.
```


rouault:
Answer from Roger Lott:

> If the deprecation of the rolled-back transformations is causing you problems, then I feel that the behaviour is not what I would expect and needs updating. The pipeline should not be regressing to using ensemble-based ETRS89 as its primary route. All of the national realizations are relatable to ETRF2000 and transformations (often but not necessarily null) from the national realization to this should be the preferred path. So (ignoring the vertical component) BD72 > ETRF89-BEL [BEREF2011] > ETRF2000 > ETRF89-NLD [AGRS2010] > Amersfoort. But don't hard code ETRF2000 as the hub – in the future EUREF may recommend using a more recent realization of ETRFyy, in which case we plan to add a series of transformations from national realizations to this. How do you find/create this pipeline? The text in section C of the document circulated ahead of the mid-May meeting, now incorporated as section §6.2.1 in [EPSG Guidance Note 7-7](https://www.iogp.org/bookstore/product/epsg-guidance-note-7-7-etrs89-in-the-epsg-dataset/), can be used.
>
> For the Slovakia height transformation, if you have an historic record of using the now deprecated ETRS89 + Baltic 1957 height to ETRS89 + EVRF2007 height (1) (EPSG:8363), the deprecation trail associated with EPSG:8363 takes you to its replacement SKTRF09 + Baltic 1957 height to SKTRF09 + EVRF2007 height (2) (EPSG:11523, we had to truncate the name as using the 'standard syntax' caused the string length to exceed our 80-character data type). But I imagine that you (quite properly) completely ignore deprecated records. However, EPSG:11523 has an alias, Baltic 1957 height to EVRF2007 height (2).
>
> The argument for deprecating the geoid model transformations is that from a geodetic perspective the vertical CRS values emerging from the application of the model are sensitive to the physical position of the ellipsoid height, so may be sensitive to datum. Not doing the same for transformations between two geodetic CRSs was because in the May meeting Esri said that deprecating would not fix their pipeline problem, so roll-back and supersession was a compromise solution.

rouault:
my answer

> Hi Roger,
>>
>> If the deprecation of the rolled-back transformations is causing you problems, then I feel that the behaviour is not what I would expect and needs updating. The pipeline should not be regressing to using ensemble-based ETRS89 as its primary route. All of the national realizations are relatable to ETRF2000 and transformations (often but not necessarily null) from the national realization to this should be the preferred path. So (ignoring the vertical component) BD72 > ETRF89-BEL [BEREF2011] > ETRF2000 > ETRF89-NLD [AGRS2010] > Amersfoort. But don't hard code ETRF2000 as the hub – in the future EUREF may recommend using a more recent realization of ETRFyy, in which case we plan to add a series of transformations from national realizations to this. How do you find/create this pipeline? The text in section C of the document circulated ahead of the mid-May meeting, now incorporated as section §6.2.1 in [EPSG Guidance Note 7-7](https://www.iogp.org/bookstore/product/epsg-guidance-note-7-7-etrs89-in-the-epsg-dataset/), can be used.
>
> PROJ can only compute a pipeline with at maximum one intermediate datum . So before 13.001, BD72 to Amersfoort was fine since there was only ETRS89 as intermediate. Now we would need to go through 3 (ETRS89-BEL, ETRF2000, ETRS89-NLD).  As in computer science, pretty much is possible in theory, but in practice I won't go to that level of sophistication (ok, maybe if AI is wiped off from the surface of the Earth, ultra-capitalism as well, and looking at stars without being disturbed by one satellite every 2 minutes becomes possible again, I might make an effort). So this use case will remain broken.
>>
>> For the Slovakia height transformation, if you have an historic record of using the now deprecated ETRS89 + Baltic 1957 height to ETRS89 + EVRF2007 height (1) (epsg:8363), the deprecation trail associated with epsg:8363 takes you to its replacement SKTRF09 + Baltic 1957 height to SKTRF09 + EVRF2007 height (2) (epsg:11523, we had to truncate the name as using the 'standard syntax' caused the string length to exceed our 80-character data type). But I imagine that you (quite properly) completely ignore deprecated records. However, epsg:11523 has an alias, Baltic 1957 height to EVRF2007 height (2).
>
> Yes I don't consider deprecated records. And aliases aren't used in that particular situation (would be quite fragile and yet another complication)
>
> Thanks for the explanations

Issue #4097, opened 2024-03-21
Support for EPSG methods 9666 and 1049


yportier:
[P6 I=J+90 seismic bin grid coordinate operation](https://epsg.io/9666-method) (EPSG:9666) and  
[P6 I=J-90 seismic bin grid coordinate operation](https://epsg.io/1049-method) (EPSG:1049)  
are both special cases of the Affine Geometric Transformation (method code 9623)  
They are fully described in [EPSG Guidance Note 7-2](https://www.iogp.org/wp-content/uploads/2019/09/373-07-02.pdf) (p.142-147) and as such should be easy to implement.  
I'd like to help but have no idea where to start.

rouault:
cf https://github.com/OSGeo/PROJ/pull/4094/commits/4d64a35d934697c58c0e3a63aba334687f1b6d08#diff-55a7b23971462a1329140262661d2e756961dfc3bd5df17a34e471ba6c095b6c for some potential inspiration

yportier:
In `singleoperation.cpp`, I am defining the parameters for the Affine Parametric Transformation, but I need first to shift the coordinates so can I apply two affine steps like this ?

```
formatter->addStep("affine");
formatter->addParam("xoff", I0);
formatter->addParam("yoff", J0);
formatter->addStep("affine");
formatter->addParam("xoff", X0);
formatter->addParam("s11", SF * Mx * cos(theta) / inc_SX);
formatter->addParam("s12", SF * My * sin(theta) / inc_SY);
formatter->addParam("yoff", Y0);
formatter->addParam("s21", -SF * Mx * sin(theta) / inc_SX);
formatter->addParam("s22", SF * My * cos(theta) / inc_SY);
```

jjimenezshaw:
One option is to do some maths to get a single affine translation

This is the normal formula
p2 = T + S*p1

But you want to do something like this, right?
p2 = T + S*( T1 + p1)

You can calculate another translation like this
p2 = (T + S*T1) + S*p1

yportier:
Is this what you are suggesting ?

formatter->addStep("affine");
formatter->addParam("xoff", X0 + I0 * SF * Mx * cos(theta) / inc_SX + J0 * SF * My * sin(theta) / inc_SY);
formatter->addParam("s11", SF * Mx * cos(theta) / inc_SX);
formatter->addParam("s12", SF * My * sin(theta) / inc_SY);
[...]

PR #4807, opened 2026-08-09
Implement EPSG:9666 and EPSG:1049 seismic bin grid operation methods

yportier:
Implements the two seismic bin grid coordinate operation methods requested in #4097, following the structure of 4d64a35 (EPSG:9656 Cartesian Grid Offsets).

Both methods reduce to a single +proj=affine step, differing only in the sign of the first matrix row. EPSG declares these operations from the map grid (source) to the bin grid (target), which is the opposite direction to how the formulas are written in guidance note 7-2, so the implemented formula is the guidance note's reverse one.

Validated against EPSG's own worked examples (transformations 6918 and 6919, conversions 32698 and 32699). The I=J−90 example uses a ftUS map grid, which exercises the unit conversion path.

Not included: the four EPSG records themselves. Importing them needs conversion_table widened from 7 to 10 parameters and other_transformation from 9 to 10, a DATABASE.LAYOUT.VERSION.MINOR bump and a full data/sql regeneration. Since all four are placeholder examples ("enter here applicable extent"), they were deliberately left out.

### DIFF
--- a/docs/source/operations/transformations/affine.rst
+++ b/docs/source/operations/transformations/affine.rst
@@ -61,25 +61,38 @@ This can be used to implement:
     * xoff to :math:`I_0 - (s11 \times E_0 + s12 \times N_0)`
     * yoff to :math:`J_0 - (s21 \times E_0 + s22 \times N_0)`
 
-  where:
+  where, naming each quantity after the EPSG parameter that carries it:
 
-    * :math:`s_I = Inc_I / (k \times M_I)` and :math:`s_J = Inc_J / (k \times M_J)`
-      are the number of bin grid units per map grid unit along the I and J axes,
-      with :math:`M_I` and :math:`M_J` the bin widths on the I and J axes,
-      :math:`Inc_I` and :math:`Inc_J` the bin node increments on the I and J
-      axes, and :math:`k` the scale factor of the bin grid.
-    * :math:`I_0` and :math:`J_0` are the bin grid coordinates of the origin
-      point of the bin grid, and :math:`E_0` and :math:`N_0` its easting and
-      northing in the map grid.
-    * :math:`\theta` is the map grid bearing of the bin grid J-axis, clockwise
-      being positive.
+    * :math:`I_0` and :math:`J_0` are "Bin grid origin I" (EPSG:8733) and
+      "Bin grid origin J" (EPSG:8734), the bin grid coordinates of the origin
+      point of the bin grid.
+    * :math:`E_0` and :math:`N_0` are "Bin grid origin Easting" (EPSG:8735) and
+      "Bin grid origin Northing" (EPSG:8736), the map grid coordinates of that
+      same origin point.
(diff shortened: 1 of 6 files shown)


docs/source/operations/transformations/affine.rst
```diff
@@ -49,6 +49,38 @@ This can be used to implement:
...
+    * s11 to :math:`\pm s_I \cos \theta`
+    * s12 to :math:`\mp s_I \sin \theta`
+    * s21 to :math:`s_J \sin \theta`
+    * s22 to :math:`s_J \cos \theta`
+    * xoff to :math:`I_0 - (s11 \times E_0 + s12 \times N_0)`
+    * yoff to :math:`J_0 - (s21 \times E_0 + s22 \times N_0)`
+
+  where:
+
+    * :math:`s_I = Inc_I / (k \times M_I)` and :math:`s_J = Inc_J / (k \times M_J)`
```
jjimenezshaw: I don't find `k` defined anywhere. In the PDF it is named scale factor (SF), but here it should be clarified.

src/iso19111/operation/singleoperation.cpp
```diff
@@ -3386,6 +3386,90 @@ bool SingleOperation::exportToPROJStringGeneric(
...
+        // assumed by the above coefficients.
+        auto sourceCRSProj =
+            dynamic_cast<const crs::ProjectedCRS *>(sourceCRS().get());
+        if (sourceCRSProj) {
+            formatter->startInversion();
+            sourceCRSProj->addUnitConvertAndAxisSwap(formatter, false);
+            formatter->stopInversion();
+        }
+
+        // The target CRS holds bin numbers, which are unit-less, so no unit
```
jjimenezshaw, line 3461: Is there any `round` or cast to integer somewhere? I would expect bin numbers to be a whole value.

docs/source/operations/transformations/affine.rst
yportier: Good catch — I've renamed it to SF to match the parameter table, and renamed in the code too.

src/iso19111/operation/singleoperation.cpp
yportier: That was deliberate. EPSG defines bin grid coordinate systems with a real datatype (EPSG:1051/1052, used by CRS 32596–32599) as well as an integer one (EPSG:32760/32761, used by CRS 32762–32765), both served by these same two methods — the remark on CRS 32596 states bin values may be non-integer. So whether bin numbers are whole values is a property of the coordinate system, not of the method; rounding here would be wrong for the real-valued CRSs and would break invertibility. Guidance note 7-2's formulas have no rounding operator either. I've added a comment at the affine step.

jjimenezshaw:
You can see the documentation rendered in https://osgeo-proj--4807.org.readthedocs.build/en/4807/

PR #4758, opened 2026-04-17
Feature: Snyder Polyhedral Equal Area projections

felixpalmer:
- [x] Claude Code supported my development of this PR. See our [policy about AI tool use](https://proj.org/community/ai_tool_policy.html). Use of AI tools *must* be indicated.
- [x] Tests added
- [x] Added clear title that can be used to generate release notes
- [x] Fully documented, including updating `docs/source/*.rst` for new API

### Background

Please read the updated documentation in `docs/source/operations/projections/polyhedral.rst` first as well as Snyder's original paper: **AN EQUAL-AREA MAP PROJECTION FOR POLYHEDRAL GLOBES** (CARTOGRAPHICA Vol 29 No 1 SPRING 1992 pp 10–21)

### Back-compatibility with existing `isea`

Gives same results, with comparisons in `test/gie/builtins.gie`. I would propose we remove the old `isea` code, replacing with this implementation.

### Architecture

To define a projection a polyhedron and net must be defined. The simplest example is that of the 4-sided tetrahedron along with the net definitions:

- [tetrahedron.h](https://github.com/felixpalmer/PROJ/blob/felix/polyhedral/src/projections/polyhedral/polyhedra/tetrahedron.h) specifies a list of 3D vertices and faces indexing those vertices
- [tsea.h](https://github.com/felixpalmer/PROJ/blob/felix/polyhedral/src/projections/polyhedral/nets/tsea/tsea.h) specifies a list of parents, with `0` as the root

These details are not exposed to the user via the PROJ API, instead they reference them by name, e.g. `+proj=tsea +net=star`

When building the projection, the following process followed:

- Options [parsed](https://github.com/felixpalmer/PROJ/blob/felix/polyhedral/src/projections/polyhedral.cpp#L110) and populated into `pj_polyhedral_data` [struct](https://github.com/felixpalmer/PROJ/blob/felix/polyhedral/src/projections/polyhedral/sphere.h#L27)
- [polyhedral::load_meshes](https://github.com/felixpalmer/PROJ/blob/felix/polyhedral/src/projections/polyhedral/sphere.h#L86) called with the selected polyhedron and matching net, which:
  -  [Unfolds the net](https://github.com/felixpalmer/PROJ/blob/felix/polyhedral/src/projections/polyhedral/unfold.h#L93) by starting with the root face and then attaching other faces in a 2D plane
[...]

Issue #4794, opened 2026-06-26
Support for generic Snyder Equal-Area Projection for Polyhedral Globes


jerstlouis:
As discussed in https://github.com/OSGeo/PROJ/pull/4758#issuecomment-4804215608 (including an advanced eC prototype of all the necessary pieces), it would be great to be able to support e.g.

`+proj=polyhedral +polyhedron={4,3} +net=0,1,2,3,4,5`

(specifying a Cubic Snyder Equal-Area projection).

The `+polyhedron` could support:
- common name,
- extended Schlafli symbol, or
- extended Wythoff symbol

(code already written for this in prototype).

The `+net` would support the current internal face indices, and could automatically switch between the _kis_ or  _meta_ operation based on how many faces are listed. We should validate the net, and also support a default unfolding if no net is specified (code already written for this in the prototype as well).

In addition, we should introduce a `+transform` parameter (or individual parameters as in [`+proj=affine`](https://proj.org/en/stable/operations/transformations/affine.html)) allowing e.g., to define the 5x6 rhombic/icosahedral space in the form of applying an affine transformation (rotation/shearing/scaling) to the icosahedral net, or other arbitrary rotation to the net (without a need for `+proj=pipeline` which is not well supported in many contexts).

We could have various examples of these projections, including IVEA (https://github.com/felixpalmer/PROJ/pull/3) and R. Buckminster Fuller's [Dymaxion orientation](https://en.wikipedia.org/wiki/Dymaxion_map#/media/File:Dymaxion_projection.png) (aka [Airocean](https://proj.org/en/stable/operations/projections/airocean.html)) for ISEA/IVEA keeping all land contiguous (particularly useful since `ST_Area()` for any land polygons will then yield correct values, rather than crossing interruptions with e.g, ISEA/IVEA or DSEA).

PR #4817, opened 2026-08-16
IVEA: DSEA projection onto icosahedron net

felixpalmer:
- [x] Tests added

Followup to work done in https://github.com/OSGeo/PROJ/pull/4758

### Changes

- Adds `+dual` option to `+proj=dsea` to allow unfolding onto icosahedral net aligned with `+proj=isea`
- Adds `+proj=ivea` alias for better discoverability
- Implement dual operation in Conway meta function
- Bounding box fix to include all vertices
- Docs & tests & figures


felixpalmer:
@jerstlouis could you review please?

src/projections/polyhedral/conway.h
```diff
@@ -51,9 +55,14 @@ inline void conway_meta(const Vec3 (&vertices)[NumVertices],
         }
 
         for (int k = 0; k < FanSize; k++) {
-            out[FanSize * i + k][0] = center;
-            out[FanSize * i + k][1] = fan[(k + 1) % FanSize];
-            out[FanSize * i + k][2] = fan[k];
+            Vec3 *tri = out[FanSize * i + k];
+            const bool even = (k % 2) == 0;
+            const Vec3 &vertex = fan[even ? k : (k + 1) % FanSize];
+            const Vec3 &midpoint = fan[even ? (k + 1) % FanSize : k];
```
jjimenezshaw, line 61: `even` (when `dual` is false) is swapping out[1] and [2] respect to the old code , right? Was it a bug?

src/projections/polyhedral/state.h
```diff
@@ -121,15 +125,19 @@ inline void load_meshes(pj_polyhedral_data *Q,
...
 inline Face2D face_reference_point(const pj_polyhedral_data *Q, int face_idx,
                                    DefaultOrigin kind) {
-    const int nfv = Q->face_vertex_count;
-    const int fan_size = 2 * nfv;
+    static_assert(sizeof(FaceTriangle) == 3 * sizeof(Face2D),
+                  "FaceTriangle must be three packed Face2D");
+    const int fan_size = 2 * Q->face_vertex_count;
+    const int n = 3 * fan_size;
+    const auto *corner =
+        reinterpret_cast<const Face2D *>(&Q->face_tris[fan_size * face_idx]);
```
jjimenezshaw, line 134: if `face_idx` is "too big", this is an out of bounds memory access. I know it shouldn't happen. Is it possible in any wrong configuration? Should we check it?

jjimenezshaw (approved):

docs/source/operations/projections/ivea.rst
```diff
@@ -0,0 +1,100 @@
...
+icosahedral net.
+
+Snyder's equal-area mapping :cite:`Snyder1992` applied to the twelve pentagonal
+faces of a regular dodecahedron. Each face is subdivided into 10 spherical right
+sub-triangles, yielding a total of 120.
+
+The icosahedron is unfolded into a planar net and subdivided into 20 × 6 = 120
+right sub-triangles. Each sub-triangle is mapped independently using the
+area-preserving Snyder construction to the 120 dodecahedral spherical triangles.
+
```
jerstlouis, line 21: @felixpalmer Could we please add a reference to [Slice and Dice]() here?

Something like:

> This is fully equivalent to using the icosahedron directly, but changing the vertex from which the great circles originate from the center of the faces to the vertices, applying the _Vertex-oriented great circle projection_ as described in _A “slice and dice” approach to area equivalence in polyhedral map projections_ ([Diederik van Leeuwen and Daniel Strebe, 2006](https://www.tandfonline.com/doi/abs/10.1559/152304006779500687)) to the icosahedron.

docs/source/operations/projections/ivea.rst
```diff
@@ -0,0 +1,100 @@
...
+    *Defaults to* ``isea``.
+
+.. include:: ../options/orient_lat.rst
+
+*Defaults to ~58.40° geodetic (arctan(φ) ≈ 58.2825° authalic).*
+
+.. include:: ../options/orient_lon.rst
+
+*Defaults to 11.25° on a sphere, 11.20° on an ellipsoid.*
```
jerstlouis, line 89: We don't have the legacy backward compatibility issue for IVEA here, so if it's easy and you think it makes more sense, we could have it 11.2 on the sphere as well.

But maybe it's best to stay consistent with ISEA (especially given the `+dual` option on `+proj=isea`), I don't know!

jerstlouis:
cc. @ldesousa @mpadillaruiz @ppKrauss @wrenoud

docs/source/operations/projections/ivea.rst
jerstlouis: IVEA is the last _Figure 42 – Vertex-oriented great circle icosahedron_ in that paper.
And I guess `+proj=isea +dual` yields the _Figure 39 – Vertex-oriented great circle dodecahedron_, and the tetrahedron being its own dual, `+proj=tsea +dual` yields _Figure 31 – Vertex-oriented great circle tetrahedron_.

If we had CSEA (cube) we could do _Figure 37 – Vertex-oriented great circle octahedron_, and if we had OSEA (octahedron), we could do _Figure 34 – Vertex-oriented great circle hexahedron_.

src/projections/polyhedral/conway.h
felixpalmer: Well spotted. The projection is actually symmetric about those two vertices, so it doesn't matter if they are swapped. For simplicity I left it like this

docs/source/operations/projections/ivea.rst
```diff
@@ -0,0 +1,100 @@
+.. _ivea:
+
+********************************************************************************
+Icosahedral Vertex Equal Area
```
jerstlouis: could we maybe change the title to:

Icosahedral Vertex-oriented great circle Equal Area

as the official name?

It's the name in slice and dice (Vertex-oriented great circle) with the Icosahedral designation and the clarification that it is Equal Area (like ISEA).

docs/source/operations/projections/ivea.rst
felixpalmer: I have no strong opinion here, but given that this is a drop-in replacement for ISEA I figured I would keep the same defaults

docs/source/operations/projections/ivea.rst
felixpalmer: Well.. that doesn't spell IVEA :) it is also very long which means that it will wrap in a number of places. I think it is best to keep as it

docs/source/operations/projections/ivea.rst
felixpalmer: Yes, we could in theory add the `dual` flag to all projections but I chose not to as this PR is to deliver IVEA and also I don't see the practical use of `+proj=dsea +dual` - it is just uglier version of the standard DSEA, no?

docs/source/operations/projections/ivea.rst
jerstlouis: it does if you just use the uppercase letters in the acronym, which is how the acronym was coined ;)

"Vertex-oriented great circle" is shortened as `V` ;)

It would just be easier to relate it to the Slice & Dice paper projection, and to the implementations in e.g.

https://github.com/GeoPlegma/GeoPlegma/blob/master/gp-proj/src/projections/projections/vgc.rs

https://github.com/ecere/dggal/blob/main/src/projections/icoVertexGreatCircle.ec

I get your point about being very long and wrapping though...

But there are some fairly long projection names here:

https://proj.org/en/stable/operations/projections/gstmerc.html

"Gauss-Schreiber Transverse Mercator (aka Gauss-" -- this fits in one line in the default doc page width, and might be about the same length as

"Icosahedral Vertex-oriented great circle Equal Area"

?

docs/source/operations/projections/ivea.rst
jerstlouis: @felixpalmer ah... I thought you had already added it there.

> it is just uglier version of the standard DSEA, no?

Actually I do wonder about that.

Unless we're seriously mistaken, `+proj=dsea +dual` is ISEA unfolded unto the net of the dodecahedron. The distortion figures in Slice and Dice are shown on the tetrahedron, but the tables show metrics for tetrahedron/cube(hexahedron)/octahedron/doceahedron/icosahedron.

Several metrics in Table 1 (maximum angular distortion), 2 (edge AB -- edge mid-point -> vertex) and 3 (edge BC -- vertex -> centroid) are better in the "Vertex-oriented great circle" version of the Dodecahedron (DVEA) than the Snyder projection (DSEA), while table 4 shows the distortion is 0 everywhere for Snyder and worst with V-oGC for edge CA (centroid -> edge midpoint). That should be the same vertex labeling as in @mpadillaruiz's new figures in our paper.

We need to keep in mind that those angular / scale distortion metrics are fully meaningful only for the actual cartographic projection, rather than the effect on the inverse-projection of grids drawn as straight lines in the projected space for DGGS. But perhaps there is an advantage of that vertex-oriented great circle projection on the dodecahedron for the cartographic projection?

I would be very interested to better understand those metrics!

For that alone it would be good reason to add `+dual` to the others ;)
Also you know I'm going to wish for `+proj=csea` and `+proj=osea` (or the [generic polyhedra support](https://github.com/OSGeo/PROJ/issues/4794)) next ;)

src/projections/polyhedral/state.h
felixpalmer: @jjimenezshaw I came up with https://github.com/OSGeo/PROJ/pull/4817/changes/a2d77e9ebf32bee4957ec35d661684a17919fc40 to ensure the values in the parents array can't be out of bounds. Worth adding, or am I overengineering it?

jjimenezshaw: nice

docs/source/operations/projections/ivea.rst
felixpalmer: Well, I believe the same author named it and wrote DGGAL (I believe you know him...) so a bit of a circular argument there ;) I also note that in GeoPlegma they have yet another name, VGC - hence the filename. 

Anyway, I'm changing it to "Icosahedral Vertex-oriented great circle Equal Area" as requested. My view has long been that it sufficient to describe this as a dual variant of DSEA - but I know that others see it differently, and as a genuinely useful projection I can see the argument for a dedicated name. What that name is, I'll leave to others to decide - I'm going with the suggestion of @jerstlouis, any follow-up PRs can always improve the naming or documentation if there is appetite

docs/source/operations/projections/ivea.rst
felixpalmer: Well.... I stand corrected :)

Adding `+dual` to both `dsea` and `tsea` improves them both, and the output matches the Slice&Dice paper. I've added the flag to both those projections, it is easy

dsea_net_dual
dsea_net
tsea_net_dual
tsea_net

jerstlouis: > Adding +dual to both dsea and tsea improves them both,

@felixpalmer How exactly did you assess that they are improved? Did you do some of your own measurements, or are you just going by what the paper says and/or my summary? I'm very curious about all that, as I still don't fully understand all this.

jerstlouis: Well I do see some cusps disappearing between the two versions!!

But as I said above, I think this improvement only applies to the cartographic projection, not to shape of zones resulting from inverse projection of a tessellation using straight lines in the projected space (hence why DSEA / IVEA is better than ISEA / DVEA for that purpose, and why DSEA to IVEA is also fully equivalent for that purpose -- if you can figure out how to perform the straight lines tessellation across the different faces, much simpler for hexagons on the icosahedral net than the dodecahedral net).

felixpalmer: @jerstlouis I just judged by eye, if you open the above images and switch between them it's pretty obvious - especially for TSEA

jerstlouis: @felixpalmer Yes, sorry, I did see that as per my later comment above ;)
I had not realized at first they were dual vs. non-dual.

felixpalmer:
From my side, this is good to land, thanks for the input @jjimenezshaw @jerstlouis. I will be out for the rest of August, so will be slow in responding if any more comments come up

jerstlouis:
@rouault Any chance this could possibly still make it into 9.9? :)

This is the projection (introduced by van Leuuwen and Strebe in 2006 Slice and Dice) that hexagonal DGGSs have been waiting for a practical implementation for the last 3 decades ;)

(e.g., see https://ajfriend.com/dggs_compare/ , https://docs.ogc.org/is/21-038r1/21-038r1.html#ivea7h-dggrs , https://vgrid.gishub.vn/)

Thanks!

Issue #1924, opened 2020-02-11
CMake's set(CMAKE_CXX_STANDARD 11) does not work


mwtoews:
It seems that CMake's `set(CMAKE_CXX_STANDARD 11)` approach does not work, since a configure can succeed, but the build fail.

#### Example of problem

Try to configure a CMake build with an older compiler before C++11 support, e.g., GCC 4.4 or MSVC++ 9.0, and the configure messages may look like:

```
...
-- Requiring C++11
-- Requiring C++11 - done
-- Requiring C99
-- Requiring C99 - done
...
```
no problems? I'm actually having a hard time finding an old compiler that fails during CMake's configuration stage.

However when building with MSVC++ 9.0:
```
...
[  2%] Building CXX object src/CMakeFiles/proj.dir/4D_api.cpp.obj
D:\src\PROJ\PROJ\include\proj/util.hpp(33) : fatal error C1189: #error :  Must have C++11 or newer.
NMAKE : fatal error U1077: 'C:\Users\user\AppData\Local\Programs\Common\MICROS~1\VISUAL~1\9.0\VC\bin\amd64\cl.exe' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Users\user\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\amd64\nmake.exe"' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Users\user\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\amd64\nmake.exe"' : return code '0x2'
Stop.
```
or with GCC 4.4
```
...
[  1%] Building CXX object src/CMakeFiles/proj.dir/4D_api.cpp.o
[...]

PR #4824, opened 2026-08-22
Increase CMake minimum version from 3.16 to 3.22.1

### LINKED ISSUE #4756: Minimum CMake version for PROJ 9.9 series
Every couple of years, it's generally good to increment CMake's minimum version used to build PROJ. It was 2 years ago for PROJ 9.4 that the minimum CMake was changed from 3.9 to 3.16 (#3997).

This issue is a general discussion to see if we can agree if PROJ's minimum CMake version should be increased for the PROJ 9.9.x series (anticipated September 2026), and what version it should go to. Note that [RFC 3](https://proj.org/en/stable/community/rfc/rfc-3.html) says that "software packages will always be at least two years old", so the upper limit could be [CMake 3.29](https://cmake.org/cmake/help/latest/release/3.29.html)-ish (released [March 21, 2024](https://www.kitware.com/cmake-3-29-0-available-for-download/)).

Refer to [What’s new in CMake](https://cliutils.gitlab.io/modern-cmake/newcmake/) for a handy/quick view of features from CMake releases.

For example, I see a decent case in #4753 that the minimum should be moved to 3.18 (or even 3.22). Are there other case examples that would benefit from other modern versions of CMake?

mwtoews:
This PR upgrades the CMake minimum version from 3.16 to 3.22.1. It also cleans up code that was needed to support older CMake versions.

I will consider some improvements using [`target_compile_features()`](https://cmake.org/cmake/help/latest/command/target_compile_features.html#command:target_compile_features) (discussed in #1924) in a follow-up.

Closes #4756


jjimenezshaw:
What about upgrading Ubuntu already to 24.04?

mwtoews:
I'd prefer keeping the CI testing to use the minimum supported versions. Also, I might hold this on draft to separate the CI upgrade stuff separately.... see <s>#4825</s> #4835

mwtoews:
The minimum CMake version is tested with `mingw_w64_build`, which runs `ubuntu-22.04` (jammy) with [cmake 3.22.1](https://packages.ubuntu.com/jammy/cmake). The remaining CI uses [3.28.3](https://packages.ubuntu.com/noble/cmake) to CMake 4.4.

rouault (approved):

PR #4825, opened 2026-08-22
Upgrade CI images

mwtoews:
This upgrades the CI ubuntu images from 20.04 to 22.04. Also clang 10 to clang 12, as supported by these images.

### DIFF
--- a/.github/workflows/clang_linux.yml
+++ b/.github/workflows/clang_linux.yml
@@ -54,4 +54,4 @@ jobs:
           restore-keys: ${{ runner.os }}-cache-clang-${{ matrix.id }}-
 
       - name: Run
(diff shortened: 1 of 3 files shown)


mwtoews:
A few notes:
 - clang-11 with ASAN threw many errors, unsure why. These seem to go away with clang-12, which is fine with me, but I have not investigated if the errors from clang-11 are worth looking into.
 - There are issues passing (e.g.) `CXXFLAGS=-Werror` with `find_packages()`, as this fails finding packages (the warning/error is hidden in a log file). A possible workaround is used for CMake 3.22.1 to remove this flag until after `project()`, then restored as `CMAKE_CXX_FLAGS`. This workaround can be removed at a later time to use [`-D CMAKE_COMPILE_WARNING_AS_ERROR=ON`](https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html) instead, added with CMake 3.24.

rouault (approved):

mwtoews:
While this gets a "green" CI pass, I'd rather use a solution that doesn't modify the CMakeLists.txt for a workaround.

PR #4830, opened 2026-08-27
Doc: fix LaTeX 'Erroneous nesting of equation structures' errors

### LINKED ISSUE #4831: [Docs] Helmert transform: "Erroneous nesting of equation structures" errors
### What is the bug?

In https://proj.org/en/stable/operations/transformations/helmert.html

Image

### Steps to reproduce the issue

Browse to https://proj.org/en/stable/operations/transformations/helmert.html

### Versions and provenance

Windows 10 1903 Home | Chrome 151.0.7922.174 / Firefox 154.0.1

### Additional context

_No response_

rouault:
Fixes #4831

Source: https://stackoverflow.com/a/26165916

### DIFF
--- a/docs/source/operations/transformations/vertoffset.rst
+++ b/docs/source/operations/transformations/vertoffset.rst
@@ -68,9 +68,7 @@ The :math:`Z_{dest}` destination elevation is obtained from the
 
 .. math::
 
-    \begin{align}
-        Z_{dest} = Z_{src} + \left( dh + slope_{lat} * {\rho}_0 * (\phi - {\phi}_0) + slope_{lon} * {\nu}_0 * (\lambda - {\lambda}_0) * cos(\phi) \right)
-    \end{align}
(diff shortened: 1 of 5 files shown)


docs/source/operations/transformations/vertoffset.rst
```diff
@@ -68,9 +68,9 @@ The :math:`Z_{dest}` destination elevation is obtained from the
 
 .. math::
 
-    \begin{align}
+    \begin{aligned}
```
mwtoews: This environment and the the other starting on L87 don't use any alignment, so these could actually be removed.

mwtoews (changes requested):
HTML output looks good!
It seems that that `docs/source/operations/transformations/vertoffset.rst` doesn't actually use any alignment features.

docs/source/operations/transformations/vertoffset.rst
rouault: indeed, removed

mwtoews (approved):

Issue #4730, opened 2026-04-01
proj.db location when bundling for macOS


mwestphal:
#### Example of problem

It looks like the expected location of proj.db `./share/proj` is not a valid location when installing a macOS package .dmg containing the libproj.

Indeed, macOS package uses completelly different directory architectures, such a file would be expected to be put in `./Resources`.

While proj consumer would probably be able to handle that by specifying to proj where to look for proj.db, I think this may the PROJ responsability to pick the right location, or at least to offer a CMake option to select the default behavior.

Current code `src/filemanager.cpp`:

```
    // The initial path was something like foo/bin/libproj.so"
    out += "/share/proj";
    return out; 
```

Could be something like:

```
    // The initial path was something like foo/bin/libproj.so"
#if PROJ_MACOS_BUNDLE
    out += "/Resources";
#else
    out += "/share/proj";
#endif
    return out; 
```

Let me know if this is something PROJ should be supporting.

rouault:
Foreword: I know almost nothing about Mac

> Indeed, macOS package uses completelly different directory architectures, such a file would be expected to be put in `./Resources`.

Per CMakeLists.txt, pro.db is installed in
````
set(PROJ_DATA_PATH "${CMAKE_INSTALL_FULL_DATADIR}/proj")
````
Who/what move things around afterwards?

That said in GDAL I see we have in gdal.cmake (at the root of osgeo/gdal) specific logic related to if(GDAL_ENABLE_MACOSX_FRAMEWORK) that sets
```
  set(CMAKE_INSTALL_DATADIR
      "${FRAMEWORK_SUBDIR}/Resources"
      CACHE INTERNAL "Installation sub-directory for resources" FORCE)
```

So maybe the right way to address that would be to have a similar setup in PROJ and then the subpath would be provided to src/filemanager.cpp

gdt:
There are two approaches to using proj or anything on macOS.  One is to treat it as a slightly odd regular computer, building normal programs with a prefix.  In that case, there should be no special treatment.   For example, I use pkgsrc on macOS, with `--prefix=/opt/pkg` and everything under that path is accessible (indeed, that's the point of choosing it).

Instead of building a normal POSIX program, one can choose to build native mac apps, which has different rules and things then go in very mac-ish places.

I don't particularly care how a mac app gets built, if it does, but I would ask that
   - there be no default to a mac app just because it's built on a mac.  A different build target, or a -DPROJ_BUILD_MACAPP=ON or whatever
[...]

PR #4843, opened 2026-09-07
Replace `PROJ_DATA_PATH` with `${CMAKE_INSTALL_DATADIR}/proj`

### LINKED ISSUE #3742: patch for data/CMakeLists.txt
#### Example of problem
When building proj from source  I need to configure with one prefix but do install with another. This is to ensure that configuration uses correct final paths settings but the install goes into a  "sandbox" as  it  is used for  a creation of an RPM. 
All works fine as expected  (our build system handles all spec files , etc). The only issue is the files that are generated during build in data/ are not getting installed where expected.  Here are the details:

```bash
#  configure and make, $APATH is set to needed path
mkdir builddir
cd builddir
cmake -DCMAKE_INSTALL_PREFIX=$APATH -DCMAKE_BUILD_TYPE=Release ..
make -j 8
# install, here $(ROOT) designates an additional path prefix, set in the environment 
cmake -DCMAKE_INSTALL_PREFIX=$(ROOT)/$APATH -P cmake_install.cmake
```
#### Problem description
All  files generated by build  (but those from data/) are installed as expected in `$(ROOT)/$APATH`.
However all files from data/ are getting installed in `$APATH` instead of `$(ROOT)/$PATH`

#### Expected Output
Files generated in  data/ (per data/CMakeLists.txt) need to be installed in `$(ROOT)/$APATH`

#### Environment Information
 - PROJ version 9.1.1
 - cmake 3.22.1
 - Operation System Information: Rocky 8.7

#### Installation method
From source using cmake
 
#### A simple patch fixes this issue
```bash  
--- PROJ-9.1.1/data/CMakeLists.txt	2022-12-01 01:51:00.000000000 -0800
+++ PROJ-9.1.1/data/CMakeLists.txt.new	2023-05-24 13:37:33.855510748 -0700
@@ -101,5 +101,5 @@
 )
 install(
   FILES ${ALL_DATA_FILE}
-  DESTINATION ${PROJ_DATA_PATH}
+  DESTINATION ${CMAKE_INSTALL_DATADIR}/proj
```

Here `${CMAKE_INSTALL_DATADIR}` is translated properly by cmake into `$APATH/share/`.
The `${PROJ_DATA_PATH}`  is defined  in the top level CMakeLists.txt as `set(PROJ_DATA_PATH "${CMAKE_INSTALL_FULL_DATADIR}/proj")` does not work.

mwtoews:
This replaces the CMake variable `PROJ_DATA_PATH` from an absolute path (e.g. `/usr/local/share/proj`) to a relative path `${CMAKE_INSTALL_DATADIR}/proj` (e.g. `share/proj`).

This is because [`install(FILES <file> ... [...])`](https://cmake.org/cmake/help/latest/command/install.html#files) expects that [`DESTINATION <dir>`](https://cmake.org/cmake/help/latest/command/install.html#common-options) is a relative path, and CMake 4.4 and later will show a warning for this with [`-Wauthor`](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-W). Furthermore, "it is strongly recommended that [packages] use a path that begins with the appropriate relative [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#module:GNUInstallDirs) variable", as done in this PR. I believe this allows the prefix to be adjusted at install time (which is not fully supported in this project, yet.)

When `EMBED_PROJ_DATA_PATH=ON` is used (default), the absolute path is stored for the `PROJ_DATA` define (not sure what happens at install time with this one).

The primary change in this PR was suggested in #3742 which @rouault was "not a big fan of" at the time.

Also, related to #4730


rouault (approved):

kbevers:
@mwtoews are you happy with this PR as is?

PR #4847, opened 2026-09-08
CMake: define target_compile_features with C/C++ standards

mwtoews:
This defines the high level meta features for C/C++ support based on their standard, and set using [`target_compile_features`](https://cmake.org/cmake/help/latest/command/target_compile_features.html) to the main `proj` target. It is PUBLIC, meaning that the C/C++ standard applies to compile this library, and anything that links to it. The exported target would look like this:
```
Properties for TARGET PROJ::proj:
  PROJ::proj.LOCATION = "/tmp/proj_shared_install_from_dist/lib/libproj.so.25.9.9.0"
  PROJ::proj.INTERFACE_INCLUDE_DIRECTORIES = "/tmp/proj_shared_install_from_dist/include"
  PROJ::proj.INTERFACE_LINK_LIBRARIES = <NOTFOUND>
  PROJ::proj.INTERFACE_COMPILE_FEATURES = "c_std_99;cxx_std_17"
```

Also tidy-up some of the messages while setting the CMake standard variables. These never fail, so `message(CHECK_FAIL ...)` is not needed (the failure, e.g. non-existing standard, happens before in `project()`).

Background information [here](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/).

Xref #1924

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,19 +25,19 @@ include(CheckCXXCompilerFlag)
 # Make CMAKE_CXX_STANDARD available as cache option overridable by user
 set(CMAKE_CXX_STANDARD 17
   CACHE STRING "C++ standard version to use (default is 17)")
-message(STATUS "Requiring C++${CMAKE_CXX_STANDARD}")
+message(CHECK_START "Requiring C++${CMAKE_CXX_STANDARD}")
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-message(STATUS "Requiring C++${CMAKE_CXX_STANDARD} - done")
+message(CHECK_PASS "done")
 
 # Set C99 version
 # Make CMAKE_C_STANDARD available as cache option overridable by user
 set(CMAKE_C_STANDARD 99
   CACHE STRING "C standard version to use (default is 99)")
-message(STATUS "Requiring C${CMAKE_C_STANDARD}")
+message(CHECK_START "Requiring C${CMAKE_C_STANDARD}")
 set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_C_EXTENSIONS OFF)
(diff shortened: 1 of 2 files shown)


kbevers:
@mwtoews do you want this included in 9.9.0?

mwtoews:
@kbevers yes, and nothing else queued up.

kbevers:
> @kbevers yes, and nothing else queued up.

Perfect. It looks good to me so merging and including in release notes

landryb:
fwiw, this strangely breaks configure in mapnik (4.1.2):
```
-----
CMake Error in CMakeLists.txt:
  No known features for C compiler

  ""

  version .
-----
``` 
(see https://marc.info/?l=openbsd-ports-cvs&m=178982741117532&w=2 for the full log)

i've reverted this commit (which added `'INTERFACE_COMPILE_FEATURES "c_std_99;cxx_std_17"'` to `proj-targets.cmake`) and that unbreaks configure for mapnik. this is puzzling..

i've looked and found assimp/assimp#4257 which looks very similary, but blindly adding `set(CMAKE_C_STANDARD 99)` to mapnik's cmake goo in https://github.com/mapnik/mapnik/blob/master/CMakeLists.txt#L136 didn't help. cmake can be cryptic :)

i'll have to retry with latest mapnik but afaict not much changed there wrt cmake/standards.

mwtoews:
Thanks for report @landryb 
We do try to test and catch these types of issues with [post install tests on a generic C++ app](https://github.com/OSGeo/PROJ/blob/master/test/postinstall/cpp_app/CMakeLists.txt) but this one might have slipped thru. I'll investigate and follow-up in a new issue and/or PR when I find the time. Another patch idea would be to replace `PUBLIC` with `PRIVATE`.

stefanb:
@landryb, the workaround for mapnik is in:
* https://github.com/Homebrew/homebrew-core/pull/313034

Probably worth reporting to mapnik or addressing in proj.

